Issue #5, opened 2023-06-02
Abstract out import logic, bash utils


fvankrieken:
All repos currently have their own code to import datasets. We'll need to 
- check for any idiosyncracies, see if certain data products actually need special logic
- write common code in some sort of utils folder
- update code in individual folders

Bash utils might as well go here too if possible since they're fairly similar in structure in our codebases. Anything to do with python imports should likely be pulled out as well, though I'm not sure if we're doing any imports in python (outside of my [testing branch](src/dataloading.py) moving ztl to prefect). Might be worth adding that code here (the python utils, not the prefect testing). @damonmcc offhand do you know if we're importing data from do/edm-recipes via python anywhere else in our codebase?

fvankrieken:
Notes on idiosyncracies:
- `import_public` is a very common function to grab sql files from `edm-recipes`
   - cbbr assumes that a zipped sql file is available for download and grabs that rather than raw sql

fvankrieken:
Notes on changes:
- we commonly have `urlparse` function which parses data from connection string. Rename to `parse_connection_string`

PR #7, opened 2023-06-02
global config.sh first draft

fvankrieken:
Start of #5 - figure it'd make sense to merge this now as a "first draft" of a global `config.sh` so that when I actually start moving things over to using it we can see changes more easily in review.

So far, I've gone through all 8 `config.sh` files and made a first attempt of one file that should meet the needs of all repos, assuming I make changes outline [here](https://github.com/NYCPlanning/data-engineering/issues/5#issuecomment-1573810141). As you can see in the files changed, no actual changes to data product functionality yet

### DIFF
--- a/bash_utils/config.sh
+++ b/bash_utils/config.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 
+DATE=$(date "+%Y-%m-%d")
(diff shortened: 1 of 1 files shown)


fvankrieken:
I will not be squashing merges! Especially since I want to merge this into main then keep working on the same branch

damonmcc (approved):

PR #8, opened 2023-06-02
Move repos over to shared bash utils

### LINKED ISSUE #5: Abstract out import logic, bash utils

### LINKED ISSUE #6: Create "master" dev container
This will take a bit of testing, and we also need to think about longer term how to make sure that it's ergonomic for more "lightweight" testing dev containers for development. But for now, our products are similar enough that we likely can have one top-level specification for a dev container for the whole repository

fvankrieken:
Main scope of this is #5 for cbbr and pluto, but will have to close #6 (in a first iteration) to get there most likely. Workflow is currently working through data products one at a time and building locally, moving over to the new "global" bash utils and fixing errors as they pop up

Todo list in #5 is tracking current state of testing


.github/workflows/pluto_build.yml
```diff
@@ -35,44 +35,51 @@ jobs:
       AWS_S3_ENDPOINT: ${{ secrets.AWS_S3_ENDPOINT }}
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-    working-directory: ./db-pluto/pluto_build
     steps:
```
fvankrieken, line 38: I checked in an invalid file to main in my initial commit to get these actions on main - this is just correcting/reverting something that I tried. working-directory can only be specified in steps, not job-wide

fvankrieken:
Pluto built successfully!

https://github.com/NYCPlanning/data-engineering/actions/runs/5205041860

https://github.com/NYCPlanning/db-pluto/issues/447

bash_utils/config.sh
```diff
@@ -21,18 +20,28 @@ function set_env {
             export $(cat $envfile | sed 's/#.*//g' | xargs)
         fi
     done
+    # TODO this is a terrible spot for this, docker setup would make more sense but then .env must be loaded
+    mc config host add spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
```
damonmcc, line 24: i feel you with this not feeling great, but seems like a good enough approach since `spaces` get used kind of like an envar and (like you noted) repeatedly loading `.env` later hurts more

bash_utils/config.sh
```diff
@@ -173,9 +182,8 @@ function shp_export {
 }
 
 
-#colp, facdb
-function fgdb_export {
-    urlparse ${BUILD_ENGINE}
+function fgdb_export_no_docker {
```
damonmcc, line 185: `fgdb_export_no_docker` is defined twice. guessing the first one is the intended one?

bash_utils/config.sh
```diff
@@ -21,18 +20,28 @@ function set_env {
...
+  trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
+  # echo an error message before exiting
+  trap 'echo "\"${last_command}\" command filed with exit code $?."' EXIT
 }
 
 
 function run_sql_file {
-    local filename = ${1}
-    psql ${BUILD_ENGINE} --set ON_ERROR_STOP=1 --file ${filename}
+    psql ${BUILD_ENGINE} --set ON_ERROR_STOP=1 --single-transaction --file "$@"
```
damonmcc, line 39: nice. is the `single-transaction` flag useful because files almost always have more than one command and this rolls them back on an error?

db-zoningtaxlots/sql/area_zoningdistrict_create.sql
```diff
@@ -18,8 +18,6 @@ CREATE TABLE validzones AS (
 ); 
 CREATE INDEX validzones_geom_idx ON validzones USING GIST (geom gist_geometry_ops_2d);
 
-VACUUM ANALYZE;
-
```
damonmcc, line 22: just curious why you dropped these but kept them in pluto files?

damonmcc (changes requested):

bash_utils/config.sh
fvankrieken: Exactly. This will be a moot point if we move to dbt but for now this makes sense for developer QOL - can just restart a script at a specific sql file call if something goes wrong

bash_utils/config.sh
fvankrieken: Oops. Yup

bash_utils/config.sh
fvankrieken: Main thing here is that it re-adds the space when this is called each time. But this works fine, just redundant. If we keep doing some work to make the dev containers more optimized can do that then - I might make an issue for this just so I don't forget

db-zoningtaxlots/sql/area_zoningdistrict_create.sql
fvankrieken: Scratch that last point

Issue #4, opened 2023-06-02
Sort out github actions


fvankrieken:
All github actions are currently unable to run due to them being in individual data products' folders rather than a top level `.github` folder. We'll need to both move them and figure out if there's a decent way to have a convenient organizational scheme so that triggering via workflow dispatch isn't a nightmare

PR #10, opened 2023-06-08
2023 06 centralize GitHub files

fvankrieken:
#4 

This isn't meant to make them all valid - the underlying code may still not work, and there might be issues with the working directory of these actions not being the data product subfolders. But wanted to get these into main so that while working on specific data products, actions can be triggered via workflow dispatch.

It's maybe a little risky to be putting broken actions in main. We could not have workflow dispatches for them, but then we can't trigger them in dev branches...

My solution is 
- comment out all non-`workflow-dispatch` triggers
- put "(TODO)" at the beginning of their names. They'll separate out alphabetically from other workflows so all invalid workflows are in one chunk.

Lemme know if you have thoughts on a more elegant solution


fvankrieken:
Also put "z"s at the beginning of workflow names so that they'll sort last alphabetically on left of Action pane in github

.github/workflows/stale_issues.yml
```diff
@@ -1,12 +1,11 @@
-name: Mark stale issues and pull requests
+name: (TODO) Mark stale issues and pull requests
 
 on:
-  schedule:
-  - cron: "0 0 * * *"
+#  schedule:
+#  - cron: "0 0 * * *"
```
fvankrieken, line 6: This one theoretically could be used now, but figured we should maybe talk about if we actually want this

damonmcc (approved):

PR #13, opened 2023-06-09
Main docker image for build + colp bash changes

fvankrieken:
Continuing #5 for COLP, though then ran into docker issue so worked on #18 as well. 

changes are
- renaming of bash utils folder (open to suggestions, just didn't like what I had it as originally). Added python requirements compile/install scripts from template repo
- Added a shared python folder as well which eventually can hold utils but for now just has "global" requirements.
- docker folder - currently one dockerfile as our "main" build image. It has python requirements, gdal-bin, postgres cli, geosupport. Ends up a bit under 900MB. We could go a direction of having a couple - one without geosupport for example - and using appropriate ones in github actions for builds, use the "main" one for devcontainer (which I just called "base" for now, happy to change to something a bit more informative). The docker publish can only be built via github actions at the moment (or on your local machine but NOT in the devcontainer). 
  - The published image is used in the dev container, colp build workflow, and devdb build workflow (as a check to make sure it still builds. And it does!)
- update colp scripts to use shared bash utils, not invoke docker for geosupport

[COLP build](https://github.com/NYCPlanning/data-engineering/actions/runs/5404737929)
[DevDB build](https://github.com/NYCPlanning/data-engineering/actions/runs/5404389147)
[Docker image publishing](https://github.com/NYCPlanning/data-engineering/actions/runs/5404729190)


fvankrieken:
Blocked by #15

.github/workflows/cbbr_build.yml
```diff
@@ -45,7 +45,6 @@ jobs:
           python-version: "3.11.1"
 
       - name: install dependencies ...
-        working-directory: db-cbbr/cbbr_build
```
fvankrieken, line 48: Not sure how this ended up in this PR, seems like a git mistake by me but this is an unnecessary line so 🤷

fvankrieken:
Run using "shared" build workflow https://github.com/NYCPlanning/data-engineering/actions/runs/5405240057/jobs/9820511100

.github/workflows/cbbr_build.yml
damonmcc: this may be required since one of the lines is `pip3 install -r requirements.txt`

docker/publish.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "$DOCKER_PASSWORD" | docker login -u $DOCKER_USER --password-stdin
+# Build image - Once we reach some sort of MVP, maybe worth starting versioning. For now, just latest
+docker build \
+    --tag $DOCKER_IMAGE_NAME:latest .
+docker push $DOCKER_IMAGE_NAME:latest
+
+rm Dockerfile
+rm -rf bash
+rm -rf python
```
damonmcc, line 24: probably doesn't affect anything, but are these needed? maybe especially when it's run locally?

damonmcc:
love it! just left a few comments and feel like the PR title could be more descriptive. maybe something like "add base docker image and update colp"?

docker/base
```diff
@@ -0,0 +1,16 @@
...
+
+RUN curl -O https://dl.min.io/client/mc/release/linux-amd64/mc \
+    && chmod +x mc \
+    && mv ./mc /usr/bin/ 
+
+COPY bash/ bash/
+COPY python python/
+
+RUN bash/compile_python_packages.sh python \
+    && bash/install_python_packages.sh python \
```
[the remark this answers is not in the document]
alexrichey: yeah, wouldn't we want to go solely off the (version controlled) `requirements.txt` when building the image?

db-colp/bash/04_export.sh
```diff
@@ -1,40 +1,39 @@
...
 rm -rf output
 mkdir -p output && (
     cd output
 
     echo "Exporting COLP"
-    CSV_export colp
-    SHP_export colp POINT
-    FGDB_export colp POINT
+    csv_export colp
+    csv_export colp POINT
```
alexrichey, line 11: is this intended to be csv and not shp?

alexrichey (dismissed):
LGTM - had a question about a potential copy-pasta, but other than that, 👍 
(Also, quick nit, but in your editor, can you enable adding new-lines at the end of files?)

docker/base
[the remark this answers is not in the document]
fvankrieken: Makes sense

docker/publish.sh
fvankrieken: They're really just temporary parts of the build, I figure leave the workspace like I found it. The reason for not callin the Dockerfile "Dockerfile" was that we might for now want a few dockerfiles in there, and since there's not a lot of overhead just having the files at the same folder level rather than subfolders made sense to me

db-colp/bash/04_export.sh
fvankrieken: Ooh thank you, that's not a great little change

docker/publish.sh
fvankrieken: Though I see your point if this is mainly for running in an action

PR #15, opened 2023-06-12
2023 06 fvk pluto upload tweaks

fvankrieken:
Pluto has a slightly more specific fgdb export - two different tables are exported as different layers. Working to provide that functionality in a slightly more generalized fashion.

~~As part of this, should also make sure all fgdb exports are working as expected~~ only colp and facdb also use fgdb export, and they haven't been converted over yet so that can happen when they are. FacDB needs `-s_srs EPSG:4326 -t_srs EPSG:2263` as additional arg, colp seems more standard


.devcontainer/Dockerfile
```diff
@@ -10,7 +10,10 @@ ARG USER_GID=$USER_UID
 RUN if [ "$USER_GID" != "1000" ] || [ "$USER_UID" != "1000" ]; then groupmod --gid $USER_GID vscode && usermod --uid $USER_UID --gid $USER_GID vscode; fi
 
 RUN apt update && export DEBIAN_FRONTEND=noninteractive \
-    && apt -y install postgresql-client python3-pip python3-distutils jq zip unzip git bash-completion
+    && apt -y install postgresql-client python3-pip python3-distutils jq zip unzip git bash-completion locales
+
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en
```
fvankrieken, line 16: I'm having issues running stuff in the new container without this. Haven't seen this with other containers using the same base image so it's a bit odd to me but this resolves it

fvankrieken:
looks like we cannot get 3.6 via apt/apt-get, only 3.4. I found some docs that said to use ppa:ubuntugis/ppa as source to get latest gdal-bin, but seems that's not needed any more, but that regardless of source the latest available is 3.4. So need a slightly more manual install if we're not using a docker image

fvankrieken:
Testing using gdal docker image here

https://github.com/NYCPlanning/data-engineering/actions/runs/5268191473

.github/workflows/pluto_build.yml
```diff
@@ -30,7 +31,7 @@ jobs:
         ports:
           - 5432:5432
     env:
-      BUILD_ENGINE: postgresql://postgres:postgres@localhost:5432/postgres
+      BUILD_ENGINE: postgresql://postgres:postgres@postgres:5432/postgres
```
fvankrieken, line 34: When using a container along with the postgres service, this is how its exposed

damonmcc (approved):
lgtm! GIS was able to use the new OpenFGDB format?

db-pluto/pluto_build/bin/setup.sh
```diff
@@ -1,13 +1,21 @@
...
+    apt-get install -y curl zip jq unzip postgresql-client
+    #wget http://download.osgeo.org/gdal/3.6.4/gdal-3.6.4.tar.gz
+    #tar -xvzf gdal-3.6.4.tar.gz && (
+    #    cd gdal-3.6.4
+#
+#        ./configure --prefix=/usr 
+#        make
+#    )
+#    rm -rf gdal-3.6.4
+#    rm gdal-3.6.4.tar.gz
```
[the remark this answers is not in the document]
fvankrieken: Ah yeah I would say not worth preserving

PR #22, opened 2023-06-16
delete duplicate archive, fix typo in max_bg_procs

fvankrieken:


damonmcc:
@fvankrieken was there an issue with a build or are these just things you noticed in the code?

PR #25, opened 2023-06-20
improve dataset build workflow

### LINKED ISSUE #12: Norms around github workflows
Bit of a brainstorming issue but wanted to get some thoughts down

Would love to standardize github actions for builds a bit, whether in their individual yml files or if we can abstract some out and build a product based on arguments supplied. But really would love some consistency and clarity around these product-building workflows. Stuff like
- options to run specific parts of build process - just data loading/ingestion, or just through build but no export to DO/EDM_DATA
- options to run whole process but "clean up" at end - don't overwrite any data in DO or EDM_DATA and remove any exports, but still test "entire" pipeline for product
Some products have this fairly flushed out, others just run the whole pipeline. Others have some overlap in scope of the different builds - ztl's exporting is sort of split between the "build" script and the "qaqc" script

This could become a bit of a moot point depending on what we prioritize - might be worth putting this time just in to general infrastructure (or specifically around the build process, work around getting everything happening in a remote db/engine that we've talked about). But could also be an easy-ish win.

@damonmcc your thoughts around testing (or really the part of that which interacts with github actions, how we automate it) is a bit distinct from this but also fits into the title of this issue at least

damonmcc:
resolves #12 
follow-up to PR #19 

changes
- finish adding CBBR build in centralized build workflow
- add PLUTO build to centralized build workflow
- remove the "is a minor release build" input to PLUTO build
  - by using the version set in `version.env` to infer whether it's a major or minor release
- ensure changes to the centralized github action files (`build.yml`, `tests.yml`) trigger the Template Datasets tests
- add dev container extension for formatting yaml files

tests
- [pluto build](https://github.com/NYCPlanning/data-engineering/actions/runs/5327347377) w/o exporting

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ on:
         description: "A short note about this build (optional)"
         type: string
         required: false
-      do_export:
+      will_export:
         description: "Export build outputs"
         type: boolean
         required: true
@@ -36,7 +36,7 @@ jobs:
     with:
       logging_level: ${{ inputs.LOGGING_LEVEL }}
       build_note: ${{ inputs.BUILD_NOTE }}
-      do_export: ${{ inputs.DO_EXPORT }}
+      will_export: ${{ inputs.will_export }}
   cbbr:
     if: inputs.dataset_name == 'cbbr'
     uses: ./.github/workflows/cbbr_build.yml
@@ -44,4 +44,4 @@ jobs:
     with:
       logging_level: ${{ inputs.LOGGING_LEVEL }}
       build_note: ${{ inputs.BUILD_NOTE }}
-      do_export: ${{ inputs.DO_EXPORT }}
+      will_export: ${{ inputs.will_export }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,6 +60,5 @@ jobs:
     uses: ./.github/workflows/pluto_build.yml
     secrets: inherit
     with:
-      logging_level: ${{ inputs.logging_level }}
       minor_version: true
       run_export: ${{ inputs.will_export }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ on:
         options:
           - template
           - cbbr
+          - pluto
       logging_level:
         description: "Logging level"
         type: choice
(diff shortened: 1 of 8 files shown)


.github/workflows/cbbr_build.yml
```diff
@@ -74,9 +74,9 @@ jobs:
       - name: Export CBBR Output to DO (edm-publishing)
         working-directory: db-cbbr/cbbr_build
         run: ./04_export.sh
-        if: inputs.do_export == true
+        if: inputs.will_export == true
```
fvankrieken, line 77: very minor, but don't need `== true`!

fvankrieken: Also I didn't realize you didn't need these in `${{ inputs.foo }}` format

.github/workflows/build.yml
```diff
@@ -27,21 +24,40 @@ on:
...
+          echo "${{ toJSON(github.event.inputs) }}"
   template:
     if: inputs.dataset_name == 'template'
     uses: ./.github/workflows/template_build.yml
     secrets: inherit
     with:
-      do_export: ${{ inputs.do_export }}
       logging_level: ${{ inputs.logging_level }}
       build_note: ${{ inputs.build_note }}
+      will_export: ${{ inputs.will_export }}
```
fvankrieken, line 49: Are these doing anything? because below it seems that it's still just using the inputs anyways

.github/workflows/cbbr_build.yml
damonmcc: yea I think I've only tried dropping the braces in these if clauses, pretty sure they're needed in more bash-ish places like `run: ...`

.github/workflows/pluto_build.yml
```diff
@@ -48,13 +80,13 @@ jobs:
         shell: bash
         working-directory: ./db-pluto/pluto_build
         run: ./01_dataloading.sh
-        if: inputs.minor_version == false
+        if: needs.version.outputs.is_minor_version == 'false'
```
fvankrieken, line 83: I think longer term we might just want to specify version name at top level, then have this sort of logic within the pluto scripts themselves. But for now think this is fine

.github/workflows/build.yml
damonmcc: this line is only for the job to build the Temple dataset. in the jobs below,  the input is being referenced because the jobs are all independent

fvankrieken: Ohh sorry. Gotcha

fvankrieken:
One postmortem note - saw you were using this, and devdb archiving does too, saw this warning on my job

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

fvankrieken:
You mind if I update and get in my devdb PR? Seems minor and straightforward

damonmcc:
@fvankrieken go for it! [here's](https://github.com/NYCPlanning/db-zap-opendata/blob/main/.github/workflows/single_visible.yml#L44) the only example so far where I've made the change

PR #26, opened 2023-06-22
2023 06 fvk devdb gis enhancements

### LINKED ISSUE #21: DevDB Enhancements for GIS
Continuing [this project](https://github.com/NYCPlanning/edm-overview/issues/905) 

Project plan outlined [here](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BA8DA99B5-0FCC-488B-BF48-90AA087AD8BE%7D&file=Project%20statement%20-%20HousingDB%2023Q2%20enhancements.docx&wdOrigin=TEAMS-ELECTRON.teams.chiclet&action=default&mobileredirect=true&cid=e46b6d08-4864-49fb-8f2c-76e6d06e00f1)

Changes to outputs need to be made in addition to current outputs, so current thought is to include a zipped "publish" folder within the current output folder that has necessary tweaks to schemas, data filtering/changes, etc

Still in-progress list of todos (as in still missing items as scoping out continues)
- Project-level files
  - [x] scope out changes needed
  - [x] generate output versions of housingdb with various filters
  - [x] export as both shp and csv
- Unit change summary files
  - [x] update schemas to match outputs (not 100% but framework for exporting different column names is there)
  - [x] filter "current year" data based on build version (22Q4 should not have 2023 data, 23Q2 should)
  - [x] for aggregates by geographic area, include those with 0 rows
  - [x] output shapefiles in addition to csvs
  - [x] See section 6 in [Housing DB Data Production instructions_20220909_update.docx](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/DevDB/Housing%20DB%20Data%20Production%20instructions_20220909_update.docx?d=w40f84a847de64715a1f047d9899c0114&csf=1&web=1&e=VokmQX) to know how to address NULL values and calcualte fields.  Information is highlighted in yellow.

fvankrieken:
Closes #21. 

Latest build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/5380095359)

I've commented on various files to try to provide context. Basically two main changes
- aggregate files (unit change summary files)
  - changes in aggregate logic to reuse code, include geom for shapefiles 
  - export of aggregates with slightly different schema for gis
  - export of aggregate shapefiles
- project-level files
  - create some views that recreate the GIS output files (housingdb with various filters)
  - export both csvs and shp files for these

Writing this, I realize I didn't finalize the schema for the unit change summary file outputs. But that's slightly minor - code isn't really going to change just column headers


.devcontainer/Dockerfile
```diff
@@ -10,7 +10,7 @@ ARG USER_GID=$USER_UID
 RUN if [ "$USER_GID" != "1000" ] || [ "$USER_UID" != "1000" ]; then groupmod --gid $USER_GID vscode && usermod --uid $USER_UID --gid $USER_GID vscode; fi
 
 RUN apt update && export DEBIAN_FRONTEND=noninteractive \
-    && apt -y install postgresql-client python3-pip python3-distutils jq zip unzip git bash-completion locales
+    && apt -y install postgresql-client libpq-dev python3-pip python3-distutils jq zip unzip git bash-completion locales
```
fvankrieken, line 13: Needed for psycopg2

.devcontainer/docker-compose.yml
```diff
@@ -21,17 +21,16 @@ services:
     # (Adding the "ports" property to this file will not forward from a Codespace.)
 
   db:
-    image: postgis/postgis:12-3.0-alpine
+    image: postgis/postgis:14-3.3-alpine
```
fvankrieken, line 24: We should really get everything using this image - 15 can't be installed on ubuntu via apt yet (for pg_dump etc), so 14 is most convenient

bash_utils/config.sh
```diff
@@ -165,6 +165,24 @@ function csv_export {
...
+        "\COPY (SELECT '\"' || STRING_AGG(attname, '\",\"' ORDER BY attnum) || '\"'\
+        FROM pg_attribute\
+        WHERE attrelid = 'public.${table}'::regclass\
+            AND attname not in (${columns})\
+            AND attnum>0
+        ) TO STDOUT;")
+    run_sql_command \
+        "\COPY (\
+            SELECT ${select_columns} FROM ${table}\
+        ) TO STDOUT DELIMITER ',' CSV HEADER;">${output_file}.csv
```
fvankrieken, line 182: Don't love this entirely because in general it probably makes more sense (say in a dbt world) to simply make tables/views for all exports. But there's a specific case in which it's useful - tables where we export both a shapefile (with geom, shape_area, shape_leng. Or just geom), and a csv where we want to drop spatial fields. So in this case, it was nice to not make duplicate tables just without the geometries. Interested in your thoughts

db-developments/bash/03_aggregate.sh
```diff
@@ -0,0 +1,16 @@
...
+display "Creating spatial aggregate tables"
+#python3 python/aggregate.py sql/aggregate/spatial.sql 2010 block
+#python3 python/aggregate.py sql/aggregate/spatial.sql 2010 tract
+python3 python/aggregate.py sql/aggregate/spatial.sql 2020 block
+python3 python/aggregate.py sql/aggregate/spatial.sql 2020 tract
+python3 python/aggregate.py sql/aggregate/spatial.sql 2010 commntydst
+python3 python/aggregate.py sql/aggregate/spatial.sql 2010 councildst
+#python3 python/aggregate.py sql/aggregate/spatial.sql 2010 nta
+python3 python/aggregate.py sql/aggregate/spatial.sql 2020 nta
+python3 python/aggregate.py sql/aggregate/spatial.sql 2020 cdta
```
fvankrieken, line 16: This was previously a function in db-developments/devdb.sh

db-developments/bash/05_export.sh
```diff
@@ -0,0 +1,101 @@
...
+    wait
+
+    display "CSV Export Complete"
+    echo "[$(date)] ${VERSION}" > version.txt
+)
+
+zip -r output/output.zip output
+
+wait 
+display "Upload Complete"
```
fvankrieken, line 101: Really just renamed this file since I added aggregate as a step and bumped it from 04 to 05 but github thinks otherwise

db-developments/python/aggregate.py
```diff
@@ -0,0 +1,105 @@
...
+            output_column_internal=output_column_internal,
+            additional_columns=additional_columns,
+            group_by=group_by,
+            join_table=join_table,
+            right_join_column=right_join_column)
+        
+        sql_rendered = Template(sql_rendered).render(decade=decade)
+    
+    with engine.begin() as connection:
+        connection.execute(text(sql_rendered))
```
fvankrieken, line 105: renamed from "yearly.py" with a bunch of changes. All the geom sql files were very repetitive so tried to abstract out main logic. This might be a tad sloppy overall (or more hard to follow - don't know if I love just having all of this in dictionaries). Would like your thoughts on this file/approach overall

db-developments/sql/_create.sql
```diff
@@ -43,44 +43,3 @@ CREATE TABLE housing_input_hny (
...
-    bctcb2010 text,
-    bct2010 text,
-    puma text,
-    pumaname text,
-    nta text,
-    ntaname text,
-    commntydst text,
-    councildst text
-);
-\COPY lookup_geo FROM 'data/lookup_geo.csv' DELIMITER ',' CSV HEADER;
```
fvankrieken, line 86: None of these were used elsewhere in the repo

db-developments/sql/_export.sql
```diff
@@ -11,13 +11,13 @@ WHERE (Date_Complete::date <=  :'CAPTURE_DATE'
 		AND Date_Filed::date <=  :'CAPTURE_DATE'));
 
 -- EXPORT HousingDB
-DROP TABLE IF EXISTS EXPORT_housing;
-SELECT * INTO EXPORT_housing
+CREATE VIEW export_housing AS 
+SELECT *
```
fvankrieken, line 15: These are very redundant tables that are pretty large, so thought it would just make sense to make them views instead and save the space. They still query quite fast

db-developments/sql/aggregate/spatial.sql
```diff
@@ -0,0 +1,90 @@
+DROP TABLE IF EXISTS aggregate_{{ geom }}_{{ decade }} CASCADE;
```
fvankrieken, line 1: This file replaces the 6 or so spatial aggregate files (tract.sql etc). Called/formatted by aggregate.py

bash_utils/config.sh
damonmcc: I feel you. I like this approach! especially using tables like `pg_attribute`. for future us, a quick comment above the `local select_columns= ...` line to explain what it's doing would be helpful

db-developments/bash/03_aggregate.sh
damonmcc: wanna delete the commented out lines or keep and note as deprecated?

db-developments/python/aggregate.py
```diff
@@ -0,0 +1,105 @@
...
+    # Get current year
+    current_year = datetime.today().year
+
+    # SQL Template
+    with open(filename, 'r') as f:
+        sql = f.read()
+
+    try:
+        geom=sys.argv[3]
+    except:
```
damonmcc, line 71: a little risky not specifying the "expected" exception. would it be `except IndexError:`?

fvankrieken: Good point. I'll check and implement

db-developments/bash/config.sh
```diff
@@ -2,9 +2,8 @@
 source ../bash_utils/config.sh # expected to be run from db-developments folder
 
 # Setting Environmental Variables
-set_env .env version.env
+set_env ../.env version.env
 DATE=$(date "+%Y-%m-%d")
-set_error_traps
```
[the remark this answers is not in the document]
fvankrieken: No, I had done that with some specific testing (for a real reason I cannot remember) and then left

db-developments/bash/03_aggregate.sh
fvankrieken: They are deleted now, i originally commented before they were

db-developments/python/aggregate.py
```diff
@@ -0,0 +1,105 @@
...
+from sqlalchemy import text
+from utils import engine
+
+geoms = {
+    "block": {
+        "source_column": "bctcb{{decade}}",
+        "output_column": "bctcb{{decade}}",
+        "additional_columns": [
+            ("geoid", "cenblock{{decade[:-2]}}")
+        ],
```
damonmcc, line 15: `additional_columns` seems to always be a list with one single tuple item in it. since it's use in the sql query is:
```sql
{%- for column in additional_columns %}
```
does that end up happy as:
```sql
{%- for column in ("geoid", "cenblock20") %}
```
for example?

db-developments/python/aggregate.py
damonmcc: definitely a nice approach to abstract stuff to a single function. I could image a namedtuple _maybe_ being nicer than a dictionary, but probably not worth the effort in this PR

.devcontainer/docker-compose.yml
damonmcc: good call, `pg_dump` is needed for some builds and may come in very handy for future data warehouse stuff. noted this in the "standardize environment" issue

db-developments/python/aggregate.py
fvankrieken: Ah. No, each tuple refers to one column which both has a source column name and then desired name. So

```
SELECT
{%- for column in [("geoid", "cenblock20")] %}
    {{ column[0] }} AS {{ column[1] }},
{% endfor %}
```
Becomes
```
SELECT
    geoid AS cenblock20,

File: db-developments/python/aggregate.py (hunk 6 above)
Comment (fvankrieken) [reply db-developments/python/aggregate.py:105]: I hope/think things will become a little nicer doing this with dbt instead of just using jinja with python 🤞

File: db-developments/bash/03_aggregate.sh (hunk 4 above)
Comment (damonmcc) [reply db-developments/bash/03_aggregate.sh:16]: I think they're still there? [link to the file on this branch](https://github.com/NYCPlanning/data-engineering/blob/2023-06-fvk-devdb-gis-enhancements/db-developments/bash/03_aggregate.sh)

Comment (fvankrieken) [reply db-developments/bash/03_aggregate.sh:16]: Oh oops - I removed them on my sequel branch. Leaving them for now then. Sorry!

File: db-developments/python/aggregate.py (hunk 12 above)
Comment (damonmcc) [reply db-developments/python/aggregate.py:14]: ooo I see, and every other use of it is like
```sql
{%- for column in additional_columns %} 
        {{ column[1] }},
```

db-developments/bash/03_aggregate.sh
fvankrieken: Nevermind, just did it in this branch

damonmcc (approved):

PR #31, opened 2023-06-29
DevDB: incorporate new Affordable Housing Production data

### LINKED ISSUE #30: DevDB: incorporate new Affordable Housing Production data
### notes
- `hpd_hny_units_by_building` currently has 6K rows
- new historical data has 14K rows and far fewer columns
  - 8K rows `WHERE reporting_construction_type = 'New Construction'`
- ...

damonmcc:
resolves #30


.github/workflows/developments_datasync.yml
```diff
@@ -46,25 +46,28 @@ jobs:
...
+    container: 
+      image: nycplanning/base:latest
+      env:
+        BUILD_ENGINE: postgresql://postgres:postgres@postgis:5432/devdb
+        EDM_DATA: ${{ secrets.EDM_DATA }}
+        AWS_S3_ENDPOINT: ${{ secrets.AWS_S3_ENDPOINT }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     services:
       db:
```
fvankrieken, line 59: This needs to line up with host name in connection string (currently "postgis" on line 53 - which sorry, I changed from our standard "db". Figured it was more informative)

fvankrieken: And by "this" I mean the name of this service, currently "db"

.github/workflows/developments_datasync.yml
```diff
@@ -67,19 +62,31 @@ jobs:
...
+        run: |
+          sudo apt update
+          curl -O https://dl.min.io/client/mc/release/linux-amd64/mc
+          chmod +x mc
+          sudo mv ./mc /usr/bin
+          mc alias set spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
+
       - name: Load to Database (workflow_dispatch)
         if: github.event_name == 'workflow_dispatch'
         working-directory: db-developments
```
alexrichey, line 77: Just fyi, you can set the default working dir or shell at the job level, e.g. [here](https://github.com/NYCPlanning/data-engineering/pull/41/files#diff-b2d695d3b864a94b1de11ae76cbf468bab8f2e856db07aff558306a3dbe58f7aR21)

db-developments/python/geocode_dob.py
```diff
@@ -18,20 +18,45 @@ def geocode(input):
...
-            street_name=sname, house_number=hnum, borough=borough, mode="regular"
+            street_name=sname,
+            house_number=hnum,
+            borough=borough,
+            mode="regular",
         )
         geo = parse_output(geo)
-        geo.update(dict(uid=uid, mode="regular", func="1B", status="success"))
+        geo.update(
+            dict(
```
alexrichey, line 28: might be a nice time to clean these up and just use a dictionary literal, e.g. {"uid"=uid...}

db-developments/python/geocode_hpd_historical.py
```diff
@@ -0,0 +1,122 @@
...
+    engine = create_engine(os.environ["BUILD_ENGINE"])
+
+    # read in housing table
+    select_query = """
+            SELECT * 
+            FROM hpd_historical_units_by_building
+            WHERE reporting_construction_type = 'New Construction'
+            AND project_name <> 'CONFIDENTIAL';
+            """
+    with engine.begin() as conn:
```
alexrichey, line 93: curious if there's a reason to start a transaction, rather than just using the `with engine.connect()` method?

db-developments/python/geocode_hpd_historical.py
```diff
@@ -0,0 +1,122 @@
...
+        df = pd.read_sql(
+            text(select_query),
+            conn,
+        )
+
+    # get the row number
```
alexrichey, line 99: is that what this is doing?

alexrichey (approved):
LGTM! I had a few nits, which you're always welcome to ignore.

.github/workflows/developments_datasync.yml
```diff
@@ -67,19 +62,31 @@ jobs:
       BUILD_ENGINE: postgresql://postgres:postgres@localhost:5432/postgres
     steps:
       - uses: actions/checkout@v3
+
```
fvankrieken, line 65: Not changed in this pr, but up on line 46 we should have postgis be 14-3.3 for consistency with other processes

.github/workflows/developments_datasync.yml
```diff
@@ -67,19 +62,31 @@ jobs:
...
+      - name: install dependencies ...
+        working-directory: db-developments
+        run: |
+          sudo apt update
+          curl -O https://dl.min.io/client/mc/release/linux-amd64/mc
+          chmod +x mc
+          sudo mv ./mc /usr/bin
+          mc alias set spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
```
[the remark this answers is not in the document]
fvankrieken: Oh this isn't using the docker image. It definitely could - and then not invoke docker to run geocode/geosupport commands

.github/workflows/developments_datasync.yml
```diff
@@ -91,8 +98,9 @@ jobs:
       - name: Export to csv
         working-directory: db-developments
         run: |
-          ./devdb.sh output hny_geocode_results csv
-          ./devdb.sh output dob_geocode_results csv
+          # ./devdb.sh output hny_geocode_results csv
```
fvankrieken, line 101: These meant to be commented out?

db-developments/devdb.sh
```diff
@@ -42,12 +42,14 @@ function output {
...
-    get_version $2
+    local version=$(get_version $2)
+    echo "version of ${2} is ${version}"
     docker run --rm\
         -e AWS_S3_ENDPOINT=$AWS_S3_ENDPOINT\
         -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID\
         -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY\
-        -e AWS_S3_BUCKET=$AWS_S3_BUCKET\
+        -e AWS_S3_BUCKET=$recipes_bucket\
+        -e CI=true\
```
fvankrieken, line 52: Nice. I'm curious if we can just use the library archive action in CI rather than relying on these calls but that should be out of scope here I think anyways. So that answers my above note on maybe moving whole job to using published image

fvankrieken: Actually shouldn't this be `CI=$CI`? Could still be run locally

fvankrieken: Empty string for CI is fine for data library to recognize that it's not being run in CI, so that should be fine

db-developments/python/geocode_hpd_historical.py
```diff
@@ -0,0 +1,122 @@
...
+    print("Geocoding HPD Historical...")
+    # Multiprocess
+    with Pool(processes=cpu_count()) as pool:
+        it = pool.map(geocode, records, 10000)
+
+    print("Geocoding finished, dumping to postgres ...")
+    df = pd.DataFrame(it)
+    df.to_sql(
+        "hpd_historical_geocode_results",
+        con=engine,
```
fvankrieken, line 118: Regardless of whether we do in transaction per Alex's comment, would be good to use the connection rather than engine here to avoid that intermittent weird pandas/sqlalchemy error

db-developments/sql/_hny_union.sql
```diff
@@ -0,0 +1,49 @@
...
+DROP TABLE IF EXISTS HPD_units_by_building;
+SELECT 
+    *
+INTO 
+    HPD_units_by_building
+FROM (
+    SELECT * FROM hpd_hny_units_by_building
+    UNION
+    SELECT * FROM _INIT_HPD_historical_units_by_building
+) all_hpd_units;
```
fvankrieken, line 27: Just a bit nitpicky, but this can be a bit more succinct with just something like
```sql
CREATE TABLE HPD_units_by_building AS
SELECT * FROM hpd_hny_units_by_building
UNION
SELECT * FROM _INIT_HPD_historical_units_by_building
```

fvankrieken:
Lot of boilerplate geocode code. I know you're following the pattern, but I might take a look and refactor a bit - seems like there's some that could go into a more shared python file around using it, going from a source data frame to an end-result dataframe with some options around what's attempted, how things are renamed

damonmcc:
@fvankrieken 
> Lot of boilerplate geocode code. I know you're following the pattern, but I might take a look and refactor a bit after this gets merged - seems like there's some that could go into a more shared python file around using it, going from a source data frame to an end-result dataframe with some options around what's attempted, how things are renamed

no doubt, geocoding is one of the biggest things I'm looking forward to abstracting to shared utils. [this implementation in CBBR](https://github.com/NYCPlanning/db-cbbr/blob/main/cbbr_build/library/geocode.py) I think is the closest thing to a "clean slate" I've had with it so far. both here and there I was trying to balance "starting over" with "same approach as before"

db-developments/python/geocode_hpd_historical.py
damonmcc: I prefer `with engine.connect()` as well, I was just mimicking `geocode_hpd_hny.py`

fvankrieken: I think I probably introduced that in `geocode_hpd_hny.py` then... and no, there's no good reason in these cases, we're not doing anything that needs a transaction

db-developments/python/geocode_hpd_historical.py
damonmcc: nope! lol this is also inaccurate in `geocode_hpd_hny.py` so this is copypasta

.github/workflows/developments_datasync.yml
damonmcc: nope! thanks for catching that

damonmcc:
@alexrichey @fvankrieken incorporated most feedback. all great ideas! but some felt out of scope for this PR

fvankrieken (approved):

PR #32, opened 2023-06-29
DevDB output tweaks for GIS/Bytes, some docker/git config

fvankrieken:
Incorporating @AmandaDoyle 's [feedback](https://github.com/NYCPlanning/data-engineering/issues/21#issuecomment-1611613416) in #21 
- project-level output (_export.sql, export.sh): restrict columns for export. This is mostly done in sql at the level of view creation. However, there's one column (Job_Inactv) that is not public but is used to subset the data for exports, so it's kept in the views (as to not list every column in the dataset 5 times), then dropped when csv export is done.
- unit_change_summary files:
  - remove all geometric fields from csv export. I had a typo - in `spatial.sql` did not select columns with double quotes (preserving capitalization), but at export tried to exclude capitalized columns
  - capitalize output columns for CDTA, NTA for bytes output because that's what's expected
  - change logic around max year in columns to be based on version rather than date of running
- alter `shp_export` to allow for additional options to be fed to `ogr2ogr` call (to set output srs and export in 2263 for all bytes outputs)
- 
As well as docker/gha tweaks, see comment below

[Last build succeeded](https://github.com/NYCPlanning/data-engineering/actions/runs/5468754244)


fvankrieken:
So I've come up against a slight issue running our code inside a docker image. Basically, the checkout action runs with user id 1001 while any subsequent commands in the container are run as root. This causes a mismatch in some cases of ownership in folders, which git is unhappy about when we try to use it from the command line in the container.

We use git as part of our build process because we export to folders based on branch name in digital ocean s3. See [here](https://github.com/NYCPlanning/data-engineering/blob/main/bash/utils.sh#L258).

This has led to this error when [trying to call ](https://github.com/NYCPlanning/data-engineering/actions/runs/5467647209/jobs/9954227758#step:9:320)that function
```
fatal: detected dubious ownership in repository at '/__w/data-engineering/data-engineering'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/data-engineering/data-engineering
```
You can see the owner of folders at this point [here](https://github.com/NYCPlanning/data-engineering/actions/runs/5467647209/jobs/9954227758#step:9:301). Note the different owner of the root workspace directory.

Two fixes to this: first is specify user as 1001 in [github action](https://github.com/NYCPlanning/data-engineering/actions/runs/5467921957/workflow#L22)
```
jobs:
  build:Take
    runs-on: ubuntu-22.04
    container: 
      image: nycplanning/base:latest
      options: --user 1001
      env: ...
```

Other is to declare the working directory as a safe directory globally, and then run this in a shell script. We already have to run an "init" sort of step in the github actions, so 
```
- name: Add s3 endpoint ...
   run: mc alias set spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
```
becomes
```
- name: Finish container setup ...
```
And we don't specify userid. So slightly cleaner yml IMO. Then the script looks like
```bash
#!/bin/bash

mc config host add spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4

if [[ -z "$CI" ]]; then
    workspace=/workspace
else
    workspace=/__w/data-engineering/data-engineering
fi 
echo $workspace
git config --global --add safe.directory $workspace
```
For now

alexrichey:
Fun stuff! `git config --global --add safe.directory` seems 👍

fvankrieken:
Re-running job after rebase fun to make sure I didn't screw anything up

https://github.com/NYCPlanning/data-engineering/actions/runs/5468754244

damonmcc (approved):

Issue #33, opened 2023-07-03
some data products aren't built using centralized action and DB


damonmcc:
## goals
- for all data products in the mono repo, we'd like to:
  - ensure they can currently be built here
  - standardize their build processes to use our persistent DB

## notes
- we now have [a single action to build all data products](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/build.yml). to be called by it, a data product's build action must be changed so that the inputs align
- currently, it isn't configured to build  `factfinder`, `zap`
- this repo's wiki is used to track which products are being built here ([Data Catalog wiki page](https://github.com/NYCPlanning/data-engineering/wiki/Data-Catalog)), but it's likely out-of-date and not of long-term value to track in this way (e.g. if this issue is resoled, is the `Builds in monorepo` column still useful?)

## tracking product DB use
- [ ] cbbr
- [ ] checkbook
- [ ] colp
- [x] cpdb
- [x] developments
- [ ] facilities
- ~[ ] factfinder (doesn't use a postgis DB?)~
- [x] knownprojects
- [x] pluto
- [x] template
- [x] zap-opendata
- [ ] zoningtaxlots

fvankrieken:
Another requirement - update the run name based on the input dataset. It's nice to be able to easily filter from the workflow runs page by dataset

<img width="409" alt="image" src="https://github.com/NYCPlanning/data-engineering/assets/9454672/8120c6f6-4f49-421c-882a-7d6e4fbe4eb6">

fvankrieken:
See https://github.com/NYCPlanning/db-gru-qaqc/actions/runs/5146890103/workflow for an example

damonmcc:
we may want to remove the `workflow_dispatch` trigger from all product build actions to force the use of the centralized action

PR #36, opened 2023-07-05
Add Dataset Name to build action

alexrichey:
per
https://github.com/NYCPlanning/data-engineering/issues/33#issuecomment-1618857375, adds the dataset to the run-name


fvankrieken:
Tough review!

Off the top of my head I wouldn't say we need anything else - you can already filter by branch when looking at a given workflow and that's really the only other one I would want

fvankrieken (approved):

damonmcc (approved):

PR #38, opened 2023-07-06
DevDb: add internal project-level exports, clean up other new exports

fvankrieken:
Continuing #21 

For project-level files (`HousingDBpost2010...` files), we needed both an internal export for distribution on the m drive as well as the external export to go on bytes of the big apple. I created a view for each export, with columns for each of the different exports in a separate file (internal vs external - the different files within each one have the same schema with different filtering of rows). 

Nested views (views on top of views) are maybe a potential criticism, but the queries are very simple and performance is not an issue

Latest action [here](https://github.com/NYCPlanning/data-engineering/actions/runs/5532403003). Exports look good


db-developments/bash/05_export.sh
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash
 source bash/config.sh
+source sql/constants/export_schemas.sh
```
fvankrieken, line 3: Should this be stored somewhere else or named differently? This seemed like easiest way to not repeat a ton of code for the external tables by just storing the list of columns as a bash variable. See `_export.sql` to see why this was useful

damonmcc: definitely nice to define things in one place!

I can't think of a better/simpler way to define and pass a column list to a bash-invoked sql file via `-v internal_columns="$internal_columns"`

but other potential file formats to store variables that come to mind are `.toml` and `.yaml`, but I don't know how straightforward it'd be to set variables in a bash script using those file formats

damonmcc: future us: I think this is easier to do in either dbt + jinja or python land

fvankrieken: Agreed! Would like to get it out of bash eventually

damonmcc (approved):

alexrichey (approved):
👍

PR #39, opened 2023-07-06
Move CPDB to shared utils

fvankrieken:
Continuing #5 

Latest build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/5547386109)

Failing spending action is put in an `archive` folder in github workflows. Happy to put somewhere else, but seems worth making not runnable while it's really not doing what it's supposed to.

One little outstanding question, but just a pinning of a version for a data source, see below. Other than that, I think it's good to go


fvankrieken:
Failing spending action is horribly broken. It relies on the bigquery-archived checkbook nyc data, which hasn't been properly loading in at least a year and a half. Getting an odd that doesn't entirely make sense, but tough to troubleshoot because the source data in bq is spread across thousands of tables - one per day - which get unioned (bq does this pretty easily). But seems maybe there's some schema issue across these tables, even though they all should have the same schema.

This data does make it into the cpdb export, but the csv has blank columns for 2022 and 2023 due to the data being way out of date. Feel like maybe we shelve fixing this action for now while we figure out what we want to do around that spending data.

fvankrieken:
And the checkbooknyc data is archived to bigquery by logic [here](https://github.com/NYCPlanning/db-airflow/blob/main/dags/checkbooknyc_capital_spending.py)

db-cpdb/bash/01_dataloading.sh
```diff
@@ -2,39 +2,40 @@
...
+import_recipe dcp_councildistricts &
+import_recipe nypd_policeprecincts &
+import_recipe fdny_firecompanies &
 
 # Building and lot-level info
-import dcp_mappluto_wi &
-import dcp_facilities 20210811 &
-import doitt_buildingfootprints &
+import_recipe dcp_mappluto_wi &
+import_recipe dcp_facilities 20210811 & # why pinned to this version?
```
fvankrieken, line 24: @AmandaDoyle @damonmcc any thoughts on why facdb is pinned to this version in cpdb build?

damonmcc: hmm I'm not sure, worth checking with Amanda

my only idea is that it may be joined onto some other source data that isn't updated very frequently, so an older FacDB version is more useful than the latest?

.github/workflows/cpdb_build.yml
```diff
@@ -1,19 +1,37 @@
...
+      run_export:
+        type: boolean
+        required: true
+  workflow_dispatch:
+    inputs:
+      run_export:
+        description: "Run Export Step and Upload to DO"
+        type: boolean
+        required: true
+        default: false
```
damonmcc, line 15: is it worth adding CPDB to `build.yml` and dropping the `workflow_dispatch` trigger here?

db-cpdb/README.md
```diff
@@ -5,7 +5,7 @@
 ## Instructions
 
 1. All the relevant commands for running CPDB is wrapped in the cli bash script `./cpdb.sh`, please read the file for more details.
-2. The capital spending scraping process should be done right after we load `fisa_capitalcommitments` to data library. A separate bash script will import  `fisa_capitalcommitments` to bigquery and we will create export the capital spending table `cpdb_capital_spending` via a bigquery command (see `bash/11_spending.sh` for more details)
+2. The capital spending scraping process should be done right after we load `fisa_capitalcommitments` to data library. A separate bash script will import  `fisa_capitalcommitments` to bigquery and we will create export the capital spending table `cpdb_capital_spending` via a bigquery command (see `bash/06_spending.sh` for more details)
```
damonmcc, line 8: was the conclusion from standup yesterday that bigquery would no longer be a part of the archive/build process? maybe this readme section could be deleted?

looks like the `bq` commands are untouched in `06_spending.sh` and are never called since the workflow is getting moved to an archive folder

db-cpdb/bash/01_dataloading.sh
AmandaDoyle: Looks like we just pinned FacDB to the latest published version.  We do not need to pin to a specific version of FacDB 
https://github.com/NYCPlanning/db-cpdb/commit/3ff76c0f600365cc20a1493a4decf24e5b436bfa

fvankrieken: Gotcha. I'm just goin to unpin then, default to latest if that sounds good

alexrichey (dismissed):
Makes sense to me!

fvankrieken:
Had to resolve a conflict in build.yml, but nothing else has changed

alexrichey (approved):
👍

damonmcc (approved):

Issue #43, opened 2023-07-12
ZTL workflow maintenance


fvankrieken:
One bug and some little needed maintenance tasks for zoning tax lots github action workflow files.
- [x] [Dataloading action is failing](https://github.com/NYCPlanning/data-engineering/actions/workflows/zoningtaxlots_dataloading.yml). This is a step separate from the build, one that uses the [published library archive action](https://github.com/NYCPlanning/action-library-archive) to archive various needed inputs for ztl build. GIS generally runs this action since they are responsible for updating the source datasets. It's failing because the filepath for template files was not updated when I initially went over ztl in the monorepo to get it building.
- [x] build action is not using build docker image. Workflow file should be updated to use nycplanning/build-base image (see [pluto](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/pluto_build.yml#L49) for example, also for the env setup step)
- [x] build and dataloading actions both should be updated to ubuntu 22.04, checkout action v3
- [x] build action should be updated to postgis 14-3.3
- [x] `close issue` job of dataloading can be deleted, as well as other logic around triggering workflow via issue
- add ztl as an option in the generic `build` workflow file. this requires both
  - [x] changes to the "generic" build action as well as
  - [x] addition of `workflow call` as option for ztl build workflow file

And as part of the above
- [ ] qc outputs of a build after changes have been made

fvankrieken:
Facdb might be the better example for a workflow file, at least in terms of how it sets up the job parameters (prior to the running of steps).

https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/facilities_build.yml#L21C14-L24C43

PR #46, opened 2023-07-14
Use run defaults for shell, working-directory for pluto/cbbr/colp builds, use docker images for pluto inputs

fvankrieken:
Little bit of maintenance on workflow files

For cbbr/colp/pluto
- use run defaults
- make sure versions for actions, postgis, ubuntu, etc consistent
- cbbr use docker image for build

Just pluto
- use docker image for input processing, including removing docker calls to geosupport. Need to test all to make sure they work properly

All 3 builds running fine in actions

Pluto inputs a little harder to judge, since two of them "fail" regularly due to missing source data retrieved via ssh. This is a known issue. I wrote a ticket, #47 

[Numbldgs](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_numbldgs.yml) however is working now


fvankrieken:
Actually not ready for review - was going to just do pluto but I'll quickly normalize some of the other data products that I'd already moved over to shared utils

fvankrieken:
Leaving ZTL out of this as leaving #43 for the fellows

.github/workflows/cbbr_build.yml
```diff
@@ -15,10 +15,22 @@ on:
...
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: ./db-cbbr/cbbr_build
+    container:
+      image: nycplanning/build-base:latest
+    env:
```
fvankrieken, line 25: So having env at the container level (which I did originally when moving to using containers as that was what most examples looked like) makes variables inaccessible to `use`d actions (see data library action in pluto input ymls below), but declaring them at the job level as we have in the past works fine for both parts of the job in the container and these other calls. So think we should go back to norm of env at job level

fvankrieken:
cbbr has an unrelated issue, going to work on that in a separate branch/pr. This imo is good to go, with caveat that the pluto input jobs that don't have source data are failing and hard to test. But this PR doesn't change that and that should be addressed separately

fvankrieken:
Ah I messed up the rebase. Alright lemme fix that

fvankrieken:
> Ah I messed up the rebase. Alright lemme fix that

Fixed

fvankrieken:
Just merged main in, going to rebuild these three datasets one more time

fvankrieken:
I have certainly violated my own principles for this slightly circuitous PR. But good to go!

.github/workflows/cbbr_build.yml
```diff
@@ -15,10 +15,22 @@ on:
...
+      BUILD_ENGINE: postgresql://postgres:postgres@postgis:5432/postgres
+      EDM_DATA: ${{ secrets.EDM_DATA }}
+      AWS_S3_ENDPOINT: ${{ secrets.DO_S3_ENDPOINT }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.DO_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.DO_SECRET_ACCESS_KEY }}
     services:
-      postgres:
-        image: postgis/postgis:12-3.0-alpine
+      postgis:
+        image: postgis/postgis:14-3.3-alpine
```
damonmcc, line 33: no downstream changes needed when this service's name changes?

damonmcc (approved):
lgtm!

.github/workflows/cbbr_build.yml
fvankrieken: The build_engine reflects this change

damonmcc: oops sorry didn't see the text change within the line change

PR #48, opened 2023-07-14
added to-do list on readme, top-level folder for db-checkbook (eventu…

athursland:
Created a db-checkbook folder in the monorepo where Dea and I will begin merging units of our work from the last two sprints on the Historical Liquidations project, added a brief to-do list

### DIFF
--- a/db-checkbook/README.md
+++ b/db-checkbook/README.md
@@ -1,0 +1,11 @@
+# Checkbook NYC Exploratory Work and Join to CPDB
+
+## To do:
+- upload relevant Checkbook NYC data to Digital Ocean
+- write the code to pull down data from Digital Ocean
+- Dea and Ali make their own branches and start working on smallest units of extractable code from our original notebooks, checking it into our projects, implementing, then submitting PR requests
+- mimic utility of methods from db-facilities/facdb/cli.py (init, dataloading, build, etc) using facdb terminology
+- make a data product-level requirements doc if needed, and update top-level requirements doc (requirements.in) in monorepo with python modules that are generally applicable to the team's work
+
+## Eventually to do: 
+- (eventually) add recipe for updating Checkbook NYC data from website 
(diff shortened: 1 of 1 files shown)


db-checkbook/.gitignore
alexrichey, line 1: I don't know that we'll need this. Hopefully .gitignore at the project root should take care of us. So maybe ✂️

db-checkbook/README.md
```diff
@@ -0,0 +1,11 @@
+# Checkbook NYC Exploratory Work and Join to CPDB
+
+## To do:
+- upload relevant Checkbook NYC data to Digital Ocean
+- write the code to pull down data from Digital Ocean
+- Dea and Ali make their own branches and start working on smallest units of extractable code from our original notebooks, checking it into our projects, implementing, then submitting PR requests
+- mimic utility of methods from db-facilities/facdb/cli.py (init, dataloading, build, etc) using facdb terminology
```
alexrichey, line 7: Was chatting with @fvankrieken, and our thought is that it might be easiest to (for now) just mimic the terminology of facdb, but putting the actual data manipulation steps in separate, sequential files, so that you two can avoid any merge conflicts. So you'd end up with a few python scripts like 
- 01_dataloading.py
- 02_build.py
etc. Though it's up to you!

db-checkbook/README.md
```diff
@@ -0,0 +1,11 @@
+# Checkbook NYC Exploratory Work and Join to CPDB
+
+## To do:
+- upload relevant Checkbook NYC data to Digital Ocean
+- write the code to pull down data from Digital Ocean
+- Dea and Ali make their own branches and start working on smallest units of extractable code from our original notebooks, checking it into our projects, implementing, then submitting PR
+- mimic utility of methods from db-facilities/facdb/cli.py (init, dataloading, build, etc) using facdb terminology
+- make a data product-level requirements doc if needed, and update top-level requirements doc (requirements.in) in monorepo with python modules that are generally applicable to the team's work
+
```
alexrichey, line 9: I suppose to Dea's question on chat, there's an additional step for uploading the data to s3 at the end.

fvankrieken: Which we would definitely want to include!

db-checkbook/README.md
athursland: that sounds good to me - where in the current folder structure would it be most appropriate to have those scripts? python folder, or?

alexrichey: python folder sounds good to me. Sticks with convention

alexrichey (approved):
LGTM!

PR #49, opened 2023-07-14
Ar fac refac

alexrichey:
There's one thing I'm not super happy with here, and it has to do with including the `dcpy` shared code. Invoking python as a module from project root feels a weird, but it's the easiest way to include `dcpy` shared library. Also, the dash in something like `db-facilities` makes it unusable as a python module, meaning you can't do absolute path'ing within the data product unless you run it from within the data product folder.

The other option is to modify the $PYTHONPATH to add `dcpy` in the container. It feels cleaner, but this works for now. In any case, I'm open to ideas. (Later Edit: @damonmcc noted that we could just change the paths for these products 🤔)

One additional note: can remove the pip install step from the Github action build after the Docker image has been published. Long term there's probably a nicer way to handle the publishing that would obviate this (e.g, building an image per branch)


dcpy/connectors/psql/__init__.py
damonmcc, line 1: since so far there's one file per module, would these be simpler as `psql.py`, `s3.py`, and `recipes.py` rather than `__init__.py` files in named folders? I think their uses ([example](https://github.com/NYCPlanning/data-engineering/blob/3fd0ce93cc67cbf4d3d98729a2d903c487041710/db-facilities/facdb/utility/export.py#L3)) would then change slightly to be something like

```python
from dcpy.connectors.psql import insert_copy

...

df.to_sql(
    name,
    con=ENGINE,
    if_exists="replace",
    index=False,
    dtype={
        "geo_1b": dialects.postgresql.JSON,
        "geo_bl": dialects.postgresql.JSON,
        "geo_bn": dialects.postgresql.JSON,
    },
    method=insert_copy,
)
```

damonmcc (dismissed):
LGTM! and looks like [this build](https://github.com/NYCPlanning/data-engineering/actions/runs/5579998387) succeeded?

just noted a potential approach to files in `dcpy`, but maybe it's better to hold off and think about that when the modules are actually being used in different builds

dcpy/connectors/psql/__init__.py
alexrichey: @damonmcc That's a good question... I actually think the import/implementation would change at all outside of dcpy. Just fewer directories in the end. Seems cleaner for now, so I'll change that

.github/workflows/facilities_build.yml
```diff
@@ -47,34 +46,30 @@ jobs:
...
       - name: Run Container Setup
-        run: ../bash/docker_container_setup.sh
+        run: ./bash/docker_container_setup.sh
 
-      - name: Build Python Poetry deps and CLI
-        run: |
-          pip3 install poetry==1.5.1 &&\
-          python3 -m poetry config virtualenvs.create false --local && python3 -m poetry install --no-dev
+      - name: Install python deps
+        run: pip install -r ./python/requirements.txt
```
damonmcc, line 52: per your note in the PR description, this is to use changes made to `requirements.txt` before the published docker image `nycplanning/build-geosupport:latest` incorporates those changes

yea I remember having this sort of chicken-egg problem before. definitely like your idea of building/using an image per branch. I think an approach at my last place was to make package changes in standalone PR's, but that had some issues

dcpy/connectors/psql/__init__.py
alexrichey: Done!

damonmcc (approved):

PR #55, opened 2023-07-18
Use sys.path to make `dcpy` accessible to python

alexrichey:
Using sys.path.append is a little dirty, but it's the easiest way to make our shared code available in the products themselves. Other products will need to implement something similar in their top-level __init__.py

Also refactors paths in db-facilities to make reflect that we're executing python from inside the product folder. Also implements absolute paths for constant imports.

### DIFF
--- a/db-facilities/facdb/__init__.py
+++ b/db-facilities/facdb/__init__.py
@@ -1,17 +1,22 @@
 import os
+import sys
 from pathlib import Path
 from dotenv import dotenv_values
 
 
 _module_top_path = Path(__file__).resolve().parent
-_config = dotenv_values(_module_top_path.parent / "version.env")
+_product_path = _module_top_path.parent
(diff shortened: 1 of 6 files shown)


db-facilities/facdb/__init__.py
```diff
@@ -1,17 +1,22 @@
...
 
 
 _module_top_path = Path(__file__).resolve().parent
-_config = dotenv_values(_module_top_path.parent / "version.env")
+_product_path = _module_top_path.parent
+_proj_root = _product_path.parent
 
-VERSION_PREV = _config["VERSION_PREV"]
+# Make `dcpy` available
+sys.path.append(str(_proj_root))
```
alexrichey, line 12: This is the important line, FYI

damonmcc (approved):
LGTM!

PR #58, opened 2023-07-18
Db-checkbook adding in data loading functions

DeaBardhoshi:
Adding two functions to read in the data from Digital Ocean.


.gitignore
```diff
@@ -130,3 +130,6 @@ dmypy.json
 
 .vscode
 *.DS_Store
+
+
+db-checkbook/historical_cpdb_data/
```
athursland, line 135: I think if you upload the input data to the .library file it will be ignored - so maybe we should move the historical cpdb data there? (Although this totally works too)

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,39 @@
...
+
+# reading in the checkboob daa
+def read_from_csv(name: str) -> pd.DataFrame:
+    df = pd.read_csv(
+        "https://edm-recipes.nyc3.cdn.digitaloceanspaces.com/datasets/test_nypl_libraries/20210122/test_nypl_libraries.csv", dtype=str, index_col=False
+    )
+    return df
+
+def read_from_gpd(path):
+    df = gpd.read_file('https://edm-recipes.nyc3.cdn.digitaloceanspaces.com/datasets/dcp_cpdb/2018_adopted/dcp_cpdb.sql', geometry = 'geometry')
```
athursland, line 31: I like naming geopandas df's `gdf` to differentiate but that's also just a personal preference

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,39 @@
...
+  #  s3_client = client()
+   # available_buckets = [
+    #    bucket["Name"] for bucket in s3_client.list_buckets()["Buckets"]
+    #]
+   # print(f"This S3 client is a {type(s3_client)}")
+   # print(f"This S3 client has access to buckets: {available_buckets}")
+
+
+# reading in the checkboob daa
+def read_from_csv(name: str) -> pd.DataFrame:
```
athursland, line 24: maybe we should specify the url on digital ocean as a param as well so this function can be generalized to different csvs (noting that the url is currently a placeholder)

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,39 @@
-from dcpy.connectors.s3 import client
+#from dcpy.connectors.s3 import client
```
athursland, line 1: going off of another comment I think that eventually we'll incorporate the client function from dcpy to grab the data from digital ocean

db-checkbook/build_scripts/01_dataloading.py
fvankrieken: A couple notes on aligning this with our bash functionality in terms of how its used
- store the "base" url as a literal (i.e. at top of file for now, have `recipes_url="https://...` or something like that
- use name to format string in read_csv
- have `version` as an optional arg with default value `latest`. In this case at the moment, the version is `20210122`

db-checkbook/dcp_cpdb.yml
fvankrieken, line 1: Looks like this is an empty file that can be removed

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,40 @@
...
+
+
+
+def read_s3(digital_ocean_filepath, save_file_path):
+
+    #digital_ocean_filepath example "datasets/bpl_libraries/20210208/bpl_libraries.csv"
+
+    s3_client = s3.client()
+    file = s3_client.download_file('edm-recipes', digital_ocean_filepath, save_file_path)
+    return file
```
fvankrieken, line 22: Looks like this function is more what you're going to use than the read_from_csv? At least in terms of moving a file from s3 to local storage.

Since our recipe file paths are pretty consistent, maybe take dataset name, extension, and version all as arguments similar to my suggestion for `read_from_csv`? And again, version could be optional and default to "latest". And at that point, you could construct the DO_filepath from these arguments.

This could also use a more descriptive name - it's downloading a file specifically from edm-recipes, so name should reflect that.

Lastly, not sure what boto3 `download_file` actually returns ([docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/download_file.html) are even unclear) - since this is more a function that has a side-effect than output, I might not return anything (@alexrichey / @damonmcc thoughts?)

db-checkbook/build_scripts/02_build.py
```diff
@@ -1,5 +1,8 @@
 import pandas as pd
 
+def run_categorization(df):
```
athursland, line 3: best practice might be to open a separate PR for changes made to the build script and keep the scope of this PR as small as possible (limiting to dataloading)

fvankrieken:
Hi @DeaBardhoshi - is this PR ready for review/merging, or are you still working on it?

DeaBardhoshi:
Hi @fvankrieken,  the current functions pulling in data are ready to be reviewed.

PR #60, opened 2023-07-19
Ar add recipe connector

alexrichey:


alexrichey:
@fvankrieken @damonmcc  I'd started down this route of porting some recipe stuff to python for facdb, but I'm having second thoughts about having duplicated logic in bash and py. Thoughts? Does it make sense to have the python equivalent for grabbing recipes? Guessing that the logic won't change very often, so many divergence isn't really an issue.

fvankrieken:
I think so - personally longer term I wouldn't mind moving fully to python for this stuff

damonmcc:
totally support doing it in python too. and yea divergence doesn't seem like an issue

the bash logic doesn't change often and I suppose when we de want a build to use different data laoding logic, would be nice to do that in python and maybe treat the bash one as the "this is legacy, don't change me" approach?

alexrichey:
@damonmcc and @fvankrieken I was holding off getting this in, so I could also implement it in facdb. But I think this is ready to go, and I'll implement in facdb after we've released. (that way I refactor in facdb and be confident I haven't broken anything)

fvankrieken (approved):

damonmcc (approved):

dcpy/connectors/data_library.py
```diff
@@ -0,0 +1,59 @@
...
+                CREATE TABLE source_data_versions (
+                schema_name character varying,
+                v character varying);"""
+            )
+        )
+
+
+def import_recipe(
+    recipe_name: str,
+    *,
```
alexrichey, line 27: @damonmcc (and @athursland and @DeaBardhoshi ) mentioned this to Finn, but a convention I like in functions with many arguments is adding the `*`, which forces everything after to be a keyword argument. So I can't invoke this fn like
```
import_recipe('my_recipe_name', 'latest', 'my_path', False)
```
Instead when I want to override a default, I have to specify it 
```
import_recipe('my_recipe_name',  set_version=False)
```
Does wonders for readability in the implementation.

PR #62, opened 2023-07-19
Add db-factfinder

fvankrieken:
Just dumping in without changes before working on the latest update.

That's a lot of lines... gonna poke around and see if we definitely want all of this in here


damonmcc (approved):

fvankrieken:
It's the big metadata files mapping input columns to outputs. Not actually large in terms of storage

PR #68, opened 2023-07-21
db-checkbook dataloading functions from DO for Checkbook NYC and CPDB

DeaBardhoshi:
Added two new functions to dataloading for getting data from Digital Ocean (one for CPDB files + one for Checkbook), adding @athursland


db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,33 @@
...
+# example: datasets/dcp_cpdb/2018_adopted_polygons/
+# dataset name: datasets/dcp_cpdb
+# version: 2018_adopted, 2019_adopted, 2020_adopted, 2021_adopted, 2022_adopted, 2023_executive
+# type geom: _polygons, _points
+
+def read_s3_edm_recipes_cpdb(dataset_name, version, type_geom, save_file_path):
+    digital_ocean_filepath = f'{dataset_name}/{version}{type_geom}/dcp_cpdb.shp'
+    s3_client = s3.client()
+    file = s3_client.download_file(base_bucket, digital_ocean_filepath, save_file_path)
+    return file
```
fvankrieken, line 23: This is maybe more specific than I'd hope, but didn't realize we had the complication of geometry separated by data, so this seems like roughly the best path for now.

Longer term would like just one file per cpdb version, but that would require merging of shapefiles. For now, use this, and I'll write up a data library ticket to get these in as single shapefiles and we can put that on hold for now

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,16 +1,33 @@
...
+
+
+def read_edm_recipes_nyc_checkbook(name):
+    file_name = f'{base_url}/{name}'
```
fvankrieken, line 27: couple questions here - is this going to work as planned? From my understanding, you'd need to feed in `"datasets/nycoc_checkbook/latest/nycoc_checkbook.csv"` as the name, which seems like maybe we should only need to feed in `"nycoc_checkbook"`.

Similarly, for function name - this doesn't seem specific to nycoc checkbook, but rather is reading in any csv of any recipe, so it should be named as such.

And version should be specified as an argument, just like for the cpdb function.

Let me know if I'm making any incorrect assumptions about the data being pulled in

DeaBardhoshi: For version as an argument, should it be implemented when we have the Checkbook API pulling also set up? At the moment we are working with a single dataset, but we could also just do a default of version = "" for now.

fvankrieken: Hmm that's a good point, I was assuming the checkbook data would just be one file but it won't necessarily be that way in the future...

Let's go ahead and make that assumption for now, and for this go with what I'm suggesting. I think that's how the checkbook data in recipes will be most useful in the future anyways, even if it requires some aggregating in the data library step (and thinking about not being super wasteful with storage since this is one of the bigger source data sets in terms of rows)

fvankrieken: So "latest" still being default version

damonmcc (approved):
pair coded on this today along with Alex and Ali

PR #75, opened 2023-07-24
capital spending data pipeline - category assignment 

athursland:
@DeaBardhoshi 
Creating functions for category assignment based on columns of interest and final category assignment using high-sensitivity fixed asset method 

## todo:


db-checkbook/build_scripts/02_build.py
```diff
@@ -84,18 +91,62 @@ def _join_checkbook_geoms(df: pd.DataFrame, cpdb_geoms: gpd.GeoDataFrame) -> gpd
     :param cpdb_geoms: final versions of archived CPDB geometries from every year, and the most recent geometry for the current year
     :return: CPDB geometries left-joined onto Checkbook NYC data 
     """
-    merged = df.merge(cpdb_geoms, how='left', left_on='FMS ID', right_on='maprojid', indicator=True)
-    merged = df.drop('Unnamed: 0', axis=1, inplace=True) # TODO: figure out how to avoid this column appearing in the first place!
+    merged = df.merge(cpdb_geoms, how='left', left_on='fms_id', right_on='maprojid', indicator=True)
+    #merged = df.drop('Unnamed: 0', axis=1, inplace=True) # TODO: figure out how to avoid this column appearing in the first place!
```
fvankrieken, line 95: nitpick - is this issue solved? If so remove comment

db-checkbook/build_scripts/02_build.py
```diff
@@ -84,18 +91,62 @@ def _join_checkbook_geoms(df: pd.DataFrame, cpdb_geoms: gpd.GeoDataFrame) -> gpd
...
     gdf = gpd.GeoDataFrame(merged, geometry='geometry')
     return gdf
 
-# ----  TODO: category assignment on BC, CP, and high-sensitivity Fixed Asset approach ----
+def _df_to_sql(df: pd.DataFrame) -> None:
+    table_name = 'capital_projects'
+    df.to_sql(table_name, ENGINE, if_exists='replace', index=False)
```
fvankrieken, line 101: Hmm interesting that `df.to_sql` doesn't fail here in the same way that `pd.read_sql_table` was below, given that it's just being fed the engine.

But still a minor suggestion here - this is a pretty specific function, in that it's loading a dataframe to the `capital_projects` table, and it's a one-liner to boot (if we just feed `'capital_projects'` to `df.to_sql` directly). So I would scrap it, and call df.to_sql on line 117 before the for loop, using `conn` as the second arg instead of `ENGINE`

fvankrieken (approved):

PR #77, opened 2023-07-25
Db checkbook Data Loading Fixes

DeaBardhoshi:
Fixing the data pipeline: 01_dataloading (fixed the reading from DO functions), tagging @athursland.

Notes on the S3 Loading Function: 
- put a checker in place to see if DO key for folder doesn't end with "/"
- part of the files (2017 points and polygons and 2023 points) were not loading properly prior to this fix


db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,29 +1,62 @@
...
+    prefix = 'datasets/dcp_cpdb/'
+
+    bucket = s3_resource.Bucket(BASE_BUCKET)     
+    for obj in bucket.objects.filter(Prefix = prefix):
+        key = obj.key.replace(prefix, '')
+        if not os.path.exists(LIB_DIR / os.path.dirname(key)):
+            os.makedirs(LIB_DIR / os.path.dirname(key))
+        bucket.download_file(obj.key, LIB_DIR / key)
+
+    return ''
```
fvankrieken, line 39: Very minor, but I would just `return` rather than return an empty string

db-checkbook/build_scripts/01_dataloading.py
```diff
@@ -1,29 +1,62 @@
...
 BASE_URL = "https://edm-recipes.nyc3.cdn.digitaloceanspaces.com"
 
+_curr_file_path = Path(__file__).resolve()
+LIB_DIR = _curr_file_path.parent.parent / '.library'
+load_dotenv(_curr_file_path.parent.parent.parent / '.env') 
 
 
-def read_s3_edm_recipes_cpdb(version, type_geom, save_file_path):
+
+def download_s3_edm_recipes_cpdb():
```
fvankrieken, line 17: @damonmcc @alexrichey - having some difficulties with the cpdb zipped shapefiles (aren't happy being opened by gdal as part of data library or by geopandas here, so ) so we've gone a slightly unconventional route. @DeaBardhoshi had manually put the shapefiles in DO/edm-recipes as directories. This folder then recursively downloads the `dcp_cpdb` folder in DO/edm-recipes into the local filesystem, and geopandas then calls `read_file` on the folders themselves (containing all the relevant files for a shapefile) rather than a single file

__MACOSX/._cpdb_dcpattributes_pts_23
fvankrieken, line 1: This directory and files in it should be deleted

fvankrieken:
Does this whole pipeline work? I haven't pulled and tested myself yet, but don't see changes to `build` here, is the build script set up to read in the loaded cpdb folders?

athursland:
> Does this whole pipeline work? I haven't pulled and tested myself yet, but don't see changes to `build` here, is the build script set up to read in the loaded cpdb folders?

_merge_cpdb_geoms() in 02_build.py uses the subdirectories created in the dataloading step to pull in the cpdb data!

db-checkbook/build_scripts/01_dataloading.py
alexrichey: maybe an enhancement for later, but a generalized version of this might be nice for `dcpy`

db-checkbook/build_scripts/01_dataloading.py
fvankrieken: Yeah that's in 02_build

damonmcc (approved):

PR #81, opened 2023-07-26
edit and run python package compile script

damonmcc:
no changes were made to `/python/requirements.in`, so all package version changes seem to be due to new patch versions or due to the compile script being run in python 3.9 rather than 3.10 causing changes to package dependencies

since I deleted `/db-template/requirements.txt`, I had to change relevant github actions to use the top-level packages. I did this by having them use `nycplanning/build-geosupport:latest`

### DIFF
--- a/db-template/python/geospatial.py
+++ b/db-template/python/geospatial.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 import contextily as cx
 import geopandas as gpd
 import pandas as pd
@@ -71,7 +73,7 @@ def pad_map_bounds(
 def map_simple(
     data: gpd.GeoDataFrame,
     projection: str = NYC_PROJECTION,
-    basemap: str | TileProvider = DEFAULT_BASEMAP,
+    basemap: Union[str, TileProvider] = DEFAULT_BASEMAP,
     map_config: dict = DEFAULT_MATPLOTLIB_MAP_CONFIG,
 ) -> Axes:
     axes = data.plot(
--- a/.github/workflows/template_build.yml
+++ b/.github/workflows/template_build.yml
@@ -24,6 +24,12 @@ jobs:
   build:
     name: Build Dataset
     runs-on: ubuntu-22.04
+    container:
+      image: nycplanning/build-geosupport:latest
+    defaults:
+      run:
+        shell: bash
+        working-directory: ./db-template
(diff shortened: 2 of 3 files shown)


python/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.9
```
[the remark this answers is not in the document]
fvankrieken: Intriguing. Yeah dev container is currently pinned to 3.9 because that's what `docker-geosupport` image is based on, so for consistency's sake both build images are using 3.9 as well. Would be happy to bump if we want though

fvankrieken (approved):

fvankrieken:
Do you want to either drop requirements files in template repo or bump there too to make dependabot happy?

python/requirements.txt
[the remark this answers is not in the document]
damonmcc: down to stick with 3.9 for now

db-template/python/geospatial.py
```diff
@@ -71,7 +73,7 @@ def pad_map_bounds(
 def map_simple(
     data: gpd.GeoDataFrame,
     projection: str = NYC_PROJECTION,
-    basemap: str | TileProvider = DEFAULT_BASEMAP,
+    basemap: Union[str, TileProvider] = DEFAULT_BASEMAP,
```
damonmcc, line 76: `|` to indicate multiple types in a typehint is only support in python versions >=3.10

db-checkbook/checkbook.db
alexrichey, line 1: @damonmcc 
make sure to rebase this out

damonmcc: oh I think I did just that as you were typing this comment? lol the checkbook.db file is gone now

alexrichey: It's still in the first commit... I _think_ that if you squash merge it'll never touch `main`, but if you merge commit, it might. Not sure! I'd just rebase it out.

damonmcc: thanks for clicking into the commits and catching this!

alexrichey (approved):
LGTM

PR #83, opened 2023-07-26
testing outputs of db-checkbook build 

athursland:
After running our pipeline from start to finish we realized that the output file shape was different than expected, so adding some basic tests to ensure that the output we export to DO is correct. 

tagging @DeaBardhoshi 
## to do: 
- [x] create test file containing functions to test the output of running `build.py`
- [x] ensure that the source data is the expected size (`nycoc_checkbook.csv` and all of the cpdb shapefiles) 
- [x] test the results of merging the cpdb geometries
- [x] test cleaning checkbook nyc data 
- [x] test results of the groupby on `fms_id` of the checkbook data
- [x] test the results of left joining the cpdb geometries onto the checkbook nyc dataset


athursland:
Note for future: to improve testing for this pipeline, the functions in `build.py` should be re-written to be more modular, and in such a way that makes it simpler to run tests on sample data. For example, _merge_cpdb_geoms() takes no arguments and grabs the cpdb data from the `.library` folder, which makes it strenuous to manually pass test data to verify the outputs of this function.

athursland:
also, definitely want to return to this and add testing for high sensitivity fixed asset category assignment

db-checkbook/test/test_build_output.py
```diff
@@ -1,19 +1,68 @@
...
+    assert not cpdb_df['maprojid'].duplicated().any()
+
+def test_null_geometry():
+    cpdb_gpd_list_test = generate_cpdb_test_data()
+    cpdb_df = _merge_cpdb_geoms(cpdb_gpd_list_test)
+    assert np.where(cpdb_df['geometry'].isnull())
+
+# --- results of join
+
+# @pytest.mark.skip(reason='TODO after major refactor')
```
damonmcc, line 61: could delete the comment?

db-checkbook/test/test_build_output.py
```diff
@@ -1,19 +1,68 @@
...
+def test_merge_cpdb_geoms():
+    cpdb_gpd_list_test = generate_cpdb_test_data()
+    expected_result = generate_expected_cpdb_join()
+    result = _merge_cpdb_geoms(cpdb_gpd_list_test)
+    assert result.equals(expected_result)
+
+# --- high-sensitivity fixed asset CATEGORY ASSIGNMENT
```
damonmcc, line 68: this is a section header for TODO tests?

db-checkbook/test/test_build_output.py
```diff
@@ -1,19 +1,68 @@
...
+# @pytest.mark.skip(reason='TODO after major refactor')
+def test_unique_maprojid(): 
+    cpdb_gpd_list_test = generate_cpdb_test_data()
+    cpdb_df = _merge_cpdb_geoms(cpdb_gpd_list_test)
+    assert not cpdb_df['maprojid'].duplicated().any()
+
+def test_null_geometry():
+    cpdb_gpd_list_test = generate_cpdb_test_data()
+    cpdb_df = _merge_cpdb_geoms(cpdb_gpd_list_test)
+    assert np.where(cpdb_df['geometry'].isnull())
```
damonmcc, line 57: what's this meant to test?

damonmcc (dismissed):
love it! just left a couple comments for little things but they're not too important

db-checkbook/test/test_build_output.py
athursland: sanity check to make sure none of the CPDB geometries are null. Since our input data from cpdb is only capital projects with geometries, I figured that if any of the rows lack geometries, it would indicate a problem with the source data - might be unnecessary though?

damonmcc: gotcha. this might be passing no matter what:  as far `assert` is concerned,  I think `np.where` returns a "truthy" value no matter what. an empty numpy array wouldn't be "false" enough for this to fail

but no worries maybe better to come back and iterate on this than hold it up!

damonmcc (approved):

PR #86, opened 2023-07-26
DB Checkbook fixed dataloading

DeaBardhoshi:
Fixing 01_dataloading functions + added *.db to gitignore, tagging @athursland

Notes on the S3 Loading Function:

put a checker in place to see if DO key for folder doesn't end with "/"
part of the files (2017 points and polygons and 2023 points) were not loading properly prior to this fix
(resubmitting PR without checkbook.db)


alexrichey (approved):
LGTM

Issue #50, opened 2023-07-14
23Q2 DevDB UPDATE


AmandaDoyle:
## Update code

- [ ] Add most recent full/half year to aggregate tables

## Update source data

- [ ] Update version.env to:

```bash
# DevDB versions for latest release
VERSION=22Q4 #Current Build Version of DevDB
VERSION_PREV=22Q2 #Previous Build Version of DevDB

# DevDB data details
CAPTURE_DATE= #Reference date for the DOB records (i.e. 22Q4 is a snapshot of all DOB records filed before 01-01-2023)
CAPTURE_DATE_PREV= #Reference date for the DOB records (i.e. 22Q4 is a snapshot of all DOB records filed after 2022-07-01)

# external tools versions
GEOSUPPORT_DOCKER_IMAGE_VERSION=22.3.0

# input spatial boundaries versions
GEOSUPPORT_VERSION=22c
DCP_MAPPLUTO_VERSION=
DOITT_ZIPCODE_VERSION=
DOF_VERSION=

# input data versions
COUNCIL_MEMBERS_VERSION=
DOE_ZONES_VERSION=
DOE_SUBDISTRICTS_VERSION=
HNY_VERSION=
DOB_DATA_DATE=
DOB_NOW_APPS_VERSION=
DOB_NOW_PERMITS_VERSION=
DOB_COFOS_VERSION=
DOITT_BUILDINGS_VERSION=
DOITT_BUILDINGS_HISTORICAL_VERSION=
```

#### City Administrative Boundaries
- [ ] `dcp_cdboundaries`  
- [ ] `dcp_cb2010`  
- [ ] `dcp_ct2010` 
- [ ] `dcp_cb2020` 
- [ ] `dcp_ct2020` 
- [ ] `dcp_school_districts` 
- [ ] `dcp_boroboundaries_wi`
- [ ] `dcp_councildistricts` 
- [ ] `dcp_firecompanies` 
- [ ] `dcp_policeprecincts` 
- [ ] `doitt_zipcodeboundaries` 
- [ ] `dof_shoreline`
#### General

- [ ] `dcp_mappluto_wi`
- [ ] `dof_shoreline` updated with zoningtaxlots, safe to ignore
- [ ] `council_members` [check opendate](https://data.cityofnewyork.us/City-Government/Council-Members/uvw5-9znb)
[...]

PR #88, opened 2023-07-27
fix DevDB corrections

### LINKED ISSUE #84: DevDB: not all corrections were applied in 23Q2 build
`corrections_not_applied.csv` is blank and many corrections are unapplied (e.g. units are showing as null that we can see in the manual corrections table we have previously corrected)

an example of this may be `104303004,classa_init,,9,,4/21/2022,VB` from `manual_corrections.csv`

it seams like only corrections related to the fields `latitude`, `longitude`, and `remove` are being applied. there are many variations of the sql notice `NOTICE: ( bin ) is not a valid field for function apply_correction to ( _INIT_devdb )` in [the raw build logs](https://github.com/NYCPlanning/data-engineering/commit/fa758665d3dd0038dc69fc1a27255e0cd351554c/checks/15292269898/logs). used the query below to check which fields had corrections applied:
```sql
select
  field, corr_applied
from build_workflow_dispatch_main.corrections_reference
group by field, corr_applied
order by field, corr_applied;
```

that notice is raised by [our sql procedure](https://github.com/NYCPlanning/data-engineering/blob/main/db-developments/sql/_procedures.sql#L77-L82) `apply_correction`

## likely cause

[the procedure is hard-coded](https://github.com/NYCPlanning/data-engineering/blob/main/db-developments/sql/_procedures.sql#L96) to use the `public` schema to create the list of valid columns for a given table, but the build is no longer taking place in that schema

the sql procedure `apply_corrections` uses the table `information_schema.columns` to:
1. determine which columns are in each table and would be valid to correct
2. not apply any if the column isn't in the list
3. raise a sql notice

```sql
...
WHERE table_schema = 'public'
AND table_name = lower(_table) INTO  _valid_fields;
...
```

since DevDB no longer uses the default DB schema `public` to build tables, this list was empty. per #50, DevDB now uses the edm-data DB and a schema which is unique to the branch + workflow trigger (e.g. `build_workflow_dispatch_fix_devdb_correction`)

## some options to fix

1. change corrections logic to not need the list of columns in a table. if a correction is attempted on a column that doesn't exist it should either fail, warn, or we should deprecate the correction (in-progress in #88)
2. change corrections logic to use the build schema rather than `public`. I tried this by passing an environment variable when calling `_procedures.sql`, but it didn't work and seems as though procedure definitions can't ingest variables defined as `:VARIABLE_NAME` like normal queries can
3. revert the use of a specific schema in the EDM-DATA DB
4. revert the use of EDM-DATA DB (in-progress in #95)

damonmcc:
resolves #84 

fixed by passing the build schema to `apply_corrections` (@fvankrieken got this working on [`fvk-devdb-corrections`](https://github.com/NYCPlanning/data-engineering/tree/fvk-devdb-corrections))

[successful build](https://github.com/NYCPlanning/data-engineering/actions/runs/5716162406/job/15487056722)

files sizes of corrections outputs:
image

## notes

the sql procedure `apply_corrections` was using the table `information_schema.columns` to:
1. determine which columns are in each table and would be valid to correct
2. not apply any if the column isn't in the list
3. raise a sql notice

```sql
...
WHERE table_schema = 'public'
AND table_name = lower(_table) INTO  _valid_fields;
...
```

since DevDB no longer uses the default DB schema `public` to build tables, this list was empty. per #50, DevDB now uses the edm-data DB and a schema which is unique to the branch + workflow trigger (e.g. `build_workflow_dispatch_fix_devdb_correction`)

### DIFF
--- a/db-developments/sql/_init.sql
+++ b/db-developments/sql/_init.sql
@@ -94,12 +94,12 @@ CORRECTIONS:
 	date_statusr
 	date_statusx
 */
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'stories_prop');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'bin');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'bbl');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_lastupdt');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_filed');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusd');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusp');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusr');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusx');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'stories_prop');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'bin');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'bbl');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_lastupdt');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_filed');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_statusd');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_statusp');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_statusr');
+CALL apply_correction(:'build_schema', '_INIT_devdb', '_manual_corrections', 'date_statusx');
--- a/db-developments/sql/_procedures.sql
+++ b/db-developments/sql/_procedures.sql
@@ -80,6 +80,7 @@ If a field does not exist in the table or is a geometry field, notice gets raise
 */
 DROP PROCEDURE IF EXISTS apply_correction;
 CREATE OR REPLACE PROCEDURE apply_correction (
+    _schema text, 
     _table text, 
     _corrections text,
     _field text,
@@ -92,12 +93,13 @@ DECLARE
     _reason text;
     _valid_fields text[];
 BEGIN
+    RAISE NOTICE 'Attempting to apply corrections for field % to %.%', _field, _schema, _table;
+    
     SELECT array_agg(column_name) FROM information_schema.columns
-    WHERE table_schema = 'public' 
+    WHERE table_schema = _schema
     AND table_name = lower(_table) INTO  _valid_fields;
 
     IF (_field = any(_valid_fields)) AND (_field NOT IN ('latitude','longitude')) THEN
-        RAISE NOTICE 'Applying Corrections for %', _field;
         FOR _job_number, _old_value, _new_value, _reason IN 
             EXECUTE FORMAT($n$
                 SELECT job_number, old_value, new_value, reason 
@@ -117,7 +119,7 @@ BEGIN
             );
         END LOOP;
     ELSE
-        RAISE NOTICE '( % ) is not a valid field for function apply_correction to ( % )', _field, _table;    
+        RAISE EXCEPTION '( % ) is not a valid field for function apply_correction to ( % )', _field, _table;    
     END IF;
 END
 $BODY$ LANGUAGE plpgsql;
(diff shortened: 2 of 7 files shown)


fvankrieken:
Yeah definitely seems like if a correction isn't valid it should still go into not applied. Unless we'd want a third output to have "applied", "not applied" (but valid), and "not valid" just as some sort of qc

fvankrieken:
Actually with how the correction code is written I'm going to disagree with that - for an "invalid" correction field, it will now go into that clause, return an empty query result for the for loop, then exit silently, right? And not ever make it into a call of the `correction` function. The original code wasn't perfect either - similar thing would happen if you tried to correct a column of table A but provided a column that's in table B but not A. But regardless not having the valid columns still encourages silent failures from what I can tell

damonmcc:
[looks like](https://github.com/NYCPlanning/data-engineering/blob/add-checkbook-pr-tests/db-developments/sql/_procedures.sql#L70) a correction only can be in not_applied if the procedure `correction` is reached

`IF (_field = any(_valid_fields))` in `apply_correction` was preventing that, so `corrections_not_applied.csv` was blank. now it's 18.6 KB. it was 12.7 KB in the 22Q4 release

fvankrieken:
Okay I'm slightly wrong in how I described it I think - let me rephrase and summarize in terms of how things end up in `corrections_applied` and `corrections_not_applied`

Old approach does two key things in apply_corrections - checks valid columns, then looks for potential applications (and attempts to execute). For valid columns if potential applications are found, each potential application is either applied or not, with rows ending up in both `corrections_applied` and `corrections_not_applied`. However, if no potential applications are found, nothing is logged at all. This seems like a problem. Also, if the column is invalid, a notice is raised - seems that it'd be preferable to log in `corrections_not_applied` instead.

New approach, the only difference now is we're not filtering by valid column. This will lead to a SQL error rather than raising a notice. I think we'd rather log in `corrections_not_applied` than have the build error out?

fvankrieken:
Actually that's still not quite right, sorry. Let me keep staring at this code

db-developments/sql/_init.sql
```diff
@@ -94,12 +94,12 @@ CORRECTIONS:
...
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'bin');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'bbl');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_lastupdt');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_filed');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusd');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusp');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusr');
-CALL apply_correction('_INIT_devdb', '_manual_corrections', 'date_statusx');
\ No newline at end of file
+CALL apply_correction('{{BUILD_SCHEMA}}', '_INIT_devdb', '_manual_corrections', 'stories_prop');
```
fvankrieken, line 97: I believe args need to be in format `:'BUILD_SCHEMA'`. I think this is just passing that string literal at the moment, hence why no schema is being used and therefore no corrections are being applied

fvankrieken: Ah yeah they're not being applied either. Only lat/long corrections which have a different mechanism

damonmcc: yea I tried `:'BUILD_SCHEMA'` first and [the build failed](https://github.com/NYCPlanning/data-engineering/actions/runs/5693851614/job/15433833407)

fvankrieken: Looks like a mismatch of variable name? `BUILD_ENGINE_SCHEMA` in the `psql` call, `:'BUILD_SCHEMA;` in the sql file

damonmcc: that's what I get for trying to shorten the variable name

damonmcc: `corrections_applied.csv` is still 59 B. added a sql NOTICE to see what schema it's using and re-running

db-developments/sql/_procedures.sql
```diff
@@ -26,6 +26,7 @@ Procedure to apply a single correction:
 - If applicable, applies correction and adds to 'corrections_applied' table
 - If not applicable, adds to 'corrections_not_applied'
 */
+--\set build_schema `echo "$BUILD_ENGINE_SCHEMA"`
```
fvankrieken, line 29: Can drop this now

db-developments/sql/_procedures.sql
```diff
@@ -120,4 +123,4 @@ BEGIN
         RAISE NOTICE '( % ) is not a valid field for function apply_correction to ( % )', _field, _table;
```
fvankrieken, line 123: Let's make this `RAISE EXCEPTION` so that if we're calling this with bad values that the call fails?

fvankrieken (approved):

PR #91, opened 2023-07-27
Facilities Database Build Changes

alexrichey:
This a grab-bag of tweaks needed to get Facilities building again. Context for each change is in the commit body.


db-facilities/facdb/sql/dsny_electronicsdrop.sql
```diff
@@ -2,13 +2,13 @@ DROP TABLE IF EXISTS _dsny_electronicsdrop;
...
-    street as streetname,
+    NULL as addressnum,
+    NULL as streetname,
     address,
     NULL as city,
     zipcode,
-    NULL as boro,
-    LEFT(borocd, 1) as borocode,
+    borough as boro,
+    NULL as borocode,
```
fvankrieken, line 11: Glanced at the dataset, this one is a bit of a doozy in terms of changes it seems.

LEFT(bbl, 1) should likely be a more steady way to get borocode, I don't think that should be dropping from the dataset. And if it does, it's its own column so it will present more issues

db-facilities/facdb/sql/dsny_specialwastedrop.sql
```diff
@@ -4,8 +4,8 @@ SELECT uid,
     name as facname,
     parsed_hnum as addressnum,
     parsed_sname as streetname,
-    address,
-    city,
+    NULL as address,
+    NULL as city,
```
fvankrieken, line 8: These seem to still be there, looking at the files in recipes

alexrichey: You're right, definitely shouldn't have dropped these. Will remove this change and re-run

db-facilities/facdb/sql/dsny_electronicsdrop.sql
alexrichey: Thanks! Will give that a shot.

fvankrieken (approved):

PR #93, opened 2023-07-27
adding ability to manually change zoning values

AmandaDoyle:
Adding in manual changes to research file, script to apply changes, and line to run the script

### DIFF
--- a/db-pluto/pluto_build/data/pluto_input_research.csv
+++ b/db-pluto/pluto_build/data/pluto_input_research.csv
@@ -27254,3 +27254,5 @@ bbl,field,old_value,new_value,Type,reason,version
 ,ownername,NEW YORK STATE DEPARTMENT OF TRANSPORTAT,NYS DEPARTMENT OF TRANSPORTATION,1,Normalized owner name,23v2
 ,ownername,NYS COMPTROLLER OFFICE,NYS COMPTROLLER'S OFFICE,1,Normalized owner name,23v2
 ,ownername,NYS DEPT OF TRANSPORTATIO,NYS DEPARTMENT OF TRANSPORTATION,1,Normalized owner name,23v2
+4158620001,zonedist2,,M1-4/R6,2,issue calculating zoning distrcts,23v2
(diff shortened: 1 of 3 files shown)


fvankrieken (approved):

db-pluto/pluto_build/data/pluto_input_research.csv
```diff
@@ -27253,4 +27253,6 @@ bbl,field,old_value,new_value,Type,reason,version
 ,ownername,NYCHA,NYC HOUSING AUTHORITY,1,Normalized owner name,23v2
 ,ownername,NEW YORK STATE DEPARTMENT OF TRANSPORTAT,NYS DEPARTMENT OF TRANSPORTATION,1,Normalized owner name,23v2
 ,ownername,NYS COMPTROLLER OFFICE,NYS COMPTROLLER'S OFFICE,1,Normalized owner name,23v2
-,ownername,NYS DEPT OF TRANSPORTATIO,NYS DEPARTMENT OF TRANSPORTATION,1,Normalized owner name,23v2
\ No newline at end of file
+,ownername,NYS DEPT OF TRANSPORTATIO,NYS DEPARTMENT OF TRANSPORTATION,1,Normalized owner name,23v2
+4158620001,zonedist2,,M1-4/R6,2,issue calculating zoning distrcts,23v2
+4158620001,spdist1,,MX-21,2,issue calculating zoning distrcts,23v2
```
fvankrieken, line 27258: @AmandaDoyle looking at the sql below, this isn't being applied because it matches column 3 here, `old_value`, with the value in the table.

We could
- change the sql to not check for equality of old value and the value to be replaced
- support nulls - not sure if this csv is getting loaded as whitespace or null (and not sure what the values look like during the build), but `'' = ''` returns `true` in postgres while `null = null` returns false. So we could change `AND b.old_value = a.zonedist2` to `AND (b.old_value = a.zonedist2 OR b.old_value IS NULL)` (or check if they're both null for that last clause - depends how forgiving we want to be)
- add the correct old value to the input_research file - this is if these are actually replacing values, not nulls. But if I remember right, we are adding them specifically because they are turning up null in QA

PR #94, opened 2023-07-27
add utils to upload file, upload/download folder

fvankrieken:
A couple points

- do we want to use the `rich` printing/progress I copied over from data library? Essentially all `upload_file` is is a wrapper around boto3 client built-in upload with this progress. If we like this we could do a similar thing for download
- download_folder is essentially ported over from @DeaBardhoshi and @athursland 's work in db-checkbook
- ~~should acl have default? This is what we had in data library but it seems maybe it's good to make ourselves be explicit if we're making things being pushed public rather than defaulting to it~~ going with no default - acl must be passed explicitly. We could have a small wrapper in `recipes` if we really want one that defaults to public but this will likely be used to push publishing data as well
- ~~I have a funny mish-mash of os.path stuff and pathlib.Path~~ better now I think

### DIFF
--- a/dcpy/connectors/s3.py
+++ b/dcpy/connectors/s3.py
@@ -85,17 +85,17 @@ def download_folder(
 
 
 def upload_folder(
-    bucket:str,
-    local_folder_path:str,
-    upload_path:str,
+    bucket: str,
+    local_folder_path: str,
+    upload_path: str,
     acl: str = "public-read",
     metadata: dict = {},
-    include_foldername: bool=True,
+    include_foldername: bool = True,
 ):
     local_folder_path = Path(local_folder_path)
     if not local_folder_path.exists() or (not local_folder_path.is_dir()):
         raise NotADirectoryError(f"'{local_folder_path}' is not a folder.")
-    for path_object in local_folder_path.rglob('*'):
+    for path_object in local_folder_path.rglob("*"):
         if path_object.is_file():
             if not include_foldername:
                 key = upload_path / Path(*path_object.parts[2:])
--- a/dcpy/connectors/s3.py
+++ b/dcpy/connectors/s3.py
@@ -28,16 +28,16 @@ def client(
     )
 
 
-## adapted from data library
-## https://github.com/NYCPlanning/db-data-library/blob/main/library/s3.py#L37
 def upload_file(
     bucket: str,
     path: str,
     key: str,
     acl: str = "public-read",
-    metadata: dict = {},
+    metadata: dict = None,
 ) -> dict:
-    """Uploads a single file to AWS S3."""
+    """Uploads a single file to AWS S3.
+    adapted from data library
+    https://github.com/NYCPlanning/db-data-library/blob/main/library/s3.py#L37"""
     with Progress(
         SpinnerColumn(spinner_name="earth"),
         TextColumn("[progress.description]{task.description}"),
--- a/dcpy/connectors/s3.py
+++ b/dcpy/connectors/s3.py
@@ -53,7 +53,8 @@ def upload_file(
             f"[green]Uploading [bold]{path.name}[/bold]", total=size
         )
         extra_args = {"ACL": acl}
-        if metadata: extra_args["Metadata"] = metadata
+        if metadata:
+            extra_args["Metadata"] = metadata
         response = client().upload_file(
             path,
             bucket,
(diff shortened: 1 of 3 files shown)


dcpy/connectors/s3.py
```diff
@@ -8,9 +18,83 @@ def client(
...
+
+
+## adapted from data library
+## https://github.com/NYCPlanning/db-data-library/blob/main/library/s3.py#L37
+def upload_file(
+    bucket: str,
+    path: str,
+    key: str,
+    acl: str = "public-read",
+    metadata: dict = {},
```
alexrichey, line 38: So this is a bizarre Python'ism (thanks Guido) but you'd typically pass None here due to [dangerous-default-values](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html

alexrichey: also, if you want to put the context above the fn in the docstring, that'd be 👍

fvankrieken: Good to know

fvankrieken:
Made a couple tweaks around paths and reformatted - more ready for review now. I'll add any more changes in additional commits, will likely just squash this one

dcpy/connectors/s3.py
```diff
@@ -8,9 +18,87 @@ def client(
...
+    return response
+
+
+def download_folder(
+    bucket, prefix, export_path, include_prefix_in_export=False
+) -> None:
+    """given bucket, prefix filter, and export path, download contents of folder from s3 recursively"""
+    export_path = Path(export_path)
+    if prefix[-1] != "/":
+        raise NotADirectoryError("prefix must be a folder path, ending with '/'")
```
alexrichey, line 70: Maybe I'm missing where we defined this Exception? Also, you might just use an assert here, unless you were planning to specifically handle `NotADirectoryError`s somewhere?

dcpy/connectors/s3.py
```diff
@@ -8,9 +18,87 @@ def client(
...
+            ExtraArgs={"ACL": acl, "Metadata": metadata},
+            Callback=lambda complete: progress.update(task, completed=complete),
+        )
+    return response
+
+
+def download_folder(
+    bucket, prefix, export_path, include_prefix_in_export=False
+) -> None:
+    """given bucket, prefix filter, and export path, download contents of folder from s3 recursively"""
```
alexrichey: So does this just download a folder and it's non-folder children? Recursively makes me think you'd be recursing to download a full folder tree? (seems fine either way... maybe safer the first way, so you don't accidentally download a full bucket)

alexrichey: for later readers, yes, it is recursive because of how prefixes work

dcpy/connectors/s3.py
fvankrieken: It's a built-in python exception, just figured give the most specific error possible. I'm personally fine with asserts as well, just depends a bit what we like for stylistic norms around errors

alexrichey: didn't realize it was a built-in. Seems fine!

dcpy/connectors/s3.py
fvankrieken: Chatted in person but for posterity (aka @damonmcc ) - list_objects does find ALL objects with the given path prefix. So downloading a whole bucket (or first 1000, believe is default for list_objects) is certainly possible.

dcpy/connectors/s3.py
```diff
@@ -8,9 +18,88 @@ def client(
...
+            if include_prefix_in_export:
+                export_path = export_path / prefix
+            if not (export_path / key_directory).exists():
+                os.makedirs(export_path / key_directory)
+            client_.download_file(bucket, obj["Key"], export_path / key)
+
+
+def upload_folder(
+    bucket: str,
+    local_folder_path: str,
```
alexrichey, line 89: do you think it makes sense just ask for `pathlib.Path`s in these methods?

fvankrieken: Yeah I went back and forth a little myself. I assume you think yes?

fvankrieken:
download_folder is currently a tad broken, hold off on further reviews

fvankrieken:
Functions working as expected! All set from my end

alexrichey (approved):
Good stuff!

alexrichey:
Oh, one other thing. Can you add `rich` to requirements.txt/in?

PR #59, opened 2023-07-18
build DevDB 23Q2 using EDM-DATA db

damonmcc:
resolves #50 

## todos
- [x] archive `dob_cofos`, `dob_now_permits`, `dob_now_applications`
- [x] `psql:sql/_status_q.sql:47: ERROR:  invalid input syntax for type date: "00:00.0"` in [latest run](https://github.com/NYCPlanning/data-engineering/actions/runs/5602877945/jobs/10248712223#step:7:613)
- [x] keep runnin [[in-progress builds](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml)]
- [ ] ...

## changes
- update source data versions in `versions.env`
- use new common build github action to build DevDB
- ...

## notes
- latest outputs are in S3 at `edm-publishing/db-developments/update-devdb-23Q2/23Q2/`
- the DevDB data sync & geocoding [action](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/developments_datasync.yml) was silently failing. appeared to be due to `local version=$(get_version( $2)` in `devdb.sh` being empty ([run logs](https://github.com/NYCPlanning/data-engineering/actions/runs/5570531785/jobs/10174908216#step:10:114)). with changes in this PR, the error seems to be fixed ([run logs](https://github.com/NYCPlanning/data-engineering/actions/runs/5601017037/jobs/10244718031#step:10:113))

PR #95, opened 2023-07-28
Devdb: use a temporary db to build

### LINKED ISSUE #84: DevDB: not all corrections were applied in 23Q2 build

damonmcc:
resolves #84 

to avoid any changes to the logic of how corrections are applied, this reverts some changes from #59 which changed the build DB from the schema `public` in a temporary DB created by the github action to a unique schema in the persistent EDM-DATA DB

[successful build](https://github.com/NYCPlanning/data-engineering/actions/runs/5693086692)

files sizes of relevant output files:
image


damonmcc:
@fvankrieken I tried passing the schema to the sql procedure and it didn't work (empty `corrections_applied.csv` noted in https://github.com/NYCPlanning/data-engineering/pull/88)

with that and the ballooning disk usage of `edm-data`, seems like "reverting" to using a temporary DB may be the best approach for now

fvankrieken:
@damonmcc can we close this?

PR #96, opened 2023-07-28
Db checkbook Exporting Functions

DeaBardhoshi:
Created functions to upload final output to S3, create metadata for folder structure, write a version txt file and source data csv file. Adding @athursland


fvankrieken (changes requested):

fvankrieken:
Can you generate outputs in DO again, with the new publishing path?

DeaBardhoshi:
> Can you generate outputs in DO again, with the new publishing path?

just did!

db-checkbook/build_scripts/03_export.py
```diff
@@ -1,6 +1,57 @@
...
+    file_version = open(_curr_file_path.parent.parent / 'output'/ 'version.txt',"w+")
+    file_version.write(str(today))
+    return 
+
+def upload_final_file(final_file, digital_ocean_file) -> None:
+    """ final_file: str of output csv 
+        digital_ocean_file: str name of file in digital ocean to save  
+    """
+
+    s3_client = s3.client('s3',
```
fvankrieken, line 31: Very minor note - this client doesn't need any args provided, these are the same as the defaults (and I'm surprised the positional `s3` didn't cause an error), so this can just be `s3.client()`. Once you push that I'll approve and we can get this merged!

fvankrieken (approved):

PR #103, opened 2023-08-01
Expose Jupyter Notebook default port in container

alexrichey:
It's nice to run a jupyter notebook in docker to have access to pg, dockerized python deps, and gdal, etc.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -33,6 +33,8 @@ services:
       POSTGRES_PASSWORD: postgres
     expose: 
       - "5432"
+    ports:
+      - "8888:8888"
 
 volumes:
   postgres-data:


alexrichey:
question for @fvankrieken and @damonmcc - Are you using Jupyter Notebookes, and if so how are you running them? Wondering if this change is necessary.

fvankrieken:
I am - I just have a folder in my repo/devcontainer where I keep them and run them in vscode (basically just open it like any other file and run it right there)

fvankrieken (approved):

fvankrieken:
TBH I have not looked into the details, but from interacting with it basically seems that it starts its own server in the background, in whatever environment your in. You also can add external/existing servers but I haven't done that. Don't think it's horribly hard - at first glance I think roughly what you've done here would cover the changes to the container needed, if you had a server running locally outside of the dev container

PR #104, opened 2023-08-01
Move data products into subfolder

fvankrieken:
Definitely look at individual commits, otherwise it'll be a nightmare

- Commit 1: moved data products to `products` folder
- Commit 2
  - updated filepaths to reflect this
  - some minor bash cleanup (use pathlink to consistently reference `bash/utils.sh` from product-specific `config.sh` files)

@alexrichey and @damonmcc - this is a bit of a nightmare obviously. A couple notes

1. @alexrichey motivation was primarily around moving in qaqc app to this repo to be able to start parsing out additional python utilities (and making use of what's been added to dcpy). With expanding scope of repo, would make sense to move data products into a subfolder. @damonmcc and I had liked something along the lines of [this](https://gitlab.com/gitlab-data/analytics/-/tree/master/) as a long term goal, though it doesn't need to be that specifically. Also our data products aren't quite just "transformations" right now, hence `products` for the folder name.
2. Should we maybe have a huddle to talk about what we're thinking in terms of longer-term repo structure? Might be good before making any big tweaks, to try to minimize having to do this again
3. Does it make sense to just try to bite the bullet and do this all at once? Might be some gnarly collisions - I would hope just rebasing main into a feature after this would work, but I don't think it'll be quite that simple. Maybe this waits until we hit a point with minimal active dev branches and then do this change, or maybe we move data products in one or a couple at a time when we know there's no other active development on them.


fvankrieken:
@alexrichey and @damonmcc - this is a bit of a nightmare obviously. A couple notes

1. @alexrichey motivation was primarily around moving in qaqc app to this repo to be able to start parsing out additional python utilities (and making use of what's been added to dcpy). With expanding scope of repo, would make sense to move data products into a subfolder. @damonmcc and I had liked something along the lines of [this](https://gitlab.com/gitlab-data/analytics/-/tree/master/) as a long term goal, though it doesn't need to be that specifically. Also our data products aren't quite just "transformations" right now, hence `products` for the folder name.
2. Should we maybe have a huddle to talk about what we're thinking in terms of longer-term repo structure? Might be good before making any big tweaks, to try to minimize having to do this again
3. Does it make sense to just try to bite the bullet and do this all at once? Might be some gnarly collisions - I would hope just rebasing main into a feature after this would work, but I don't think it'll be quite that simple. Maybe this waits until we hit a point with minimal active dev branches and then do this change, or maybe we move data products in one or a couple at a time when we know there's no other active development on them.

fvankrieken:
Actually also it looks like some files got dropped when I pulled latest main in - not sure how that happened. Might need to do this over a bit.

fvankrieken:
> Actually also it looks like some files got dropped when I pulled latest main in - not sure how that happened. Might need to do this over a bit.

Actually nevermind on that - just files that by our current .gitignore should be ignored but were checked in previously

fvankrieken:
I would love dropping the `db-` prefixes from the folders personally

alexrichey:
> I would love dropping the `db-` prefixes from the folders personally

I guess that would affect the publishing folders, so maybe that's something we'd implement product by product

fvankrieken:
It wouldn't need to, but it would lead to at least a mismatch between the two

.github/workflows/test.yml
```diff
@@ -29,9 +29,9 @@ jobs:
               - .github/workflows/build.yml
               - db-template/**
```
damonmcc, line 30: should this be `products/db-template/**`?

fvankrieken: Yup. Done

.github/workflows/archive/cpdb_spending.yml
```diff
@@ -37,7 +37,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: install dependencies ...
-        run: ../bash/docker_container_setup.sh
+        working-directory: ./
```
damonmcc, line 40: when no bash defaults are set for the job, is `working-directory: ./` needed?

fvankrieken: This does have a default working directory, though that wasn't updated! I'll do another pass on my local files

fvankrieken: Fixed

damonmcc:
🎉

PR #107, opened 2023-08-01
add build process for db-checkbook

damonmcc:
## changes
- add a github action to call the three `build_script` files aka build stages: dataloading, build, export
- rename `build_script` files to be importable / callable (e.g. no numeric prefix in filename)
- add `__init__.py` to set common constants and create the library and output folders
- minor changes to `build_script` files to fix bugs, use `dcpy` utils, and ensure calling them as modules works as intended
- update README
- standardize the export-related input in the common build action: `run_export` rather than `will_export`

## notes
- [successful build](https://github.com/NYCPlanning/data-engineering/actions/runs/5740463258) and screenshot of outputs in `edm-publishing`
  - image

- at first I wrote a python script to call the scripts but since we often separate export from everything before it, decided to just call each script separately


.github/workflows/checkbook_build.yml
```diff
@@ -0,0 +1,40 @@
...
+    name: Build Dataset
+    runs-on: ubuntu-22.04
+    container:
+      image: nycplanning/build-geosupport:latest
+    defaults:
+      run:
+        shell: bash
+        working-directory: products/db-checkbook
+    steps:
+      - uses: actions/checkout@v3
```
athursland, line 26: more of a question than feedback, but what is this line doing?

.github/workflows/build.yml
```diff
@@ -29,7 +30,7 @@ on:
         description: "A short note about this build (optional)"
         type: string
         required: false
-      will_export:
+      run_export:
```
athursland, line 33: I am a fan of the verb change

products/db-checkbook/build_scripts/build.py
```diff
@@ -195,21 +193,26 @@ def assign_category(row):
     gdf['final_category'] = gdf[cols].apply(lambda row: assign_category(row), axis=1)
     return gdf
 
-def run_build() -> pd.DataFrame:
+def run_build() -> None:
     """
     :return: historical spending data
     """
+    print("merge and group source data ...")
```
athursland, line 200: do we want to keep these print statements? (I am pro-print statement, just wondering if leaving these here was intentional or not)

.github/workflows/build.yml
```diff
@@ -11,6 +11,7 @@ on:
         options:
           - template
           - cbbr
+          - checkbook
```
athursland, line 14: :')

.github/workflows/checkbook_build.yml
damonmcc: it's my understanding that `actions/checkout@v3` is checking out the repo during the github action run so that every step after this has access to the files in the repo at a specific commit, by default the commit from which the run was triggered ([link to this in a build](https://github.com/NYCPlanning/data-engineering/actions/runs/5740463258/job/15558375820#step:3:19), [link to docs](https://github.com/actions/checkout))

products/db-checkbook/build_scripts/build.py
DeaBardhoshi: also pro-print statements, I think they are helpful for knowing where in the process we are

products/db-checkbook/build_scripts/export.py
DeaBardhoshi, line 1: Do we want to add the function that makes the txt file for the versions of the CPDB datasets?

damonmcc: I think that's what we're getting with `source_data_versions.csv` which is being exported to S3. here's what the csv looks like when I download it:
image

products/db-checkbook/build_scripts/build.py
damonmcc: these I'd meant to keep, they can definitely get to be a lot but with build pipelines I especially like logging when a major stage of it has started

I resist putting these types of "status" print statements in smaller functions in favor of putting them in the higher level calling functions

products/db-checkbook/build_scripts/export.py
DeaBardhoshi: Yes, you're right. I realized we made an extra function just logging the date of the build (originally called "create_version_text_file") which it seems might not be necessary?

damonmcc: yea since the exports are organized into versioned folders, it didn't seem worth having an extra text file with one string in it

fvankrieken: Counterpoint to that is when we have a "latest" folder, it's useful to have the `version.txt`. Open to other alternatives, but for now I think it's maybe best to leave that

fvankrieken (approved):

athursland (approved):

DeaBardhoshi (approved):

PR #74, opened 2023-07-24
ensure all products can build in monorepo

damonmcc:
resolves #72, resolves #3

## changes
- ZTL
  - update paths in github actions and fix the use of `pg_dump` in 02_build.sh
  - [dataloading passing](https://github.com/NYCPlanning/data-engineering/actions/workflows/zoningtaxlots_dataloading.yml)
  - [build passing](https://github.com/NYCPlanning/data-engineering/actions/runs/5649591665)
- KPDB
  - deprecate the use of the [KPDB data repo](https://github.com/NYCPlanning/db-knownprojects-data) in favor of a private S3 bucket
  - disable export step and code until use of private S3 bucket is implemented (not implemented yet due to build issues, see notes)
  - incorporate all data loading / transformation code from the KPDB repo
  - simplify build action and use common bash utils
  - [build failing](https://github.com/NYCPlanning/data-engineering/actions/runs/5753922994/job/15598117213#step:6:17) at expected point (when using `dcp_projects` columns)

## notes
- this is meant to be focused on "does it build in the mono repo?" this doesn't have to improve builds by using shared bash utils, using shared containers, or being called via the common build action (but it might)
- KPDB: many columns that are expected to be in the source dataset `dcp_projects` aren't there because we only archive the public version, which has less columns than the private version. I'll create a separate issue for this
- on related issues, I'll flag which tasks have been completed by this PR

Issue #11, opened 2023-06-09
ZTL - fix versioning, align s3 folder layout with other repos


fvankrieken:
ZTL uses `%Y/%m/01` as date format for versioning, as written [here](https://github.com/NYCPlanning/data-engineering/blob/main/db-zoningtaxlots/bash/config.sh#L145) (if link is broken, changes have been merged and it's one of the only lines in the file). In s3 this creates unnecessary subfolders by month and day. This same version format is used in `EDM_DATA` for archives which get used in ztl build to generate qaqc outputs. I would propose
- [x] change version to `%Y-%m-01` to align a bit better with other products
- [x] add logic in interaction with `EDM_DATA` data to do something more postgres friendly than having dashes (maybe just remove dashes), both for writing and reading
- [x] update old exports in DO to use this format
- [ ] update historical `EDM_DATA` versions, both in table names and column values in the 3 qc tables
- [ ] make sure any downstream customers are fine. I think they would just get it from `latest` folder which doesn't have this issue

Beyond that, maybe worth going to the branch/date format we use for most repos, though maybe that can wait for another issue where we aim to align all data products, because many differ just barely (output/no output, whether latest is in main or one level above it, etc)

athursland:
Looks like run instructions in the README for ZTL is outdated too - will update in PR for this issue

damonmcc:
@fvankrieken @athursland

with Ali's PR https://github.com/NYCPlanning/data-engineering/pull/109 merged, do we wanna drop the tasks around changing old files and tables and close this?

fvankrieken:
I think we should update s3 folders to align with this

damonmcc:
renamed and moved all folders in `edm-publishing/db-zoningtaxlots` to have the pattern `YYYY-MM-01/output/...`

PR #109, opened 2023-08-02
fix versioning, align s3 folder layout in ZTL with other repos

athursland:
resubmitting draft PR after noticing I opened the last one before pulling the changes to main yesterday

Fixing portions of #11, pulled from list of gfi, tagging @DeaBardhoshi 

### what changed
- Modified date formatting to use hyphens instead of slashes to align more closely with other DE data products 
- Added an environment variable to config.sh to define SQL table names for version / previous version of ZTL dropping the hyphens
- updated references to version/version_prev in the build and qaqc bash scripts to reflect these changes

### what is out of scope for this PR
- re #11 these changes do not modify the existing historical versions of ZTL on DO... so merging this PR does not close the issue

### notes
Build failed when I tried testing these changes locally, which may have to do with this issue that there's [currently a PR for](https://github.com/NYCPlanning/data-engineering/pull/74).


damonmcc:
looks like there's a merge conflict in 02_build. you'll want to keep the version from main

athursland:
> looks like there's a merge conflict in 02_build. you'll want to keep the version from main

done!

fvankrieken:
A little late to the party, but looks good - will definitely want to test how this impacts the qaqc process (make sure that these versions don't make it into column names, values in rows that get used in qaqc app), but that's not exactly easy to test at the moment outside of a normal build

PR #114, opened 2023-08-03
Move QAQC app into monorepo, remove dokku

fvankrieken:
Relatively straightforward - to see actual changes to functionality, go commit-by-commit. In these we have:

1. code copied in from qaqc repo
2. modification to deploy via git subtree (not changing DO infrastructure) - now defunct but leaving for posterity. This logic successfully deployed to the dokku app from the qaqc repo
3. remove dependency on dokku, build/publish/deploy docker image to new droplet


apps/qaqc/bash/droplet_setup.sh
```diff
@@ -0,0 +1,6 @@
+doctl compute droplet create \
```
fvankrieken, line 1: This file is not used, nor did I actually run it, but this is the command that would recreate the droplet with doctl (cli for DO) set up. Not sure if we feel like this would be worth having in here but included for now

fvankrieken:
This is essentially done, could use review on commit number 3 which is main functionality (and shouldn't change)

Todo
- [ ] finish up DO setup (and maybe modify that little doctl script I dumped into the repo, if we want it)
- [ ] grab @alexrichey 's latest commits to the qaqc app

Need to check a bit around ssls, if there's pre-config I want in DO like using apps vs just creating droplet, or if maybe we want a setup script checked in to this repo for getting the instance 100% ready to go

damonmcc:
@fvankrieken
it does seem nicer to have a script to set things up, but I guess if with DO apps we can get that + other stuff we need like SSLs maybe that bash script isn't worth the potential confusion?

.github/workflows/qaqc_deploy.yml
```diff
@@ -4,26 +4,47 @@ on:
...
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          sparse-checkout: apps/qaqc
-      - name: Deploy
-        run: apps/qaqc/bash/deploy.sh 
+          sparse-checkout: |
+            apps/qaqc
+            bash
+            dcpy
```
damonmcc, line 22: just curious if a sparse checkout is still needed. we've definitely limited the idea to "QAQC app should use `dcpy`" so no worries, but maybe access to product-level files would be valuable in the future?

apps/qaqc/entrypoint.sh
damonmcc, line 1: does this affect local deployment? at the moment, this script and port `5000` are mentioned in the README

fvankrieken: Ah yeah that would need to be updated in the dev container (exposing the port)

There's no real huge rationale for dropping the 5000... but I don't really see an argument for keeping it either? Streamlit defaults to 8501 and I feel like maybe we should just go with that?

fvankrieken: Ah other remaining point - need to merge python requirements files

alexrichey:
Good stuff! Seems like a nice starting point to iterate off. I can move my changes over once this is merged

alexrichey (dismissed):

apps/qaqc/entrypoint.sh
damonmcc: gotcha, I like the idea of using Streamlit's default port

on merging requirements files, is that needed or nice-to-have? It'll definitely be important when QAQC starts using `dcpy`

damonmcc (dismissed):

apps/qaqc/entrypoint.sh
fvankrieken: It's needed if it's going to be run locally. Which for now is maybe a nice-to-have?

alexrichey:
This also raises some fun questions about directory structure. Ideally (maybe?) a lot of the QAQC code lives with the product itself. So maybe the QAQC app imports that from the products? (which would be a good argument to ✂️ the `db-` prefix)

apps/qaqc/entrypoint.sh
damonmcc: hmmm if not here than maybe running locally be the best next thing to add, since the only other way to see changes is deploying?

.github/workflows/qaqc_deploy.yml
```diff
@@ -4,26 +4,64 @@ on:
...
+            if [[ \$pid ]]; then
+              docker stop \"\$pid\"
+              docker rm \"\$pid\"
+            fi
+            docker run -p 8501:8501 -d --restart always\
+                -e AWS_S3_ENDPOINT=$AWS_S3_ENDPOINT\
+                -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY\
+                -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID\
+                -e SQL_ENGINE_EDM_DATA=$SQL_ENGINE_EDM_DATA\
+                $DOCKER_IMAGE_NAME"
```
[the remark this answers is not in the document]
alexrichey: @fvankrieken gotcha, good to hear. I still think that we probably want to investigate using Digital Ocean Apps. Just thinking that we'd get access to all sorts of nice stuff... zero-downtime deploys, alerts, automatic restarts if the app goes down, easy subdomain management, log tailing, configuration in the GUI rather than env vars (and maybe access to secrets via DO?) and a nice CLI that would probably do most of the hairy deployment stuff with a single command. (Just remembering my time with AWS Elastic Beanstalk, and assuming DO has something similar)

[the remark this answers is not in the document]
fvankrieken: Agreed. Still think I'm going to get this merged roughly as-is for now to do some of the work I had originally intended to do on the qaqc app, but will definitely come back to this, maybe next week

alexrichey (approved):
Good stuff! Plan sounds good to me

damonmcc (approved):

PR #121, opened 2023-08-04
check and format db-checkbook and db-template files

damonmcc:
for products that are being checked, [here's](https://github.com/NYCPlanning/data-engineering/actions/runs/5764031074/job/15627122402) what it looks like when python code fails the format check on a PR

fyi [documentation](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html) on `black`

### DIFF
--- a/products/db-checkbook/build_scripts/build.py
+++ b/products/db-checkbook/build_scripts/build.py
@@ -6,31 +6,36 @@ from sqlalchemy import create_engine, text
 
 from . import LIB_DIR, OUTPUT_DIR, SQL_QUERY_DIR
 
-DB_URL = 'sqlite:///checkbook.db'
+DB_URL = "sqlite:///checkbook.db"
 ENGINE = create_engine(DB_URL)
 
-def _read_all_cpdb_geoms(dir = LIB_DIR) -> list:
+
+def _read_all_cpdb_geoms(dir=LIB_DIR) -> list:
     """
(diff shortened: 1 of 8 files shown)


alexrichey:
I suppose we could implement this at the product level. However, I think it might be more useful to pipe in the list of changed files, regardless of which project they belong to. (ie make this a repo-level check)

products/db-checkbook/build_scripts/build.py
```diff
@@ -6,31 +6,36 @@
 
 from . import LIB_DIR, OUTPUT_DIR, SQL_QUERY_DIR
 
-DB_URL = 'sqlite:///checkbook.db'
+DB_URL = "sqlite:///checkbook.db"
```
athursland, line 9: this is purely a comment  but reading [the current Black style documentation](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) is so thorough it's wild - never really thought about the aesthetic merits of favoring double quotes over single quotes for strings lol

fvankrieken: If you want a rollercoaster of a discussion on it

https://github.com/psf/black/issues/118

damonmcc: definitely thorough, imagine trying to make so many devs happy with a common style and (pretty much) succeeding!

PR #122, opened 2023-08-04
check and format all python files

damonmcc:
per [initial failed check](https://github.com/NYCPlanning/data-engineering/actions/runs/5765130965/job/15630502827#step:4:2202), 43 files were reformatted locally in the devcontainer by running `black .`

this replaces #121 in order to cover all products and other python files (great idea @alexrichey). it also doesn't limit the check to only the files changed in a branch/PR to avoid complexity and lean into the auto-formatting of our files sooner rather than later


.github/workflows/test.yml
```diff
@@ -12,9 +12,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check_formatting:
+    name: Check formatting of files
+    runs-on: ubuntu-latest
+    container:
+      image: nycplanning/build-geosupport:latest
```
fvankrieken, line 19: use base? There's not a huge difference in size but can use a smaller image

Actually, if we just want a very small one that certainly will work https://hub.docker.com/r/pyfound/black/tags

fvankrieken: Looks like there's also an official gha
https://black.readthedocs.io/en/stable/integrations/github_actions.html

damonmcc: I'll switch to the base image. I think I prefer using our container + CLI command rather than a marketplace action. looks like in both cases you end up using CLI arguments and using the later seems to add a layer of separation/abstractions without much benefit

fvankrieken (approved):

.github/workflows/test.yml
fvankrieken: Also the argument (in our image vs black image) that we might add more to this job later. I still like the marketplace action, but lgtm

alexrichey: Yeah, as long as our base image is small enough to run quickly. I'm guessing the GHA would be much quicker

alexrichey (approved):
Nice. LGTM

Issue #112, opened 2023-08-02
more summary statistics in db-checkbook


damonmcc:
- currently we produce summary statistics of the db-checkbook output
- it'd be nice the have those same statistics on the CPDB source data

PR #127, opened 2023-08-07
db-checkbook: refine and export summary stats

damonmcc:
related to #112 

this is meant to refine the last build script (formerly `04_summarization.py`) to ensure it's called during the build action, creates and export relevant info to S3, and to tweak the formatting of printed messages

changes after PR review:
- better isolate the exporting and printing of summary stats
- replace repeated `print` statements with `report_summary_stat()` which standardizes and returns the formatted string to print
- export printed summary sats to a log file `build_summarization.log` using the `logging` package
- declare filenames in `__init__.py`
- build db-checkbook in PR tests

[summarization in run log](https://github.com/NYCPlanning/data-engineering/actions/runs/5801072514/job/15724535671?pr=127#step:6:9)

[summarization in run log before PR review](https://github.com/NYCPlanning/data-engineering/actions/runs/5787205980/job/15683620821#step:6:11)

exports in S3:
Screenshot 2023-08-08 at 2 47 34 PM


athursland (approved):
possible enhancement to this might be saving the summary statistics to a csv or text file so that we might easily visualize the outputs later, rather than just printing? but this is great, LGTM!

fvankrieken:
> possible enhancement to this might be saving the summary statistics to a csv or text file so that we might easily visualize the outputs later, rather than just printing? but this is great, LGTM!

It actually is, sort of hidden in there. I understand you were just moving code @damonmcc and not writing functionality but it might quickly be worth splitting out that logic slightly, the saving file vs extensive printing

That can also wait - some of these printing methods could go into dcpy if they're informative and generic enough

products/db-checkbook/build_scripts/summarization.py
```diff
@@ -0,0 +1,115 @@
...
+SUMMARY_STATS_FILENAME = "historical_spend_stats.csv"
+
+# format the display of large numbers in dataframes to use commas and 2 decimal places
+pd.options.display.float_format = "{:,.2f}".format
+
+
+def categorization_summarization_statistics(df: pd.DataFrame) -> None:
+    projects_with_geometry = df[df["has_geometry"] == True]
+
+    print("--- Value counts: 'final_category' ---")
```
alexrichey, line 16: Having so many print statements feels a little noisy. It might be nice to have just one function that collects all these summary stats into either a dictionary, e.g. 
```
{'final_category': df["final_category"].value_counts() ... }
```
or a list of tuples
```
[('final_category', df["final_category"].value_counts()), ...]
```
And then you can just print that out at the end?

products/db-checkbook/build_scripts/summarization.py
```diff
@@ -0,0 +1,171 @@
...
+
+from . import (
+    OUTPUT_DIR,
+    BUILD_OUTPUT_FILENAME,
+    SUMMARY_STATS_DESCRIBE_FILENAME,
+    SUMMARY_STATS_LOG_FILENAME,
+)
+
+
+logging.basicConfig(
```
alexrichey, line 13: dcpy?

products/db-checkbook/build_scripts/summarization.py
```diff
@@ -0,0 +1,171 @@
...
+    level=logging.INFO,
+    format="%(message)s",
+    handlers=[
+        logging.FileHandler(OUTPUT_DIR / SUMMARY_STATS_LOG_FILENAME),
+        logging.StreamHandler(),
+    ],
+)
+
+_percent_of_total_records = lambda selected_data, all_data: round(
```
alexrichey, line 22: Any reason to use lambdas here instead of def? [FYI](https://www.flake8rules.com/rules/E731.html

alexrichey: I'm not sure I'd want to enforce that yet before discussing the various rules. I do think it would be beneficial for folks to enable a PEP8 linter in their editors though!

alexrichey:
@damonmcc I think it's a great idea to upload logfiles for each of our builds. On the flip side, in this case where we just have some QA outputs, that does make them unparseable. (Ie. we can't load these QA outputs into `csv`s and compare with previous versions.) Would it make more sense to skip the logging, and just write a csv with the summary stats?

damonmcc:
> @damonmcc I think it's a great idea to upload logfiles for each of our builds. On the flip side, in this case where we just have some QA outputs, that does make them unparseable. (Ie. we can't load these QA outputs into `csv`s and compare with previous versions.) Would it make more sense to skip the logging, and just write a csv with the summary stats?

agree that text isn't the best way to store structured/relational data, but some thoughts:
- to QA, we can load the data and compare it with previous versions rather than load build-time artifacts
- **generating QA outputs during a build** is the approach in many products right now (e.g. PLUTO) but it seems to have significant limitations compared to generating QA outputs **at the time of QA**:
  - limited to the checks defined at build time, the logic behind previous checks are virtually inaccessible, locks future QA into the structure/choices of earlier QA just to ensure you can append to a csv
- this feels out of scope for this PR, these changes we just intended to improve what Dea started. if we want to spec out useful QA data structures, we should probably take into consideration other data products and their QAQC in a more thoughtful way than in this PR?

alexrichey (approved):
Cool - Looks good. I do think it would be good to remove all the logging, and just dump a csv (which you could also print). Code would be cleaner, and you'd have a usable artifact at the end. But not sure it really warrants spending more time.

Issue #47, opened 2023-07-14
PLUTO inputs - more graceful handling of missing source data


fvankrieken:
Two github actions for pluto - PTS and CAMA - grab files via ssh. The files at the source are put there manually, so these cron jobs fail. I would say we should
- remove schedule trigger, only run these manually (or some sort of other trigger but I don't know what else we would do currently)
- add more informative logging/error messages when run fails because of missing source data

@AmandaDoyle just want to ping you for thoughts here

fvankrieken:
I think rather than run manually, for now it would be good to keep on a schedule, but have a check that either cancels the job if file is not found, or simply doesn't run the final steps/jobs (I like the explicit cancel if possible rather than skipping, but I don't know if that's possible)

fvankrieken:
@damonmcc fyi

fvankrieken:
Creating a bit of a journal/todo list from poking around with @DeaBardhoshi 
- @DeaBardhoshi has grabbed a marketplace action that logs existence of file into github step outputs
- bit of work needed right now just to get jobs running, somewhat external from this task
  - [ ] I will get secrets for repo where they need to be so that job can run
    - at that point, we can re-run the job to see what an "expected" failure
  - [ ] @DeaBardhoshi to expanding this logic
    - [ ] work around the `pluto` ssh logic to check file existence on ssh server
    - [ ] if statements in all steps after checking for file existence, not running if file isn't found

fvankrieken:
So it might be helpful to get the ssh logic working on your local computer, that will make testing easier. I've sent you info around ssh host etc. 

This [`ssh_ls`](https://github.com/NYCPlanning/data-engineering/blob/main/products/db-pluto/pluto_build/bash/ssh.sh#L2C3-L2C3) function will be very helpful to getting some code that works. I would avoid using `-q` option for any ssh/sftp command you use, as it will not display errors/warnings, and right now in your actions, it would be good to know more specifically what the output of your commands are and having an example of successfully finding the file on the server in addition to "successfully" failing to find it.

Luckily, PTS should find the file successfully right now - there is a file that exists right now on the server, `Prod_FromDOF/PTS_Propmast.gz`, that you should be able to properly log that it exists in the github action.

PR #129, opened 2023-08-07
Pluto Input Error Logging

DeaBardhoshi:
## Changes:
- Added a file checker to pluto_input_pts and pluto_input_cama to output whether a file exists or not for each file

Closes Issue #47 

Relevant Runs:
- https://github.com/NYCPlanning/data-engineering/actions/runs/5797906374/job/15714498759
- https://github.com/NYCPlanning/data-engineering/actions/runs/5797908542/job/15714506358


fvankrieken:
Can you add links to workflow runs showing desired behavior?

fvankrieken:
So the pts file actually exists currently, so that one should be passing. I'll add some notes to the issue

fvankrieken (dismissed):
I've left comments both here and on #47, happy to chat if more direction is needed

fvankrieken:
[Cama run cancelling due to missing file](https://github.com/NYCPlanning/data-engineering/actions/runs/5871053497)

[PTS run succeeding](https://github.com/NYCPlanning/data-engineering/actions/runs/5871054516)

.github/workflows/pluto_input_cama.yml
```diff
@@ -41,12 +41,33 @@ jobs:
...
+
+      - name: Setup ssh
+        run: ./pluto ssh setup
+
+      - name: Download file
+        id: download
+        run: |
+          (./pluto ssh download cama) && true
+          output=$?
+          if [ $output -ne 0 ] && [ $output -ne 2 ]; then
```
fvankrieken, line 53: exit code 2 -> file not found error.

Anything else, should actually exit

.github/workflows/pluto_input_cama.yml
```diff
@@ -41,12 +41,33 @@ jobs:
...
-        
-      - name: install ssh key
+
+      - name: Setup ssh
+        run: ./pluto ssh setup
+
+      - name: Download file
+        id: download
+        run: |
+          (./pluto ssh download cama) && true
```
fvankrieken, line 51: The `&& true` is a bit hacky, but it makes bash not exit - within pluto cli we use `set -e` so just running this command causes the workflow run to fail

damonmcc (approved):
really nice having the grey cancel icon rather than the red fail

just a question for later: would it be worth combining the workflows into one? if they each take a while to download though, it might not be worth it

fvankrieken:
Then the cancelling logic doesn't make senes - PTS is always there, CAMA only periodically

PR #134, opened 2023-08-07
use a github secret in db-template tests

damonmcc:
this was originally an exercise to troubleshoot other dev work, but this seems informative enough to squash & merge

## motivation
- when an action is triggered via `workfloa_call`, how can it be passed a github secret to use it as an environment variable?
- called actions don't have access to secrets by default?

## series of changes and results of tests in this branch
1. [a KeyError](https://github.com/NYCPlanning/data-engineering/actions/runs/5789853710/job/15691806582#step:5:33) when using an undeclared envar
2. [works](https://github.com/NYCPlanning/data-engineering/actions/runs/5789903190/job/15691952005#step:5:33) when the envar is hardcoded
3. [envar is blank](https://github.com/NYCPlanning/data-engineering/actions/runs/5789953209/job/15692094240#step:5:70) when trying to use a github secret
4. [works](https://github.com/NYCPlanning/data-engineering/actions/runs/5789987702/job/15692186716#step:5:34) when secrets are inherited from the caller action

### DIFF
--- a/.github/workflows/template_test.yml
+++ b/.github/workflows/template_test.yml
@@ -25,6 +25,8 @@ jobs:
   branch_unit_tests:
     name: Unit tests
     runs-on: ubuntu-22.04
+    env:
+      TOY_SECRET: toy_secret_value
     container:
       image: nycplanning/build-geosupport:latest
     defaults:
--- a/.github/workflows/template_test.yml
+++ b/.github/workflows/template_test.yml
@@ -26,7 +26,7 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-22.04
     env:
-      TOY_SECRET: toy_secret_value
+      TOY_SECRET: ${{ secrets.TOY_SECRET }}
     container:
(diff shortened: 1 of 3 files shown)


damonmcc:
@fvankrieken

fvankrieken:
A couple other fun notes
- called actions appear to only be able to inherit repo secrets, not org secrets (via `secrets: inherit`, I'm guessing you could pass more explicitly). See error [here](https://github.com/NYCPlanning/data-engineering/actions/runs/5790083946) 
- In above case, trying to access a secret which doesn't exist results in error, but when trying to access a repo secret that's not passed down via inheritence, it just silently becomes an empty string

alexrichey:
@damonmcc Thanks for the research. I'm curious what the issue was that prompted this?

fvankrieken:
My struggles yesterday - was having some issues due to using workflow call and not inheriting secrets

fvankrieken (approved):

PR #136, opened 2023-08-08
Update path to bash utils

fvankrieken:


products/db-pluto/pluto_build/bash/config.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 FILE_DIR=$(dirname "$(readlink -f "$0")")
```
fvankrieken, line 3: If we want this to refer to the file itself always (and not where this is being sourced from), we could do $BASH_SOURCE[0]

alexrichey (approved):

products/db-pluto/pluto_build/bash/config.sh
damonmcc: so here `FILE_DIR` is whatever directory the script that's sourcing `pluto_build/bash/config.sh` is in? so it could be `/pluto_build/bash`, `/pluto_build`, or somewhere else?

since we're only using it to navigate up to the repo's root, would the working directory would be a more stable path to use? in the build workflow, I think it'd always be `products/db-pluto/pluto_build`

Issue #125, opened 2023-08-04
KPDB failing to use `dcp_projects` source data


damonmcc:

damonmcc:
update:
- KPDB is successfully using ZAP source data
- currently build fails during export during SQL to Shapefile step, due to inability to coerce `GEOMETRYCOLLECTION` values to `MULTIPOLYGON`

PR #146, opened 2023-08-09
use shared utils in KPDB

### LINKED ISSUE #5: Abstract out import logic, bash utils

damonmcc:
resolves #5 

KPDB is the last data product to use shared bash utils. after switching to shared utils and deleting unused bash functions, `db-knownprojects/bash/config.sh` was empty so deleted it!

[KPDB build](https://github.com/NYCPlanning/data-engineering/actions/runs/5812720611/job/15758716692#step:5:82) with successful dataloading (build fails for other reasons to be fixed in #125)


alexrichey (approved):

products/db-knownprojects/bash/config.sh
```diff
@@ -32,26 +32,6 @@ function max_bg_procs {
...
-  name=$1
-  version=${2:-latest}
-  version=$(curl -s https://nyc3.digitaloceanspaces.com/edm-recipes/datasets/$name/$version/config.json | jq -r '.dataset.version')
-  echo "$name version: $version"
-  curl -O https://nyc3.digitaloceanspaces.com/edm-recipes/datasets/$name/$version/$name.sql
-  run_sql_file $name.sql
-  rm $name.sql
-}
-
 function CSV_export {
```
fvankrieken, line 35: These exports should also be moved over! This csv export seems to be same as in shared utils. Same goes for shp export. Not sure at first glance about archive.

Above, urlparse and max_bg_processes can also be deleted

damonmcc: thanks! turns out the whole config file can go!

products/db-knownprojects/bash/04_export.sh
```diff
@@ -28,25 +28,25 @@ mkdir -p output
...
+        csv_export review_dob &
+        csv_export corrections_applied &
+        csv_export corrections_not_applied &
+        csv_export corrections_zap &
+        csv_export corrections_dob_match &
+        csv_export corrections_project &
+        csv_export corrections_main
         wait
         
         Compress combined.csv
```
fvankrieken, line 52: This `Compress` function is coming from deleted `config.sh`

fvankrieken (approved):

Issue #143, opened 2023-07-25
weekly ZAP CRM export is failing sporadically


damonmcc:
[link to failed run](https://github.com/NYCPlanning/db-zap-opendata/actions/runs/5639290010)

`dcp_projects` appears to [fail](https://github.com/NYCPlanning/db-zap-opendata/actions/runs/5741983217/job/15563168052#step:5:45) with `Exception: data has current_milestone of 1cb27780-be0d-ed11-82e3-001dd8068d18 but expanded of ***`

not sure why `dcp_projectmilestones` fails. usually if `dcp_projects` fails, others do

damonmcc:
haven't seen this happen in the many runs during recent dev, but can't be sure it was fixed

gonna lower the priority of this issue and move it to Backlog

damonmcc:
closing since it hasn't happened in a while. may have been caused by trying to append to a sql table from a previous run rather than deleting and recreating it

PR #147, opened 2023-08-09
ensure zap workflows work

### LINKED ISSUE #152: ZAP project-bbl record count anomaly
Over a million records are included in the zapprojectbbl.csv output from June 6th, 2023 as well as July 3, 2023. This is a **significant** record count increase from previous opendata cycles (5/8/23 version had 66,869 records). 

GIS suspects this likely is a filtering issue that is leading to internal records being included in the final outputs. 

@damonmcc @AmandaDoyle @fvankrieken

damonmcc:
resolves #152 and related to #143

## cause of record count anomaly issue
- there appeared to be duplicates in the raw Project and Project BBL data pulled from the CRM (e.g. Project ID `2023K0228` and BBL `3050637501`)
- these duplicates were caused by the appending of CRM records to existing SQL tables
  - each dataset is downloaded as multiple json files
  - during the `combine` stage of the `Runner`, all downloaded json files [are written to a sql table](https://github.com/NYCPlanning/data-engineering/blob/main/products/db-zap-opendata/src/runner.py#L133-L139) using `if_exists="append"`
  - all sql queries in this data product use [a DB schema arg](https://github.com/NYCPlanning/data-engineering/blob/main/products/db-zap-opendata/src/runner.py#L38) passed to the `Runner` so that data is isolated by build environment in a persistent database `edm-zap`
    - examples of DB schemas are `weekly_export_schedule_main`, `weekly_export_workflow_dispatch_zap_runnable`, `test_data_actual_pr_147`
- **[we weren't deleting the table](https://github.com/NYCPlanning/data-engineering/blob/main/products/db-zap-opendata/src/runner.py#L107-L117) if it exists before writing the first json file to sql**

## changes
- add ZAP product to centralized test workflow
- rename and fix secrets in ZAP workflows
- ensure an existing sql table is deleted before ingesting new CRM data
- remove a redundant `INNER JOIN` from an early intermittent `dcp_projectbbls` table (it's used to limit BBL records to publicly visible projects, doesn't have to happen twice)
- add an assertion to an integration test to ensure a single project and a BBL associated with it have the expected number of rows in their respective tables after `Runner.combine()`

## run notes
- most recent CRM export run on this branch: [link](https://github.com/NYCPlanning/data-engineering/actions/runs/5870648856)
- The file sizes from the 8/15 build are very close to those from the valid 7/17 build and have the following row counts:
  - Projects file: 32,428
  - Project BBLs file: 67,289

## dev notes
- inspecting integration test results helped identify the problem. the number for rows for a single project was increasing after each test (e.g. [5 rows](https://github.com/NYCPlanning/data-engineering/actions/runs/5869800018/job/15915549659#step:5:266) became [6 rows](https://github.com/NYCPlanning/data-engineering/actions/runs/5869866740/job/15915739481#step:5:266))
- ZAP devcontainer/image requires google cloud services so I held off on switching to our standard one. but we don't use the bigquery tables at all so I commented out all uses of it and we can revisit this

### DIFF
--- a/.github/workflows/zap_build_mapzap.yml
+++ b/.github/workflows/zap_build_mapzap.yml
@@ -1,4 +1,4 @@
-name: Build MapZAP
+name: ZAP - MapZAP (Placeholder)
 
 on:
   workflow_dispatch:
--- a/.github/workflows/zap_test.yml
+++ b/.github/workflows/zap_test.yml
@@ -1,25 +1,25 @@
-name: Branch tests
+name: ZAP - Tests
 run-name: Tests (${{ github.event_name }})
 
 on: 
-  workflow_dispatch:
-
-concurrency:
-  group: ${{ github.head_ref || github.run_id }}
-  cancel-in-progress: true
+  workflow_call:
 
 env:
-  ZAP_DOMAIN: ${{ secrets.ZAP_DOMAIN }}
-  TENANT_ID: ${{ secrets.TENANT_ID }}
-  CLIENT_ID: ${{ secrets.CLIENT_ID }}
-  SECRET: ${{ secrets.SECRET }}
-  ZAP_ENGINE: ${{ secrets.ZAP_ENGINE }}
+  ZAP_DOMAIN: ${{ secrets.ZAP_CRM_DOMAIN }}
+  TENANT_ID: ${{ secrets.ZAP_CRM_TENANT_ID }}
+  CLIENT_ID: ${{ secrets.ZAP_CRM_CLIENT_ID }}
+  SECRET: ${{ secrets.ZAP_CRM_SECRET }}
+  ZAP_ENGINE: ${{ secrets.SQL_ENGINE_EDM_DATA_SERVER }}/edm-zap
   TEST_SCHEMA_SUFFIX: pr_${{ github.event.pull_request.number || 'workflow_dispatch' }}
 
 jobs:
   branch_unit_tests:
     name: Unit tests
     runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: products/db-zap-opendata
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3
@@ -37,6 +37,10 @@ jobs:
   branch_integration_tests:
     name: Integration tests
     runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: products/db-zap-opendata
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v3
@@ -65,6 +69,7 @@ jobs:
       - name: Build and run dev container task
         uses: devcontainers/ci@v0.3
         with:
+          subFolder: products/db-zap-opendata
           runCmd: |
             echo "Dev Container is in CI? $CI"
             gcloud version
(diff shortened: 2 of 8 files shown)


.github/workflows/zap_build_mapzap.yml
```diff
@@ -1,4 +1,4 @@
-name: Build MapZAP
+name: ZAP - MapZAP (Placeholder)
```
damonmcc, line 1: this workflow should've been labeled as a placeholder already. it lives in bigquery and the exploratory builds of it were first run manually via sequential queries and then via dbt on the command line

.github/workflows/zap_test.yml
```diff
@@ -65,6 +69,7 @@ jobs:
       - name: Build and run dev container task
         uses: devcontainers/ci@v0.3
         with:
+          subFolder: products/db-zap-opendata
```
fvankrieken, line 72: Is this test still needed if zap doesn't have its own dev container at this point?

products/db-zap-opendata/src/visible_projects.py
```diff
@@ -101,8 +101,7 @@ def make_staging_table(sql_engine, dataset_name) -> None:
                     %(source_table_name)s.dcp_userinputborough as unverified_borough,
                     %(source_table_name)s.dcp_userinputblock as unverified_block,
                     %(source_table_name)s.dcp_userinputlot as unverified_lot
-             from %(source_table_name)s INNER JOIN dcp_projects_recoded
-            on SUBSTRING(%(source_table_name)s.dcp_name, 0,10) = dcp_projects_recoded.project_id);
+            FROM %(source_table_name)s);
```
fvankrieken, line 104: Did you do any testing on the results of this? Main case I could see is that this is a specific filtering of the source data to only dcp recoded projects

damonmcc: the final outputs (the "visible" versions of Projects and BBLs tables) are as expected based on row counts and file sizes

this filtering was redundant since [the same filtering](https://github.com/NYCPlanning/data-engineering/blob/0232a12a6f414bb321e47b4f41381b8f9849d091/products/db-zap-opendata/src/visible_projects.py#L188-L189) takes place when it's actually needed: when BBL records have to be limited to DCP projects with "General Public" visibility

fvankrieken: Gotcha. I'm a little more confused after looking through that whole file more, though for reasons unrelated to this PR, just the general logic in here. 

Are the `_staging` tables used at all in export? Like for our own use? Or really just temporary, or what

.github/workflows/zap_test.yml
damonmcc: needed, no. cool, yes.

[the devcontainer files](https://github.com/NYCPlanning/data-engineering/tree/main/products/db-zap-opendata/.devcontainer) are still here and since I only disabled the BigQuery stages of the ZAP export and MapZAP hasn't been converted from BigQuery to Postgres yet, it didn't seem worth deleting everything yet

fvankrieken: I hope we can convince @AmandaDoyle that coolness of codebase should be one of our KPIs. Maybe the only one!

products/db-zap-opendata/src/visible_projects.py
damonmcc: `_staging` tables aren't exported. they're used to select and rename the columns we want for the subsequent `_recoded` and `_visible` tables

it goes `_crm` -> `_staing` -> `_recoded` -> `_visible`

[previously](https://github.com/NYCPlanning/db-zap-opendata/pull/45/files#diff-e6905d6f9c665e3efc78cde7015333a2048c899407361715dfbd7a40cdfd015aL28), the logic and columns were more confusing so I added the use of intermediate tables to try and better separate stages of the export

fvankrieken (approved):

fvankrieken:
Oops - meant to ask for this in this PR, but would you mind consolidating requirements files, or at least bumping zap's requirements so that dependabot is happy?

PR #149, opened 2023-08-09
Enable Pluto Minor Builds

alexrichey:
These changes gracefully enable minor version builds, that carry over versioning from a parent build. It rewrites the dataloading step to use a recipe file.


products/db-pluto/build-minor.yml
```diff
@@ -0,0 +1,23 @@
+name: Pluto Minor
+base_build: build.yml
+version_type: minor
+product: db-pluto
+recipe:
```
fvankrieken, line 5: Now THIS is a recipe

fvankrieken:
> Eventually I think this functionality should include a full plan step, where `latest` is resolved to an actual version. Then we can replace the `source_versions` postgres table with this functionality.

Very much agree with this - also this is a very basic thing, but just excited even about just actually defining inputs for all data products - if you notice in the [qaqc code](https://github.com/NYCPlanning/data-engineering/blob/main/apps/qaqc/src/source_report_utils.py#L22) there's a function to figure this out slightly more implicitly from the build outputs.

Honestly, no notes for now! I'll keep mulling things over, but this is a really awesome start

dcpy/connectors/edm/recipes.py
```diff
@@ -70,7 +70,7 @@ def import_recipe(
     recipe_version = config["dataset"]["version"]
     sql_script_path = fetch_sql(recipe_name, recipe_version, local_library_dir)
 
-    postgres.exec_file_via_shell(postgres.BUILD_ENGINE_RAW, sql_script_path)
+    postgres.execute_file_via_shell(postgres.BUILD_ENGINE_RAW, sql_script_path)
```
fvankrieken, line 73: Oops. Also, if you saw, I went ahead and just deleted the part of this function doing anything with the source versions table - figured since this isn't even being used yet, we won't plan around something we're going to phase out

dcpy/connectors/edm/recipes.py
```diff
@@ -77,3 +83,73 @@ def import_recipe(
...
+        ).to_dict()["version"]
+
+    for rec in recipe["inputs"]:
+        if "version" not in rec:
+            if recipe["missing_versions_strategy"] == "use_latest_release":
+                rec["version"] = previous_versions[rec["name"]]
+            else:
+                rec["version"] = "latest"
+
+    if fully_resolve_versions:
```
fvankrieken, line 141: Is there a reason we wouldn't want to fully resolve versions?

dcpy/connectors/edm/recipes.py
```diff
@@ -77,3 +118,171 @@ def import_recipe(
...
+        if import_table_name is not None:
+            logger.info(f"Renaming table {recipe_name} to {import_table_name}")
+            con.execute(
+                text(f"ALTER TABLE {recipe_name} RENAME TO {import_table_name};")
+            )
+            con.commit()
+
+
+def _recipe_datasets(recipe):
+    return recipe.get("inputs", {}).get("datasets", []) or []
```
fvankrieken, line 131: nit, but is the `or` redundant with the two `get`s?

products/pluto/pluto_build/00_setup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+source ./bash/config.sh
+set_error_traps
+
+# DROP all tables
+if [[ $1 == "drop" ]]; then
```
fvankrieken, line 6: since this file only drops all tables, should we just get rid of this if clause

fvankrieken:
And I would say you don't need to get checked-in recipes into csvs as part of this PR

fvankrieken (approved):
This looks great! Excited to start using it

dcpy/connectors/edm/recipes.py
alexrichey: @fvankrieken somewhat frustratingly not, in the case when the value is set to None (as opposed to the key being missing)
```
recipe = {"inputs": {"datasets": None}}
recipe.get("inputs", {}).get("datasets", []) # -> None
```

fvankrieken: Ohh, gotcha

products/pluto/pluto_build/00_setup.sh
alexrichey: @fvankrieken Good call! Updated

dcpy/connectors/edm/publishing.py
```diff
@@ -14,38 +14,42 @@
 ## TODO - these functions need some cleaning
 ## ideally they don't need to called individually, but we either have consistent folder structure
 ## Or have metadata in json/etc about structure of outputs per dataset
-def get_dataset_branch_path(dataset: str, branch: str, version: str):
-    return f"{BASE_URL}/{dataset}/{branch}/{version}/output"
-
-
-def get_dataset_path(dataset: str, version: str):
-    return f"{BASE_URL}/{dataset}/{version}/output"
+def get_dataset_path(dataset: str, *, branch: str, version="latest", published=False):
```
damonmcc, line 17: nit: should the `*` be after the last param without a default value? so
```python
def get_dataset_path(dataset: str, branch: str, *, version="latest", published=False):
````

File: dcpy/connectors/edm/recipes.py [hunk 7]
```diff
@@ -1,15 +1,59 @@
...
+
 from dcpy import DCPY_ROOT_PATH
 from dcpy.utils import s3
+from dcpy.utils.logging import logger
 from dcpy.utils import postgres
+from dcpy.utils import versions
+
+# In order to keep things sane, maybe we should allow recipes to import
+# publishing but not the other way around?
+from . import publishing
```
damonmcc, line 22: I like this for now, but maybe long-term we could move all `publishing` functions that are used here to something like `dcpy.utils.s3`? and same for anything `publishing` might import from `recipes`?

dcpy/utils/logging.py
damonmcc, line 1: love it, not used anywhere yet though?

dcpy/connectors/edm/publishing.py
alexrichey: That would have the effect of making it a required positional variable, which we probably don't want to do. All the product pathing is in flux atm, so maybe we revisit once all releases are in the `publish` folder

dcpy/connectors/edm/recipes.py
alexrichey: I actually think (and @fvankrieken might also have thoughts) that there's a lot in `recipes` that isn't strictly about recipes, e.g. importing stuff into a database. I think a lot of this build stuff really belongs elsewhere. Maybe something like `dcpy.products` which would be responsible for the steps to build a product. Thoughts? @damonmcc

dcpy/utils/logging.py
damonmcc: oops, thanks!

products/pluto/recipe.yml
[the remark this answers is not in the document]
alexrichey: that's a really good question. Technically not, unless you're generating a bumped version. Ie going from 23v2.1 -> 23v3. Might be nice to nail that down in Cerberus.

dcpy/connectors/edm/recipes.py
damonmcc: true, or maybe `dcpy.builds`

damonmcc (approved):

dcpy/connectors/edm/recipes.py
fvankrieken: Yeah eventually that might make sense. For now I'm fine with it in `recipes` and keeping this file as related to everything starting a build, including the recipes datasets as well as grabbing build metadata

fvankrieken (approved):

PR #150, opened 2023-08-10
fix zap tests failing on all PRs

damonmcc:
follow up to #145

ZAP tests [are failing](https://github.com/NYCPlanning/data-engineering/actions/runs/5814017939) on all PRs because they haven't been changed yet to run in the monorepo

these tests shouldn't be running yet and, when they do, should only be triggered by the centralized test.yml workflow


alexrichey (approved):

damonmcc:
this fix is moot since we merged #126 

no urgency to change `workflow_dispatch` to `workflow_call` on `main`

PR #154, opened 2023-08-11
PFF tweaks for decennial manual update

fvankrieken:
Continuation of #67. Per convo with EM, some changes around schema. 
- don't use metadata.json as input. These files were written as part of the api-pulling code. When they were written, the decennial census had not released as much data as they did for the latest update (far fewer columns before - 20ish vs 200ish). 
  - don't filter columns based on inclusion in old metadata files
  - don't rename new column headers to align with old, but use new instead
- generate new metadata files
  - these are used by OSE/AE in their ingestion. From glancing at [their code](https://github.com/search?q=repo%3ANYCPlanning%2Flabs-factfinder-api+-%3E%3E+%27base_variable%27&amp%3Btype=code&type=code), it seems like they only need `pff_variable`, `base_variable`, `domain`, and `category`. 
 
Need to think a little bit about long-term health here. It's a little odd to both have the metadata be defined by us but also semi-auto-generated from excels from housing. It also might make sense to export these metadata files to DO and not keep in the repo at the moment, since they really are outputs now rather than metadata inputs - also, currently I believe OSE/AE is grabbing them from the repo. if we're just generating them, we can export them to DO instead and not need to commit changes as part of a build. For now I'll do both (commit changes and export to DO), but worth revisiting.

### DIFF
--- a/dcpy/utils/json.py
+++ b/dcpy/utils/json.py
@@ -1,0 +1,8 @@
+import json
+import pandas as pd
+
+
+def df_to_json(df: pd.DataFrame):
+    """Converts pandas dataframe to pretty json
+    Prints as json array rather than standard to_json"""
+    return json.dumps(json.loads(df.to_json(orient="records")), indent=4)
--- a/products/db-factfinder/factfinder/data/decennial/2010/metadata.json
+++ b/products/db-factfinder/factfinder/data/decennial/2010/metadata.json
@@ -1,204 +1,827 @@
 [
     {
-        "pff_variable": "decennial_pop",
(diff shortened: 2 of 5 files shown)


damonmcc (approved):
looks like [this run](https://github.com/NYCPlanning/data-engineering/actions/runs/5837381826) produced the expected outputs?

fvankrieken:
> looks like [this run](https://github.com/NYCPlanning/data-engineering/actions/runs/5837381826) produced the expected outputs?

Oops - yup

PR #156, opened 2023-08-14
Continue moving qaqc digital ocean utils to dcpy

fvankrieken:


apps/qaqc/src/cpdb/helpers.py
```diff
@@ -48,22 +44,21 @@ def get_data(branch, previous_branch, previous_version) -> dict:
...
-    client = DigitalOceanClient(bucket_name=BUCKET_NAME, repo_name=REPO_NAME)
-
     for t in tables["analysis"]:
-        rv[t] = client.csv_from_DO(url=construct_url(branch, t, sub_folder="analysis/"))
-        rv["pre_" + t] = client.csv_from_DO(
-            url=construct_url(
-                previous_branch, t, previous_version, sub_folder="analysis/"
-            )
+        rv[t] = publishing.read_csv(
+            DATASET, f"{branch}/latest/output", f"analysis/{t}.csv"
```
fvankrieken, line 49: maybe for clarity with how this function is being used now, `version` arg should be called `output` since really it makes most sense to be pointing to output folder.

dcpy/connectors/edm/publishing.py
```diff
@@ -97,3 +99,26 @@ def upload(
         if latest:
             ## much faster than uploading again
             s3.copy_file(BUCKET, str(key), str(prefix / "latest" / output.name), acl)
+
+
+def read_csv(dataset, version, filepath, **kwargs):
```
fvankrieken, line 104: I'm calling this middle arg "version" but it's blurring some lines - really, right now it's a path to an output folder. Longer term, we need to agree a bit about standardizing how we're outputting data and what a "version" vs a "build" is (dates vs branches etc) but for now this seemed relatively fine

dcpy/connectors/edm/publishing.py
```diff
@@ -97,3 +99,26 @@ def upload(
...
+
+
+def read_shapefile(dataset, version, filepath):
+    with s3.get_file_as_stream(BUCKET, f"{dataset}/{version}/{filepath}") as stream:
+        gdf = gpd.read_file(stream)
+    return gdf
+
+
+def get_zip(dataset, version, filepath):
+    stream = s3.get_file_as_stream(BUCKET, f"{dataset}/{version}/{filepath}")
```
fvankrieken, line 122: No `with` clause here because a zipfile requires stream to remain open.

dcpy/connectors/edm/publishing.py
damonmcc: I like the choice of "version", it's hard to make a function & it's args accurate for all products before we standardize folder structure

damonmcc (approved):
love it! this is all about cleanup after merging repos so I don't wanna be picky about the architecture of the QAQC pages or anything like that

PR #159, opened 2023-08-15
`db-checkbook` - layer in Parks Properties as source of geometries

### LINKED ISSUE #113: db-checkbook - layer in parks properties dataset as another source for geometries
- currently CPDB is the only source of geometries in db-checkbook
- fuzzy string matching to other sources (like FacDB) would likely increase the number of capital projects with a geometry

athursland:
Resolves #113 

### changes: 
- Wrote a SQL query in `db-checkbook/sql` to join MultiPolygons from Parks Properties to capital projects in CSDB based on partial string matches between `signname` and `eapply` column values in Parks Properties and `budget_code` and `contract_purpose` in CSDB
- Added fn to `dataloading.py` for reading in Parks Properties source data from s3
- Added fn to `build.py` for calling parks SQL query 
- Slight enhancements to `build.py` like switching from using a SQLite engine to a Postgres engine to align with the other data products, and checking for duplicate agencies while aggregating in `_group_checkbook()`
- Altered ordering of SQL category assignment queries in `categorization.sql` to prioritize assigning parks to fixed assets. The ordering was incorrect before–we cascade down the three categories in order of priority `['Fixed Asset', 'ITT, Vehicles, and Equipment', 'Lump Sum']` but the query for assigning parks to 'Fixed Asset' came last for some reason. 
- added `geoalchemy2` to `requirements.in`, recompiled python package installation + rebuilt docker image (`geoalchemy2` is necessary support to `sqlalchemy` for writing GeoDataFrames to SQL and back while preserving geometries) 
- Defined env in build and test workflows, added postgis service to both workflows

### outcomes:
- Added 322 geometries to CSDB, so % of projects mapped to geometries: 23.1 --> 25.1
- number of projects mapped to geometries: 3,867 -> 4,189


products/db-checkbook/parks/Parks_Properties.csv
athursland, line 1: going to move this to .library before submitting a final PR

fvankrieken: Let's make sure we squash this pr when merging then

products/db-checkbook/sql/parks.sql
```diff
@@ -0,0 +1,40 @@
...
+            upper(a.contract_purpose) LIKE '%' || upper(b.NAME311) || '%' OR
+            upper(a.contract_purpose) LIKE '%' || upper(b.SIGNNAME) || '%'
+        )
+    )
+
+UPDATE csdb 
+SET 
+    geometry = master.multipolygon,
+    has_geometry = 1,
+    final_category = 'Fixed Asset'
```
fvankrieken, line 37: Is this a valid assignment?

products/db-checkbook/sql/parks.sql
```diff
@@ -0,0 +1,40 @@
...
+        upper(a.agency) LIKE '%DPR%')
+        AND a.geometry IS NULL
+        AND (
+            upper(a.budget_code) LIKE '%' || upper(b.EAPPLY) || '%' OR
+            upper(a.budget_code) LIKE '%' || upper(b.ADDRESS) || '%' OR
+            upper(a.budget_code) LIKE '%' || upper(b.NAME311) || '%' OR
+            upper(a.budget_code) LIKE '%' || upper(b.SIGNNAME) || '%' OR
+            upper(a.contract_purpose) LIKE '%' || upper(b.EAPPLY) || '%' OR 
+            upper(a.contract_purpose) LIKE '%' || upper(b.ADDRESS) || '%' OR
+            upper(a.contract_purpose) LIKE '%' || upper(b.NAME311) || '%' OR
```
fvankrieken, line 28: I'm not sure we need these other fields - from what I can tell, EAPPLY and NAME311 seem like duplicates of SIGNNAME (other than a case I saw where EAPPLY is blank). We also definitely need the 
```sql
WHERE
      upper(b.signname) <> 'PARK' AND
      upper(b.signname) <> 'LOT' AND
      upper(b.signname) <> 'GARDEN' AND
      upper(b.signname) <> 'TRIANGLE' AND
      upper(b.signname) <> 'SITTING AREA'
```

from looking at the data. Multiple "Park"s and "Garden"s which would be matching on to tons of projects that they shouldn't

image

fvankrieken:
Other thing to work on is code to pull in csv from recipes. Check out dcpy.connectors.edm.recipes and see if something there works for you, or if not, go ahead and add it there!

.github/workflows/checkbook_build.yml
```diff
@@ -8,6 +8,7 @@ on:
         required: true
 
 env:
+  BUILD_ENGINE: postgresql://postgres:postgres@localhost:5432/postgres
```
fvankrieken, line 11: You're going to run into one more issue here - these jobs/docker images don't have a running database by default. Again, go ahead and look to other workflows as examples, and ping me if you have any questions!

fvankrieken (approved):

PR #162, opened 2023-08-16
Add ztl to main build workflow

### LINKED ISSUE #43: ZTL workflow maintenance

fvankrieken:
Closes #43

### DIFF
--- a/.github/workflows/zoningtaxlots_build.yml
+++ b/.github/workflows/zoningtaxlots_build.yml
@@ -10,13 +10,15 @@ on:
 jobs:
   build:
     name: Building ...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     defaults:
       run:
         shell: bash
         working-directory: products/db-zoningtaxlots
+    container:
+      image: nycplanning/build-base:latest
     services:
-      db:
+      postgis:
         image: postgis/postgis:14-3.3-alpine
         env:
           POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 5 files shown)


.github/workflows/zoningtaxlots_build.yml
damonmcc, line 1: part of the linked issue is to have this use the build docker image, I think by adding `container: image: nycplanning/build-base:latest` and dropping the `install dependencies` step

damonmcc:
just made #163 to fix that template unit test and make it more stable

.github/workflows/zoningtaxlots_build.yml
fvankrieken: Oh oops - thank you

fvankrieken:
Successful build (minus upload)
https://github.com/NYCPlanning/data-engineering/actions/runs/5891162968

.github/workflows/zoningtaxlots_build.yml
```diff
@@ -23,30 +29,26 @@ jobs:
...
+        run: ./bash/docker_container_setup.sh
+        
       - name: 1. dataloading ...
         run: ./ztl.sh dataloading
 
       - name: 2. build ...
         run: ./ztl.sh build
 
-      - name: 3. qaqc ...
+      - name: 3. qaqc, export, and upload ...
```
fvankrieken, line 52: Would have split these into multiple bash scripts but for ztl they're kind of intertwined

PR #166, opened 2023-08-17
Checkbook qaqc map

DeaBardhoshi:
## Changes:

Added a streamlit-folium map showing the geometries (both points and polygons) of the CSDB projects we mapped.

Screen Shot 2023-08-17 at 12 43 26 PM
Screen Shot 2023-08-17 at 12 43 06 PM
Screen Shot 2023-08-17 at 12 43 39 PM


fvankrieken:
Looks like this wasn't branched off latest main - you've got some conflicts with your own code in checkbook.py. Can you pull latest main, merge into your branch, resolve commits, and push again?

fvankrieken:
Or rebase main into your branch - whichever you feel more comfortable with

fvankrieken:
Maps look great at a first glance. Are the blue waypoints just for point geometries, or both points and polygons?

apps/qaqc/src/checkbook/checkbook.py
```diff
@@ -0,0 +1,108 @@
...
+
+def get_geometry_data(df):
+    geometry_data = df.groupby("has_geometry", as_index=False).sum()
+    return geometry_data
+
+def output_map(df):
+    df_clean = df.dropna(subset=['geometry'])
+    geometries = geopandas.GeoSeries.from_wkt(df_clean['geometry'])
+    gdf = geopandas.GeoDataFrame(df_clean, geometry=geometries, crs="EPSG:4326")
+    gdf_no_id = gdf.to_json(drop_id = True)
```
fvankrieken, line 22: Seems like this isn't used anywhere

apps/qaqc/src/checkbook/checkbook.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def get_geometry_data(df):
+    geometry_data = df.groupby("has_geometry", as_index=False).sum()
+    return geometry_data
+
+def output_map(df):
+    df_clean = df.dropna(subset=['geometry'])
+    geometries = geopandas.GeoSeries.from_wkt(df_clean['geometry'])
+    gdf = geopandas.GeoDataFrame(df_clean, geometry=geometries, crs="EPSG:4326")
```
fvankrieken, line 21: Unsure about this, so wondering if you tested - does specifying `geometry="geometry"` without defining the geoseries above work? Or do you specifically need the `GeoSeries.from_wkt` to read it properly

fvankrieken:
Also, looks like you need to run `black` on your code for formatting.

Once that and you've got latest main in your branch, let's go ahead and just deploy from your dev branch, using this action

https://github.com/NYCPlanning/data-engineering/actions/workflows/qaqc_deploy.yml

And choosing your branch. Seems low-risk enough

apps/qaqc/src/checkbook/checkbook.py
DeaBardhoshi: I believe just using the "geometry" column without using GeoSeries.from_wkt was giving me an error

fvankrieken:
Though also for me, with streamlit_folium installed the map doesn't actually render

fvankrieken:
> Though also for me, with streamlit_folium installed the map doesn't actually render

Sorry - that might have been something on my end actually, now it seems like it's loading. Just takes a bit

fvankrieken:
So first subjective feedback 
- would be great to make the map itself bigger. Depends on what st_folium can do, but using more of the page when in a wider browser would be great
- default zoom to map could be set to more NYC-appropriate - a bit zoomed out at the moment
- test around removing visual clutter of waypoints when decently zoomed out

DeaBardhoshi:
> So first subjective feedback
> 
> * would be great to make the map itself bigger. Depends on what st_folium can do, but using more of the page when in a wider browser would be great
> * default zoom to map could be set to more NYC-appropriate - a bit zoomed out at the moment
> * test around removing visual clutter of waypoints when decently zoomed out

Thanks Finn! I believe the first two should be easier to implement; I'll start doing some digging on how many markers the map shows based on the zoom level

DeaBardhoshi:
> So first subjective feedback
> 
> * would be great to make the map itself bigger. Depends on what st_folium can do, but using more of the page when in a wider browser would be great
> * default zoom to map could be set to more NYC-appropriate - a bit zoomed out at the moment
> * test around removing visual clutter of waypoints when decently zoomed out

It seems like adding a layer that conditionally appears in a certain zoom level would be a potential solution for 3 (i.e. at the smallest zoom maybe only have the most expensive projects)

fvankrieken:
Got things running in streamlit, reorganized code slightly, and changed aesthetics to somewhat align with capital planning explorer, though maybe we don't want to commit to that kind of brand synergy

image

apps/qaqc/src/checkbook/checkbook.py
```diff
@@ -31,17 +17,10 @@ def checkbook():
...
-    grouped_by_geometry_all = get_geometry_data(data)
-
-    st.header("Check Amounts Per Category")
-    st.bar_chart(grouped_by_category_all, x="final_category", y="check_amount")
-
-    agency = st.selectbox(
+    agency = st.sidebar.selectbox(
         "Agency Selection: ",
         (
+            "All",
```
damonmcc, line 23: just a future thought: would be cool to generate the list of options from the values in the data

fvankrieken: Agreed. First though there's some data cleanup in the checkbook process that's needed right now, as these are lists in the dataset (aggregated across checks)

damonmcc (approved):

damonmcc:
maybe squash rather than rebase?

Issue #100, opened 2023-08-01
KPDB enhancements part 1


damonmcc:
[Project Statement](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7B8203D2FE-4776-488F-8BF7-F956A8F87C1E%7D&file=Project%20statement%20-%20KPDB%202023Q3%20enhancements.docx&action=default&mobileredirect=true) in sharepoint

create allocation tables for CD and other geographies, incorporate new data, and make changes to outputs

building with these enhancements will effectively be a revision to the KPDB 2023 release

damonmcc:
currently blocked by #125

PR #168, opened 2023-08-17
refresh KPDB build

### LINKED ISSUE #125: KPDB failing to use `dcp_projects` source data

damonmcc:
resolves #125
related to #100 

## scope
- ensure KPDB can be built in the monorepo

## changes
- add columns to ZAP exports for KPDB
- test KPDB on PRs
- use edm-data for KPDB build
- update KPDB build
  - quality of life
  - don't use the old [KPDB](https://github.com/NYCPlanning/db-knownprojects) and [KPDB data](https://github.com/NYCPlanning/db-knownprojects-data) repos
  - use private S3 for non-public source data and export
  - skip redundant pgdump in data loading
- use correct ZAP column names in KPDB

## notes
- the SCA aggregation script wasn't being called as part of the build before these changes (and still isn't being called)
- [ZAP export runs](https://github.com/NYCPlanning/data-engineering/actions/workflows/zap_export.yml)
- [latest KPDB build run](https://github.com/NYCPlanning/data-engineering/actions/runs/6110889817/job/16584945271)
- screenshot of files exported to private S3 bucket
  - image


fvankrieken:
Can you revert s3 path of `db-zap-opendata` to `db-zap` in `products/zap-opendata/zap.sh`? Line 40

.github/workflows/knownprojects_build.yml
```diff
@@ -17,41 +17,42 @@ jobs:
...
         run: ./bash/01_dataloading.sh
 
-      - name: 2. build ...
+      - name: build ...
         run: ./bash/02_build.sh
-      
-      # TODO upload outputs to a private S3 bucket
-      - name: 3. export ...
+
+      # TODO fix SCA aggregation
```
fvankrieken, line 49: What's the thought moving forward here? Get it building without this in main for now so enhancements can be worked on? But we'll still need this as part of the update with the enhancements I'm assuming

.github/workflows/knownprojects_test.yml
```diff
@@ -0,0 +1,12 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  branch_test_build:
+    name: Build test run
+    uses: ./.github/workflows/knownprojects_build.yml
+    with:
+      run_export: true
```
fvankrieken, line 11: If we're exporting as part of a test, can we add a step to delete the outputs?

products/knownprojects/bash/01_dataloading.sh
```diff
@@ -3,16 +3,22 @@ source ../../bash/utils.sh
 set_error_traps
 max_bg_procs 5
 
+if [ -n "${BUILD_ENGINE_SCHEMA}" ]; then
+    echo "Dropping and creating build schema '$BUILD_ENGINE_SCHEMA'"
+    run_sql_command "DROP SCHEMA IF EXISTS ${BUILD_ENGINE_SCHEMA} CASCADE;"
+    run_sql_command "VACUUM (ANALYZE);"
```
fvankrieken, line 9: I don't think we need the analyze here as we don't need to compute any new statistics on tables, but I would say we should do `VACUUM FULL` to make sure we're freeing up storage space on the db

damonmcc: sure I'll drop ANALYZE

unfortunately I was using FULL at first and was getting timeouts

.github/workflows/knownprojects_test.yml
damonmcc: would it be even better to split the export logic enough so that a test doesn't everything except upload?

I'd move everything that's currently before the upload into the "build" stage and set `run_export: false`

products/knownprojects/python/__init__.py
```diff
@@ -11,14 +11,18 @@
...
 S3_SOURCE_HOUSING_TEAM_DIR = "dcp_housing_team/db-knownprojects"
+S3_OUTPUT_DIR = "db-kpdb"
 
 DATA_PATH = _product_path / "data"
 RAW_DATA_PATH = DATA_PATH / "raw"
 PROCESSED_DATA_PATH = DATA_PATH / "processed"
 CORRECTIONS_DATA_PATH = DATA_PATH / "corrections"
 
+CURRENT_DIR = os.getcwd()
+OUTPUT_DIR = f"{CURRENT_DIR}/output"
```
alexrichey, line 24: just curious, do we actually want these to change based on where you run python? (as opposed to using one of the fixed paths, e.g. `_product_path / "output"`

products/knownprojects/python/download.py
```diff
@@ -14,13 +14,13 @@
 def download_s3_source_file(file_name: str) -> None:
     s3_filepath = f"{S3_SOURCE_HOUSING_TEAM_DIR}/raw/{file_name}"
     save_file_path = f"{RAW_DATA_PATH}/{file_name}"
-    client().download_file(S3_SOURCE_BUCKET, s3_filepath, save_file_path)
+    client().download_file(S3_BUCKET, s3_filepath, save_file_path)
```
alexrichey, line 17: just a nit... thinking we should make s3.client() private, which will nudge folks to just use our s3.download_file

products/knownprojects/python/upload.py
```diff
@@ -1,90 +1,25 @@
...
-    print(f"Created PR with PR number: {pr.number}")
+from . import S3_BUCKET, OUTPUT_DIR, S3_OUTPUT_DIR
 
 
 if __name__ == "__main__":
-    # List all files under output
     SENDER = sys.argv[1]
     basepath = os.getcwd()
-    file_list = glob.glob(basepath + "/output/**", recursive=True)
+    output_folder = basepath + "/output/"
```
alexrichey, line 14: `OUTPUT_DIR` constant from __init__?

alexrichey, line 14: ✂️

.github/workflows/knownprojects_test.yml
damonmcc: the test now does everything except uploading (see [latest test run](https://github.com/NYCPlanning/data-engineering/actions/runs/6111936414/job/16588301689#step:7:243))

products/knownprojects/python/__init__.py
damonmcc: per this and other comments, simplified/dropped a lot of unused old stuff!

products/knownprojects/sql/combine.sql
```diff
@@ -414,30 +414,220 @@ FROM (
         NULL::numeric AS inactive,
         NULL::text AS no_classa
     FROM (
-        SELECT * FROM _dcp_application
+        SELECT
+            source,
```
alexrichey, line 418: Save us, dbt!

.github/workflows/knownprojects_build.yml
fvankrieken: Wasn't done in carto, we ran it as part of the last build

products/knownprojects/sql/dcp_application.sql
```diff
@@ -17,49 +17,15 @@ DROP TABLE IF EXISTS _dcp_application;
 WITH
 zap_translated AS (
     SELECT
-        dcp_name,
-        dcp_projectname,
-        dcp_projectbrief,
-        dcp_projectdescription,
-        (CASE
```
alexrichey, line 24: Curious, we don't need to do this data standardization anymore?

alexrichey (approved):
I don't quite follow all the changes, but if it works, then 👍!

products/knownprojects/sql/_procedures.sql
```diff
@@ -76,8 +76,10 @@ DECLARE
     _new_value text;
     _valid_fields text[];
 BEGIN
+    RAISE NOTICE 'Attempting to apply corrections for field % to %.%', _field, _schema, _table;
+    
     SELECT array_agg(column_name) FROM information_schema.columns
-    WHERE table_schema = 'public' AND table_name = _table INTO _valid_fields;
+    WHERE table_schema = _schema AND table_name = _table INTO _valid_fields;
```
fvankrieken, line 82: One day, we will have consistent applications of corrections across data products

products/knownprojects/sql/dcp_application.sql
damonmcc: nope! this is one of many columns that have always been standardized [during the ZAP export](https://github.com/NYCPlanning/data-engineering/blob/fix-kpdb-build/products/zap-opendata/src/visible_projects.py#L12)

since we now archive the standardized (aka recoded) ZAP export rather than the raw CRM export, that's what KPDB uses

fvankrieken: My guess would be that it's moved upstream, to the zap data processing? @damonmcc ?

products/knownprojects/sql/combine.sql
damonmcc: cannot overstate how much this super union tripped me up (thinking I was fixing all geometries in the final table when it was only one of the tables pre-union)

products/knownprojects/sql/dcp_application.sql
```diff
@@ -68,24 +34,24 @@ zap_translated AS (
...
-                        dcp_certifiedreferred, dcp_dcptargetcertificationdate
+                        certified_referred, dcp_dcptargetcertificationdate
                     )::timestamp,
                     'YYYY/MM/DD'
                 )
         END) AS dcp_certifiedreferred,
+        (CASE
+            WHEN completed_date IS NULL THEN NULL
+            ELSE TO_CHAR(completed_date::timestamp, 'YYYY/MM/DD')
+        END) AS completed_date,
```
fvankrieken, line 54: Nit, but you don't need the case (for this or the similar clause above)

`select to_char(null::timestamp, 'YYYY/MM/DD')` will return `NULL`

products/knownprojects/sql/dcp_application.sql
```diff
@@ -203,7 +169,7 @@ SELECT DISTINCT
 
     dcp_projectbrief,
 
-    dcp_projectdescription,
+    -- dcp_projectdescription,
```
fvankrieken, line 172: What's up with this one?

products/knownprojects/sql/dcp_application.sql
damonmcc: word. I didn't add this though, it was on an earlier line and probably only moved via sqlfluff

fvankrieken: Wait what the heck really? Why is this computed column reordered differently than others? Bizarre

products/knownprojects/sql/dcp_application.sql
damonmcc: just noticed that too! it's [a ZAP column I added](https://github.com/NYCPlanning/data-engineering/pull/168/files#diff-afd801c435253a3a9eec4615111e15d379274569ffce04a0d863af73b7f8ff12R44) to the standardized export that KPDB now uses

this should be uncommented

products/knownprojects/sql/dcp_application.sql
damonmcc: hmmmm I was gonna say maybe it sorts alphabetically based on the column used, but that doesn't seem to be it

maybe I moved it and sqlfluff doesn't care since all the CASE statements are still grouped together?

fvankrieken (approved):

PR #172, opened 2023-08-21
add sqlfluff and compile requirements

damonmcc:
will publish docker images after merge to main


fvankrieken (approved):

alexrichey (approved):
👍 Will eventually be a good idea to separate out the dev dependences, e.g. sqlfluff, ipykernel, black but this seems fine for now.

fvankrieken:
I was just thinking about that - more on the side of splitting out envs given that we're adding more packages around visualization to qa~~qc~~, while our builds certainly don't need folium etc.

PR #173, opened 2023-08-21
[POC] add sql file linting to PR tests and format all sql files

damonmcc:
changes
- configure and use `sqlfluff` to format all sql files
- add a sql linting check to the PR test workflow

todos
- [ ] finish trying to get test passing using `fix_even_unparsable = True`
- [ ] revert a file's formatting and try doing it with `fix_even_unparsable = False` and `templater = placeholder` to handle psql variables
- [ ] try using `pyproject.toml` rather than `.sqlfluff` (better list values and will allow for one file for other configurations like pytest)

notes
- sqlfluff fix failing on some files with `Loop limit on fixes reached [10]` (e.g. `projects_dollars_mapped_categorized_sponsor.sql`)


products/db-cpdb/sql/_create_export.sql
```diff
@@ -1,11 +1,13 @@
...
 
 DROP TABLE IF EXISTS cpdb_dcpattributes_poly;
 SELECT * INTO cpdb_dcpattributes_poly
-FROM cpdb_dcpattributes 
-WHERE ccpversion = :'ccp_v' 
-AND ST_GeometryType(geom)='ST_MultiPolygon';
\ No newline at end of file
+FROM cpdb_dcpattributes
+WHERE
+    ccpversion =: 'ccp_v'
```
fvankrieken, line 12: This doesn't look like it's going to run

fvankrieken: Of note: https://github.com/sqlfluff/sqlfluff/pull/1709

damonmcc: hm I wonder why it finds them unappeasable then

an example is `DROP TABLE IF EXISTS :TABLE` in `/db-pluto/pluto_build/sql/export_mappluto_gdb.sql`

damonmcc: if it's breaking queries, I may have to set `fix_even_unparsable` back to the default of `False` and ignore all files with psql variables in them

PR #175, opened 2023-08-23
format and add test for KPDB sql files

damonmcc:
changes
- add a `pyproject.toml` file to configure `sqlfuff` (rather than a `.sqlfluff` file. we get better list value formatting and will allow for one file for future configurations like `pytest`)
  - excluded rules were chosen to avoid substantial untestable changes to queries in this PR / first pass
- call `sqlfluff` in PR tests on KPDB files only
- format KPDB sql files using `sqlfluff fix` and manual edits


alexrichey:
Nice... I'd actually added a pyproject file locally, just to configure my python linter. Mind adding the following? 
```
[tool.pydocstyle]
ignore = ["D100", "D101", "D102", "D107", "D104", "D213", "D407", "D413"]

[tool.black]
```

alexrichey:
> @alexrichey what affect does the empty `[tool.black]` section have?

@damonmcc It's usually just to inform an editor to use black. Mine requires it

damonmcc:
@alexrichey added to pyproject.toml

alexrichey (approved):
@damonmcc looks good... curious if you were able to run these just to check that the formatting isn't invaliding the files somehow?

products/db-knownprojects/sca_aggregate/boundaries_es_zone.sql
```diff
@@ -12,332 +12,459 @@ drop table if exists aggregated_es_zone_project_level;
 drop table if exists longform_es_zone_output;
```
fvankrieken, line 12: This file was made all lowercase sql keywords, while others are uppercase. @alexrichey has theory that maybe it does it per file, depending on capitalization of first keyword? Would be nice to test, and maybe see if we can specify that in the template/settings just to make sure it's always consistent.

products/db-knownprojects/sca_aggregate/create_zap_projects.sql
```diff
@@ -1,35 +1,40 @@
...
 DROP TABLE IF EXISTS dcp_projectbbl_subset;
 CREATE TABLE dcp_projectbbl_subset AS
-SELECT 
-     -- Extract the first 9 characters of the "dcp_name" column and rename it as "dcp_name" to match 
-     SUBSTRING(dcp_name FROM 1 FOR 9) AS dcp_name,
-     -- Rename dcp_bblnumber to bbl 
-     dcp_bblnumber AS bbl
+SELECT
+    -- Extract the first 9 characters of the "dcp_name" column and rename it as "dcp_name" to match 
+    dcp_bblnumber AS bbl,
```
fvankrieken, line 6: Did fluff reorder this query or was that you?

fvankrieken:
Thanks for putting the time into this, very happy with it

products/db-knownprojects/sca_aggregate/boundaries_es_zone.sql
alexrichey: @fvankrieken @damonmcc https://docs.sqlfluff.com/en/stable/rules.html#rule-capitalisation.keywords 

Looks like we can just set the policy to upper case.

products/db-knownprojects/sca_aggregate/create_zap_projects.sql
damonmcc: pretty sure this was fluff. I'll look for the "rule" but I think a convention is to put function's last so 

```
SUBSTRING(dcp_name FROM 1 FOR 9) AS dcp_name,
dcp_bblnumber AS bbl
```

becomes

```
dcp_bblnumber AS bbl,
SUBSTRING(dcp_name FROM 1 FOR 9) AS dcp_name
```

alexrichey (approved):
nice!

products/db-knownprojects/sca_aggregate/create_zap_projects.sql
damonmcc: found the rule! it's [`structure.column_order`](https://docs.sqlfluff.com/en/stable/rules.html#rule-structure.column_order)

"Select wildcards then simple targets before calculations and aggregates"

damonmcc:
@alexrichey 
> looks good... curious if you were able to run these just to check that the formatting isn't invaliding the files somehow?

to follow up: it isn't a test of all the queries because the build is "broken", but it is [currently failing](https://github.com/NYCPlanning/data-engineering/actions/runs/5976894603/job/16215716801) at the same spot as it was before the reformatting

PR #177, opened 2023-08-23
Simplify folder names

fvankrieken:


alexrichey:
@fvankrieken there's a hypothetical scenario I'm worried about, where a product infers it's name (and s3 bucket) from its folder name. Just want to confirm that's never the case. cc @damonmcc

fvankrieken:
The bash function `upload` require s3 folder (dataset name, with "db-") explicitly. I certainly haven't written any python functions which get dataset name from working directory/path

fvankrieken:
Also we have a conflict - maybe git will be less happy with changes to files in these folders than I had hoped

damonmcc:
yea the conflict seems weird, that files doesn't seem so unique but it [was changed recently](https://github.com/NYCPlanning/data-engineering/commit/ed1fb0a81517120291edf873495e875c84515a38). maybe just needs a rebase??

alexrichey (dismissed):

damonmcc (approved):

fvankrieken:
🫡

PR #178, opened 2023-08-24
export internal ZAL CRM files to edm-private

damonmcc:
related to #XXX

todos
- [ ] ...


.github/workflows/zap_export.yml
```diff
@@ -74,6 +71,9 @@ jobs:
           sudo mv ./mc /usr/bin
           mc config host add spaces $AWS_S3_ENDPOINT $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY --api S3v4
```
fvankrieken, line 72: Thoughts on doing a bit of zap maintenance as part of this PR?
- merge python reqs with overall repo reqs, publish new build images if needed
- use build images
- bump to ubuntu 22.04
- remove python/minio setup steps

fvankrieken: Ah there's a data library step below. That could be done still with a `use` step of the published action

.github/workflows/zap_export.yml
```diff
@@ -116,11 +116,14 @@ jobs:
...
+      - name: Export to edm-private and edm-publishing
         env:
           VERSION: ${{ steps.version.outputs.version }}
         run: |
+          ./zap.sh upload_crm_do ${{ matrix.entity }} $VERSION
+
           if ${{ matrix.open }};
           then
-          ./zap.sh upload_do ${{ matrix.entity }} $VERSION
+          ./zap.sh upload_internal_do ${{ matrix.entity }} $VERSION
```
fvankrieken, line 127: Should this step be happening for non-open datasets? Since this is the internal upload function

.github/workflows/zap_export.yml
damonmcc: good idea, in-progress

.github/workflows/zap_export.yml
damonmcc: for non-open datasets, the CRM upload function and Data Library archiving above this section is actually all we need

"internal" and "visible" exports only relate to the open datasets

.github/workflows/zap_export.yml
```diff
@@ -88,39 +81,52 @@ jobs:
...
-          echo "exporting ${{ matrix.entity }}.csv"
-          docker run --rm \
-            -e AWS_S3_ENDPOINT=$AWS_S3_ENDPOINT\
-            -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID\
-            -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY\
-            -e AWS_S3_BUCKET=$AWS_S3_LIBRARY_BUCKET\
-            -v $(pwd)/templates:/templates\
-            -v $(pwd)/.output:/.output\
-            nycplanning/library:ubuntu-01d9e5eddf65dbdea3eb0c500314963b5ec6246a library archive -f /templates/${{ matrix.entity }}.yml -v $VERSION -s -l -o csv --compress
+      # - name: Archive to data library - PGdump
```
fvankrieken, line 94: Want to just drop these comments if it's working as desired now?

.github/workflows/zap_export.yml
fvankrieken: So just to clarify one more time - "non-open" datasets only go to recipes through data library, while the others go to publishing and private (private being the "internal" one)

fvankrieken (approved):
Nice. Thanks for taking the time to update the run environment!

.github/workflows/zap_export.yml
damonmcc: yup! wanted to keep those details handy until it was working

fvankrieken (approved):

.github/workflows/zap_export.yml
damonmcc: exactly

Issue #180, opened 2023-08-25
db-factfinder Space in decennial male variable 


TangoYankee:
# Description
The `male` variable within the decennial census has a mysterious space that is preventing the labs-factfinder frontend from linking to the data source. I attempted to manually encode the space in the front end. However, it has not resolved the behavior.

## Requested action
Remove the space

fvankrieken:
`male p` should be corrected to `malep` as well right? Just for consistenty (as with `fem` and `femp`)

TangoYankee:
> `male p` should be corrected to `malep` as well right? Just for consistenty (as with `fem` and `femp`)

yes, please

PR #181, opened 2023-08-25
Fvk pff tweaks

### LINKED ISSUE #179: db-factfinder variables missing from decennial metadata files
# Description
Please add the following variables to the `decennial/2020/[version]/metadata.json`.

Spouse, OpSxS, SmSxS, UMrdPtnr, OpSxUMrd, SmSxUMrd, Child, ChildU18, GrndCh, GrndChU18, OthrRel, NonRel, Fam, NFamHH, NFmHHU65, MrdCplHH, MrdCh18, CoCplHH, CoCplCh18, MaleHHNS, MnSPChU18, FemHHNS, FnSPChU18

## Context
~~These variables appear for 2020 but not 2010. There was a brief discussion on strategies other than adding the variables to a single metadata file for each data source. However, the ACS data already has instances where variables in one year are not found in another year. Consequently, the ingest pipeline for labs-factfinder-api is already designed to handle variables that occur in one year and not another. This is likely something we will want to reconsider in future iterations of the pipeline. But for the immediate data needs, I would prefer to keep the current file structure.~~

Removed the context because it was based on faulty assumptions and was adding confusion.

fvankrieken:
Closes #179, #180 

For 179, there was a bug that I'd introduced - hadn't realized that `object` types in pandas could be mixed types. So in the data dictionary, if a variable was present for `2010, 2020` it was interpreted correctly but for the couple cases where the year was `2020` it was interpreted as a float and converted to `NaN` by `.str.split(", ")` rather an array of years.

For 180, figured just keep it minimal - replace column names exactly rather than any sort of regex or trimming whitespace, just in case anything down the road requires anything different

[Successful build](https://github.com/NYCPlanning/data-engineering/actions/runs/5979687092) - I've manually QA'd outputs. Made small fix to commit 2 this morning and tested changes locally

### DIFF
--- a/products/factfinder/factfinder/data/decennial/2010/metadata.json
+++ b/products/factfinder/factfinder/data/decennial/2010/metadata.json
@@ -1,974 +1,204 @@
 [
     {
-        "pff_variable": "Pop1",
-        "base_variable": "Pop1",
-        "category": "POPULATION, SEX, AGE, AND DENSITY",
-        "domain": "decennial"
-    },
-    {
-        "pff_variable": "Male",
-        "base_variable": "Pop1",
-        "category": "POPULATION, SEX, AGE, AND DENSITY",
-        "domain": "decennial"
-    },
-    {
-        "pff_variable": "Fem",
-        "base_variable": "Pop1",
-        "category": "POPULATION, SEX, AGE, AND DENSITY",
-        "domain": "decennial"
-    },
-    {
-        "pff_variable": "PopU5",
-        "base_variable": "Pop1",
-        "category": "POPULATION, SEX, AGE, AND DENSITY",
-        "domain": "decennial"
-    },
-    {
-        "pff_variable": "Pop5t9",
-        "base_variable": "Pop1",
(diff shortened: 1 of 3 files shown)


alexrichey (dismissed):
@fvankrieken just confirming that the metadata changes are intentional? Everything else looks good

damonmcc (approved):

fvankrieken:
> @fvankrieken just confirming that the metadata changes are intentional? Everything else looks good

They are, they've been re-generated. I might just remove them from git actually - before they were both and input to us and to AE's work, now we're just generating them and uploading to DO so it doesn't make a ton of sense to check into the repo

fvankrieken:
@alexrichey since the rest of the files in those folders are sort of archived from when we used them as inputs, I've reverted the checked-in metadata files for decennial to those versions (should we ever go back to pulling from an API) and the new generated ones are just pushed to S3 without checkin now

Successful job here, outputs have been qa'd
https://github.com/NYCPlanning/data-engineering/actions/runs/6002251634

damonmcc (approved):

PR #186, opened 2023-08-29
Implement mypy and fix static checking violations

alexrichey:
Running `mypy` over `dcpy` returns mostly type nit-picking, and also some real issues. We should consider implementing an automated mypy type check in CI, whenever files in `dcpy` are changed.

### DIFF
--- a/.github/workflows/dcpy_test.yml
+++ b/.github/workflows/dcpy_test.yml
@@ -1,0 +1,22 @@
+name: DCPY Tests
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  type_check:
+    name: Static Type Check (mypy)
+    runs-on: ubuntu-22.04
+    container:
+      image: nycplanning/build-geosupport:latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: (Temp) Install stubs # can remove once these make it into base docker image
+        run: pip install -r python/requirements.txt
+
+      - name: Static Type Check (mypy)
+        run: mypy ./dcpy
--- a/dcpy/utils/postgres.py
+++ b/dcpy/utils/postgres.py
@@ -5,20 +5,23 @@ import os
 import pandas as pd
 from psycopg2.extensions import AsIs
 from sqlalchemy import create_engine, text
+from typing import Optional
 
 
 BUILD_ENGINE_RAW = os.environ["BUILD_ENGINE"]
 build_engine = create_engine(BUILD_ENGINE_RAW)
 
 
-def execute_query(query: str, placeholders: dict = None) -> None:
+def execute_query(query: str, placeholders: Optional[dict] = None) -> None:
     if placeholders is None:
         placeholders = {}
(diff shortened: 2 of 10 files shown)


fvankrieken:
Thanks for this, a lot of this is my doing. I would be very in favor of automatic checks, even if it's just adding to the overall requirements for now.

dcpy/utils/postgres.py
```diff
@@ -5,20 +5,23 @@
...
 build_engine = create_engine(BUILD_ENGINE_RAW)
 
 
-def execute_query(query: str, placeholders: dict = None) -> None:
+def execute_query(query: str, placeholders: Optional[dict] = None) -> None:
     if placeholders is None:
         placeholders = {}
     with build_engine.connect() as connection:
-        connection.execute(statement=text(query), params=placeholders)
+        connection.execute(statement=text(query), parameters=placeholders)
```
alexrichey, line 19: FYI, this was one of the bugs. I haven't tested usages of this fn. We could benefit from some integration testing here!

alexrichey:
> I would be very in favor of automatic checks, even if it's just adding to the overall requirements for now.

@fvankrieken (cc @damonmcc ) happy to just add these reqs to the standard requirements.[in/txt]. Thoughts?

damonmcc:
yea the standard requirements seems good for now. love type checking!

fvankrieken (dismissed):

.github/workflows/dcpy_test.yml
```diff
@@ -0,0 +1,25 @@
...
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install mypy
+        run: pip install mypy
+
+      - name: (Temp) Install stubs
+        run: pip install -r python/requirements.txt
```
damonmcc, line 22: why not add `mypy` to the requirements too?

pyproject.toml
```diff
@@ -34,4 +34,8 @@ capitalisation_policy = "upper"
 [tool.pydocstyle]
 ignore = ["D100", "D101", "D102", "D107", "D104", "D213", "D407", "D413"]
 
-[tool.black]
\ No newline at end of file
+[tool.black]
+
+[[tool.mypy.overrides]]
+module = "geopandas.*" # no stubs available
+ignore_missing_imports = true
```
damonmcc, line 41: to confirm, we don't want `ignore_missing_imports` to be a global mypy setting?

I'm not familiar with configuring mypy but I think at the moment, this setting is module-specific to `geopandas` and it's submodules

alexrichey: @damonmcc Correct, we only want this to affect geopandas, which has no type stubs available (at least not in our current version). It's currently the only dcpy lib where that's the case.

.github/workflows/dcpy_test.yml
alexrichey: ah, good catch. Adding.

alexrichey: @damonmcc Added

damonmcc (approved):

fvankrieken (approved):

Issue #92, opened 2023-07-27
PLUTO - zoning data correction


AmandaDoyle:
For tax lot 4158620001 add in MX-21 for the Special District and M1-4/R6 for the zoning 2 district.  This should be done before 23v2.1.  

DE should investigate how to adjust the thresholds for assigning these so that a manual correction is not needed, if it's possible to do so without having negative impacts elsewhere.

damonmcc:
re-opening to use this issue for a change / enhancement to the thresholds

PR #193, opened 2023-08-31
Fvk update ztl sql

fvankrieken:
Commit-by-commit will be easier. 4th commit is sqlfluff but also changes to pyproject.toml, pinning sqlfluff to my fork in requirements, and recompiling requirements

Bit of a follow up to #92 - since the changes were made for Pluto, they should presumably be made to ztl as well.

While I was at it, figured I'd fluff ztl. This required [adding a bit of logic to sqlfluff](https://github.com/fvankrieken/sqlfluff/commit/0148345857e9571c0dfda9187bc9fe9ab40df80b) around quoted variables, so we're pinned to my fork now

[ztl build](https://github.com/NYCPlanning/data-engineering/actions/runs/6040760875/job/16392275939)


products/zoningtaxlots/bash/02_build.sh
```diff
@@ -3,16 +3,15 @@ source bash/config.sh
 
 echo "Run build"
 
-run_sql_file sql/create_priority.sql &
+run_sql_file sql/create_priority.sql
```
fvankrieken, line 6: removing making these background processing - not horribly long to run and it gets around `set -e` when things go wrong (case in point I deleted the file area_zoningdistrict_create.sql without removing the `run_sql_file` command belowand the job ran just fine)

damonmcc (approved):

PR #198, opened 2023-09-06
pluto: use max proximity code per billing-bbl

alexrichey:
In the case of condos, there are sometimes multiple prox codes per bbl in a given billing bbl. Previously when we were assigning proxcodes, the choice of proxcode was not deterministic. This changes to use the most proximate condo unit. E.g.
```
+-----+--------------+-----------+
| bbl |  billing bbl |  proxcode |
+-----+--------------+-----------+
|   1 |          123 |         1 |
|   2 |          123 |         2 |
+-----+--------------+-----------+
```

billing bbl for `123` will use proxcode 2

### DIFF
--- a/products/pluto/pluto_build/sql/cama_proxcode.sql
+++ b/products/pluto/pluto_build/sql/cama_proxcode.sql
@@ -3,7 +3,7 @@
 -- select proxcode for lot from record where bldgnum is 1
 -- all bbl reocrds have at least one record with a bldgnum = '1'
 WITH dcpcamavals AS (
-    SELECT DISTINCT
+    SELECT
         primebbl AS bbl,
         (CASE
             WHEN proxcode = '5' THEN '2'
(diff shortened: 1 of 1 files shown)


AmandaDoyle:
I approve of this change.  I'm flagging that we'll need to update the ProxCode definition in the [data dictionary](https://s-media.nyc.gov/agencies/dcp/assets/files/pdf/data-tools/bytes/pluto_datadictionary.pdf) to reflect this change.

alexrichey:
I did some QA between data sets to assess impact of this change, and it appears to be very, very minor. I generated one Pluto dataset with this fix, and one without, both comparing against 23v2 for QA. 

[23v2.2 with fix](https://cloud.digitalocean.com/spaces/edm-publishing?path=db-pluto%2Fdraft%2Far-pluto-proxcode-fix%2F2023-09-21%2F)
[23v2.2 without fix](https://cloud.digitalocean.com/spaces/edm-publishing?path=db-pluto%2Fdraft%2Ftemp-for-mimic-pluto-build%2F)

They're basically identical. The number of `proxcode` changes against 23v2 stays the same - which felt suspicious at first, until I remembered that in every discrepancy we looked at, the `proxcode`s in the later version were higher. So I _think_ this is what we would expect. 

In fact this fix only appears to affect 1 row, and only for the following fields: `comarea, resarea, otherarea, lottype, bsmtcode`. Even more curiously, the version with fix is more similar to 23v2. This may, I hate to say it, be another source of indeterminacy. But since it's a single row, I'm not sure it's high priority to chase down.

To sum up, from this experiment, I can't say for _certain_ that this resolves the issue - to do that, I'd need to replicate the conditions (code, database, etc.) that generated 23v2, and implement the fix there. But I feel confident that this will keep the `proxcodes` from varying between releases, so I think this is probably worth merging.

fvankrieken (approved):

damonmcc (approved):

PR #200, opened 2023-09-06
Bump postgis in local tooling

alexrichey:
Allow an override to the postgis devcontainer image, and bump default version to 15-3.3.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -23,7 +23,7 @@ services:
     # (Adding the "ports" property to this file will not forward from a Codespace.)
 
   db:
-    image: postgis/postgis:14-3.3-alpine
+    image: ${POSTGIS_IMAGE}
     restart: unless-stopped
     volumes:
       - postgres-data:/var/lib/postgresql/data
--- a/.gitignore
+++ b/.gitignore
@@ -109,7 +109,7 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
+/.env
 .venv
 env/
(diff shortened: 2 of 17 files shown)


alexrichey:
@fvankrieken Yeah, seems like our images already have v15 `postgresql-client` installed just via `apt-get install`?

fvankrieken:
Oh really? When I last checked it was 14, guess apt has 15 now. Lovely!

fvankrieken (approved):

.gitignore
```diff
@@ -109,7 +109,7 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
+# .env
```
damonmcc, line 112: comment or delete? and would we rather whitelist `.devcontainer/.env` and still ignore all other `.env` files?

.devcontainer/docker-compose.yml
damonmcc, line 1: since `env_file` here is still `../.env`, does it find the new one in `/.devcontainer`? on my machine the dev container only uses a file at the top level

.gitignore
alexrichey: oops, right. Left it to ignore .env in the root folder.

.devcontainer/docker-compose.yml
alexrichey: Yup, compose looks for a file called `.env` in the same folder. Could specify with a CLI arg, but wanted to to leave behavior the same for our "default" users (i.e. the two of you)

damonmcc (approved):

PR #203, opened 2023-09-06
pluto: add yq to base images

alexrichey:
yq is the yaml analogue to jq. We use it for the pluto builds to parse the recipe file.


alexrichey:
@damonmcc or @fvankrieken - IIRC, the base image is rebuilt and pushed to the registry after merge to main?

damonmcc (approved):
@alexrichey AFIK there's no rebuild trigger on merge to main, so you'll have to manually run it

PR #208, opened 2023-09-07
add new aggregations to KPDB

### LINKED ISSUE #100: KPDB enhancements part 1

damonmcc:
resolves #100 

## for reviewers
- the large new sql files in [`/aggregate/`](https://github.com/NYCPlanning/data-engineering/tree/f665ddca6e560cc9c87cf004eae3b43a9ce2f655/products/knownprojects/aggregate) are copies of an SCA aggregation sql file where the only changes are the geographic boundary input table/columns and the name of the final output table

## changes
- add sql files to aggregate KPDB by more geographic boundaries:
  - 2020 Census Tract (CT)
  - 2020 Neighborhood Tabulation Areas (NTA)
  - Community District (CD)
  - Community District Tabulation Areas (CDTA)
- load source data needed for new aggregations
- combine new and existing SCA aggregation into a common build stage
- revise KPDB build stages to be `dataloading`, `build`, `aggregate`, `export`, `upload`
- improve existing SCA aggregation files
  - drop intermediate tables at end of query
- name the upload folder in the private S3 bucket using the build engine schema name and the build date
- output the table `source_data_versions` and include internal source data versions in it

## notes
- Capital Planning team is reviewing the latest outputs. but with product builds (and maybe enhancements too?) we've started to err on the side of calling a task complete & merging code when data's been handed off for review
- these aggregation queries are **not optimized**. they were originally written/run in CARTO's SQL interface and copied to  the KPDB repo. I created an issue to optimize them later : #230 
- this is currently the only product uploading to the private S3 bucket. it doesn't use our new draft/publish approach and doing so would require refactoring of those utils to not be hard-coded to use the `edm-publishing` bucket
- this currently runs a full KPDB build and uploads to S3 on relevant changes in PRs. was very helpful during this dev but could undo that if it feels like overkill
- the [KPDB build](https://github.com/NYCPlanning/data-engineering/actions/runs/6158541533) that was shared with Housing and Capital Planning for review


products/knownprojects/bash/03_aggregate.sh
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Create ZAP Project Many BBLs table"
+run_sql_file aggregate/create_zap_projects.sql
+
+## Do SCA aggregation
+echo "Create the longfrom SCA Aggregate Tables..."
+
+# Preprocess the tables to standardize geometry column name 
+echo "Preprocess column names to standardize"
+run_sql_file aggregate/sca/preprocessing.sql
```
fvankrieken, line 16: is there a big reason for the sca subfolder at this point? If these are all just similar aggregations by different geographies

products/knownprojects/python/extractors.py
```diff
@@ -8,81 +8,69 @@
...
 
 @ETL
-def edc_dcp_inputs() -> gpd.geodataframe.GeoDataFrame:
-    filename = "edc_shapefile_20221118.zip"
-    df = gpd.read_file(f"zip://{RAW_DATA_PATH}/{filename}")
-    # print(df)
+def dcp_planneradded(filename: str) -> pd.DataFrame:
+    df = pd.read_csv(f"{RAW_DATA_PATH}/{filename}", dtype=str)
+    df.rename(columns={"WKT": "geometry"}, inplace=True)
     return df
```
fvankrieken, line 74: It seems like the extractors can really just be abstracted out to one or two functions at this point, no? Given that filename is now an arg, this etl isn't exactly specific to "dcp_planneradded" - it's just pd.read_csv with a rename (which I believe could even be done with a kwarg to read_csv)

damonmcc: maybe. there seems to be a little bit of source-specific stuff like [here](https://github.com/NYCPlanning/data-engineering/blob/f665ddca6e560cc9c87cf004eae3b43a9ce2f655/products/knownprojects/python/extractors.py#L18). that line could be done to all source data, but I was aiming for minimal unrelated change

and I think it'd have to be 3 functions to cover the 3 file types we get from the Housing team ([list of filenames](https://github.com/NYCPlanning/data-engineering/blob/f665ddca6e560cc9c87cf004eae3b43a9ce2f655/products/knownprojects/python/__init__.py#L25)). or could be one function with branching logic based on the filename's extension?

products/knownprojects/bash/03_aggregate.sh
damonmcc: the only reasons I can think of were
- to make it a little easier to test the new files. since these all take so long, disabling all SCA aggregation was easier when they're in a subfolder and have their own [`preprocessing.sql`](https://github.com/NYCPlanning/data-engineering/blob/kpdb-geographies/products/knownprojects/aggregate/sca/preprocessing.sql)
- unlike the new files, these are kinda more specific and are related to each other because they're meant to be shared with the SCA
- per that last one, they're exported and [zipped](https://github.com/NYCPlanning/data-engineering/blob/kpdb-geographies/products/knownprojects/bash/04_export.sh#L61-L62) separate from the new, more general, aggregations

PR #212, opened 2023-09-08
pluto: fix failing build

alexrichey:
Sets a default for the recipe file to use in a build when called (this was already set on dispatch). We should probably just default in `dcpy` to use `recipe.yml` in the product dir

### DIFF
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -6,6 +6,9 @@ on:
       run_export:
         type: boolean
         required: true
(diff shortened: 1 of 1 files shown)


alexrichey:
FYI, [here's a build](https://github.com/NYCPlanning/data-engineering/actions/runs/6122879264/job/16619593713) that's progressed past the failing step.

fvankrieken (approved):

damonmcc (approved):

PR #222, opened 2023-09-11
Allow building PLUTO Minor on load of Zoning data

alexrichey:
[This](https://github.com/NYCPlanning/data-engineering/issues/224) is an example of an auto-generated issue. 
Quick note that the version wasn't being passed correctly. That's fixed now, and it will look like this: `New Product Build Pluto:{{version}}: Build Failed`

Also, I decided against having a separate workflow for this, for the moment, since this is solely a PLUTO use-case.


.github/workflows/pluto_build.yml
```diff
@@ -107,3 +116,29 @@ jobs:
       - name: Export and Upload
         run: ./06_export.sh
         if: inputs.run_export
+
```
alexrichey, line 119: I'll say, this does feel a little misplaced. It would be nice if this were more generic and reusable, but since this is a single use-case for now, I left it here.

damonmcc (approved):

PR #191, opened 2023-08-31
Fvk draft and publish in publishing

fvankrieken:
Continuing #174. Also closes #195 

Some datasets required a bit of retooling. Successful builds:
- [cbbr](https://github.com/NYCPlanning/data-engineering/actions/runs/6090203285/job/16524611543)
- [colp](https://github.com/NYCPlanning/data-engineering/actions/runs/6100743393/job/16555682972)
- [cpdb](https://github.com/NYCPlanning/data-engineering/actions/runs/6090206258/job/16524620348)

Running some other builds now just as a check

Code used to migrate data put in #174, I ran from jupyter notebook in my devcontainer. Happy to archive this in another way if desired, but it seems pretty straightforward.

### Review
Definitely going to be easier to go commit-by-commit. The first 5 are all more straightforward. After these 5 I migrated data, then did the slightly meatier work of both updating the publishing api (for getting data, rather than uploading/publishing which was already done) and adjusting calls to publishing from the qa app. The draft/publish logic is a little clunky at best here, would love thoughts and feedback

#### Commits
1. renaming and addition of git utils in dcpy
2. mainly adding metadata functionality to both utils.s3 and connectors.edm.publishing, as well as some bug fixes. 
3. self-explanatory
4. I needed a function to publish old data with a little more flexibility. I wrote this in `edm.publishing` and then called this function from a jupyter notebook, see code comments in #174. Also added in a CLI - I had a bad one before but seem to have lost it in a rebase error in my last PR
5. update bash utils to use new upload functionality (to draft folder only)
6. getting data from publishing
    - rework publishing functions to get data from s3
    - add one component to be used across all products in qa to select publish/draft and version/build.
    - rework individual products to use said component and get data in consistent manner. There's a fair amount more work here to be done but in the sake of getting this merged and everything going through the draft/publish cycle by the time Sasha really get's started up, would like to hold off on that for now (and hopefully get this merged by EOD Monday at latest)
     - two exceptions are GRU and EDDE - EDDE's data is not migrated since it's not incorporated to monorepo yet. GRU is a more custom interface for the GR(U) team. But at least now these are the only two product pages that reference s3, the rest only reference edm.publishing
7. dataclasses!
8. address comment from Alex
9. "
10. thorough type signatures in edm.publishing as well as expanded docstrings

PR #225, opened 2023-09-12
object-oriented approach for publishing utils

fvankrieken:
Follow up to #191 - from discussions with @alexrichey we thought it could theoretically be cleaner to have interactions with products/drafts go through these objects representing them. This was a first pass. Not looking necessarily to merge, but just wanted to use this as a starting point for discussion for now. 

Don't worry too much about changes in the qa app, meat is really all in `dcpy.connectors.edm.publishing`

### DIFF
--- a/apps/qa/src/colp/helpers.py
+++ b/apps/qa/src/colp/helpers.py
@@ -6,12 +6,12 @@ from dcpy.connectors.edm import publishing
 PRODUCT = "db-colp"
 
 
-def get_data(product_key: publishing.ProductKey) -> dict[str, pd.DataFrame]:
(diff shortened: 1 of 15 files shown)


damonmcc:
I think I prefer the initial implementation since it feels like it's only as OOP as it needs to be. I do generally prefer functional over OOP since it often feels like re-inventing the wheel/overkill, but it definitely seems nice to reduce passing around tons of arguments!

and the initial implementation seems to be a nice mix to start with: static functions that operate on an object. maybe some refactoring of the QA app will be a good way to work out how these utils could work best?

PR #229, opened 2023-09-13
Fvk cleanup publishing s3 calls

fvankrieken:
in #225, realized it'd also be nice to use only s3 utils within publishing (and not requests assuming that certain files would be public-read).

Commits:
1. above-mentioned issue
2. move `product_path` to a property of the keys. This makes sense and is nice within the `publishing` module as it really is a property of the key, but maybe shouldn't be included because then we need to expose it to anyone using a product key. Fine either way, can either leave this commit in or just drop it
3. fix some import paths causing issues on prod for pluto and edde at the moment

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -32,6 +32,10 @@ class ProductKey:
             BUCKET, f"{self.product_path}/version.txt"
         ).splitlines()[0]
 
+    @property
+    def product_path(self) -> str:
+        return f"{self.product}/publish/{self.version}"
+
 
 @dataclass
 class DraftKey(ProductKey):
@@ -51,12 +55,9 @@ class DraftKey(ProductKey):
     def label(self):
         return f"{self.product} - {self.version} - {self.build}"
 
-
-def __get_product_path(product_key: ProductKey) -> str:
-    if isinstance(product_key, DraftKey):
-        return f"{product_key.product}/draft/{product_key.build}"
-    else:
-        return f"{product_key.product}/publish/{product_key.version}"
+    @property
+    def product_path(self) -> str:
+        return f"{self.product}/draft/{self.build}"
 
 
 def get_latest_version(product: str) -> str:
@@ -100,7 +101,7 @@ def upload(
     max_files: int = 20,
 ) -> None:
     """Upload build output(s) to draft folder in edm-publishing"""
-    draft_path = __get_product_path(draft_key)
+    draft_path = draft_key.product_path
     meta = build_metadata.generate()
     if output_path.is_dir():
         s3.upload_folder(
@@ -180,7 +181,7 @@ def publish(
     By default, keeps draft output folder"""
     if publishing_version is None:
         publishing_version = draft_key.version
-    source = __get_product_path(draft_key)
+    source = draft_key.product_path
     target = f"{draft_key.product}/publish/{publishing_version}/"
     s3.copy_folder(BUCKET, source, target, acl, max_files=max_files)
     if not keep_draft:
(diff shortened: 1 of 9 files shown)


alexrichey (approved):
I like the concept of product key more and more... I think the one change/minor refactor I'd suggest is simplifying the `*Key` classes to be pure dataclasses (ie no magic methods overrides or annotations) and moving anything complicated (e.g. fetching a version) into a static constructor

damonmcc (dismissed):

fvankrieken:
> I like the concept of product key more and more... I think the one change/minor refactor I'd suggest is simplifying the `*Key` classes to be pure dataclasses (ie no magic methods overrides or annotations) and moving anything complicated (e.g. fetching a version) into a static constructor. (This is more of a comment for your previous PR, but wanted to get it in before we move on to other things)

The one complication with moving the fetching version into a constructor is still that a DraftKey could be made before said instance of a draft product actually exists, meaning the version fetching would fail. Maybe that's getting into a "DraftKey" vs just a "Draft" as a concept, where they Key maybe really shouldn't have version at all since it's just a uid essentially.

alexrichey:
@fvankrieken Gotcha, so my understanding of the `DraftKey` or `ProductKey` is that it just encapsulates product path info on DO (regardless of whether the product has been published). As opposed to representing an actual dataset. Is that right? 

If so, maybe ✂️ing the `logged_version` makes sense, unless that's required? Then one other thought is instead of having a `label` fn, it's pretty standard practice to just override the str() dunder

```
@dataclass
class ProductKey:
    product: str
    version: str

    @property
    def path(self):...
    
    def __str__(self):...


@dataclass
class DraftKey(ProductKey):
    build: str

    @property
    def path(self):...

    def __str__(self):...
```

fvankrieken:
> @fvankrieken Gotcha, so my understanding of the `DraftKey` or `ProductKey` is that it just encapsulates product path info on DO (regardless of whether the product has been published). As opposed to representing an actual dataset. Is that right?
> 
> If so, maybe ✂️ing the `logged_version` makes sense, unless that's required? Then one other thought is instead of having a `label` fn, it's pretty standard practice to just override the str() dunder
> 
> ```
> @dataclass
> class ProductKey:
>     product: str
>     version: str
> 
>     @property
>     def path(self):...
>     
>     def __str__(self):...
> 
> 
> @dataclass
> class DraftKey(ProductKey):
>     build: str
> 
>     @property
>     def path(self):...
> 
>     def __str__(self):...
> ```

Re logged_version, it is sort of needed if we want `version` for `DraftKey` (which is needed if it's extending ProductKey). Which since they're slightly different, maybe it makes sense if one is not extending the other. Might make sense to call these `PublishKey` and `DraftKey`, which are extensions of `ProductKey` (or have `ProductKey` class really be an alias for the union of the two key types. Really just need it for these functions that can take either)

fvankrieken:
For the latter, this would be valid I believe

```python
ProductKey = Union["PublishKey", "DraftKey"]

### or in 3.10
ProductKey: TypeAlias = "PublishKey" | "DraftKey"

def get_source_data_versions(product_key: ProductKey) -> pd.DataFrame:
    ...
```

alexrichey:
@fvankrieken Is the distinction between `logged version` and `version` for the case when it's a "latest" folder? I'm just thinking that `logged_version` feels like an attribute of a product itself, whereas these *Key objects seem more to represent path info. 

I wonder if it would make sense to be more specific than `Key` . e.g. calling these DraftPath and PublishedPath. 

DraftPath has a `product` and a `build`
PublishedPath has a `product` and a `version`

Then you can override their __str__ , and implement a `url()` method for each? (and you could either have a parent class with a `url` method, or just use a Union type as you suggested)

dcpy/connectors/edm/publishing.py
```diff
@@ -18,65 +17,55 @@
...
-
-def _get_product_url(product_key: ProductKey) -> str:
-    return f"{BASE_URL}/{_get_product_path(product_key)}"
+def get_version(product_key: ProductKey) -> str:
+    """Given product key, gets version
+    Assumes existence of version.txt in output folder
+    """
+    return s3.get_file_as_text(BUCKET, f"{product_key.path}/version.txt").splitlines()[
+        0
+    ]
```
fvankrieken, line 61: This is the silliest decision `black` as ever made

damonmcc (approved):

PR #231, opened 2023-09-14
use Template DB as an end-to-end test (part 1/2)

### LINKED ISSUE #192: use Template Database as an end-to-end test
## motivations

we'd like to be able to run an end-to-end test of our build pipeline

this probably should
- include the major stages of all builds: ~Extract~, Load, Transform, Export
  - Extract is currently done by data-library, our builds expect source data to have been extracted/archived
- cover some the most common approaches seen in data products
- have realistic and fun/interesting outputs 📊

this probably shouldn't
- be expected to test every `dcpy` function or every single approach seen in our data products (e.g. a SQL procedure to apply corrections)
- take long to build
- force significant changes to the rest of the codebase (yet)

## thoughts

this is meant to test our data platform, not a particular data product

seems like we're most interested in covering the Load and Transform stages. Extract is performed by [`data-library`](https://github.com/NYCPlanning/db-data-library/)

started [a TDB wiki page](https://github.com/NYCPlanning/data-engineering/wiki/Product:-TDB) to start ideas for this fictional data product and perhaps be a model for what we'd like data product wikis to be

future stuff:
- be our most "up-to-date" product by being a codebase where we can implement our latest favored build approaches

## progress
- [x] part 1: add test structure and minimal Template DB build
- [ ] part 2: expand and test Template DB build

damonmcc:
related to #192

## approach
- as much as possible, avoid changes to `dcpy` and other repo-level code
- focus on testing the Template DB build process and it's use of high-level code like `recipes`
- design this with future product build generalization in mind (e.g. to move product build logic out of github actions to python, we may like new `dcpy` or `build` modules)
- get feedback on this testing approach here and add the rest of the Template DB product in a follow-up PR

## changes
- refresh Template DB product's github actions to align with monorepo standards and to remove unecessary steps
- add end-to-end test in TDB product so that it [fails in an expected way](https://github.com/NYCPlanning/data-engineering/actions/runs/6200692552/job/16835896730#step:11:74) (because the TDB build didn't do anything yet)
- add TDB build script for Load stage to make test pass

## notes
- [the first sensical end-to-end test failure](https://github.com/NYCPlanning/data-engineering/actions/runs/6199072353/job/16830856980#step:9:34)


.github/workflows/template_test.yml
```diff
@@ -3,52 +3,48 @@ on:
...
     uses: ./.github/workflows/template_build.yml
     with:
       run_export: false
       logging_level: DEBUG
       build_note: a test build
     secrets: inherit
+
+  # * end-to-end test must run in build workflow until persistent DB is used
+  # end_to_end_tests:
+  #   name: Run end-to-end tests
```
damonmcc, line 50: planning to move the test here when the build DB is in edm-data rather than a temporary github action service

products/template/build_scripts/load.py
damonmcc, line 1: chose this build approach for now because:
- easy to call the whole build stage module in a github action
- having logic in functions rather then `__main__` makes it more modular and testable

products/template/recipe.yml
damonmcc, line 1: just a first pass, these will likely change in Part 2 when this "product" is more fleshed out and transformations are added

products/template/recipe.yml
```diff
@@ -0,0 +1,24 @@
+name: Template DB
```
alexrichey, line 1: I think that for this, it might be nice to use some very, very small datasets (as opposed to Pluto). And Would be nice to test resolution of missing versions to latest, so maybe ✂️ one of the versions keys?

products/template/build_scripts/load.py
```diff
@@ -0,0 +1,22 @@
...
+
+def load_source_data():
+    recipe_lock = recipes.plan_recipe(Path(RECIPE_PATH))
+    with open(RECIPE_LOCK_PATH, "w", encoding="utf-8") as f:
+        logger.info(f"Writing recipe lockfile to {RECIPE_LOCK_PATH}")
+        yaml.dump(recipe_lock, f)
+    with open(RECIPE_PATH, "r", encoding="utf-8") as f:
+        recipe_lock = yaml.safe_load(f)
+    logger.info("Importing Recipes")
+    recipes.import_recipe_datasets(recipe_lock)
```
alexrichey, line 18: can also spit out the versions csv too!

products/template/tests/test_build.py
```diff
@@ -0,0 +1,44 @@
...
+        "dpr_greenthumb",
+        "lpc_historic_districts",
+        "lpc_landmarks",
+    ]
+    built_in_tables = [
+        "spatial_ref_sys",
+        "geography_columns",
+        "geometry_columns",
+    ]
+    all_build_db_tables = sorted(get_schema_tables(BUILD_SCHEMA))
```
alexrichey, line 25: Should we just filter out the built-in tables from `get_schema_tables`? We'd never want to reference them, I assume. Looks like there's just one other usage of `get_schema_tables`, so no risk.

products/template/build_scripts/load.py
```diff
@@ -0,0 +1,22 @@
+import os
```
alexrichey, line 1: Unfortunately, can't comment on empty files. But you can ✂️ the empty __init__, right?

products/template/build_scripts/__init__.py
fvankrieken, line 1: @alexrichey you can absolutely comment on empty files!

products/template/build_scripts/load.py
damonmcc: hope so, I was swatting down `dcp not found` errors so I'll try without this one

.github/workflows/template_build.yml
```diff
@@ -44,36 +47,28 @@ jobs:
...
-      - name: Setup build environment
-        working-directory: ./
-        run: ./bash/docker_container_setup.sh
+          python -m products.template.python.run_logging
 
-      - name: 1) Data loading
+      - name: Load
         run: |
-          echo "PLACEHOLDER running Data loading"
+          python3 -m products.template.build_scripts.load
```
alexrichey, line 61: I think we'd probably want default behavior to be running these from the product folder itself (in this case `./products/template`. Otherwise your imports will be janky when you run as a module, e.g. you'd have to do something like `from products.template.build_scripts import load` or only use relative imports. 

I think. Python imports are confusing.

products/template/recipe.yml
damonmcc: see https://github.com/NYCPlanning/data-engineering/pull/231/commits/a533e64051b0656b92ed4ff9db7becf8ce4eadf0

products/template/tests/test_build.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+from dcpy.utils.postgres import get_schema_tables
+from .. import BUILD_SCHEMA
+
+
+# test build stage Load
+@pytest.mark.end_to_end()
+def test_load_db_tables():
+    expected_source_tables = [
```
alexrichey, line 9: if you wanted to be fancy, you could read the recipes.yml and extract the expected datasets

alexrichey (approved):
Once the pathing stuff is sorted, this looks 👍. Good stuff!

damonmcc:
@alexrichey 
> Once the pathing stuff is sorted, this looks 👍. Good stuff!

think I've got it!

PR #232, opened 2023-09-14
refactor recipes/publishing to use pydantic + typer

alexrichey:
Refactors edm.connectors recipes and publishing to use pydantic for class serialization/deserialization, and typer for CLI. I'd recommend reviewing commit by commit. 

For publishing, I wrapped our existing API with typer methods mainly because of typer defaults. Typer resolves defaults really nicely if you call it via CLI. If you call via python but omit an arg, typer will return a `typer.models.OptionInfo` which isn't something we want.


dcpy/connectors/edm/recipes.py
```diff
@@ -101,35 +100,29 @@ def fetch_sql(name: str, version: str, local_library_dir):
...
-    recipe_name: str,
-    *,
-    version="latest",
-    local_library_dir=LIBRARY_DEFAULT_PATH,
-    import_table_name: Optional[str] = None,
-):
+def import_dataset(dataset: Dataset, *, local_library_dir=LIBRARY_DEFAULT_PATH):
     """Import a recipe to local data library folder and build engine."""
-    config = get_config(recipe_name, version)
+    config = get_config(dataset.name, dataset.version)
```
fvankrieken, line 105: `version` field is optional for a `Dataset` defaulting to `None`, while `get_config` expects a (non-optional) string and won't coerce `None` to `"latest"` but will return a weird path

dcpy/connectors/edm/recipes.py
```diff
@@ -140,9 +140,11 @@ def plan_recipe(recipe_path: Path) -> Recipe:
     # Determine the recipe version
     if recipe.version is None and recipe.version_strategy is not None:
         if recipe.version_strategy == VersionStrategy.bump_latest_release:
+            if recipe.version_type is None:
+                raise Exception("Recipe needs a 'version_type' to bump")
             prev_version = publishing.get_latest_version(recipe.product)
             recipe.version = versions.bump(
-                prev_version, bumped_part=recipe.version_type
+                prev_version, bumped_part=str(recipe.version_type)
```
fvankrieken, line 147: If you use an `else` after the exception, will mypy know that this is a `str` rather than `Optional[str]`? I assume that's what the casting is converting it from here. Would find that preferable personally, just to not have to cast

fvankrieken:
A couple small comments. Really like it!

dcpy/connectors/edm/recipes.py
alexrichey: @fvankrieken Well I'll be. It does. Good job `mypy`!

dcpy/connectors/edm/recipes.py
alexrichey: Good call... Will address that.

alexrichey: @fvankrieken resolved [here](https://github.com/NYCPlanning/data-engineering/pull/232/commits/50fc9862f3c26ab06cc02fbe054d675a3a31e95f)

dcpy/connectors/edm/recipes.py
```diff
@@ -77,210 +82,189 @@ def get_latest_version(name):
...
+    if lock_file is not None:
+        with open(lock_file, "w", encoding="utf-8") as f:
+            logger.info(f"Writing recipe lockfile to {str(lock_file.absolute())}")
+            yaml.dump(recipe.model_dump(), f)
+    else:
+        print(recipe)
+
+
+@app.command("import")
+def import_datasets(recipe_file: Path = _typer_recipe_file_opt):
```
fvankrieken, line 234: Just making sure I understand intended CLI functionality - this function is only supposed to take in a "planned" recipe, right? since it calls `recipe_from_yaml` which just reads a file without looking up any versions for sources, then calls `import_dataset` which requires that versions are not `"latest"`. So then in python or in bash, you should plan, then call this with the name of the "planned" lockfile

alexrichey: @fvankrieken Yes, that's the gist. The idea is just that we should resolve versions to actual versions (not latest) before we do any importing. Though honestly, not sure the lockfile is something we want to keep doing.

damonmcc: seems nice, like `requirements.in `-> `requirements.txt`

fvankrieken: Cool. Just might also be good to somehow specify that this "import" command doesn't just take any recipe, but a "planned" recipe

alexrichey: @fvankrieken @damonmcc Made it a little more explicit [here](https://github.com/NYCPlanning/data-engineering/pull/232/commits/99484dcd9c0cb928de4f11a37c7772d779c3df57)

fvankrieken (approved):

damonmcc (approved):

Issue #17, opened 2023-06-12
approach to database for builds


damonmcc:
## task list
- [x] #192 
- [x] #269
- [x] #307

an issue for brainstorming

with the understanding that we like the idea of a single Postgres DB to serve as our data warehouse for building all data products, some thoughts:
- the current DB named `edm-data` is persistent and hosted on Digital Ocean. happy to use it
- in Postgres, the hierarchy of "database objects" is Server -> Databases -> Schemas -> Tables
- cross-database queries are not possible
- potential design:
  - three databases: `Test`, `Staging`, `Production`
  - test builds would target the Test DB (local, CI)
  - normal builds would target the Staging DB (local, CI)
  - staged data could be "promoted" to the Production DB by copying it (via `pg_dump`)

## potential data warehouse / build engine design

- use 3 levels of DBs to isolate dev and production data
- the highest level would be the "endpoint" for both exporting and querying build outputs. all data there would have been "promoted" after QAQC
- record build details when writing to a DB
- Test
  - meant to be isolated versions of Staging
  - used by default for non-release builds
  - structure
    - permutations of Staging schemas (with prefixes based on the environment)
- Staging
  - meant to be the production build engine
  - used for release builds
  - structure
    - a schema for all source tables and a table for source details
    - a schema for product details and build details
    - a schema for each product with all releases
- Publishing
  - populated by copying from Staging
  - structure
    - a schema for each product

## plan for trying this out
- document desired database structure(s)
- document high-level build sequence needed to use data warehouse
- change a data product to use this (db-template?)
- change at least one more data product in order to test using common source data (db-zoningtaxlots, db-checkbook?)

damonmcc:
follow-up thoughts on testing:
- we probably don't want a lot a test databases
- to allow isolated testing for multiple developers/PRs in the same repo:
  - maybe create and name schemas in the Test DB to have a prefix/suffix unique to the source of the testing
  - e.g. `test.pluto_pr_80`, `test.pluto_workflow_dispatch_branch_name`, `test.ztl_local_damon`

damonmcc:
so far the thoughts above are all about the Transform part of ELT. some thoughts on the Load part:
- reminder that cross-DB queries aren't possible. so if source data isn't already in a Transformation DB (e.g. `edm-date.staging`), it'd have to be copied into it before a build. maybe we like that? the postgres utility [`pg_dump`](https://www.postgresql.org/docs/current/app-pgdump.html) would work!
[...]

PR #234, opened 2023-09-18
test dcpy and add PostgresClient

### LINKED ISSUE #137: `git_branch()` creates subfolders on PRs
when building `db-checkbook` in PR tests, this folder structure was created during [`run_export()`](https://github.com/NYCPlanning/data-engineering/blob/main/products/db-checkbook/build_scripts/export.py#L17):
Screenshot 2023-08-09 at 10 54 28 AM

I think the expected behavior is:
- for push/workflow_dispatch, folder structure like `branch_name` -> `version_1`, `version_2`, `latest`
- for pull request trigger, folder structure like `127_merge` -> `version_1`, `version_2`, `latest`

in addition to getting the expected `git_branch()` string in the PR case, it may be more useful to have that result be something like `pr_prnumber` or `pr_branch_name`? the word `merge` doesn't seem useful

damonmcc:
related to #192 and #17 
resolves #137 

refactors and tests some dcpy utils in preparation for building products in a shared persistent postgres DB and expanding the Template DB end-to-end testing

## motivation
- to standardize how we build products in our persistent postgres DB, we'd like to use DB schemas to organize and isolate build tables (a DB schema is a group of tables. is this pretty much [namespaces](https://www.dremio.com/wiki/namespaces/)?)
- in order to avoid having to specify the schema in all SQL queries (would be a huge pain) we can set the "search path" of a DB connection so that all queries use the specified schema
- currently, [our postgres utils](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/utils/postgres.py#L11-L12) use a single environment variable to create a `sqlalchemy` Engine at the top of the module
  - this forces us to construct the search path in bash in a github action ([example](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/knownprojects_build.yml#L36)) rather than in python
  - our postgres uitls that use the psql CLI need a string rather than an Engine
  - it seems we'd prefer to create and reuse a single `sqlalchemy` Engine during the lifetime of a build (also seems to be [the recommended approach](https://docs.sqlalchemy.org/en/20/core/connections.html#basic-usage))

## changes
- move and call dcpy tests
- fix TDB use of recipes
- add PostgresClient class to manage DB connection and most postgres utils
- add git.run_name git.event_name
- refactor recipes to use PostgresClient
- update the QA tests and page that used deleted postgres utils


dcpy/utils/postgres.py
```diff
@@ -8,41 +8,170 @@
...
+            FROM pg_class c
+            JOIN pg_namespace n ON n.oid = c.relnamespace
+            WHERE c.relname = ':table_name'
+            AND n.nspname = ':table_schema';
+            """,
+            {"table_schema": AsIs(table_schema), "table_name": AsIs(table_name)},
+        )
+        return int(row_counts["row_count"][0])
+
+    # ! DEPRECATED in favor of dcpy.connectors.recipes
```
damonmcc, line 141: could just delete this if preferred

dcpy/connectors/edm/recipes.py
```diff
@@ -248,7 +256,10 @@ def import_datasets(planned_recipe_file: Path = _typer_recipe_file_opt):
     """
     logger.info("Importing Recipe Datasets")
     recipe = recipe_from_yaml(planned_recipe_file)
-    [import_dataset(ds) for ds in recipe.inputs.datasets]
+    pg_client = postgres.PostgresClient(
+        schema=git.run_name(),
```
alexrichey, line 260: @damonmcc Let's extract this to be an env var

.github/workflows/dcpy_test.yml
```diff
@@ -15,8 +15,23 @@ jobs:
...
+      BUILD_ENGINE_SERVER: PLACEHOLDER_VALUE
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Run python tests
+        run: |
+          python3 -m pytest ./dcpy/test/ -s --verbose --verbose --cov-config=pyproject.toml --cov .
```
damonmcc, line 37: this gives us [a test coverage report](https://github.com/NYCPlanning/data-engineering/actions/runs/6250559203/job/16969711745#step:4:31) which can be configured to be more accurate ([example](https://github.com/NYCPlanning/data-engineering/actions/runs/6250559203/job/16969711076#step:4:66)) using the pyproject.toml ([example config](https://github.com/NYCPlanning/data-engineering/blob/main/products/template/pyproject.toml#L9))

I didn't add such configurations in this PR

dcpy/utils/postgres.py
fvankrieken: in the words of @alexrichey, ✂️

dcpy/connectors/edm/recipes.py
damonmcc: see [be9a76f](https://github.com/NYCPlanning/data-engineering/pull/234/commits/be9a76f1b4995afb4ea58d15b68baaf2a89e2e8b)

was tempting to add an argument to `import_datasets()` but that felt like a violation of it's intent as a CLI command

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,163 @@
...
+        )
+        self.engine = create_engine(
+            self.engine_uri,
+            isolation_level="AUTOCOMMIT",
+        )
+        self.create_schema(self.schema)
+
+    def execute_query(self, query: str, placeholders: Optional[dict] = None) -> None:
+        if placeholders is None:
+            placeholders = {}
```
fvankrieken, line 59: Nit of nits, but connection.execute can just take `parameters=None` (`None` is the default value), so we don't need this

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,163 @@
...
+        with self.engine.connect() as connection:
+            connection.execute(statement=text(query), parameters=placeholders)
+
+    def execute_select_query(
+        self,
+        query: str,
+        placeholders: Optional[dict] = None,
+    ) -> pd.DataFrame:
+        if placeholders is None:
+            placeholders = {}
```
fvankrieken, line 69: Same as above

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,163 @@
...
+            "geometry_columns",
+        ]
+        table_names = [
+            table_name
+            for table_name in all_table_names
+            if table_name not in postgis_tables
+        ]
+        return table_names
+
+    # def get_table_columns(self, table_schema: str, table_name: str) -> list:
```
fvankrieken: Why are these commented out for now?

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,142 @@
...
+            "geometry_columns",
+        ]
+        table_names = [
+            table_name
+            for table_name in all_table_names
+            if table_name not in postgis_tables
+        ]
+        return table_names
+
+    def get_table_columns(self, table_schema: str, table_name: str) -> list:
```
fvankrieken, line 122: table_schema should maybe have default of the schema of the client? This would simplify calls in qa app to just `client.get_table_columns(f"{table_name}")`

Applies to other functions as well

apps/qa/src/source_report_utils.py
```diff
@@ -102,9 +99,9 @@ def compare_source_data_row_count(source_report_results: dict) -> dict:
 
 
 def create_source_data_schema() -> None:
-    schema_names = get_schemas()
+    schema_names = pg_client.get_schemas()
     if QAQC_DB_SCHEMA_SOURCE_DATA not in schema_names:
-        schema_names = create_sql_schema(table_schema=QAQC_DB_SCHEMA_SOURCE_DATA)
+        schema_names = pg_client.create_schema(table_schema=QAQC_DB_SCHEMA_SOURCE_DATA)
```
fvankrieken, line 104: client creates schema when initialized, right? So this can be cut out

dcpy/utils/postgres.py
damonmcc: I'm thinking maybe no argument and no default since (for now) the client is designed to be limited to a single schema and `public`?

fvankrieken: I think there are definitely possible cases in the future where we'd like to query across schemas, in general I would prefer the client is scoped to a specific database but there's nothing limiting, say, referencing other schemas explicitly in a query, so we don't necessarily need to impose that limitation on some of these functions.

products/template/tests/test_build.py
```diff
@@ -15,11 +22,11 @@ def test_load_db_tables():
...
         "lpc_historic_districts",
-        "lpc_landmarks",
+        "nyc_landmarks",
     ]
-    actual_source_tables = get_schema_tables(BUILD_SCHEMA)
-
-    assert actual_source_tables == expected_source_tables
+    actual_tables = pg_client.get_schema_tables(BUILD_SCHEMA)
+    for table in expected_source_tables:
+        assert table in actual_tables
```
alexrichey, line 26: Any reason not to just throw them in a set and assert on equality?

damonmcc: after a full build, there are more tables than just source tables in the schema.

could do other set operations but I'm not sure that the logs of a failed test would list the actual and expected values. when I did something like `assert [e in actual for e in expected]` the pytest logs show something like `<Iterable ... >` instead of the actual values

dcpy/utils/postgres.py
```diff
@@ -8,9 +8,15 @@
 from typing import Optional
 
 
+def _format_schema_name(schema: Optional[str] = None):
```
alexrichey, line 11: nit: probably wouldn't make this optional. I feel like it should just raise an exception if schema is "" or None.

alexrichey:
Apologies, Still reading. Will give this a better look tomorrow morning!

dcpy/utils/postgres.py
damonmcc: agreed. but since the majority of product builds still use the default `public` schema in their build DB, I didn't wanna force a change in all workflows to declare the schema yet, or distribute this check/formatting to multiple places

the function may not even be worth it since `git.run_name` should always cover the issue of dashes from branch names, so I dropped it in [8da4ecd](https://github.com/NYCPlanning/data-engineering/pull/234/commits/8da4ecd78ff79327fc87b77bfb1d3ec0746323ff)

dcpy/connectors/edm/recipes.py
```diff
@@ -109,33 +113,39 @@ def fetch_sql(ds: Dataset, local_library_dir):
...
+    postgres.execute_file_via_shell(pg_client.engine_uri, sql_script_path)
+
+    with pg_client.engine.begin() as con:
+        con.execute(text(f"ALTER TABLE {dataset.name} SET SCHEMA {pg_client.schema};"))
 
-    with postgres.build_engine.connect() as con:
         con.execute(
             text(f"ALTER TABLE {dataset.name} ADD COLUMN data_library_version text;")
         )
-        con.commit()
```
alexrichey, line 126: @damonmcc these are now taken care of by autocommit mode?

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,131 @@
 from typing import Optional
 
 
-BUILD_ENGINE_RAW = os.environ["BUILD_ENGINE"]
-build_engine = create_engine(BUILD_ENGINE_RAW)
+def generate_engine_uri(server_url: str, database: str, schema: Optional[str] = None):
```
alexrichey, line 11: nit: should the schema default just be "public"?

dcpy/connectors/edm/recipes.py
damonmcc: yup!

dcpy/utils/postgres.py
damonmcc: a default sounds good!

I think I didn't make it a PostgresClient method because functions like `postgres.execute_query_via_shell` and `postgres.execute_file_via_shell` can _only_ use an engine URI, so thought this function would be useful outside of PostgresClient

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,131 @@
...
+        self.engine_uri = generate_engine_uri(
+            server_url=server_url, database=database, schema=schema
+        )
+        self.engine = create_engine(
+            self.engine_uri,
+            isolation_level="AUTOCOMMIT",
+        )
+        self.create_schema()
+
+    def execute_query(self, query: str, placeholders: Optional[dict] = None) -> None:
```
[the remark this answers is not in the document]
damonmcc: way back when, `execute_select_query` and `execute_query` were created to call different types of queries

`execute_query` is currently used for stuff like `DROP SCHEMA IF EXISTS :schema_name CASCADE` and `VACUUM (ANALYZE)` where no return is expected

I don't see any examples of connection.execute returning anything, so I'm not sure that it does?

`execute_select_query` uses `pd.read_sql` for queries expecting a return value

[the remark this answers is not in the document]
damonmcc: oops I forgot about `execute.fetchall()`

dcpy/utils/postgres.py
```diff
@@ -8,41 +8,131 @@
...
+    cmd = f"psql {engine_uri} -v ON_ERROR_STOP=1 {sql_statement}"
     if os.system(cmd) != 0:
         raise Exception(f"Command has errors! {cmd}")
 
 
+class PostgresClient:
+    def __init__(
+        self,
+        schema: Optional[str] = None,
+        database: str = os.environ["BUILD_ENGINE_DB"],
```
alexrichey, line 35: So just a note... these defaults are resolved when the module is imported, not when the class is instantiated. So this will throw an exception in a product like facdb, where we just use the `postgres.insert_copy` function, unless we set `BUILD_ENGINE_DB` and `BUILD_ENGINE_SERVER` there as well

dcpy/connectors/edm/recipes.py
```diff
@@ -12,9 +13,10 @@
 
 from dcpy import DCPY_ROOT_PATH
 from dcpy.utils import s3
-from dcpy.utils.logging import logger
 from dcpy.utils import postgres
+from dcpy.utils import git
```
alexrichey, line 17: ✂️

.github/workflows/dcpy_test.yml
```diff
@@ -15,8 +15,23 @@ jobs:
...
         run: mypy ./dcpy
+
+  unit_tests:
+    name: Unit tests
+    runs-on: ubuntu-22.04
+    container:
+      image: nycplanning/build-geosupport:latest
+    env:
+      BUILD_ENGINE_DB: PLACEHOLDER_VALUE
+      BUILD_ENGINE_SERVER: PLACEHOLDER_VALUE
```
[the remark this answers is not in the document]
fvankrieken: Ohh - are those actually materialized at time of module loading? As in, even if the defaults are never accessed, `os.environ["BUILD_ENGINE_DB"]` is executed and "stored" as the default arg?

dcpy/utils/postgres.py
fvankrieken: Ah yeah this addresses my question just asked [here](https://github.com/NYCPlanning/data-engineering/pull/234/files#r1333285247)

alexrichey:
@damonmcc What's the rationale for moving `dcpy_tests` into `dcpy`? I can think of a few reasons against, the primary one being that the tests shouldn't be part of the API for `dcpy`, and we wouldn't want our pure module code to have a dependencies upon dev dependencies, ie test frameworks. Also, now any module in `dcpy` can freely import `dcpy.test.*` which, while I don't think anybody _would_ do that, it's conventional to make that a little more difficult. Also, you typically will want to easily distribute your package sans tests, (ie in our case, just copying the `dcpy` folder)

So to pick an outside example, [boto3](https://github.com/boto/boto3) has a strict separation between the two.

alexrichey (changes requested):
I love the PostgresClient implementation. So much simpler to work with. 

For changes, I think unless there's compelling reason, we should put the tests module back outside. Then there's the env vars in the client init, which I think will cause some problems.

damonmcc:
@alexrichey appreciate the module vs tests details! do you like the old name of `dcpy_test` or shortening it to `tests`?

TLDR: no compelling reason, will move em back! 

I felt inclined to put it in the dcpy directory because this repo isn't only dcpy and the only tests in that directory were for dcpy. in the boto3 example, everything in that repo is related to that single python package, so isn't it kinda of `boto3/boto3` and `boto3/tests`?

from a quick look at a random local virtual environment I have, [pandas](https://github.com/pandas-dev/pandas/tree/main/pandas) doesn't have a strict separation. I see the tests in my file system. same for packages like [jsonschema](https://github.com/python-jsonschema/jsonschema/tree/main/jsonschema), scipy, sklearn, matplotlib, numpy, and others

dcpy/utils/postgres.py
damonmcc: yea I'd rather not force setting envars when they won't be used.

per [my comment above](https://github.com/NYCPlanning/data-engineering/pull/234#discussion_r1331836745), this is the current approach in our `s3.client` class. but looks like every function in the module would instantiate a client, so maybe the approach is good to keep using there?

alexrichey:
@damonmcc I suppose this is all pretty build-system specific, but I think you're touching on something correct though - it would be nice to have the tests grouped with the package, not in the project root. And maybe the stuff @fvankrieken is doing with setuptools would allow us to exclude the `dcpy.test` folder in builds.

fvankrieken:
> @damonmcc I suppose this is all pretty build-system specific, but I think you're touching on something correct though - it would be nice to have the tests grouped with the package, not in the project root. And maybe the stuff @fvankrieken is doing with setuptools would allow us to exclude the `dcpy.test` folder in builds.

We could have `src` and `tests` within the `dcpy` folder

alexrichey:
There's also [this](https://github.com/tweag/python-monorepo-example/tree/main/libs) approach, which has a somewhat telescoping folder structure, but is nicely structured to group functionality while also preserving namespaces. It would make me a little bit sad to introduce `src` into our module names, e.g. `dcpy.src.utils.*`. 

But it's neat that we can do this stuff now, via setuptools

dcpy/utils/postgres.py
damonmcc: @alexrichey see [41a0d8f](https://github.com/NYCPlanning/data-engineering/pull/234/commits/41a0d8f77038249f2abbc61b90732c1b387dd088)

my typing conundrum was because I was trying to pass along `database` rather than `self.database`

damonmcc:
@alexrichey @fvankrieken

just a ping. hope the new commits cover everything, probably not worth doing a `dcpy.src` folder change in this PR

probably best to do a squash & merge with this one?

dcpy/utils/postgres.py
```diff
@@ -7,42 +7,143 @@
...
+        database: Optional[str] = None,
+        server_url: Optional[str] = None,
+    ):
+        self.schema = schema
+        self.database = database if database else os.environ["BUILD_ENGINE_DB"]
+        self.server_url = (
+            server_url if server_url else os.environ["BUILD_ENGINE_SERVER"]
+        )
+
+        self.engine_uri = generate_engine_uri(self.server_url, self.database, schema)
```
alexrichey, line 54: @damonmcc It's a little confusing to have both an `engine_uri` and a `server_url` attribute. Should we just inline this `generate_engine_uri` call?

alexrichey (approved):
Have one suggested edit to the PostgresClient, but other than that, great stuff!

dcpy/utils/postgres.py
damonmcc: true. having an `engine_uri` attribute seems more useful than `server_url`, like if a build creates a PostgreClient but then has to execute sql via shell, it could look like `execute_file_via_shell(pg_client.engine_uri, file_path)`

so maybe something like this then:
```python
self.engine_uri = generate_engine_uri(
    server_url if server_url else os.environ["BUILD_ENGINE_SERVER"],
    self.database,
    schema
)
self.engine = create_engine(
    self.engine_uri,
    isolation_level="AUTOCOMMIT",
)
```

damonmcc: see [8f15ee3](https://github.com/NYCPlanning/data-engineering/pull/234/commits/8f15ee38f341a8005dbab3baa4ced2ade3922c9c)

alexrichey (approved):

fvankrieken (approved):

Issue #18, opened 2023-06-14
Environment norms/standardization


fvankrieken:
Another brainstorming issue. Prior to our monorepo, we were generally moving towards dev containers. Do we want one main dev container that we use for all parts of the repo? This also then ties in to environments for github actions - are we pulling from a custom built docker image? Running setup.sh scripts?

Some thoughts 1 vs multiple dev container definitions

Pros of 1 dev container definition/Cons of multiple:
- don't have to build new containers for different projects, once you've built for one you don't need to for others
- eliminate some potential redundancies of having the same things installed in multiple containers, storage concerns
- overall simplicity - keep everything using same versions of things, etc

Cons (or some pros of having multiple container definitions)
- More flexibility per project with multiple dev containers
- Maybe less alignment with cloud environment - for github actions, we probably want to be using minimal overhead for producing datasets, not pulling in some massive image
- Potentially more flexibility if we go with multiple - geosupport has a big image, would be nice to not necessarily use in our "standard" image
- If we have 1 mega-container, it might be slightly tougher to keep track of what's actually needed for each project (easy to work around, write more explicit documentation, but still something to think about
- potentially more annoying to iterate - if it's a long time to pull/build and we need to test new changes, slower cycle

Random thoughts around answers
- Have one general mega-container for local development, have geosupport, latest gdal, etc. It seems as long as we pull from latest gdal, we have all the functionality we need with just a few lines of bash (rather than needing to use a gdal docker image) so wouldn't be hard to just add that on to geosupport docker image.
- publish maybe 3-4 images to dockerhub. These can be used in github actions, as well as if someone needs to pull something lighter weight for their dev container
  - a "base" image - no gdal, no geosupport, but has mc, python (and the packages we use) and other things our most basic pipelines depend on
  - gdal (base + gdal-bin, gdal python)
  - geosupport (current geosupport one maybe fine? Or add on mc)
  - "full" - gdal, geosupport, etc

@damonmcc

fvankrieken:
Couple tangential notes on gdal - 
- looks like we cannot get 3.6 via apt/apt-get, only 3.4. I found some docs that said to use ppa:ubuntugis/ppa as source to get latest gdal-bin, but seems that's not needed any more, but that regardless of source the latest available is 3.4. So need a slightly more manual install if we're not using a docker image

fvankrieken:
Series of issues at end of this PR sort of summarize issue

https://github.com/NYCPlanning/data-engineering/pull/15

[...]

PR #236, opened 2023-09-18
Docker Maintenance

fvankrieken:
Related to discussions in #18

Couple main notes here
- bring in https://github.com/NYCPlanning/docker-geosupport repo to monorepo
- add "dev" docker image to list of images we publish regularly (build-base, build-geosupport, docker-geosupport)
- Consolidate logic around building of images
- Update some logic around publishing of qa app. Keeping that one a bit more isolated since it is a somewhat distinct entity (and is a containerized app rather than just general-use container), but it should use the same constraints.txt file for python packages

I would go commit-by-commit for review. Biggest one is commit 5 - that's the one that focuses on publishing logic, modularity, etc. That's where I would like the most feedback as well, went in circles a bit about how to make this a bit modular. Now, docker folder/publishing process looks like this
- one `publish.sh` file in docker folder. Get's called with arg which corresponds to docker image, and contains logic around what pre-build steps need to occur (looking up latest geosupport version for example)
- one folder per docker image. 
  - Dockerfile
  - Requirements.txt
    - each of these is just a list of packages. We have a "main" requirements.in still in the `python` folder, which gets compiled and then edited slightly to a "main" constraints.txt file which pins version for all other requirements.txt files. Currently, these are the requirements in these docker folders and in the qa app folder 
  - setup.sh
- `setup.sh`s refer to config.sh in parent directory for shared functionality, constants
- build images just contain build-related python packages (and those required for dcpy), dev image has things like mypy, sqlfluff, black
- any image which has geosupport (docker-geosupport, build-geosupport, dev) will all be versioned by geosupport version as well as keeping `latest` tag up to date

[Facdb run](https://github.com/NYCPlanning/data-engineering/actions/runs/6254149211) using new image (build-geosupport)
[Pluto run](https://github.com/NYCPlanning/data-engineering/actions/runs/6253947733) using new image (build-base)

There's also a companion db-gru-qaqc pr: https://github.com/NYCPlanning/db-gru-qaqc/pull/262. 3.11 as base container broke something with the [poetry install](https://github.com/NYCPlanning/db-gru-qaqc/actions/runs/6239021291), so I regenerated poetry lock file with 3.11. [Passing run there after update](https://github.com/NYCPlanning/db-gru-qaqc/actions/runs/6239353035)


bash/utils.sh
```diff
@@ -335,3 +335,48 @@ function archive {
...
+    docker push $image_name:$version
+}
+
+
+function build_and_publish_versioned_docker_image {
+    local path_to_docker_file=$1
+    local image_name=$2
+    local version=$3
+    shift 3
+    #if docker_tag_exists $image_name $version; then
```
fvankrieken, line 369: will uncomment before merging

bash/utils.sh
```diff
@@ -335,3 +335,48 @@ function archive {
...
+    echo "$DOCKER_PASSWORD" | docker login -u $DOCKER_USER --password-stdin
+}
+
+
+function docker_tag_exists() {
+    docker manifest inspect $1:$2 >/dev/null
+}
+
+
+function build_and_publish_docker_image {
```
fvankrieken, line 350: rationale for these two is that the bottom one is likely better practice - if we really have a more rigid versioning (like geosupport), would be good to not change an image once it's published. But that being said, sometimes flexibility is good so I have this one too.

dcpy/connectors/__init__.py
fvankrieken, line 1: These are needed when installing dcpy with pip - otherwise these folders are not auto-discovered as submodules

bash/compile_python_packages.sh
```diff
@@ -18,3 +18,5 @@ rm --force ${path_to_requirements}/requirements.txt
 echo -e "🛠 compiling from ${path_to_requirements}/requirements.in"
 CUSTOM_COMPILE_COMMAND="./${RELATIVE_SCRIPTPATH} ${path_to_requirements}" python3 -m piptools compile --output-file=${path_to_requirements}/requirements.txt ${path_to_requirements}/requirements.in
 echo -e "✅ done compiling ${path_to_requirements}/requirements.txt"
+
+sed -e 's/\[[^][]*\]//g' $path_to_requirements/requirements.txt > constraints.txt
```
fvankrieken: This converts cases like `coverage[toml]==7.3.1` to `coverage==7.3.1`. The former is a valid requirement, pinnin coverage to 7.3.1 and installing option `toml` of the package, but it's not valid in a constraints file used in pip install

.github/workflows/publish_docker.yml
```diff
@@ -1,21 +1,89 @@
...
+        default: false
+      docker-geosupport:
+        description: "Build 'docker-geosupport' image"
+        type: boolean
+        required: true
+        default: false
   
 jobs:
-  build:
+  build-base:
```
fvankrieken: These are done explicitly instead of with matrix because I simply could not figure out how to conditionally run matrix jobs in a simple manner - it's not supported by default

.github/workflows/test.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - main
+  schedule:
+    - cron: "0 0 * * *"
```
fvankrieken, line 10: Figured it'd be good to be running nightly just in case some breaking changes make it in (two PRs in a short time which never get rebased into each other before testing/merging). This file needed a bit of maintenance to make this work - basically, on PRs only changed things get tested, while anything else (workflow dispatch/schedule) runs everything

docker/dev/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+source config.sh
+set -e
+
+apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install $DEV_APT_PACKAGES
+
+install_geosupport
+
+python3 -m pip install -e ./dcpy -r requirements.txt
```
alexrichey, line 10: @fvankrieken Maybe I don't quite follow... But shouldn't we be doing an editable install of ./dcpy on the version in the mounted volume? 

In general, love this approach. Would allow us to stop futzing with the python path just to include dpcy

fvankrieken: Yeah, good catch I was literally about to comment on this line that this is the one that maybe is a little off.

Mainly because 
1) you are right, editable install here doesn't really make sense in terms of mechanism, and
2) this is the "complete" image for now... i.e. the one that also gets used for tests. So ideally it'd just work out of the box, but actually as I write this, tests should be using the dcpy in the codebase anyways

So it seems like as a solution
1) don't install here
2) install as part of "docker setup script" that currently gets run as first step of many jobs

Per quick irl discussion, going to do the same for build containers

alexrichey:
@fvankrieken One other concern is that it looks like `dcpy`'s init file references .env files at the project root. Maybe we remove that, since the path would be non-existent when dcpy lives in site-packages
```
DCPY_ROOT_PATH = Path(__file__).resolve().parent
REPO_ROOT_PATH = DCPY_ROOT_PATH.parent

load_dotenv(REPO_ROOT_PATH / ".env")
```

fvankrieken:
python3 -m pip install -e ./dcpy is causing a weird issue - symlink in /home/vscode/.local/lib/python3.11/site-packages/easy-install.pth is to /workspace/dcpy instead of /workspace. Editing it manually fixes imports

fvankrieken:
> @fvankrieken One other concern is that it looks like `dcpy`'s init file references .env files at the project root. Maybe we remove that, since the path would be non-existent when dcpy lives in site-packages
> 
> ```
> DCPY_ROOT_PATH = Path(__file__).resolve().parent
> REPO_ROOT_PATH = DCPY_ROOT_PATH.parent
> 
> load_dotenv(REPO_ROOT_PATH / ".env")
> ```

It doewsn't return an error if the path isn't valid. I'm personally fine with leaving as is, how would you prefer to have this handled? .envs are currently used only in local dev, so we could have some other way of ensuring that changes to .env are available in pipelines, but maybe would specify on a by-pipeline basis then?

alexrichey (dismissed):
Awesome

alexrichey (dismissed):

alexrichey:
> It doewsn't return an error if the path isn't valid. I'm personally fine with leaving as is, how would you prefer to have this handled? .envs are currently used only in local dev, so we could have some other way of ensuring that changes to .env are available in pipelines, but maybe would specify on a by-pipeline basis then?

@fvankrieken Yeah, if it doesn't cause issues, then I'm fine as well. Another non-urgent issue would be here:
https://github.com/NYCPlanning/data-engineering/blob/ca1c81c07c03b93d26e29f80b82a20750939e1b2/dcpy/connectors/edm/recipes.py#L25
where we'd potentially be installing recipe datasets in `.../site-packages/.library`. Probably want to just default to the directory from which you execute python.

damonmcc (dismissed):

damonmcc:
thanks for pulling this off! re [this](https://github.com/NYCPlanning/data-engineering/pull/236#discussion_r1333080191) and [that](https://github.com/NYCPlanning/data-engineering/pull/236#issuecomment-1731898201), wouldn't object to doing those in a follow-up PR

fvankrieken:
Cleaned up in a rebase and got all tests using appropriate images

alexrichey (approved):

PR #237, opened 2023-09-18
fix potential infinite loop in retreiving workflows

fvankrieken:


dcpy/connectors/github.py
```diff
@@ -55,17 +55,20 @@ def _get_workflow_runs_helper(url, params=None):
 
 def get_workflow_runs(
     repo, workflow_name=None, items_per_page=100, total_items=100, page_start=0
-):  ## 100 is to be max
+):  ## 100 is to the max
```
damonmcc, line 58: nit: `100 is the max` ?

fvankrieken: TO THE MAX

fvankrieken: fixed

damonmcc (approved):

PR #238, opened 2023-09-19
Updates for 23C GRU QA

fvankrieken:
Commit 1 gets devdb publish via data library working
- [working action](https://github.com/NYCPlanning/data-engineering/actions/runs/6237296463)

Commit 2 cleans up some display stuff in gru qaqc page
1. displaying geosupport version above table to make it clear these results are only for the chosen geosupport version
2. expanding elements saying that no check has been run rather than just having empty cells

Before:
image

After:
image

### DIFF
--- a/.github/workflows/developments_publish.yml
+++ b/.github/workflows/developments_publish.yml
@@ -1,16 +1,18 @@
 name: DevDB - 🚀 Publish
 
 on:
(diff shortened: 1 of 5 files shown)


damonmcc (approved):
nice and love screenshots!

PR #249, opened 2023-09-25
format sql files in CPDB

damonmcc:


fvankrieken:
There's the slightest bit of logic change in projectcategorization.sql - instead of huge OR clause, use `like any(ARRAY["string1", "string2"... "string100"])`, mainly so that the fluffer doesn't freak out

fvankrieken:
@damonmcc you can't "review" since this was your PR originally, but it should be good to go

pyproject.toml
```diff
@@ -8,6 +8,7 @@ exclude_rules = [
     "references.qualification",
     "structure.unused_cte",
     "structure.column_order",
+    "structure.subquery",
```
fvankrieken, line 11: Sqlfluff doesn't like subqueries within the `JOIN` parts of a query. I'm open to enforcing that, but [this file](https://github.com/NYCPlanning/data-engineering/blob/5e93eab1aae7f8c536ab5cc9cf622aad1ed4c437/products/cpdb/sql/analysis/projects_dollars_mapped_categorized_sponsor.sql) (and another similar one) were violating. The joins could be done away with by using a bunch of functions in the select clause (`SELECT COUNT(*) FILTER (WHERE typecategory = 'Fixed Asset') AS fixedassett`) but I don't know if that's particularly more readable. Could also do CTEs instead, but again, don't know if a long list of CTEs here would improve legibility

pyproject.toml
```diff
@@ -27,6 +28,8 @@ max_line_length = 120
 # e.g. -1 means use all processors but one. 0 means all cpus.
 processes = -1
 
+large_file_skip_byte_limit = 40000
```
fvankrieken, line 31: Default is 20k. There's one file that was 31k

products/cpdb/sql/budget.sql
fvankrieken, line 1: These deleted files (budget, commitments, projects) were replaced with "budget_fisa.sql", etc some time ago

products/cpdb/sql/projectscategorization.sql
```diff
@@ -5,810 +5,705 @@
...
 WHERE (
-upper(description) LIKE '%VEHICLE%'
-OR upper(description) LIKE '%AMBULANCE%'
-OR upper(description) LIKE '%BUSES%'
-OR upper(description) LIKE '%TRUCK%'
-OR upper(description) LIKE '%TRAILER%'
-OR upper(description) LIKE '%VANS%'
+    upper(description) LIKE any(
+        ARRAY[
+            '%VEHICLE%',
```
fvankrieken, line 13: Maybe worth just putting these in a separate file to be read in? Or in the `data` folder and loaded as a table? That actually maybe makes the most sense.

alexrichey (dismissed):
Cool - I think I get the gist. Wasn't sure what, if anything, besides formatting happened in commit [3](https://github.com/NYCPlanning/data-engineering/pull/249/commits/f2352ae512d0ddf9b0d2a911da1b199e064a2568) but I assume it was nothing too dramatic.

fvankrieken:
Commit 4 finalized now and working properly

damonmcc:
approved!

fvankrieken (approved):

PR #251, opened 2023-09-26
Incorporate data library

fvankrieken:
Definitely should be reviewed by commit

1. Dump in data library repo, with no edits other than removing generated docs
2. Cleanup - move actions to runnable location, remove .gitignore, etc
3. Remove doc generation code
4. Add support for gdal python bindings in dev container
5. Switch base image from osgeo to python 3.11 slim like our other images, and associated needed maintenance
6. black formatting
7. mypy maintenance - get it passing for data library and add it to tests

Tested by running locally as well as running cloud actions
- [workflow run that published image to "ubuntu-staging" in dockerhub](https://github.com/NYCPlanning/data-engineering/actions/runs/6344202000)
- [workflow run that pulled "ubuntu-staging" image to use data library](https://github.com/NYCPlanning/data-engineering/actions/runs/6354443793)


data-library/library.sh
```diff
@@ -16,7 +16,46 @@ function library_execute {
...
+
+    export CPLUS_INCLUDE_PATH=/usr/include/gdal
+    export C_INCLUDE_PATH=/usr/include/gdal
+}
+
+function poetry_version {
+    echo $(poetry show $1) | sed -r 's/^.*version : ([0-9\.]+) .*$/\1/'
+}
+
+function poetry_lock {
```
fvankrieken, line 49: This is what should be used now to regenerate lock file - ensures that python gdal version matches what's installed.

fvankrieken:
Only one outstanding question really - the html docs. I currently added them to the `workspace/docs` folder. They're a lot but they are nice, they'll make something like [this](https://nycplanning.github.io/db-data-library/library/index.html) semi-automatically (... I think? Not hugely familiar with github.io). It's a lot of files. But it seems fine. 

Other thing is how they get generated. Right now this workflow generates it and commits. Primary use of the workflow is to publish docker image. That's sort of convenient - when we're updating docker image is likely when it makes sense to update docs. However, we don't like commits directly to main... not sure of most elegant way to do this, other than maybe make it a bit manual

fvankrieken:
> Only one outstanding question really - the html docs. I currently added them to the `workspace/docs` folder. They're a lot but they are nice, they'll make something like [this](https://nycplanning.github.io/db-data-library/library/index.html) semi-automatically (... I think? Not hugely familiar with github.io). It's a lot of files. But it seems fine.
> 
> Other thing is how they get generated. Right now this workflow generates it and commits. Primary use of the workflow is to publish docker image. That's sort of convenient - when we're updating docker image is likely when it makes sense to update docs. However, we don't like commits directly to main... not sure of most elegant way to do this, other than maybe make it a bit manual

Maybe open a PR to main with updated docs?

fvankrieken:
Doing some tinkering with commit 7+ but rest is stable

bash/compile_python_packages.sh
```diff
@@ -14,6 +14,9 @@ python3 -m pip install --upgrade pip-tools wheel
 echo -e "🛠 deleting ${path_to_requirements}/requirements.txt"
 rm --force ${path_to_requirements}/requirements.txt
 
+# Set GDAL verion
+sed -i -e "s/GDAL==.*$/GDAL==$(gdal-config --version)/g" ${path_to_requirements}/requirements.in
+
```
damonmcc, line 19: is the GDAL version here a result of `config.sh`?

fvankrieken: No, that's from the gdal binary (`gdal-bin`)

fvankrieken:
Think we're all set now - [running action from #7](https://github.com/NYCPlanning/data-engineering/actions/runs/6356749223) (with push) generated #260, this seems like automated-enough doc generation for me

bash/compile_python_packages.sh
damonmcc: ah and `gdal-bin` is listed in `config.sh` and installed via `setup.sh` which is called in every Dockerfile

damonmcc:
for the docs, seems nice to keep them as-is and revisit later to maybe do the same for `dcpy`

maybe we'd wanna generate on merges to main? I'm pretty sure a github.io site is just a reflection of a repos files on a chosen branch (usually main)

damonmcc:
seeing a merge conflict with `pyproject.toml`

alexrichey (approved):
good stuff! Re the docs, my preference would be to just delete them, since this is more of a CLI tool than a module. Would be neat to do something similar with `dcpy` though

alexrichey (approved):

damonmcc (approved):

damonmcc:
@fvankrieken any followup TODOs for this around scheduled actions in the old repo or the github marketplace action?

PR #255, opened 2023-09-28
finish creation of Template DB

### LINKED ISSUE #192: use Template Database as an end-to-end test

damonmcc:
resolves #192 

## goals
- add transform and export stages to TDB so that it can be a minimal model of our product build approaches
- create a mock data product that isn't very complex and is potentially fun/interesting to DE
- create a foundation for refining/expanding TDB later

## changes
- add a [Product design](https://github.com/NYCPlanning/data-engineering/wiki/Product:-TDB#product-design) section to the TDB wiki to brainstorm the design of this mock product
- change the TDB end-to-end tests to reflect the product design
- use new `postgres.PostgresClient` and the persistent DB server in the build
  - had to create a `db-template` DB in `edm-date`. named the DB to align with our product names in the `edm-publishing` S3 bucket
- shorten the list of source datasets
- add the transformation stage
- add the export stage
- lint all TDB sql files in PR tests
- update the [TDB wiki](https://github.com/NYCPlanning/data-engineering/wiki/Product:-TDB) to be more about product details and less about internal/dev
- update the [TDB readme](https://github.com/NYCPlanning/data-engineering/blob/template-db-build/products/template/README.md) to be more about internal/dev details and less about product details


products/template/tests/test_build.py
damonmcc, line 1: these are the end-to-end tests. they illustrate the tables and files which are expected to exist after a build

products/template/sql/templatedb.sql
damonmcc, line 1: this sql scripts generates the final primary table

products/template/seeds/boroughs.csv
damonmcc, line 1: this is a csv for Borough-related mappings. it's inspired by the concept of [seed files](https://docs.getdbt.com/docs/build/seeds#overview)

alexrichey (approved):
This is great. Looking forward to making use of this to test changes. I do wonder if we want to delete the existing s3 artifacts at the beginning of the test, to avoid false positives for successive test runs on a branch. (unless you're doing that already, and I missed it) But it's not urgent.

damonmcc:
@alexrichey good call! I got false positives when I reset and pushed one commit at a time expecting more tests to fail

I'll note this and other enhancements in a follow-up issue (duplicative landmarks records, determining the boroughs of green spaces, etc.)

fvankrieken (approved):

PR #257, opened 2023-09-28
Fvk pluto publish mvt

fvankrieken:
Just a basic implementation for now

- gha that takes in product, filepath to output to translate to MVT, min zoom and max zoom for generated mvt files
- publishing cli target to download file
- DCPY function that calls vector translate. For initial rough draft wanted to just keep it specific to this and not more generalized
- s3 upload cli target to upload a folder

[Workflow run](https://github.com/NYCPlanning/data-engineering/actions/runs/6344991950) that uploaded to de-sandbox

### DIFF
--- a/.github/workflows/publish_generate_mvt.yml
+++ b/.github/workflows/publish_generate_mvt.yml
@@ -63,4 +63,4 @@ jobs:
             --folder-path ${{ inputs.product }}_mvt \
             --s3-path mvt/${{ inputs.product }}/${{inputs.version}}/ \
             --acl private \
-            --copy-contents
+            --contents-only
--- a/.github/workflows/publish_generate_mvt.yml
+++ b/.github/workflows/publish_generate_mvt.yml
@@ -1,0 +1,66 @@
+name: 🚀 Generate MVT for a published dataset
+run-name: "🚀 Generate MVT: ${{ inputs.product }}"
+
+on:
+  workflow_dispatch:
+    inputs:
+      product:
+        description: "Which product to publish"
+        type: choice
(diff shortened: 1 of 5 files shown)


.github/workflows/publish_generate_mvt.yml
```diff
@@ -0,0 +1,66 @@
...
+          - cbbr
+          - checkbook
+          - colp
+          - cpdb
+          - developments
+          - facilities
+          - pluto
+          - zoningtaxlots
+      version:
+        description: "Published version"
```
damonmcc, line 21: really like the specificity of "Published version"

dcpy/utils/s3.py
```diff
@@ -271,3 +280,41 @@ def get_file_as_stream(bucket: str, path: str) -> BytesIO:
...
+    copy_contents: bool = typer.Option(
+        None,
+        "--copy-contents",
+        help="If true, uploads local folder into target folder. If false, uploads contents of local folder instead",
+    ),
+):
+    if copy_contents is not None:
+        include_foldername = not copy_contents
+    else:
+        include_foldername = None
```
damonmcc, line 308: if the default for `copy_contents` were False rather than None, could it then just be passed to `upload_folder`?

damonmcc (approved):

dcpy/utils/s3.py
fvankrieken: Oh that's a good point, this was clunky. Flipped the boolean logic of the argument to make it all cleaner as well

damonmcc (approved):

PR #262, opened 2023-10-02
Refactor facdb pipelines (remove decorator pattern)

alexrichey:
The decorators were causing issues when implementing dcpy recipes here, so I've gone ahead an inlined all the decorated functions. (I was able to automate most of this via the magic of emacs macros). For cleanliness, I thought a separate PR might be nice. 

Anyway, for transformation, process was something like this. For a sample fn like below:

```
@Export
@Function1B(
    street_name_field="address_street_name",
    house_number_field="address_building",
    borough_field="address_borough",
    zipcode_field="address_zip",
)
@FunctionBL(bbl_field="bbl")
@FunctionBN(bin_field="bin")
@Prepare
def my_func():
    ....
```

1. `@Prepare` -> moved the dataset-download call into the CLI, and the sanitizing into the fn body immediately after existing code 
2. `@Export`  -> moved into the CLI
3. For the others -> moved them into the function body above `export`, then reversed their order. (since decorators execute from bottom up)

You could review commit-by-commit, but it might be easiest to just review all changed files. For the pipelines, as I mentioned the process was largely automated, so I'm not sure you need to check every single function (though don't let me stop you!). I believe the control flow should be basically identical to how it was before. May not be pretty, but now it's explicit. 

I tested this by loading data. I didn't do a full build to compare to previous versions. But last time I build FacDB, it was pretty fussy, so I expect that completed loading is a sufficient test. However, happy to do a full build comparison if folks want that assurance.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,6 +79,11 @@ jobs:
               - dcpy/**
               - docker/**
               - python/**
+            facilities:
+              - products/facilities/**
+              - dcpy/**
+              - docker/**
+              - python/**
             knownprojects:
               - products/knownprojects/**
               - dcpy/**
@@ -110,6 +115,11 @@ jobs:
     needs: check_changes
     if: github.event_name != 'pull_request' || contains(needs.check_changes.outputs.path_filters, 'checkbook')
     uses: ./.github/workflows/checkbook_test.yml
+    secrets: inherit
+  facilities:
+    needs: check_changes
+    if: github.event_name != 'pull_request' || contains(needs.check_changes.outputs.path_filters, 'facilities')
+    uses: ./.github/workflows/facilities_test.yml
     secrets: inherit
   knownprojects:
     needs: check_changes
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,9 +86,9 @@ jobs:
               - python/**
             knownprojects:
               - products/knownprojects/**
-              - dcpy/**
-              - docker/**
-              - python/**
+              # - dcpy/**
+              # - docker/**
+              # - python/**
             zap:
               - .github/workflows/zap_test.yml
               - .github/workflows/zap_build_mapzap.yml
--- a/products/facilities/facdb/utility/prepare.py
+++ b/products/facilities/facdb/utility/prepare.py
@@ -57,8 +57,5 @@ def prepare(name: str) -> pd.DataFrame:
         print("reading from cached data")
         df = pd.read_pickle(pkl_path)
 
-    # Apply custom wrangler
     df["source"] = name
-    df = df.replace({np.nan: None})
-    df = df.drop_duplicates()
     return df
--- a/products/facilities/facdb/utility/prepare.py
+++ b/products/facilities/facdb/utility/prepare.py
@@ -1,6 +1,4 @@
-import json
 import os
-from functools import wraps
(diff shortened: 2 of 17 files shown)


fvankrieken:
This is much more sane. The separate functions per dataset is still interesting, but a good repo to think about how we want more "modular" code in the long term (like what specific preprocessing steps exist per dataset, including geocoding, running them more automatically at time of archival with data library, etc). Worth a brainstorm sometime.

For now, one request - can you get rid of the `None` default args for pipelines? I think that's fine at this point

Are you opposed to `mypy`ing the code for this product and getting that in tests as well? A bit unrelated to this change, but still something that we should maybe try to do a bit more religiously, especially as we start using `dcpy` more to make sure we're catching breaking changes

alexrichey:
> This is much more sane. The separate functions per dataset is still interesting, but a good repo to think about how we want more "modular" code in the long term (like what specific preprocessing steps exist per dataset, including geocoding, running them more automatically at time of archival with data library, etc). Worth a brainstorm sometime.
> 
> For now, one request - can you get rid of the `None` default args for pipelines? I think that's fine at this point
> 
> Are you opposed to `mypy`ing the code for this product and getting that in tests as well? A bit unrelated to this change, but still something that we should maybe try to do a bit more religiously, especially as we start using `dcpy` more to make sure we're catching breaking changes

Yes, good call about the `None`s. Just pushed a new commit with that change. And yeah, I'll put in a `mypy` check. I'd used it to catch bugs locally - might as well fix the issues and include the check in CI

products/facilities/facdb/utility/prepare.py
```diff
@@ -43,30 +41,21 @@ def version_from_config(name, read_version):
...
-        else:
-            # read from cache
-            print("reading from cached data")
-            df = pd.read_pickle(pkl_path)
-
-        # Apply custom wrangler
-        df = func(df)
-        df["source"] = name
-        df = df.replace({np.nan: None})
-        df = df.drop_duplicates()
```
damonmcc, line 69: this is `sanitize_df` now?

damonmcc (approved):
nice!

damonmcc (dismissed):

.github/workflows/test.yml
```diff
@@ -65,6 +65,8 @@ jobs:
               - products/cbbr/**
             checkbook:
               - products/checkbook/**
+            facilities:
+              - products/facilities/**
```
fvankrieken, line 69: and `dcpy/**`!

fvankrieken: The other option to what I did [here](https://github.com/NYCPlanning/data-engineering/pull/276/commits/4e212c67b28570f752fc955c5b8eb008da571f66) would be just mypying on all (or most of) the codebase separate from triggering product-specific tests.

alexrichey: @fvankrieken Ah, I see these have been expanded since my last rebase! And good callout... I'll copy the existing convention for now, and maybe down the line we can extract a `mypy` job that always runs against `dcpy` and certain products. My only concern would be that `mypy` is quite slow... but I'm not entirely sure how slow. Maybe we could speed it up via actions caching with mypy. Looks like there are a few published actions that will cache things.

alexrichey:
Note, the failing mypy check will succeed after rebuilding the docker image. I think.

fvankrieken (approved):

Issue #250, opened 2023-09-26
support AE pilot project


damonmcc:
per brainstorms on 9/21 and 9/25

## context
- AE is building a POC of a new app ([planning doc in sharepoint](https://nyco365.sharepoint.com/:w:/s/NYCPLANNING/itd/labs/EfZY8LZXYW9Lob5410yHhBMB3HwdfiwZ2TCN3ZVTaYAXrQ?e=uZsojM&wdOrigin=TEAMS-WEB.teams.fileLink&wdExp=TEAMS-CONTROL&wdhostclicktime=1695222032777&web=1))
- AE and DE would like to minimize the amount of data engineering performed by AE and implement an ideal AE tech stack

## details and ideas
- DE is helping with the planning and generating of source data for the app
- the planned source data is:
  - Mapbox Vector Tiles (MVTs) of MapPLUTO and Zoning Districts
  - a Postgis DB with MapPLUTO
- likely want to generate MVTs during the DE Publish stage of a product build
  - but DE doesn't build Zoning Districts, so should it be considered more of a pull from AE that generates all MVTs?

## needs scoping
- which zoom levels should the POC MVTs be generate at?
- which MapPLUTO and Zoning Districts columns should be in the MVTs?
- what data cleaning does AE need (e.g. drop invalid geometries, simplify geometries)?

damonmcc:
informed AE that MVTs can and have been pushed to the S3 bucket `de-sandbox`

will touch base w/ AE about the other desired data (tables with PLUTO data)

damonmcc:
per AE, [this Carto SQL query](https://github.com/NYCPlanning/labs-layers-api/blob/develop/data/sources/pluto.json#L8) is what they currently use to select relevant Pluto fields for MVTs

for efficiency, MVTs should have a minimal number of fields, especially non-categorical fields

fvankrieken:
Ah I can address that - once we get your little PR merged I'll add something to filter fields

damonmcc:
from meeting on 10/10

AE action items
- create DE issues with source data specifications

DE action items
-

PR #263, opened 2023-10-03
generate MVTs with specific zoom levels

damonmcc:
related to #250 

## motivation
- per chat with @TylerMatteo, zoom levels from 15-22 would be nice for prototyping

## notes
- this is meant to get the required zoom level feature working, not refactor @fvankrieken's initial approach for generating MVTs
- this includes limiting the source data fields because
  - the MVT files are too large to upload before the github action times out (6 hours)
  - it's a required feature from AE anyway (MVTs are meant for displaying geometries, not storing relational data)

## action runs
- [success](https://github.com/NYCPlanning/data-engineering/actions/runs/6565693509) using zooms 15-16 and 4 fields.  12 mins to upload
- [successfully raising gdal error](https://github.com/NYCPlanning/data-engineering/actions/runs/6395267663/job/17358522331#step:5:13) when exceeding max zoom of 22

### DIFF
--- a/.github/workflows/publish_generate_mvt.yml
+++ b/.github/workflows/publish_generate_mvt.yml
@@ -25,6 +25,16 @@ on:
         description: "Filepath within output folder to shapefile to convert"
         type: string
         required: true
+      min_zoom:
+        description: "Minimum zoom level"
+        type: string
+        required: true
+        default: '0'
+      max_zoom:
+        description: "Maximum zoom level"
+        type: string
+        required: true
+        default: '5'
 
 jobs:
   publish:
--- a/.github/workflows/publish_generate_mvt.yml
+++ b/.github/workflows/publish_generate_mvt.yml
@@ -37,7 +37,8 @@ on:
         default: '5'
 
 jobs:
-  publish:
+  build:
+    name: Build MVT files
     runs-on: ubuntu-22.04
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


fvankrieken (dismissed):

fvankrieken (dismissed):

damonmcc:
the [last run](https://github.com/NYCPlanning/data-engineering/actions/runs/6397461633) timed out after 6 hours and ~ uploaded 197,413 files at ~630MB in `de-sandbox`

will limit the number of fields and try using `upload_folder()` if it isn't already being used

.github/workflows/publish_generate_mvt.yml
```diff
@@ -49,24 +60,27 @@ jobs:
...
-      - name: Publish MVT
+      - name: Upload MVT
         run: | 
           python3 -m dcpy.utils.s3 \
-            -b de-sandbox \
+            --bucket de-sandbox \
             --folder-path ${{ inputs.product }}_mvt \
             --s3-path mvt/${{ inputs.product }}/${{inputs.version}}/ \
             --acl private \
+            --max-files 800000 \
```
fvankrieken, line 85: Was this number reached by testing

dcpy/utils/geospatial.py
```diff
@@ -1,3 +1,4 @@
+# ? move this file to dcpy.builds?
```
fvankrieken, line 1: Good question. I think we definitely want geospatial utils (including geosupport related things), but some of the interaction (or discrete steps) are definitely going to be more build related. Good here for now, but as this gets added to and matures we'll need to think about that division for geospatial-related tasks

fvankrieken (approved):
I would like an answer on the `max_files` but looks great

dcpy/utils/geospatial.py
damonmcc: and I was gonna ask in this PR if we like merging comments like this to main?

it's not a `TODO` or a `HACK`, but I made up a list in the [wiki](https://github.com/NYCPlanning/data-engineering/wiki/Operations#style-guides) for standard comment types we might like to use

.github/workflows/publish_generate_mvt.yml
damonmcc: looks like [this run](https://github.com/NYCPlanning/data-engineering/actions/runs/6395608302/job/17359648274#step:6:68) hit 617,457 when using zoom levels of 15-20

dcpy/utils/geospatial.py
alexrichey: Yeah, I think this gets at something worth teasing out (not necessarily now, though). In particular, it feels a little misplaced to have product-level information like product column-names baked into our pure utils. Problem for tomorrow, though

.github/workflows/publish_generate_mvt.yml
alexrichey: can we zip these up? Or do they need to be separate on s3?

alexrichey: Bc we're eventually just dumping them all into PostGIS, right?

damonmcc: that sounds right. but I don't think they've setup a postigs DB yet or chosen an ideal file format they'd like for the current prototyping

also the `gdal.VectorTranslate` call that generates the tiles could even just have a DB as a destination (although there may be a desire to archive all versions as files)

PR #269, opened 2023-10-03
add `dcpy.builds` and use persistent build DB for CPDB

damonmcc:
related to #17 

while converting CPDB from using a github action DB to our persistent DB, I stumbled into some changes that seem worth getting feedback on before expanding to changing other products

changes
- use the edm-data DB server for the CPDB build
- test CPDB on relevant changes in PRs
- add `dcpy.builds` modules as a way to generalize/abstract common build logic
  - `environment` for setup things
  - `load` for the ubiquitous first stage of every build
  - to avoid circular imports in `dcypy`, seems like we could think of this as `utils` -> `connectors` -> `builds`
- use `dcpy.builds` in the Template DB build


products/template/build_scripts/load.py
damonmcc, line 1: because it doesn't do anything unique in this stage, a product-level script is much shorter now!

products/cpdb/bash/01_dataloading.sh
```diff
@@ -1,9 +1,17 @@
...
 # Reference tables
-run_sql_file "sql/_create.sql"
+echo "Dropping and creating build schema '$BUILD_ENGINE_SCHEMA'"
+# create build schema
+run_sql_command \
+    "
+    DROP SCHEMA IF EXISTS ${BUILD_ENGINE_SCHEMA} CASCADE;
+    CREATE SCHEMA ${BUILD_ENGINE_SCHEMA};
+    "
 create_source_data_table
```
damonmcc, line 13: on our journey from bash to python, this could now be done via `dcpy.builds.enviroment`

that would help make future orchestration cleaner right?

dcpy/utils/s3.py
damonmcc, line 1: CPDB has more than 20 output files

.github/workflows/test.yml
```diff
@@ -87,6 +89,14 @@ jobs:
...
     if: github.event_name != 'pull_request' || contains(needs.check_changes.outputs.path_filters, 'cbbr')
     uses: ./.github/workflows/cbbr_test.yml
+    secrets: inherit
+  cpdb:
+    needs: check_changes
+    if: github.event_name != 'pull_request' || contains(needs.check_changes.outputs.path_filters, 'cpdb')
+    uses: ./.github/workflows/cpdb_build.yml
+    with:
+      run_export: true
+    secrets: inherit
```
damonmcc, line 99: did this rather than creating a `cpdb_test.yml` to run a build. I figured we want less workflow files and probably want to discourage product-level pytests

.github/workflows/cpdb_build.yml
```diff
@@ -17,27 +17,24 @@ jobs:
...
 
+      - name: Set build environment variables
+        run: |
+          # # replace dashes with underscores to create a valid postgres schema name
+          BUILD_ENGINE_SCHEMA=$(echo ${BUILD_ENGINE_SCHEMA} | tr - _)
+          echo "BUILD_ENGINE_SCHEMA=$BUILD_ENGINE_SCHEMA" >> "$GITHUB_ENV"
+          # set postgres schema search path to prioritize BUILD_ENGINE_SCHEMA
+          BUILD_ENGINE=${BUILD_ENGINE_DATABASE}?options=--search_path%3D${BUILD_ENGINE_SCHEMA},public
+          echo "BUILD_ENGINE=$BUILD_ENGINE" >> "$GITHUB_ENV"
+
```
damonmcc, line 37: `PostgresClient` would handle this if the CPDB build ran SQL queries via python

dcpy/utils/s3.py
```diff
@@ -28,6 +28,7 @@
     "public-read",
     "public-read-write",
 ]
+MAX_FILE_COUNT = 50
```
damonmcc, line 31: CPDB has more than 20 output files

products/template/build_scripts/load.py
damonmcc, line 1: because it doesn't do anything unique in this stage, this product-level script is much shorter now!

damonmcc, line 1: because it doesn't do anything unique, this product-level script is much shorter now!

dcpy/connectors/edm/publishing.py
```diff
@@ -98,7 +98,7 @@ def upload(
     draft_key: DraftKey,
     *,
     acl: s3.ACL,
-    max_files: int = 20,
+    max_files: int = s3.MAX_FILE_COUNT,
```
alexrichey, line 101: if you're just using the default, you can ✂️ this, right?

dcpy/builds/load.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+from dcpy.utils import postgres
+from dcpy.connectors.edm import recipes
+
+
+def load_source_data(
+    pg_client: postgres.PostgresClient,
+    recipe_path: Path,
+    recipe_lock_path: Path,
```
alexrichey, line 9: I think we probably always want to put the lockfile next to the recipe.yml. Maybe ✂️?

products/template/build_scripts/load.py
alexrichey: any reason to have this file now? As opposed to just invoking `python -m dcpy.builds.load` from the action??

dcpy/builds/load.py
damonmcc: done in [ca46774](https://github.com/NYCPlanning/data-engineering/pull/269/commits/ca46774cfbba5b04c35d0d75456231a564d16e4d)

could go further and construct the lockfile path in `recipes.plan` if prefered, but that'd force changes to `pluto_build.yml`

alexrichey: yeah, I'm all for it. And pluto should be changed to use this, so all for it.

products/template/build_scripts/load.py
damonmcc: I think I was just erring on the side of minimal change. since most products' `01_dataloading.sh` aren't this simple, wanted this to be a model for where that non-dcpy uniqueness could live

and invoking `python -m dcpy.builds.load` from the action would mean designing our generalized build code to be CLI tools, something I imagine we don't want to invest in so we can do python-based Orchestration™️ soon

dcpy/builds/load.py
damonmcc: mmmm looks like that'd be a more substantial change to `edm.recipes` than I'd like to do in this PR. since every recipes command after plan expects the lockfile path, they'd all have to be changed. it's a good idea but seems out of scope here

and since the pluto build action [does things](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/pluto_build.yml#L75-L95) in between `edm.recipes plan` and `edm.recipes import`, it'd take some pluto refactoring to use `builds.load.load_source_data`

products/template/build_scripts/load.py
alexrichey: This is a good topic for discussion (and def want to include @fvankrieken in it)... I'm bringing this up because I think you nailed it with this dependency formulation: `utils` -> `connectors` -> `builds`. We have CLI functionality in the edm `recipes` connector, which is probably the wrong place for it. Feels like it should be here. 

> designing our generalized build code to be CLI tools, something I imagine we don't want to invest in
I imagine we _would_ want to do that. Typer has so far made it pretty easy.

dcpy/builds/load.py
```diff
@@ -0,0 +1,23 @@
...
+    recipes.plan(
+        recipe_path,
+        recipe_lock_path,
+    )
+    recipe = recipes.recipe_from_yaml(Path(recipe_lock_path))
+    recipes.write_source_data_versions(recipe_file=Path(recipe_lock_path))
+
+    [recipes.import_dataset(dataset, pg_client) for dataset in recipe.inputs.datasets]
+
+    pg_client.create_table_from_csv(
```
alexrichey, line 20: Do we still need the pg table for source data versions?

.github/workflows/test.yml
fvankrieken: Why do we want to discourage product-level pytests?

products/template/build_scripts/load.py
fvankrieken: I agree that recipes and publishing should ideally have their CLIs scrapped, and it all be from build?. Build CLI seems useful at least short-term. Longer-term if we're in DAG-land and everything is python then it maybe becomes irrelevent

.github/workflows/cpdb_build.yml
fvankrieken: Oh well. If this is being used in a couple build files, maybe make this a bash function or script?

products/template/build_scripts/__init__.py
```diff
@@ -15,7 +15,6 @@
 PRODUCT_S3_NAME = "db-template"
```
fvankrieken, line 15: Leftover from me cleaning up dcpy installation - can you ✂️ line 13 of this file?

dcpy/builds/load.py
damonmcc: it's been nice to be able to see & query the versions

alexrichey: huh... do we need to query the versions in the db? ie. just curious if the csv isn't enough for some reason

damonmcc: being able to query the resolved recipe versions is probably useful and yea seeing the info in dbeaver when a build starts is really nice. it's kind of like a seed file/table

.github/workflows/test.yml
damonmcc: sorry yea that does sound drastic. i guess if all the `products` folders are so "transforms for this product", writing more python doesn't feel as likely as writing more SQL

no doubt we have python code in those folders at the moment though

dcpy/builds/load.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def setup_environment(pg_client: postgres.PostgresClient):
+    pg_client.drop_schema()
+    pg_client.create_schema()
+    # TODO delete draft S3 folder
+
+
+def load_source_data(product: str):
+    product_path = REPO_ROOT_PATH / f"products/{product}"
```
alexrichey, line 17: I may be wrong here, but I suspect, unfortunately, we'll need to supply the path to the product. Since `dcpy` is installed via pip, the `REPO_ROOT_PATH` might have some surprising results when `dcpy` isn't installed in editable mode. Looks like the only other usage of that constant is to source an env var... it may be time to kill it.

dcpy/builds/load.py
```diff
@@ -0,0 +1,61 @@
...
+    build_name = git.run_name()
```
alexrichey, line 18: Weren't we setting an environment var for the schema, rather than invoking git? e.g. the `BUILD_ENGINE_SCHEMA` in `knownprojects_build.yml`?

dcpy/builds/load.py
fvankrieken: Hmm. A build should be a decent test of this, no? We're doing a standard pip install at the beginning of build steps

alexrichey: @fvankrieken it might, so long as you're invoking python from outside our project root, otherwise I _think_ the python interpreter will probably just use the existing dcpy folder. (unless we're deleting that somewhere?)

Testing this locally, and if I replicate the container setup step install locally with just (in proj root)
`python3 -m pip install . -c ./python/constraints.txt`

then cd into, say, the projects folder, then invoke python, I see that 
REPO_ROOT_PATH=/workspace/.venv/lib/python3.11/

if I run python from workspace root, it ignores the installed version (which makes sense) and 
REPO_ROOT_PATH=/workspace/

fvankrieken: Ah makes sense. 

These are intended to be run in the product build folders, right? So maybe a supplied arg for path to folder, with default to `"."` (no repo root path needed)?

fvankrieken:
Couple outstanding comments but this is great. Very excited for where we're getting to with all of this

dcpy/builds/load.py
alexrichey: @fvankrieken I think one of the strengths of installing dcpy via pip is that we aren't tied to running from proj root anymore. I could also see a world in which other teams (e.g. GIS) add `dcpy` as a dependency. So I think we have no guarantees about where `dcpy` lives on a filesystem. But as Tim from AE says, that's a tomorrow issue.

For this, seems most straightforward to expect a path.

dcpy/builds/load.py
damonmcc: **TLDR:** reading and [using](https://github.com/NYCPlanning/data-engineering/blob/edm-build-db/dcpy/utils/git.py#L60) an envar in python has been nicer and more maintainable than doing the same in bash in a github action

I guess asking _why_ we're invoking git leads me to wanting to do things in python

the why seems to be "we need to know the unique name for this draft build build/environment". and our chosen approach is to use the name of the branch or the PR number. since those are git-related values, we currently use `git.py`

and in the majority of cases, we're not invoking git. we're [reading an envar](https://github.com/NYCPlanning/data-engineering/blob/edm-build-db/dcpy/utils/git.py#L37-L41) which exists because it's a [github action default variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

for better separation, `build_name` could be determined and set to an envar well before something like `load_source_data` is called? at first I had separate `load.py` and `environemt.py` files but they only had 1 function in each

alexrichey: > for better separation, build_name could be determined and set to an envar well before something like load_source_data is called?

I think this is the correct approach, with the goal of separating config from app code. Though, should we just use `BUILD_ENGINE_SCHEMA` for sake of consistency?

dcpy/builds/load.py
alexrichey: Oh, yes, I see. Totally sensible!

dcpy/builds/load.py
damonmcc: sure could! the envar name might be too specific if it'd be used to pick an S3 folder, but when it's postgres-related, "schema" is on-point

damonmcc: @alexrichey I went with [d658beb](https://github.com/NYCPlanning/data-engineering/pull/269/commits/d658beb11c898e29adf8d3b5e2d11a43076bfb36) to improve this

since AFIK, envars wouldn't persist across something like
```bash
python3 -m dcpy.builds.setup --product template
python3 -m dcpy.builds.load --product template
```
I thought it'd be nice to add `build_metadata.build_name` to use the `BUILD_ENGINE_SCHEMA` envar. and if the envar isn't set, do `git.run_name()`

if someday soon we can just do `python3 -m dcpy.builds.build --product template`, then maybe setting an envar would work. but I don't think we're there yet

.github/workflows/cpdb_build.yml
damonmcc: true! definitely trying to deprecate this approach too per [this thread below](https://github.com/NYCPlanning/data-engineering/pull/269#discussion_r1349316650)

would love to save ur idea for the follow-up PR that expands the use of the build DB to other products

.github/workflows/pluto_build.yml
```diff
@@ -72,11 +71,19 @@ jobs:
...
+          # set postgres schema search path to prioritize BUILD_ENGINE_SCHEMA
+          BUILD_ENGINE=${BUILD_ENGINE_DATABASE}?options=--search_path%3D${BUILD_ENGINE_SCHEMA},public
+          echo "BUILD_ENGINE=$BUILD_ENGINE" >> "$GITHUB_ENV"
+
       - name: Plan Build
         working-directory: ./
         run: |
-          python3 -m dcpy.connectors.edm.recipes plan -f ./products/pluto/${{ inputs.recipe_file }} -o ./products/pluto/recipe.lock.yml
-          python3 -m dcpy.connectors.edm.recipes write-source-data-versions -f ./products/pluto/recipe.lock.yml
+          python3 -m dcpy.builds.load pluto
```
alexrichey, line 86: 😍

alexrichey (approved):

alexrichey:
@damonmcc Great stuff! Thanks for being so receptive to feedback. I think that there are still a few outstanding issues, namely the pathing issue with `def load_source_data(product: str):` that we should fix sooner than later. It also seems like we need a sane way to deal with environment variables, so maybe that would be a good issue to create. It may even be a `gfi`.

dcpy/builds/load.py
alexrichey: Cool - makes sense. For env var persistance, I think it would be nice to have a dcpy module that sets them. Maybe it would emit a list of keypairs, or write a `.env` file. Agree about just doing it later.

PR #271, opened 2023-10-04
delete out of date todos

damonmcc:


bash/utils.sh
```diff
@@ -196,7 +196,6 @@ function csv_export_drop_columns {
         local schema="public"
     fi
     
-    # TODO schema isn't always public
```
damonmcc, line 199: completed via the use of `${schema}.${table}` below

dcpy/test/connectors/test_github.py
```diff
@@ -3,7 +3,6 @@
 TEST_REPO = "data-engineering"
 
 
-## TODO - these tests don't belong here since this functionality has been moved to dcpy
```
damonmcc, line 6: completed via this file being in `dcpy/test/`, it used to be in the QA app codebase

fvankrieken (approved):

PR #272, opened 2023-10-04
Upload should only upload folder contents

fvankrieken:
🤦


fvankrieken:
Introduced this bug in my pluto publish mvt stuff when I changed the inputs to `s3.upload_folder`

PR #273, opened 2023-10-04
appease mypy in qa app

fvankrieken:


dcpy/utils/s3.py
```diff
@@ -13,6 +19,16 @@
     TimeRemainingColumn,
 )
 
+ACL = Literal[
```
fvankrieken, line 22: This was to make the new s3 typing happy

dcpy/utils/s3.py
```diff
@@ -2,9 +2,15 @@
...
-from typing import Any, Optional
+from typing import Any, Literal, TYPE_CHECKING
 import boto3
+from botocore.response import StreamingBody
 from botocore.client import Config
+
+if TYPE_CHECKING:
+    from mypy_boto3_s3.client import S3Client
+else:
+    S3Client = object
```
fvankrieken, line 13: This was sort of interesting. Makes sense though - for s3 typing we need the additional `boto-stubs` package with the `s3` option. Nice to not have to include an extra package in our build environments, only dev.

fvankrieken:
This ended up being a bit meatier than I expected, mainly because I wanted to fix up a few things in dcpy and that led down a rabbit hole. A couple little things to point out

- started going with `str | None` instead of `Optional`
- using `dict` and `list` instead of `Dict` and `List` for type annotations. We currently have a mix of both. You used to have to use the latter to specify subtype (`List[str]`) but as of 3.9 the lowercase are fine (and built-in - uppercase must be imported from typing)

dcpy/utils/s3.py
damonmcc: are there any other typeing-related imports we'd also wanna put this condition around?

fvankrieken: Not that I can see

damonmcc (approved):
nice! probably wanna squash on merge?

PR #276, opened 2023-10-04
data library package/mypy maintenance

fvankrieken:
Update prompted by #265


fvankrieken:
Not ready after all - some fallout from #273 - shows need for maybe just always mypying everything (some of these errors seem to be from adding `boto-stubs[s3]` to the dev docker image which runs tests

.github/workflows/test.yml
```diff
@@ -50,30 +50,49 @@ jobs:
...
               - data-library/**.yml
+              - docker/**
+              - python/**
             build:
               - .github/workflows/test.yml
               - .github/workflows/build.yml
               - dcpy/**
               - products/template/**
+              - docker/**
+              - python/**
```
fvankrieken, line 66: The docker ones are a little funny in that they generally still require publishing a new image. But regardless in general figure it makes sense to test things if we're changing build env, be it in docker stuff or python req versions

alexrichey (approved):

PR #279, opened 2023-10-05
Fix publishin upload cli typing

fvankrieken:


damonmcc (dismissed):

fvankrieken:
> @fvankrieken This is interesting... I didn't know about the literal type. I'm guessing it would also work to put the acceptable ACLs in a set, and retrieve them in the CLI. Or use an enum value. I feel like that would be simpler, if it works. (Though if you want to move on, this seems workable.)

Yeah the reason I went this route was with `boto-stubs[s3]`, it expects ACL argument to be a literal. I'm assuming we could maybe keep it as a string with some sort of validation but this seemed to align well and make mypy happy. I would like to merge as is for now

damonmcc (approved):

alexrichey:
> > @fvankrieken This is interesting... I didn't know about the literal type. I'm guessing it would also work to put the acceptable ACLs in a set, and retrieve them in the CLI. Or use an enum value. I feel like that would be simpler, if it works. (Though if you want to move on, this seems workable.)
> 
> Yeah the reason I went this route was with `boto-stubs[s3]`, it expects ACL argument to be a literal. I'm assuming we could maybe keep it as a string with some sort of validation but this seemed to align well and make mypy happy. I would like to merge as is for now

@fvankrieken Cool - all for merging and moving on. FWIW, enums seem to work as well. Just poking around with a toy example:
```
from enum import Enum
from typing import Literal

class Greetings(str, Enum):
    hi = "hi"
    yo = "yo"

def lit(x: Literal["yo", "hi"]):
    print(x)


lit(Greetings.yo.value)  # All Good
lit("hi")  # All Good

lit("whats up")  # Fail
```

Anyhow, fun rabbit hole!

fvankrieken:
Is that a runtime error or mypy error?

alexrichey:
> Is that a runtime error or mypy error?
runtime ✅
mypy ❌

PR #283, opened 2023-10-10
Facdb Implement Recipes

alexrichey:
Implements Recipes in FacDB. 

FacDB changes:
- mostly deletions.
- Adds a dispatcher function to `pipelines` that will call methods in the module, and do all the data processing that was being done before in `cli.py` and various places. 
- Adds a recipe.yml

DCPY changes: 
Adds new functionality to recipes

```
name: Test
version: 23v1
product: db-test
inputs:
  dataset_defaults:   <---- NEW
    file_type: csv
    preprocessor:
      module: facdb.pipelines
      function: dispatch
  missing_versions_strategy: find_latest
  datasets:
  - name: has_version_from_env
    version_env_var: VERSION_PREV   <---- NEW
    file_type: csv                      <---- NEW
    preprocessor: {...}                       <---- NEW
```

Specifically this adds to recipes the ability to specify a dataset's
- file_type, 
- version_env_var 
- preprocessor

Or to just set those as defaults one level up, in `dataset.inputs`. 

Callout: We can't yet preprocess sql dumps, so this is just for csv files. 

TODO: 

- [ ] Recipes should error when a preprocessor is spec'd and the file_type is pg_dump. 
- [ ] Fix failing tests. 
- [ ] Finish test for `import_data`
- [ ] Related to above: Recipes should swap to use the PostgresClient. This will make mocking postgres functionality manageable (it currently isn't)

Question:
- This PR was to get things working in FacDB. There is a lot of cleanup, specifically deletions, that could happen. Follow-up PR? Or is this a good time to do it?


dcpy/utils/postgres.py
```diff
@@ -160,6 +160,20 @@ def create_table_from_csv(self, table_name: str, file_path: Path) -> None:
             method=insert_copy,
         )
 
+    def insert_dataframe(self, df: pd.DataFrame, table_name: str):
+        df.to_sql(
+            table_name,
+            con=self.engine,
+            if_exists="replace",
+            index=False,
+            dtype={
```
alexrichey, line 169: This is how it was in facdb. It sets a specific type for fields from geosupport. It seems harmless (and works even if those columns aren't present). But there has to be a better place for this. Suggestions welcome.

products/facilities/facdb/cli.py
```diff
@@ -6,11 +6,14 @@
 import typer
 
 from dcpy.utils import postgres
-from .utility.prepare import read_datasets_yml, prepare
+from .utility.prepare import read_datasets_yml
 from .utility.metadata import dump_metadata
-from .utility.export import export
```
alexrichey, line 11: RIP to one of my favorite lines.

products/facilities/facdb/cli.py
```diff
@@ -92,35 +87,13 @@ def run(
...
-    if there's a scripts section in datasets.yml under this dataset, the
-    sql script will also be executed.
-    facdb run -n {{ name }} to run both python and sql part\n
-    facdb run -n {{ name }} --python to run the python part only\n
-    facdb run -n {{ name }} --sql to run the sql part only\n
-    facdb run --all\n
-    """
+    """ """
     datasets = read_datasets_yml()
-    dataset_names = (
```
alexrichey, line 105: Formerly, this would search for a "script" (ie a python pipeline) and sql file, and conditionally run them. Now, we take care of the pipeline part with recipes, and just have to run the accompany sql files. This should be simplified - we should just specify to run all sql files in a given folder, as order doesn't appear to matter.

alexrichey:
@fvankrieken and @sf-dcp - This still has some work to do, but 👀 would be appreciated. There's a list of TODOs, but the shape of things is pretty settled.

products/facilities/facdb/cli.py
fvankrieken: there's something poetic about it

dcpy/connectors/edm/recipes.py
```diff
@@ -35,26 +37,56 @@ class VersionStrategy(str, Enum):
...
-class Dataset(BaseModel, use_enum_values=True):
+class DatasetType(str, Enum):
+    pg_dump = "pg_dump"
+    csv = "csv"
+
+
+_file_extensions_by_type = {"pg_dump": "sql", "csv": "csv"}
+
+
+class DataPreprocessor(BaseModel, use_enum_values=True, extra="forbid"):
```
fvankrieken, line 48: It's kinda crazy that forbid isn't the default?? Did you find a good reason for that in their docs?

dcpy/connectors/edm/recipes.py
```diff
@@ -35,26 +37,56 @@ class VersionStrategy(str, Enum):
...
+    @property
+    def file_name(self) -> str:
+        return f"{self.name}.{_file_extensions_by_type[self.file_type]}"
+
+    @property
+    def s3_key(self) -> str:
+        return f"datasets/{self.name}/{self.version}/{self.file_name}"
+
+
+class DatasetDefaults(BaseModel, use_enum_values=True):
```
fvankrieken, line 73: I understand how this is used in FacDB but I feel like in general this is complexity that I would prefer not to have in the build/recipe logic. Not 100% set on that, but it just feels like we're starting to nest more complex information in this class than is quite ideal, like maybe down the road, if this recipe is really the _start_ of a build, either the preprocessing has already happened, when we're ingesting data, or it's just part of the build. 

But again, not really relevant for this PR, just more stuff to think about as we iterate

fvankrieken: Since FacDB as is sort of relies on this structure.

fvankrieken: This is more about DataPreprocessor specifically - DatasetType obviously for now is quite useful

dcpy/connectors/edm/recipes.py
```diff
@@ -92,61 +124,68 @@ def get_latest_version(name):
     return get_config(name)["dataset"]["version"]
 
 
-def fetch_sql(ds: Dataset, local_library_dir):
-    """Retrieve SQL dump file from edm-recipes. Returns fetched file's path."""
+def fetch_dataset(ds: Dataset, local_library_dir=LIBRARY_DEFAULT_PATH) -> Path:
```
fvankrieken, line 127: LIBRARY_DEFAULT_PATH is using relative dcpy paths, is that something we want to clear out now that we're installing dcpy as a package? I do like it being in the root of the directory, maybe that's something we should be setting as an env var somehow?

dcpy/connectors/edm/recipes.py
```diff
@@ -92,61 +124,68 @@ def get_latest_version(name):
...
+    """Retrieve dataset file from edm-recipes. Returns fetched file's path."""
     target_dir = local_library_dir / "datasets" / ds.name / ds.version
-    target_file_path = target_dir / (ds.name + ".sql")
+    target_file_path = target_dir / ds.file_name
```
fvankrieken, line 130: All the stuff around dataset types is nice and clean

dcpy/connectors/edm/recipes.py
```diff
@@ -194,7 +233,14 @@ def plan_recipe(recipe_path: Path) -> Recipe:
 
     for ds in recipe.inputs.datasets:
         if ds.version is None:
-            if (
+            if ds.version_env_var is not None:
```
fvankrieken, line 236: This is for a case where we're specifying dataset versions via env var? Just want to be crystal clear

dcpy/utils/postgres.py
fvankrieken: Huh.

dcpy/connectors/edm/recipes.py
```diff
@@ -215,11 +261,21 @@ def get_source_data_versions(recipe: Recipe):
...
+        ds.file_type = ds.file_type or recipe.inputs.dataset_defaults.file_type
+
+
 def recipe_from_yaml(path: Path) -> Recipe:
     """Import a recipe file from yaml, and validate schema."""
     with open(path, "r", encoding="utf-8") as f:
         s = yaml.safe_load(f)
-    return Recipe(**s)
+        recipe = Recipe(**s)
+        _apply_recipe_defaults(recipe)
```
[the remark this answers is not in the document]
alexrichey: Yeah, I definitely prefer not mutating in general. Python doesn't make that as easy as say, F# or Clojure, at least with the class-based approach. But this has me pondering better ways to deserialize these files... Would prefer not to even have to set defaults.

In this case, since it was cumbersome to actually write this with an immutable-data approach, I wanted to make it obvious that the argument to `_apply_recipe_defaults` is being mutated, which is less obvious if we're returning a value from that function.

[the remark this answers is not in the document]
fvankrieken: Ahh, that makes sense. I didn't quite think about the mechanics of returning a new object with a single changed field. Sounds good!

dcpy/connectors/edm/recipes.py
alexrichey: yep, open to suggestion if there's a more obvious name

dcpy/connectors/edm/recipes.py
alexrichey: I think setting root path as an env var is a great idea. And since this is the only remaining usage for the *_ROOT_PATH stuff in `dcpy.__init__` it might be time to ✂️. For now I'll just remove it, and just use the working directory when `local_library_dir` isn't specified.

dcpy/connectors/edm/recipes.py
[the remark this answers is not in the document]
fvankrieken: Not to be a broken record, but fsharp is great here. If you have
```fsharp
let myObj = {
    field1 = "string"
    field2 = "otherstring"
    field3 = 1
)
```
You can just
```fsharp
let myUpdatedObj1 = { myObj with field1 = "newString" }
let myUpdatedObj2 = { 
    myObj with 
        field2 = "updatedstring"
        field3 = 2
}
```

dcpy/connectors/edm/recipes.py
alexrichey: I was very surprised by that, as forbidding by default would never lead to silent errors. The alternative does. 

It also seems like something you'd want to address early in the docs, instead of nestling in a [config](https://docs.pydantic.dev/latest/api/config/) section, where it's not super discoverable.

dcpy/connectors/edm/recipes.py
```diff
@@ -35,26 +37,56 @@ class VersionStrategy(str, Enum):
     bump_latest_release = "bump_latest_release"
 
 
-class Dataset(BaseModel, use_enum_values=True):
+class DatasetType(str, Enum):
+    pg_dump = "pg_dump"
+    csv = "csv"
+
+
+_file_extensions_by_type = {"pg_dump": "sql", "csv": "csv"}
```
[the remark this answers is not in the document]
alexrichey: much cleaner! I always forget that Enums are extensible... anyhow, changed.

alexrichey:
@fvankrieken + @sf-dcp this is ready to go. FYI, the `mypy` failure will be fixed when we rebuild the image

fvankrieken (approved):

dcpy/connectors/edm/recipes.py
```diff
@@ -35,26 +38,60 @@ class VersionStrategy(str, Enum):
     bump_latest_release = "bump_latest_release"
 
 
-class Dataset(BaseModel, use_enum_values=True):
+class DatasetType(str, Enum):
+    pg_dump = "pg_dump"
+    csv = "csv"
+
+
+_dataset_extensions = {"pg_dump": "sql", "csv": "csv"}
```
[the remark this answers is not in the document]
fvankrieken: Like `dataset.value` wasn't typed as a string?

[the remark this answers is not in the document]
fvankrieken: mypy seems happy with this

```python
class DatasetType(str, Enum):
    pg_dump = "pg_dump"
    csv = "csv"

    @property
    def extension(self) -> str:
        match self:
            case DatasetType.pg_dump:
                return "sql"
            case DatasetType.csv:
                return "csv"
```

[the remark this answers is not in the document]
fvankrieken: And below 

```python
class Dataset(BaseModel, use_enum_values=True, extra="forbid"):
    name: str
    version: str | None = None
    version_env_var: str | None = None
    import_as: str | None = None
    file_type: DatasetType | None = None

    def is_resolved(self):
        return self.version is not None and self.version != "latest"

    @property
    def file_name(self) -> str:
        return (
            f"{self.name}.{self.file_type.extension}"
            if self.file_type is not None
            else ""
        )
```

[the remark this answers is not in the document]
alexrichey: Interesting... lemme give it one more go. Thanks for investigating!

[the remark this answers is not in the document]
fvankrieken: Though I never actually tested it at runtime 🙂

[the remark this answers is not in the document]
alexrichey: Ok, I was mis-remembering. It's a runtime error... It's something to do with assigning defaults when they're not specified in the recipe. I'll poke around with a few more minutes.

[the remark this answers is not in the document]
fvankrieken: mypy is also happy with

```python
class DatasetType(str, Enum):
    pg_dump = "pg_dump"
    csv = "csv"

    @property
    def extension(self) -> str:
        match self:
            case "pg_dump":
                return "sql"
            case "csv":
                return "csv"
            case _:
                return ""
```

Though at this point why use enum and match/case rather than a dict.

I'd say leave your code as is

PR #285, opened 2023-10-10
Implement 1PW for passwords

alexrichey:


.github/workflows/facilities_build.yml
```diff
@@ -46,6 +43,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Load secrets
```
alexrichey, line 46: @damonmcc @fvankrieken @sf-dcp How we feel about copy-pasting this block all over? I've never extracted a job step into anything reusable, so I thought I'd ask the experts if it's possible/worthwhile.

damonmcc: yea I don't think steps can be be extracted to be reusable like jobs can be (e.g. all our product build actions are jobs and can be called in other workflows)

sure helps using the `1password` action already! not really any unique logic here so copy-pasting these action inputs doesn't seem too bad, plus the envars needed for an action may vary?

fvankrieken: Yeah I'd say from my brief searching in the past, this is about as reusable as it gets

alexrichey: Cool - off to the races then (ie copy-pasting this everywhere)

fvankrieken: [whatever the opposite of ✂️ is]

fvankrieken: 🍝

alexrichey: 🪨 + 📄

alexrichey:
This is all ✂️/🍝 , fyi.

fvankrieken (dismissed):

.github/workflows/test.yml
```diff
@@ -3,6 +3,7 @@ run-name: 🧪 Tests (${{ github.event_name }})
 
 on:
   workflow_dispatch:
+  workflow_call:
```
damonmcc, line 6: is this change intended? seems like it makes `test.yml` callable in another workflow but I'm not seeing it being called anywhere

alexrichey: No, good catch. Snipped.

fvankrieken (approved):

fvankrieken:
My one question around this would be is there a way we can specify some of these things without needing to be so explicit - for example, a cli call or published action that gets a specific "group" of credentials as we've defined in 1password, without needing to list each one in the github action like we're doing here

alexrichey:
@fvankrieken That's a great question - explicit listing was just what I'd seen in the docs, but there are `Secure Note`s on 1PW, which we could retrieve via the 1PW CLI, and pipe into the environment 🤔

The docs I was looking at are [here](https://developer.1password.com/docs/ci-cd/github-actions/).

alexrichey:
@fvankrieken There's also an intriguing option around using [environment files](https://developer.1password.com/docs/cli/secret-references/#use-with-environment-files). Probably wouldn't use the 1pw published action in that case, and would hook something up manually. But we could just list a group of secrets together, and resolve them in one call.

damonmcc (approved):

PR #280, opened 2023-10-06
Fvk dtm duplicate bbl fix

fvankrieken:
Fixes #278.

Essentially, bbls in dof dtm are not necessarily unique. This needed some cleanup in how we're de-duplicating in preprocessing as well as addition of essentially a primary key for the table to actually identify unique rows.

Pluto needs analogous changes, though it's a bit different - data doesn't just come from dtm, and the aggregations use pluto itself as a source. So for applying the same fix to pluto, I simply added a serial pk to the table

PR #291, opened 2023-10-12
fix pluto build use of `dcpy.builds`

damonmcc:
per Finn, this started as a fix for:

Pluto building is broken right now, two issues

- cli call in workflow doesn't work - [fixed here](https://github.com/NYCPlanning/data-engineering/pull/280/commits/cc81ea454843df89a2d5238bfc2a9bc755e71a4e)
- load currently tries to connect using db "db-{product_name}" and schema based on build env, but this isn't set up in pluto's gha at the moment since it uses a temp db, not the persisted one

and then it expanded to related improvements to `connectors.recipes` and `builds.load`

### DIFF
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -95,10 +95,6 @@ jobs:
       - name: Load Local Data
         run: ./01_load_local_csvs.sh
 
-      - name: clean out raw source data
-        working-directory: ./
-        run: python3 -m dcpy.connectors.edm.recipes purge-recipe-cache
-
       - name: building ...
         run: ./02_build.sh
 
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -8,21 +8,25 @@ on:
         required: true
       recipe_file:
         type: string
-        default: recipe.yml
+        required: true
       create_issue:
         type: boolean
-        description: Create GH Issue for build?
         default: false
   workflow_dispatch:
     inputs:
       run_export:
-        description: "Run Export Step and Upload to DO"
         type: boolean
+        description: "Run Export Step and Upload to DO"
         required: true
         default: false
       recipe_file:
-        type: string
-        default: recipe.yml
+        description: "Name of recipe file to use"
+        type: choice
+        required: true
+        default: recipe
+        options:
+          - recipe
+          - recipe-minor
       create_issue:
         type: boolean
         description: Create GH Issue for build?
@@ -80,16 +84,13 @@ jobs:
       - name: Load data
         working-directory: products/pluto
         run: |
-          python3 -m dcpy.builds.load --product pluto
+          python3 -m dcpy.builds.load --recipe_path ./${{ inputs.recipe_file }}.yml
 
       - name: Set versions env vars
         working-directory: products/pluto
         run: |
           echo VERSION=$(yq .version recipe.lock.yml) >> "$GITHUB_ENV" && \
           echo VERSION_PREV=$(yq '.inputs.datasets.[] | select(.name == "dcp_pluto") | .version' recipe.lock.yml) >> "$GITHUB_ENV"
-
-      # - name: Data setup
-      #   run: ./00_setup.sh
 
       - name: Load Local Data
         run: ./01_load_local_csvs.sh
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -39,6 +39,7 @@ jobs:
     env:
       BUILD_ENGINE_SERVER: postgresql://postgres:postgres@postgis:5432
       BUILD_ENGINE_DB: postgres
+      BUILD_ENGINE_SCHEMA: public
       EDM_DATA: ${{ secrets.EDM_DATA }}
       AWS_S3_ENDPOINT: ${{ secrets.AWS_S3_ENDPOINT }}
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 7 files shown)


fvankrieken:
Looks like it failed due to disk storage, @sf-dcp's build failed for the same reason.

We can merge this, but might also be worth just getting building on edm-data (and doubling size of edm data first to prevent maxing out storage there). Otherwise this major release might get held up more

.github/workflows/pluto_build.yml
```diff
@@ -73,26 +74,22 @@ jobs:
...
           echo VERSION=$(yq .version recipe.lock.yml) >> "$GITHUB_ENV" && \
           echo VERSION_PREV=$(yq '.inputs.datasets.[] | select(.name == "dcp_pluto") | .version' recipe.lock.yml) >> "$GITHUB_ENV"
 
-      - name: Data setup
-        run: ./00_setup.sh
+      # - name: Data setup
+      #   run: ./00_setup.sh
+
+      - name: Load data
+        working-directory: ./
```
fvankrieken, line 90: I see this was done before, but I think this should be run from the pluto directory? dcpy is installed via pip prior to this, so we don't need to use the local filesystem for it. Doesn't really make a functional difference, but it'd be consistent if in the future we're directly invoking python from the build directory.

Unless there's still code in dcpy which has logic based on relative directory

dcpy/builds/load.py
```diff
@@ -20,10 +20,7 @@ def load_source_data(product: str):
     product_path = REPO_ROOT_PATH / f"products/{product}"
     build_name = build_metadata.build_name()
 
-    pg_client = postgres.PostgresClient(
-        schema=build_name,
-        database=f"db-{product}",
-    )
+    pg_client = postgres.PostgresClient(schema=build_name)
```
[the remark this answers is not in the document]
fvankrieken: See that it's getting env var [here](https://github.com/NYCPlanning/data-engineering/pull/291/files#diff-09d7f540068920518f441207143b4a9f03cb7e876a670f5bf7a4df7823d854edR48)

.github/workflows/pluto_build.yml
damonmcc: good call, see [83afce4](https://github.com/NYCPlanning/data-engineering/pull/291/commits/83afce469e0083e946ebd19465194617b2fb54f3)

.github/workflows/pluto_build.yml
```diff
@@ -73,26 +74,22 @@ jobs:
...
+          BUILD_ENGINE=${BUILD_ENGINE_SERVER}/${BUILD_ENGINE_DB}
           echo "BUILD_ENGINE=$BUILD_ENGINE" >> "$GITHUB_ENV"
 
-      - name: Plan Build
-        working-directory: ./
+      - name: Load data
+        working-directory: products/pluto
         run: |
-          python3 -m dcpy.builds.load pluto
+          python3 -m dcpy.builds.load --product pluto
```
alexrichey, line 83: To @fvankrieken 's suggestion in the prior PR, can we ✂️ the `--product pluto` bit, and have `loads` take an optional path to a recipe file? It's a pretty trivial change, and would be nice to do it while it's still fresh. (And if we need the product name for some reason, it's in the recipe.yml itself.)

damonmcc: so `template_build.yml` would look something like
```bash
defaults:
  run:
    shell: bash
    working-directory: products/template

...

python3 -m dcpy.builds.load
```
?

and `pluto_build.yml` would look something like
```bash
defaults:
  run:
    shell: bash
    working-directory: products/pluto

...

python3 -m dcpy.builds.load --recipe ${{ inputs.recipe_filename }}
```
?

dcpy/builds/load.py
```diff
@@ -16,49 +16,43 @@ def setup_build_environments(pg_client: postgres.PostgresClient):
...
-
-    pg_client = postgres.PostgresClient(
-        schema=build_name,
-        database=f"db-{product}",
-    )
-
-    recipe_path = product_path / "recipe.yml"
-    recipe_lock_path = recipe_path.parent / "recipe.lock.yml"
+def load_source_data(recipe_path: Path):
+    recipe_lock_path = recipe_path.parent / f"{recipe_path}.lock.yml"
```
fvankrieken, line 20: Is this formatted string not going to be "recipe.yml.lock.yml"?

damonmcc: true! will fix, I guess the template and pluto builds worked anyway because the `recipe_lock_path` is just passed along and used, and not expected to have a specific format

alexrichey (dismissed):
LGTM!

damonmcc:
@alexrichey just flagging commit [4c20c26](https://github.com/NYCPlanning/data-engineering/pull/291/commits/4c20c264bff8242a876d2662f38a763839d1ad8e) in case you were reviewing before I pushed it

fvankrieken:
> @alexrichey just flagging commit [4c20c26](https://github.com/NYCPlanning/data-engineering/pull/291/commits/4c20c264bff8242a876d2662f38a763839d1ad8e) in case you were reviewing before I pushed it

Nice. I like returning the path and using that rather than "calculating" it in multiple spots, and now that `build` is meant as the cli I think it probably makes sense to remove some of the `recipe` cli stuff

dcpy/builds/load.py
damonmcc: formatted string looks fixed in this build log [here](https://github.com/NYCPlanning/data-engineering/actions/runs/6509328567/job/17680624154?pr=291#step:7:15)

damonmcc:
per pluto build failing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/6509358223/job/17680703716#step:10:20), changed the use of `purge_recipe_cache` to be called in `load_source_data` rather than via CLI in a build workflow

fvankrieken (approved):

alexrichey (approved):

PR #293, opened 2023-10-12
Add unit tests for dcpy package

sf-dcp:


dcpy/test/connectors/edm/test_publishing.py
sf-dcp, line 1: You can skip this one - writing publishing tests in next PR.

dcpy/utils/s3.py
```diff
@@ -38,7 +38,10 @@ def string_as_acl(s: str) -> ACL:
 
 
 def _make_folder(s: str):
```
sf-dcp, line 40: This function is used by `list_objects()`. The previous logic for this function gave an error if input was an empty string `""`. We would want to be able to use an empty string as input to list **all** objects in a S3 bucket.

fvankrieken:
So first step here before a real review is to get these tests fixed. These are run from the `test.yml` file, which has a `pull_request` trigger, which prompts github to ensure these tests pass before a request can be merged.

You've got a couple failing tests. 
- First is dcpy - this is actually the easier one. The image running the test doesn't have moto3 installed. Our current inelegant solution alluded to on teams is to just [re-publish the dev image](https://github.com/NYCPlanning/data-engineering/actions/workflows/docker_publish.yml) - run that workflow, from your branch, just for the "dev" image and not the others
- Second is some integration tests. These seem to be failing when calling s3 upload. Luckily for you, this has a CLI that you can test with locally. At a glance, it seems most likely that your changes to the s3 client, while working for the mocked tests, are no longer working when used in our code. So what I would do is test this locally, running upload calls. You could do this from a python notebook, or just from the command line, with something along the lines of 
  - `python3 -m dcpy.utils.s3 upload --bucket de-sandbox --folder-path ./test_data --s3-path sf-testing/ --acl private`
   de-sandbox is what it sounds like - a sandbox bucket. So no worries if anything gets put there or moved around too much

dcpy/utils/s3.py
alexrichey: Good catch! However, would `s.rstrip("/") + "/"` accomplish the same thing?

Also, looking at this now, I would sort of expect `s3._make_folder()` to actually make a folder on s3. Do you have thoughts about a potentially more accurate fn name?

dcpy/test/conftest.py
```diff
@@ -0,0 +1,34 @@
+import boto3
```
alexrichey, line 1: is it possible to move this conftest to the `dcpy/test/utils` folder? I ask because we also have integration tests in the `dcpy/test/connectors` folder, and we wouldn't want the fixtures to affect anything there. Wouldn't cause problems with the existing test for github, but would cause an issue for future s3 integration tests.

dcpy/test/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+import pytest
+
+from moto import mock_s3
+from dcpy.utils.s3 import client
+
+TEST_BUCKET_NAME = "test-bucket"
+TEST_DIR_NAME_1 = "test-dir-1"
+TEST_DIR_NAME_2 = "test-dir-2"
+TEST_FILE_NAME = "test-file"
```
alexrichey, line 11: Can we move these into the test file itself? Or do they need to be in the fixtures?

dcpy/utils/s3.py
sf-dcp: @alexrichey - No, because for some reason, when I tried listing objects at the root of a bucket (all objects) with a prefix `"/"`, it returns nothing; prefix `""` returns all objects in a bucket. 

In terms of the name, I thought same, too. I think the idea behind the function (@fvankrieken - feel free to correct me) is to convert a given prefix into a directory. Not sure what fn name I would pick for that purpose.

dcpy/test/conftest.py
sf-dcp: @alexrichey - if it's okay, I would like to revisit keeping or moving these constants after writing more tests. I may need these variables across multiple test files.

dcpy/utils/s3.py
sf-dcp: Note for myself: idea - `folderize`

alexrichey (approved):
Good Stuff! This is a really great foundation for future tests

fvankrieken (approved):

PR #296, opened 2023-10-13
fix which github secret is used for data library docker publish

fvankrieken:
See failed job here

https://github.com/NYCPlanning/data-engineering/actions/runs/6510381776

Fallout from AD updating which accounts are associated with NYCPlanning on dockerhub. The new names for vars specified here are the same as other docker secrets in the rpo


alexrichey (approved):
Would be nice to move this over to 1PW, but not urgent

PR #299, opened 2023-10-13
run less PR builds and move TDB E2E test

### LINKED ISSUE #286: Pull Requests Trigger Too Many Workflows
Opening a PR with `dcpy` changes causes about 15 workflows to run (see [here](https://github.com/NYCPlanning/data-engineering/actions/runs/6470296936/job/17566195996?pr=283) for an example build). This may not be a problem at the moment since jobs run concurrently, but I could see this getting out of control, especially as we expand test coverage / linting / etc. 

I think there's certainly value in triggering product tests whenever we modify `dcpy`, however there may be alternate ways to ensure that changes to `dcpy` don't break our products. For example, now that we pip install `dcpy`, we could pin products to a certain version (a commit hash, or a tag/release) of `dcpy`. (this isn't necessarily a suggestion - it's more to prompt us to get creative)

Opening an issue as a place to discuss. It may be that there's no real problem here, especially if we're not paying for the compute time.

ping @sf-dcp @damonmcc @fvankrieken

### LINKED ISSUE #287: use of `REPO_ROOT_PATH` in `dcpy` may not be reliable
per [this discussion](https://github.com/NYCPlanning/data-engineering/pull/269#discussion_r1349312618) during review of #269

## problem
- `dcpy.builds.load` uses `REPO_ROOT_PATH` to find a product's recipe file. this assumes a standard recipe file location and assumes where dcpy lives on a filesystem
- when installing dcpy via pip, we have no guarantees about where dcpy lives on a filesystem

damonmcc:
(I think this) resolves #287
resolves #286 

also resolves a `TODO` for moving the Template DB end-to-end test from its build to it's test workflow for better separation of that logic. this is possible because it now builds in our persistent DB


pyproject.toml
damonmcc, line 1: modified in order to delete `products/template/pyproject.toml` and so that running `pytest` for product-level tests can look like
```python
python3 -m pytest ./products/template -m "end_to_end"
```

.github/workflows/test.yml
damonmcc, line 1: in pursuit of less workflow run on PRs with `dcpy` changes

I did not try to tackle running `mypy` on everything (a proposed idea in the relevant issue)

.github/workflows/template_test.yml
```diff
@@ -1,21 +1,18 @@
 name: Template - Tests
-on: 
+on:
   workflow_call:
 
+env:
+  TOY_SECRET: ${{ secrets.TOY_SECRET }}
+  BUILD_ENGINE_SERVER: ${{ secrets.SQL_ENGINE_EDM_DATA_SERVER }}
```
alexrichey, line 7: worth moving to 1pw?

.github/workflows/test.yml
alexrichey: Is this super-verbose mode? 🙂

damonmcc: yea this is how pytest supports "higher verbosity levels" ([docs](https://docs.pytest.org/en/7.1.x/how-to/output.html#verbosity)). often worth it when asserting with lists and DataFrames in a test

alexrichey: That. Is. Wild.

alexrichey:
@damonmcc So does a change to `dcpy` still end up invoking `mypy` for the entirety of our python codebase? (ie dcpy + data-library + QA + products). Not sure if that was _really_ happening with the previous code, but IIRC @fvankrieken felt that was important. If not, I'm wondering if we should change the `mypy` invocation to include those filepaths. (maybe it's out of scope here though, esp if it balloons the job runtime)

alexrichey (approved):
Two questions, but looks good!

damonmcc:
@alexrichey yea I left the mypy improvement out of this PR because it seemed like it'd lead to many (good) changes to many product codebases that would be nicer to review on their own

looks like with the changes here, any future changes in `dcpy/**` would trigger three sets of tests: `dcpy_test`, `template_test`, and `qa_test`. so the type checking that'd be run would be:
```bash
mypy ./dcpy
mypy ./apps/qa/src
```

I made and pushed a branch to see what `mypy ./` would be like and [we get an error](https://github.com/NYCPlanning/data-engineering/actions/runs/6564515459/job/17831028165?pr=310#step:4:5) due to duplicate module names in data library and cbbr. I imagine there are more after that one lol

alexrichey:
@damonmcc Yeah... I was thinking more about targeting the products we've `mypy`ed already. Hopefully that list continues to expand.

damonmcc:
@alexrichey ahh I see! looks like it's just FacDB so far, so maybe [6fa1337](https://github.com/NYCPlanning/data-engineering/pull/299/commits/6fa1337ed479e49941b067674f552cd45e1f6fb9) is what you had in mind?

alexrichey (approved):
Once in a lifetime opportunity to merge a leet commit hash

fvankrieken:
Zap test failing on main

https://github.com/NYCPlanning/data-engineering/actions/runs/6567941997/job/17841556800

PR #304, opened 2023-10-16
finish docker container setup in publish.yml

fvankrieken:
Fix failed [publish](https://github.com/NYCPlanning/data-engineering/actions/runs/6537937853)

[Fixed workflow run](https://github.com/NYCPlanning/data-engineering/actions/runs/6538292767)

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -182,7 +182,7 @@ def publish(
     By default, keeps draft output folder"""
     if publishing_version is None:
         publishing_version = get_version(draft_key)
-    source = draft_key.path
+    source = draft_key.path + "/"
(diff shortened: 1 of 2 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -182,7 +182,7 @@ def publish(
     By default, keeps draft output folder"""
     if publishing_version is None:
         publishing_version = get_version(draft_key)
-    source = draft_key.path
+    source = draft_key.path + "/"
```
fvankrieken, line 185: This maybe isn't quite ideal but was easiest one line fix - there are a couple things that rely on the current path not including the slash. I think longer term solution would be to rigorously use `Path`s or `PurePath`s for S3 paths, I think it'll just make things a lot saner.

alexrichey (approved):

PR #267, opened 2023-10-03
delete requirements and pyproject in zap product

damonmcc:
resolves dependabot PRs #264 and #242

PR #314, opened 2023-10-23
Fix failing zap tests

### LINKED ISSUE #311: Zap tests failing on main
https://github.com/NYCPlanning/data-engineering/actions/runs/6567941997

Seems like something changed in #299 caused the break

fvankrieken:
Closes #311 

Not sure how #267 passed tests

We use "end_to_end" as a custom marker for pytests in template. Prior, zap had used "integration" for the same purpose. Our main pyproject.toml file only has the former, so I've updated zap to use it


alexrichey (approved):
Thanks!

PR #316, opened 2023-10-23
Align data library python dependencies

fvankrieken:
This (and the dependabot PRs which were data library specific) were annoying me, so here we are!

For data library, this PR
- Removes poetry as package manager/installer
- installs now in dev container and published images via pip
- moves over to shared python constraints

[Passing workflow run](https://github.com/NYCPlanning/data-engineering/actions/runs/6627059723) which just uses local code
[Passing workflow rum](https://github.com/NYCPlanning/data-engineering/actions/runs/6627668424) to publish staging docker image
[Passing workflow run](https://github.com/NYCPlanning/data-engineering/actions/runs/6627866016) which uses published (staging) docker image (added a now-dropped commit to point this workflow file at the staging image rather than "latest")


.github/workflows/data_library_single.yml
```diff
@@ -28,8 +28,7 @@ on:
 jobs:
   dataloading:
     runs-on: ubuntu-22.04
-    container:
-      image: osgeo/gdal:ubuntu-small-3.6.1
+    container: python:3.11-slim
```
fvankrieken, line 31: Still running from a real base image rather than one of ours to be able to test local changes. Didn't seem worth having an additional image with just the python requirements installed. `./library.sh setup` below takes care of the container setup

data-library/library.sh
```diff
@@ -20,42 +20,19 @@ function setup {
...
-    gdal_poetry_version=$(poetry_version gdal)
-
-    if [ $gdal_version != $gdal_poetry_version ]; then
-        echo "mismatch between installed gdal binary and gdal version in pyproject.yaml"
-        echo "update poetry gdal version to $gdal_version"
-        exit 1
-    fi
-    poetry install $option
+    
+    pip install -e . -c ${1:-constraints.txt}
```
fvankrieken, line 24: still editable install so that local file system (and template files) can be referenced - code currently uses path logic relative to current file, which long term should probably be removed.

If we got rid of logic around local file structure, would need something along the lines of having a default path of `library/templates` (or we could move the templates up a folder and just default to a `templates` folder in the running directory), and still have ability to provide full path to template rather than name of template, which is currently supported

data-library/library/archive.py
```diff
@@ -79,7 +79,7 @@ def __call__(
         """
 
         _path = f"{Path(__file__).parent}/templates/{name}.yml"
-        if name and os.path.isfile(_path):
+        if name:
```
fvankrieken, line 82: This was just giving me a confusing error before, so cleaned up the logic

Issue #282, opened 2023-10-10
low `dcpy` test coverage


damonmcc:
from [recent tests on main branch](https://github.com/NYCPlanning/data-engineering/actions/runs/6462904706/job/17545247537#step:4:41), `dcpy` is at 11% test coverage

increasing this by any amount would be valuable, especially for the most commonly used `dcpy` modules

## goals
- write tests to cover a single `utils` function
- write tests to cover a single `connectors` function

## proposed approach
- choose a function and write enough tests to ensure no lines of that function are missed in the dcpy pytest coverage report

## notes
- a function could be considered "covered" by tests if no lines of it are missed in the coverage report ([report example](https://github.com/NYCPlanning/data-engineering/actions/runs/6498068240/job/17648610618#step:4:41))

alexrichey:
@damonmcc @sf-dcp @fvankrieken Before we embark on any solutioning or implementation, can we formulate this as a problem statement? Ie can we clarify what problem we intend to solve here? Perhaps issues we've encountered, or expect to encounter.

fvankrieken:
I think a more specific case would be adding testing for connectors.publishing and recipes - these are abstracted functions which very much could see the underlying functionality change at some point at which point it would be good to have automated unit tests. We/I have changed the functionality of these a fair amount already, and I have broken it in a way that would have been caught by a test

I do also think some of the slightly more involved s3 utils (upload folder, copy folder - each of which loops through objects and call other s3 utils) would be good to have tests. Both I could see us, for example, working on performance and refactoring slightly, in which case tests would be useful. Ones like download_file or upload_file I'm not concerned about

fvankrieken:
> I think a more specific case would be adding testing for connectors.publishing and recipes - these are abstracted functions which very much could see the underlying functionality change at some point at which point it would be good to have automated unit tests. We/I have changed the functionality of these a fair amount already, and I have broken it in a way that would have been caught by a test
> 
> I do also think some of the slightly more involved s3 utils (upload folder, copy folder - each of which loops through objects and call other s3 utils) would be good to have tests. Both I could see us, for example, working on performance and refactoring slightly, in which case tests would be useful. Ones like download_file or upload_file I'm not concerned about

Counterpoint to the publishing and recipes - if the underlying functionality changes too much, we might need to write new tests anyways? But maybe if we write t

sf-dcp:
[...]

PR #319, opened 2023-10-25
Add unit tests for dcpy package (part II)

sf-dcp:
Related to #282.


dcpy/test/connectors/edm/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    file_data = mock_data_constants["TEST_DATA"]
+
+    try:
+        data_path.mkdir(parents=False, exist_ok=False)
+    except Exception as err:
+        print("❌ Unable to create test dir.")
+        raise err
+
+    try:
+        txt_file_path = os.path.join(data_path, version_file)
```
fvankrieken, line 45: Little nit - since `data_path` is a `Path` object, this can just be `txt_file_path = data_path / version_file`

dcpy/test/connectors/edm/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+            writer.writeheader()
+            writer.writerows(file_data)
+        print("Created test filesystem ✅")
+
+    except Exception as exc:
+        print("❌ Exception occured while creating test files. Deleting test dir...")
+        shutil.rmtree(data_path)
+        raise exc
+
+    yield
```
fvankrieken, line 60: Huh - I'm learning more and more about pytest fixtures! Nice that you can specify the cleanup code in the same function

alexrichey (dismissed):
This is really great - it's nice when you have a PR that forces you to re-learn something, in this case how fixtures actually work. I'll leave a few notes, but they're just enhancements. This feels totally merge'able.

dcpy/test/connectors/edm/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import shutil
+import csv
+from dcpy.connectors.edm import publishing
+
+
+@pytest.fixture(scope="module")
+def mock_data_constants():
+    constants = {
+        "TEST_DATA_DIR": Path(__file__).resolve().parent / str(uuid.uuid4()),
```
alexrichey, line 13: Just curious, why do we want this randomized?

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -1,10 +1,49 @@
...
+TEST_ACL = "bucket-owner-read"
 
+draft_key = publishing.DraftKey(product=TEST_PRODUCT_NAME, build=TEST_BUILD)
+publish_key = publishing.PublishKey(product=TEST_PRODUCT_NAME, version=TEST_VERSION)
 
-@pytest.fixture
-def get_draft_key(scope="function"):
-    return publishing.DraftKey(product=TEST_PRODUCT_NAME, build=TEST_VERSION)
+
+def test_bucket_empty(create_bucket, mock_data_constants):
```
alexrichey, line 14: It looks like these tests depend on each other? (And specifically, they have to run in specific order?) It's typically an anti-pattern in unit testing, though given our setup I don't think it will cause problems given that we run tests serially. But you typically want to ensure that each `test_` method is actually a *unit* without implicit dependencies on other units. Put another way, you'd want to write all unit tests so that they can be run in parallel.

dcpy/test/connectors/edm/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+from pathlib import Path
+import uuid  # for generating a unique temp directory name
+import os
+import shutil
+import csv
+from dcpy.connectors.edm import publishing
+
+
+@pytest.fixture(scope="module")
+def mock_data_constants():
```
alexrichey, line 11: Love this approach of dependency injection for constants

dcpy/test/connectors/edm/conftest.py
sf-dcp: @alexrichey, I wanted to make sure that there are no existing directories with the same name... Which is not a problem now, but the package could potentially evolve in the future and have new permanent directories. And since it's a temp directory that's only created for the duration of tests, the name shouldn't play a big role. What do you think?

dcpy/test/connectors/edm/test_publishing.py
sf-dcp: @alexrichey, good point! I didn't notice the dependency among the tests. I would like to address it and here is how I'm thinking of going about it: 

- revise fixtures' scope to `function`
- for `test_publish()`, call `publishing.upload` fn in the beginning (`publishing.upload` fn is tested in `test_upload`).

What do you think?

dcpy/test/connectors/edm/conftest.py
alexrichey: I'd probably use a non-random dir name, and do some teardown (ie `rm -rf`) at the beginning of the fixture. Maybe just to avoid accumulation of randomized directories in `edm` folder. (e.g. every now and then I'll do a ctrl-c in the middle of a test run... I'm not sure how gracefully pytest deals with that.) Again, not a blocker here 🙂

sf-dcp: Makes sense! I added your rec to my code.

By the way, I tried interrupting the tests with ctrl-c and it seems that pytest still does the teardown! Pleasant surprise :) 

image

dcpy/test/connectors/edm/test_publishing.py
sf-dcp: @alexrichey, yes, that's what I was thinking of...

alexrichey (approved):

PR #320, opened 2023-10-26
compile python requirements

damonmcc:
changes here are from running `./bash/compile_python_packages.sh python` from the dev container without making any changes to `python/requirements.in`

### DIFF
--- a/python/constraints.txt
+++ b/python/constraints.txt
@@ -12,7 +12,7 @@ annotated-types==0.6.0
     # via pydantic
 appdirs==1.4.4
     # via sqlfluff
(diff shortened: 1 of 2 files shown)


fvankrieken:
I think this is one you can merge without review!

Also boto has been busy, I refreshed these on Monday and they've had 2 patches since??

fvankrieken (approved):

alexrichey (approved):
Feel like these requirements compiles are becoming a bit of a pain. Perhaps we could automate a compile + commit via a github action or something. But in the meantime, totally cool just merging these without review

PR #321, opened 2023-10-26
add action to compile python requirements

fvankrieken:
#320, brought up again in @alexrichey 's comment [here](https://github.com/NYCPlanning/data-engineering/pull/320#pullrequestreview-1700032458)

Doing PR rather than direct commit because testing is nice, especially if we eventually unpin more packages as described in #320. See #324, created by [this action](https://github.com/NYCPlanning/data-engineering/actions/runs/6657880581). Currently set to run nightly, so we can just hit a button in the morning if it looks good

Speaking of which... should I unpin packages as part of this? I'm down to move to more bleeding edge stuff.

### DIFF
--- a/.github/workflows/python_compile_requirements.yml
+++ b/.github/workflows/python_compile_requirements.yml
@@ -3,6 +3,7 @@ on:
   schedule: 
     - cron: 00 00 * * *
   workflow_dispatch:
+  push:
 
 jobs:
   compile_requirements:
@@ -30,4 +31,4 @@ jobs:
           commit-message: automated compiling of python requirements
           add-paths: python/
           title: Compile python requirements
-          author: GitHub <noreply@github.com>
+          author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
--- a/.github/workflows/python_compile_requirements.yml
+++ b/.github/workflows/python_compile_requirements.yml
@@ -1,0 +1,33 @@
+name: Compile python requirements
+on:
+  schedule: 
+    - cron: 00 00 * * *
+  workflow_dispatch:
+
+jobs:
+  compile_requirements:
+    name: Compile requirements and commit changes
+    runs-on: ubuntu-22.04
+    container:
+      image: python:3.11-slim
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Install latest gdal to pin python version, setup git
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update && apt-get -y install libgdal-dev git
+
+      - uses: actions/checkout@v4
+
+      - name: Compile requirements
+        run: ./bash/compile_python_packages.sh
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v5
+        with:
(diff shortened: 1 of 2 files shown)


fvankrieken:
Ended up being just slightly tricker than I imagined, mainly because we're pinning python gdal package version to latest gdal version, and for whatever reason ubuntu is more out of date than debian with regards to gdal. So ended up taking a bit to get OS, git (and the create pr action), and gdal to all play nicely together

bash/compile_python_packages.sh
```diff
@@ -2,7 +2,7 @@
 #
 # Dev script to compile python packages from a requirements.in file to a requirements.txt file.
 set -e
-path_to_requirements="${1:-"."}"
+path_to_requirements="python"
```
fvankrieken, line 5: Figured we don't really need ability to run this from multiple places and nicer to just get this consistently, I was tired of changes between "./python/requirements.in" and "python/requirements.in" in `requirements.txt`

fvankrieken:
This will be squashed

.github/workflows/python_compile_requirements.yml
```diff
@@ -0,0 +1,33 @@
...
+    container:
+      image: python:3.11-slim
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Install latest gdal to pin python version, setup git
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update && apt-get -y install libgdal-dev git
```
fvankrieken, line 20: This was sort of interesting. I wanted to use a debian base image so that we could get latest libgdal-dev, which by default is not available on ubuntu (could only get 3.4.1 instead of 3.6.2, even with updates). Figured I'd use this python one for consistency. However, it doesn't have git installed. We need git for both compiling requirements (since some sources are from git repos - geosupport and my fork of sqlfluff) as well as the action for opening a PR.

Checkout doesn't need git installed... but if it isn't, it uses github's rest api, which doesn't actually make the checked out repo a git repo! It just copies the files. So the action to create the PR was failing due to it not actually being a git repo.

So solution is to install git before running checkout action. Kinda interesting!

.github/workflows/python_compile_requirements.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Compile requirements
+        run: ./bash/compile_python_packages.sh
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v5
+        with:
+          commit-message: automated compiling of python requirements
+          add-paths: python/
+          title: Compile python requirements
+          author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
```
alexrichey, line 33: any way to tag us as reviewers on the PR? I imagine I'd miss these being created (unless for some reason I get a notification)

alexrichey (approved):
Nice!

.github/workflows/python_compile_requirements.yml
fvankrieken: Ooh yes it is. And sorry, somehow missed this comment before merging

PR #329, opened 2023-10-30
Fix path issues in data library

fvankrieken:
## Path issue

So I caused a little issue in #316 

You can either provide the name of a dataset, the path to a yml file, or both.

The logic before was
- check if name is provided AND there exists a template within data library, use that
- if no name provided, but path provided, use path provided

I interpreted this as either a name or a path should be provided, and "cleaned up" the if/else clause. However, this caused an issue if both were provided, as the ordering tries "name" first and assumes provided "path" isn't used in that case.

So logic is now as it should be - if path is provided, use that. If it isn't and a name is provided, attempt to use a yml provided by data-library in the templates folder

## Use of templates at runtime
I also wanted to clean up referencing yml files that exist within data library when its installed. So I removed the editable flag from the installs for the docker images and modified pyproject.toml so that the yml files are packaged when it's installed

[Passing single run (not using docker image, just local install/code)](https://github.com/NYCPlanning/data-engineering/actions/runs/6695152691)

[Passing publishing of docker image](https://github.com/NYCPlanning/data-engineering/actions/runs/6695233536)

[Passing run using staging docker image](https://github.com/NYCPlanning/data-engineering/actions/runs/6694243158)

### DIFF
--- a/data-library/Dockerfile
+++ b/data-library/Dockerfile
@@ -1,10 +1,14 @@
 FROM python:3.11-slim
 
-COPY . /library/
+COPY . /library
 
-WORKDIR /library/
+WORKDIR /library
 
 RUN ./library.sh setup
+
+WORKDIR /
+
+RUN rm -rf library
 
 # GDAL for python required env variables
 ENV CPLUS_INCLUDE_PATH=/usr/include/gdal
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,10 +63,7 @@ jobs:
               - python/**
             data-library:
               - .github/workflows/data_library_test.yml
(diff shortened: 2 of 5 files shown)


data-library/Dockerfile
```diff
@@ -1,11 +1,15 @@
...
+COPY . /library
 
-WORKDIR /library/
+WORKDIR /library
 
 RUN ./library.sh setup
 
+WORKDIR /
+
+RUN rm -rf library
```
fvankrieken, line 11: It's being installed fully now so no need to keep it in the image

alexrichey (approved):
cool - thanks for fixing!

PR #332, opened 2023-11-01
move Tax Block to appropriate position

fvankrieken:
Originally changed by sqlfluff [here](https://github.com/NYCPlanning/data-engineering/commit/7f5b46e9a73626b82e3efc9a9e3ed62abffb9015#diff-3fce2a0ab0be7926d4b61a87a3b1136da5eb2a7ea8478d6b5c33f1eb3d600282). My fault for not catching the reordered columns in export

### DIFF
--- a/products/zoningtaxlots/sql/export.sql
+++ b/products/zoningtaxlots/sql/export.sql
@@ -1,22 +1,21 @@
 -- export zoning tax lot database with desired headings
 DROP TABLE IF EXISTS dcp_zoning_taxlot_export;
-CREATE TABLE dcp_zoning_taxlot_export AS (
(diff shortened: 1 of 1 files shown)


alexrichey (approved):
Thanks for fixing! This feels like a very informative use case to keep in mind for automated product QA

fvankrieken:
Yup. And while I should have caught this, it's also nuts to me that sqlfluff's default is to reorder select statements (specifically moving calculated columns to the end). Is it more readable? Maybe. Is it actually changing the logic of the query? Absolutely. Not even just downstream in exports - if you have a `UNION` this could break it

PR #339, opened 2023-11-06
Add checkbook nyc script to data library

### LINKED ISSUE #312: Add checkbook nyc (via API) to data library
[Checkbook NYC](https://www.checkbooknyc.com/spending-api) data is specified in data library via two templates
- [`cpdb_capital_spending`](https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/templates/cpdb_capital_spending.yml) - input to CPDB (being phased out)
  - [Workflow to archive](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/archive/cpdb_spending.yml), this actually calls the [logic that generates the file ](https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/bash/06_spending.sh)
- [`nycoc_checkbook`](https://github.com/NYCPlanning/data-engineering/blob/main/data-library/library/templates/nycoc_checkbook.yml)

Both currently have issues. `nycoc_checkbook` is just raw checkbook data, which is what this dataset ideally is. But it comes from a local csv rather than accessing via the checkbook nyc web api. The former is a bit more convoluted - it calls shell scripts to do some processing in bigquery, using bigquery as the source for checkbook data. It then dumps the results of these bq queries to a csv, which then gets archived with data library.

How does checkbook nyc data end up in bigquery? These two repos have the code to do it: [db-airflow](https://github.com/NYCPlanning/db-airflow/tree/main/dags) and [db-checkbooknyc](https://github.com/NYCPlanning/db-checkbooknyc/tree/main) (I actually somehow missed this in the past and only found it while looking for db-airflow). Neither has workflows run, so I'm not sure when/how it was actually being used.

So moving forward, an ideal solution would be to 
1. only have one data-library dataset for checkbook, `nycoc_checkbook`
2. use logic currently in db-airflow/db-checkbooknyc to pull data from api (via `script` source in `nycoc_checkbook` template

sf-dcp:
Resolves #312.

## Notes: 

* Various links: 

> Checkbook [API](https://github.com/NYCComptroller/Checkbook/tree/master)
> Checkbook NYC Github [link](https://github.com/NYCComptroller/Checkbook/tree/master)
> Checkbook NYC [forum](https://groups.google.com/g/checkbooknyc).

* API receives requests and sends responses in XML format.
* As noted in [here](https://github.com/NYCPlanning/data-engineering/issues/312#issuecomment-1795347678), there are multiple domains such as `Spending` & `Budget`. Response bodies slightly differ among the domains. For example, the tag name for `Spending` transactions in the xml response body is `spending_transactions` and for `Budget` - `budget_transactions`.
* There are about 37 million records in Spending checkbook API. Out of 37 million, ~2 million records are capital projects. The max number of records API returns is 20,000. This limitation needs to be accounted for when choosing an approach to query records. 
* It doesn't seem that the order of records at the source is deterministic; therefore, can't rely on `records_from` field to extract all records. This assumption is based on a call to API with `records_from = 1` parameter (see response result in the csv below). There are transactions from different years; if there was a deterministic order, we wouldn't see recent transactions in the first 20,000 records out of ~2 million total. 
[capital_spending_test.csv](https://github.com/NYCPlanning/data-engineering/files/13340569/capital_spending_test.csv)

## Proposed approach in this PR:
1. create a dictionary with request body to be sent to API. Creating a dictionary hopefully makes code more maintainable and easier to read when creating API requests. The request body may contain criteria to query on (for example, filter for capital project transactions only). 
2. Convert the request dictionary into XML string. 
3. Create a local directory for the data.
4. Query api with `spending_category = “cc”` criteria (cc - capital contracts) and make one call per one month. One month is an arbitrary time period discussed with Damon & Finn which should return under 20k records and not too few. 
5. Save monthly data from each call to a parquet files. 
6. Read in the resulting parquet files into one pandas data frame. 

## Current limitations/possible future improvements:
1. Current logic assumes there are no more than 20k records in any given month. If there are, we will only get the first 20k instead of all records in a month. <--- [ISSUE ADDRESSED]
2. Current logic scrapes the data by month-year, saves it locally into parquet files, and then reads them into a pandas df. If there is not enough disk space for data, the script will probably crash. 
3. Need to write more tests. 
4. We may want to implement incremental approach in updating checkbook data. 
5. Need to test cases where invalid criteria is provided (example, non-existent field).
6. Issues with getting a response from API (see my email to the team [here](https://groups.google.com/g/checkbooknyc/c/hgU1niDG00Y)). Latest GHA run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/6915660322/job/18814835720).

## Other discovered issues:
* The current data-library single runner template `data_library_single.yml` creates a bunch of different output formats. However, if the data doesn't have a geometry column, the GH action fails. [Link](https://github.com/NYCPlanning/data-engineering/actions/runs/6908745294/job/18798663555#step:6:110) to such example.

### DIFF
--- a/.github/workflows/data_library_single.yml
+++ b/.github/workflows/data_library_single.yml
@@ -59,5 +59,5 @@ jobs:
           compress: ${{ github.event.inputs.compress == 'true' && '--compress' || '' }}
         run: |
           library archive --name ${{ github.event.inputs.dataset }} -o pgdump $export $compress $latest $version
+          # library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
           library archive --name ${{ github.event.inputs.dataset }} -o csv $export $compress $latest $version
-        # library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
--- a/.github/workflows/data_library_single.yml
+++ b/.github/workflows/data_library_single.yml
@@ -59,5 +59,5 @@ jobs:
           compress: ${{ github.event.inputs.compress == 'true' && '--compress' || '' }}
         run: |
           library archive --name ${{ github.event.inputs.dataset }} -o pgdump $export $compress $latest $version
-          library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
           library archive --name ${{ github.event.inputs.dataset }} -o csv $export $compress $latest $version
+        # library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
--- a/data-library/library/script/nycoc_checkbook.py
+++ b/data-library/library/script/nycoc_checkbook.py
@@ -1,0 +1,455 @@
+import xml.etree.ElementTree as ET
+import requests
+import pandas as pd
+import datetime as dt
+import pytz
+import calendar
+from pathlib import Path
+import shutil
+import sys
+import time
+import random
+
+from . import df_to_tempfile
+
+from typing import Literal, TypedDict, Any
+
+api_endpoint = "https://www.checkbooknyc.com/api"
+
+TYPE_OF_DATA = Literal[
+    "Contracts",
+    "Budget",
+    "Revenue",
+    "Payroll",
+    "Spending",
+    "Spending_OGE",
+    "Contracts_OGE",
+    "Spending_NYCHA",
+    "Contracts_NYCHA",
+    "Payroll_NYCHA",
+]
+
+DEFAULT_RECORDS_FROM = "1"
+DEFAULT_MAX_RECORDS = "20000"
+
+DEFAULT_DATE_FORMAT = "%Y-%m-%d"
+DEFAULT_START_PERIOD = dt.date(2010, 1, 1).strftime(DEFAULT_DATE_FORMAT)
+DEFAULT_END_PERIOD = dt.date.today().strftime(DEFAULT_DATE_FORMAT)
+
+DEFAULT_DATA_PATH = Path(__file__).resolve().parent / "capital_spending_data"
+
+USER_AGENTS = [
+    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36",
+    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36",
+    "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Safari/605.1.15",
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 13_1) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Safari/605.1.15",
+]
+# user_agent = random.choice(USER_AGENTS)   # TODO: uncomment or remove completely
+user_agent = None
+
+
+class CriteriaValue(TypedDict):
+    name: str
+    type: Literal["value"]
+    value: str
+
+
+class CriteriaRange(TypedDict):
+    name: str
+    type: Literal["range"]
+    start: str
+    end: str
+
+
+class Scriptor:
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+    def ingest(
+        self,
+        user_agent: str | None = None,
+        type_of_data: TYPE_OF_DATA = "Spending",
+        records_from: str = DEFAULT_RECORDS_FROM,
+        max_records: str = DEFAULT_MAX_RECORDS,
+        response_columns: list[str] | None = None,
+        search_criteria: list[CriteriaValue | CriteriaRange] | None = None,
+        num_retries: int = 3,
+        data_dir: Path = DEFAULT_DATA_PATH,
+        start_period: str = DEFAULT_START_PERIOD,
+        end_period: str = DEFAULT_END_PERIOD,
+    ) -> pd.DataFrame:
+        """Fetches data from Checkbook NYC API based on input parameters, saves it
+        locally in parquet files organized by months, and returns data in a pandas DataFrame.
+        """
+        create_data_dir(data_dir)
+
+        search_criteria = validate_search_criteria(search_criteria)
+
+        month_range = generate_monthly_ranges(start_period, end_period)
+
+        for month in month_range:
+            start, end = month
+            month_criteria: CriteriaRange = {
+                "name": "issue_date",
+                "type": "range",
+                "start": start,
+                "end": end,
+            }
+            search_criteria_modified = search_criteria + [month_criteria]
+            try:
+                df = get_data(
+                    user_agent=user_agent,
+                    type_of_data=type_of_data,
+                    records_from=records_from,
+                    max_records=max_records,
+                    response_columns=response_columns,
+                    search_criteria=search_criteria_modified,
+                    num_retries=num_retries,
+                )
+                print(f"There are {len(df)} records for period {start}-{end}.")
+            except Exception as e:
+                print(
+                    f"Unable to extract data for period: {start}-{end}...Exception: {e}"
+                )
+                df = None
+
+            if df is not None:
+                filename = f"{start}-{end}_checkbook.parquet"
+                filepath = str(data_dir / filename)
+                df.to_parquet(filepath)
+
+            # pausing before next request: we don't want to overload API
+            time.sleep(random.randint(10, 15))
+
+        try:
+            data = read_parquet_files_to_df(data_dir)
+            print("Successfully ingested data.")
+        except ValueError as e:
+            print(
+                f"No data to ingest for period {start_period}-{end_period}. Error: {e}"
+            )
+            sys.exit(1)
+
+        return data
+
+    def runner(self) -> str:
+        df = self.ingest(
+            user_agent=user_agent,
+            search_criteria=[
+                {"name": "spending_category", "type": "value", "value": "cc"}
+            ],
+        )
+        local_path = df_to_tempfile(df)
+        return local_path
+
+
+def get_data(
+    user_agent: str | None,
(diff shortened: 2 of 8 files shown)


fvankrieken:
> Not sure why `Tests/Pytest-zap (pull_request)` fails. Any thoughts?

It is my doing

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+USER_AGENTS = [
+    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
+    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"
+    "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Safari/605.1.15"
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 13_1) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Safari/605.1.15"
+]
+user_agent = random.choice(USER_AGENTS)
```
fvankrieken, line 48: What's the reasoning for this?

sf-dcp: Checkbook API starts blocking requests at some point, returning a 500 server error. I found that changing a user-agent param in the request body helps to get more requests through, but it doesn't fully solve the problem.

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+        self.__dict__.update(kwargs)
+
+    def ingest(
+        self,
+        user_agent: str | None = None,
+        type_of_data: TYPE_OF_DATA = "Spending",
+        records_from: str = DEFAULT_RECORDS_FROM,
+        max_records: str = DEFAULT_MAX_RECORDS,
+        response_columns: list[str] | None = None,
+        search_criteria: list[CriteriaValue | CriteriaRange] = [],
```
fvankrieken, line 75: Bit of a small note that wouldn't cause an issue here but still is good practice - lists and dicts should not have default values in python. If they're mutated (say, call `search_criteria.append(...)` at some point), that will persist while the current session is running. So the same call to a function could have different results each time

fvankrieken: Ah actually you do mutate it. So if there are multiple calls to ingest in one session, any changes made to search criteria will persist from first call to second.

The way to avoid this is default to None, and then declare as empty list if None within the function body

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+        num_retries: int = 3,
+        data_dir: Path = DEFAULT_DATA_PATH,
+    ) -> pd.DataFrame:
+        """Fetches data from Checkbook NYC API based on input parameters, saves it
+        locally in parquet files organized by months, and returns data in a pandas DataFrame.
+        """
+        create_data_dir(data_dir)
+
+        start_period = DEFAULT_START_PERIOD
```
fvankrieken, line 84: Looking at the logic below, should these maybe just be supplied as their own args? Since they're a bit of a special case for search criteria. I like starting with search criteria as more generic - a user could supply whatever combination of search criteria they want, and we don't need to necessarily validate whatever combination they provide. But given that someone could arbitrarily provide multiple issue dates and we have special logic around ONE range of dates, I'd prefer to be explicit about that expectation.

For the checkbook api, are there other fields that support ranges? Or just issue date?

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+        raise err
+
+
+def read_parquet_files_to_df(data_dir: Path) -> pd.DataFrame:
+    """
+    Reads Parquet files from a given data directory and concatenates
+    them into a single pandas DataFrame.
+    """
+    df = pd.concat(
+        pd.read_parquet(parquet_file) for parquet_file in data_dir.glob("*.parquet")
```
fvankrieken, line 378: Just curious, did you try out pyarrow much? How did that go? This ended up being not quite as "big" data as we initially thought, but still curious on performance and how it would compare to appending as we go with pyarrow.

Again, more curious than anything! This works nicely

data-library/library/script/nycoc_checkbook.py
sf-dcp: @fvankrieken, makes sense! I will move start and end issue dates to its own args. 

Do we still want to validate that none were provided in the search criteria? I could see having conflicting issue dates if provided in 2 places...

Regarding the ranges... Yes, the other field that supports it in the `Spending` domain is `check_amount `. Note, "range" fields may vary across the six domains because columns across them are not uniform.

.github/workflows/data_library_single.yml
```diff
@@ -59,5 +59,5 @@ jobs:
           compress: ${{ github.event.inputs.compress == 'true' && '--compress' || '' }}
         run: |
           library archive --name ${{ github.event.inputs.dataset }} -o pgdump $export $compress $latest $version
-          library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
           library archive --name ${{ github.event.inputs.dataset }} -o csv $export $compress $latest $version
+        # library archive --name ${{ github.event.inputs.dataset }} -o shapefile $export $compress $latest $version
```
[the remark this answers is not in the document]
fvankrieken: This seems fine, and moving forward I would maybe lean with us keeping no shapefile export here. We never use it. Could always have it as an option if we like

python/requirements.txt
```diff
@@ -312,7 +312,9 @@ pure-eval==0.2.2
 py-partiql-parser==0.4.2
     # via moto
 pyarrow==14.0.1
```
fvankrieken, line 314: streamlit has a pyarrow dependency? Very interesting

fvankrieken: Still good to explicitly add it to requirements.in, just surprised to already see it here

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+        ]
+    if search_criteria:
+        request_dict["request"]["search_criteria"] = [
+            {"criteria": criteria_item} for criteria_item in search_criteria
+        ]
+
+    return request_dict
+
+
+def dict_to_xml_obj(request_dict: dict) -> ET.Element:
```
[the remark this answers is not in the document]
fvankrieken: Curious as to Alex and Damon's thoughts on both, my thoughts are
1. This would be good to have in dcpy, but for the moment the dcpy - data-library connection seems a little prickly, so I wouldn't include that in this PR. But I think something along the lines of `dcpy.utils.xml` would make sense longer term
2. In general my thought on docstrings is more thorough is never a bad thing. Main potential drawback is if the code is changed, you have to remember to update the example, where mypy won't catch errors for you. But this seems like a function that wouldn't change much, if it was say one of our connectors I would be a bit more hesitant just because we've been moving pretty quickly and I wouldn't be surprised if they fell out of sync by accident at some point

fvankrieken:
Sorry, never really closed the loop yesterday - 

This is great. A couple smaller notes but in general thorough and good. @damonmcc you're testing locally a bit?

@sf-dcp if you rebase latest main locally, the failing test will pass now. Plus you'll have to resolve the conflicts that have come up in python constraints/requirements from our daily compiling

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+        )
+        local_path = df_to_tempfile(df)
+        return local_path
+
+
+def get_response(
+    user_agent: str | None,
+    type_of_data: TYPE_OF_DATA,
+    records_from: str,
+    max_records: str,
```
alexrichey, line 147: @sf-dcp Are there ever cases where there are more than `max_records` (ie 20k) within a given month range? Curious what would happen in that case?

sf-dcp: @alexrichey, it seems that the average number of records per month is between 10-15k. 

If the count is greater than 20k, the current logic will return the first 20k. A potential enhancement is reading in the  total count of records for a given month (it's returned in the response body) and make additional API requests, using the `records_from` parameter. For example, let's say the total count is 35k; then we can send 2 requests where first request will return records 1-20,000 and second request - 20,001-35,000.

fvankrieken: I'd say that's an enhancement worth getting in this round. Or at a bare minimum, throwing an error if the record count is greater than 20000, but that's just kicking the issue down the road a bit. Possibility of silently missing data is one that we want to avoid

alexrichey: @sf-dcp that sounds like a great approach. Let's hope we don't find a day in which >20k checks were cut... though maybe there are other ways to filter down the data in addition

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,381 @@
...
+DEFAULT_RECORDS_FROM = "1"
+DEFAULT_MAX_RECORDS = "20000"
+
+DEFAULT_DATE_FORMAT = "%Y-%m-%d"
+DEFAULT_START_PERIOD = dt.date(2010, 1, 1).strftime(DEFAULT_DATE_FORMAT)
+DEFAULT_END_PERIOD = dt.date.today().strftime(DEFAULT_DATE_FORMAT)
+
+DEFAULT_DATA_PATH = Path(__file__).resolve().parent / "capital_spending_data"
+
+USER_AGENTS = [
```
damonmcc, line 39: I think `USER_AGENTS` is currently 1 big string without commas after each one

data-library/library/script/nycoc_checkbook.py
```diff
@@ -0,0 +1,448 @@
...
+            start, end = month
+            month_criteria: CriteriaRange = {
+                "name": "issue_date",
+                "type": "range",
+                "start": start,
+                "end": end,
+            }
+            search_criteria_modified = search_criteria + [month_criteria]
+            try:
+                df = get_data(
```
sf-dcp, line 101: `get_data` is the new function that gets ALL data for a given month (one iteration in this loop). If, for some reason, we can't get all records for a month period, an Exception is printed and the `for` loop continues to the next month.

data-library/library/script/nycoc_checkbook.py
sf-dcp: Added enhancements to the code.

alexrichey (approved):
Cool - this seems like a great checkpoint. Maybe we'll hear something more concrete from the Comptroller folks, but this is good stuff

fvankrieken:
Looks good. Per @alexrichey 's comment, i think the one final change that would be good is printing out the xml request (and maybe timestamp) when a request fails in addition to the exception. But then good to go!

PR #342, opened 2023-11-08
Fix dob_cofos for DevDB 23Q2

fvankrieken:
Needed to update `dob_cofos` properly to regenerate HDB 23Q2. Commits are
1. remove committed, hard-coded prev version for dob_cofos in template yml file. Also gave the script some tlc, it uses dcpy.connectors.edm.recipes now to read in old version for appending
2. pin to this version in `version.env`
3. remove some unnecessary lines in data library scripts


products/developments/version.env
```diff
@@ -24,6 +24,8 @@ HPD_HISTOICAL_VERSION=20230630
...
 DOB_NOW_APPS_VERSION=20230710
 DOB_NOW_PERMITS_VERSION=20230720
-DOB_COFOS_VERSION=20230719
+DOB_COFOS_VERSION=20230709
 DOITT_BUILDINGS_VERSION=20230702
-DOITT_BUILDINGS_HISTORICAL_VERSION=20230213
\ No newline at end of file
+DOITT_BUILDINGS_HISTORICAL_VERSION=20230213
+
+OTHER_DOB_SOURCES=20230827
```
damonmcc, line 31: looks like `DOB_DATA_DATE` used to do the job of `OTHER_DOB_SOURCES`. maybe good to keep either the new or the old envar?

fvankrieken: Gonna rework all of this and use a recipe.yml I think, but for now this is to re-generate with exact same sources as original 23Q2. I should at least rename this probably - it's for some other DOB sources which were unpinned before

fvankrieken: Oh I see - looked at the code and you're right, DOB_DATA_DATE is currently not being used at all. I'll fix

dcpy/connectors/edm/recipes.py
```diff
@@ -148,6 +148,22 @@ def fetch_dataset(ds: Dataset, local_library_dir=LIBRARY_DEFAULT_PATH) -> Path:
     return target_file_path
 
 
+def read_csv(
+    ds: Dataset, local_cache_dir: Path | None = None, **kwargs
+) -> pd.DataFrame:
+    if ds.file_type != DatasetType.csv:
+        raise Exception(
+            "csv output is currently only format that recipes.read_csv is implemented for"
+        )
```
damonmcc, line 157: nit: this might be nicer as something like `read_dataset` and only support csv for now

it could then be used in both `recipes.import_dataset` and `dob_cofos.py`

alexrichey (approved):

PR #350, opened 2023-11-08
Swap COLP to use recipes

alexrichey:
Well, it wasn't as easy as just adding a recipes file. There were some subsequent modifications to make sure that schemas were being used. 

Successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/6882085134)
Artifacts here [here](https://cloud.digitalocean.com/spaces/edm-publishing?path=db-colp/draft/ar-colp-use-recipes/)


damonmcc (dismissed):

bash/utils.sh
```diff
@@ -243,21 +243,27 @@ function shp_export {
 
 
 function fgdb_export_partial {
-    parse_connection_string ${BUILD_ENGINE}
+    # parse_connection_string ${BUILD_ENGINE}
```
fvankrieken, line 246: ✂️

fvankrieken (dismissed):
Two commented lines that could be snipped but lgtm!

PR #354, opened 2023-11-09
fix duplicate KPDB records

### LINKED ISSUE #309: KPDB has duplicate records
based on chat with Capital Planning and Housing teams on 10/18

## problem
- there are different types of duplicative records
- the most significant and impactful is when the `record_id` is the same across multiple records
  - e.g. `record_id = B00520132` results in 4 rows in the CDTA aggregation output
- this is all based on the KPDB build for and shared with the the Capital Planning team

## potential causes
- clustering logic
  - KPDB code has clustering logic to identify overlapping source data records and prevent over-counting of housing unit changes
    - example: 2 ZAP records which overlap but are not the same project may lead to duplication of a DOB record
    - a cluster: nearby ZAP, HPD, DOB, etc. records that represent the same project
  - there's a hierarchy of sources to prioritize and select the "truest" records: ZAP < HPD < DOB
  - a single DOB record may be attributed to 2 clusters, leading to duplication of it

damonmcc:
resolves #309 

- [x] ensure deduplicated data still has geometry data as geometry type rather than text
- [ ] ensure project deduplication doesn't have to take place before unit deduplication

## approach
- use python to deduplicate records at the end of the build stage
- only drop extra records if the values in every column are identical (other than the primary key)
  - `DataFrame.drop_duplicates(subset=columns_to_consider)`

## tests
- [build logs](https://github.com/NYCPlanning/data-engineering/actions/runs/6818960177/job/18545523794#step:8:152) showing removal of duplicate records
- SQL queries to see effects of changes
  - from schema `kpdb.build_240` without these changes
    - results of ["duplicate records" query](https://github.com/NYCPlanning/data-engineering/issues/309#issuecomment-1804267601)
      - image
    - records with example record_id `B00520132`
      - image
  - from schema `kpdb.build_354` with these changes
    - results of ["duplicate records" query](https://github.com/NYCPlanning/data-engineering/issues/309#issuecomment-1804267601)
      - image
    - records with example record_id `B00520132`
      - image

### DIFF
--- a/.github/workflows/knownprojects_build.yml
+++ b/.github/workflows/knownprojects_build.yml
@@ -18,6 +18,8 @@ jobs:
         shell: bash
         working-directory: ./products/knownprojects
     env:
+      BUILD_ENGINE_SERVER: ${{ secrets.SQL_ENGINE_EDM_DATA_SERVER }}
+      BUILD_ENGINE_DB: kpdb
       BUILD_ENGINE_DATABASE: ${{ secrets.SQL_ENGINE_EDM_DATA_SERVER }}/kpdb
       BUILD_ENGINE_SCHEMA: build_${{ github.event.pull_request.number || github.ref_name }}
       EDM_DATA: ${{ secrets.EDM_DATA }}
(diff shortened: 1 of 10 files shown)


damonmcc:
see linked issue

PR #361, opened 2023-11-13
Nightly build on main, remove builds from tests

fvankrieken:
Very open to changes as well as just discussing what we want out of these jobs! Just thought this would be a nice quality of life thing, and then some changes were either necessary or worthwhile. If people don't love aspects of it or want to discuss more, that's totally fine

Main motivation for this was that it'd be nice to build our products (on main) nightly, both as a sort of end-to-end test/qa but also so that we have tables in the persisted db and outputs generated for easy access if we need to do any checks with the latest data. Another main part was separating "tests" from actually just running build steps, so that we're not running builds as part of PRs. Just personally would prefer that because then we can be a bit more permissive when we test without needing to wait a few hours potentially before merging.

### Changes (by commit)
1. Fix `moto` mocking of `boto3` functionality. In my first iteration of this, `dcpy` tests were failing due to "invalid credentials". Apparently, the setup as defined worked in the absence of env vars (for `AWS_ACCESS_KEY_ID` etc), but did not work when they existed. I've reworked it to mock the env variables, and also got a bit of a better understanding of what's happening under the hood with `moto` - whenever you use `moto_s3` annotation or call it in a `with` block, you're mocking credentials just in the lifespan of that function or `with` block. So the `yield` statement, which essentially halts a pytest fixture (and only finishes execution after jobs dependent on fixture have stopped). Another discovery is that using a custom endpoint (i.e. NOT aws) breaks `moto`, hence getting rid of it in the fixture.
2. appease `mypy` in template db, figured it'd be nice to have this getting `mypy`-ed. Just dropped two utilities which aren't used and are redundant with `dcpy`.
3. add nightly qa job with builds and tests. Ran into an issue originally of too many `workflow_call`s - apparently there's a max of 20 total workflow calls within one main workflow. So moved some logic out of individual product test workflows and into the main one. Also condensed the `mypy` steps into one and removed duplicate `black` calls
4. dynamic matrix based on path filters! Code largely inspired by the answer [here](https://stackoverflow.com/questions/65384420/how-do-i-make-a-github-action-matrix-element-conditional). There's then also a json file which gets read in which has metadata for the specific test jobs. Bit of a proof of concept here how to have more dynamic logic within a matrix job. I'll add some comments on the specific changes about what's happening there (I know the `sed` stuff is super easy to read), but I'm pleased with it as a first pass at reducing some redundant code. Nice way to not have to duplicate the logic around onepass loading, container setup, etc.
5. facdb build was failing because it's not built in persisted db, so it was trying to drop the "public" schema.

[Successful run of new workflow](https://github.com/NYCPlanning/data-engineering/actions/runs/6872243263). KPDB build failing, but it also is on main. Manually set it to run all pytest matrix options as well.


dcpy/test/conftest.py
```diff
@@ -2,22 +2,24 @@
...
-    env variables for fake AWS credentials."""
-    with mock_s3():
-        yield client()
+@pytest.fixture(scope="session")
+def aws_credentials():
+    """Mocked AWS Credentials for moto."""
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    if "AWS_S3_ENDPOINT" in os.environ:
+        os.environ.pop("AWS_S3_ENDPOINT")
```
fvankrieken, line 17: This variable messes up boto3 mocking

.github/workflows/test.yml
```diff
@@ -73,73 +86,110 @@ jobs:
...
+        id: set-matrix
         run: |
           echo "path filters with changed files: ${{ steps.filter.outputs.changes }}"
-  dcpy:
-    needs: check_changes
-    if: github.event_name != 'pull_request' || contains(needs.check_changes.outputs.path_filters, 'dcpy')
-    uses: ./.github/workflows/dcpy_test.yml
-    secrets: inherit
-  build:
+          filter=$(echo '${{ steps.filter.outputs.changes }}' | sed 's/[][]//g' | sed -r 's/(\w+)(,|$)/"\1"\2/g')
```
fvankrieken, line 99: gets just comma delimited, quoted fields from the path-filter step. Looking now actually it looks like there's a bug. Will fix now

.github/workflows/test.yml
```diff
@@ -73,73 +86,110 @@ jobs:
...
+          matrix=$(cat .github/workflows/data/pytest.json | jq "map(. | select(.name|IN(${filter})))")
+          echo "matrix={\"include\":$(echo $matrix)}" >> $GITHUB_OUTPUT
```
fvankrieken, line 101: So basic logic is on 99 we clean the output of the filter step to better feed it into line 100.

Line 100 we take the json data from `pytest.json` and filter it to entries which have changes (based on `filter` variable). We then export this to github step output so we can set the matrix to this json in line 115

.github/workflows/test.yml
```diff
@@ -73,73 +86,110 @@ jobs:
...
+    - name: Run Container Setup
+      working-directory: ./
+      run: ./bash/docker_container_setup.sh
+
+    - name: Pytest
+      run: |
+        echo '${{ toJson(matrix.calls) }}' \
+          | sed '$ d;1d;s/^[[:space:]]*\"//;s/\",\{0,1\}[[:space:]]*$//'\
+          | while read call; do
+          pytest $call
```
fvankrieken, line 154: this is a tad clunky. The matrix (json data in `pytest.json`) has this `calls` field which include the calls to be fed to pytest. The `sed` call is cleaning the string representation coming from `toJson` - removing first and last lines, cleaning whitespace and commas, then this gets iterated through. This was done because zap has two pytest calls.

It's sort of an unhappy middle ground though - the two pytest calls for zap are nicer to have as distinct steps for clarity when one fails, and if we're forced to do one step by the limitations of a matrix strategy, maybe we might as well just do a single pytest call

.github/workflows/test.yml
```diff
@@ -73,73 +86,110 @@ jobs:
...
+      working-directory: ./
+      run: ./bash/docker_container_setup.sh
+
+    - name: Pytest
+      run: |
+        echo '${{ toJson(matrix.calls) }}' \
+          | sed '$ d;1d;s/^[[:space:]]*\"//;s/\",\{0,1\}[[:space:]]*$//'\
+          | while read call; do
+          python3 -m pytest $call
+        done
```
fvankrieken, line 155: this is a tad clunky. The matrix (json data in pytest.json) has this calls field which include the calls to be fed to pytest. The sed call is cleaning the string representation coming from toJson - removing first and last lines, cleaning whitespace and commas, then this gets iterated through. This was done because zap has two pytest calls.

It's sort of an unhappy middle ground though - the two pytest calls for zap are nicer to have as distinct steps for clarity when one fails, and if we're forced to do one step by the limitations of a matrix strategy, maybe we might as well just do a single pytest call (and just make this code a lot simpler)

fvankrieken:
TODO - something has to happen about PLUTO opening an issue on a successful build. Somehow I don't think we want those every single day

fvankrieken:
One argument against - this is going to be asking a fair amount of the database every night. If we have any long ongoing jobs, this might hamper performance. I had one run of this fail seemingly because of cpu constraints (timed out during pluto build). 

For that reason, maybe weekly would make more sense? Or do the builds sequentially instead of in parallel?

alexrichey (dismissed):
Very cool, esp the matrix strategy. I like the idea of building everything, often. Maybe it's something that gets kicked off on e.g friday night, to let it run over the weekend? Regardless, good stuff here.

damonmcc (dismissed):
🚀 good stuff! we'll see if the DB's cpu can handle it

PR #363, opened 2023-11-14
fix DB schema cleaning

### LINKED ISSUE #341: auto DB schema cleaning is deleting too many schemas
expected vs actual schema suffixes may be causing the issue
- e.g. `build_main` vs `run_main`

damonmcc:
resolves #341 

## approach
- use all three potential schema prefixes when creating the list of DB schemas to save from deletion

## background
We currently have three DB schema prefixes (see screenshots below):
- `build_` when we have to construct the schema name before we can use `utils.postgres.PostgresClient`
  - [CPDB example](https://github.com/NYCPlanning/data-engineering/blob/kpdb-drop-comments/.github/workflows/cpdb_build.yml#L21)
  - this is true for every products except Template DB [here](https://github.com/NYCPlanning/data-engineering/blob/kpdb-drop-comments/products/template/build_scripts/__init__.py#L19) because they all use bash scripts to run SQL
- `pr_` when a pure-python build is triggered by PR tests
- `run_` when a pure-python build is triggered manually

image
image

### DIFF
--- a/dcpy/builds/clean_artifacts.py
+++ b/dcpy/builds/clean_artifacts.py
@@ -7,19 +7,27 @@ from . import BUILD_REPO, BUILD_DBS
 
 
 def get_active_build_names() -> list:
-    # all remote branches
-    branches = github.get_branches(repo=BUILD_REPO)
+    # use the hard-coded "build_" prefix from github action workflow files
+    # and the event-based prefixes from build_metadata.build_name
+    branches = github.get_branches(repo=BUILD_REPO)  # all remote branches
     branch_build_names = sorted(
-        [build_metadata.build_name("branch", branch) for branch in branches]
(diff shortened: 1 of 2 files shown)


alexrichey (approved):
Cool - makes sense to me

fvankrieken:
What's the rationale for separating `build` vs `run`?

damonmcc:
@fvankrieken 

I don't think that was a deliberate separation. pretty sure we started with `build` for everything before we had `utils.postgres`

so when choosing for PR/manual I think I chose `pr`/`run` because they're both builds, and still `run` probably isn't the best choice

fvankrieken:
Should we just have one of those two for consistency? I don't see value to keeping both

PR #364, opened 2023-11-14
use branch name for DB schemas

damonmcc:
per changes and discussions in https://github.com/NYCPlanning/data-engineering/pull/363

## approach
- for all builds, use the branch to generate the DB schema name without any prefixes
- when cleaning stale build schemas, delete all that aren't existing branches

## tests
- [run](https://github.com/NYCPlanning/data-engineering/actions/runs/6908626627/job/18798306467#step:4:8) of cleaning of TDB schemas after these changes
- screenshot of schema created by tests in this PR:
  - image

## considerations
- `build_` is currently used in builds that don't use `PostgresClient`, otherwise `run_`/`pr_` are used
- there's strong interest in running all builds on `main` every night and not auto-running product test builds in PRs
  - recent discussion has raised the concern of nightly `main` builds overwriting data in QA. will address that in a follow-up issue


dcpy/connectors/edm/build_metadata.py
alexrichey, line 1: Non-urgent thought: Should this file actually be `dcpy/builds/metadata`?

sf-dcp:
My understanding is that we currently build published version products using our own branches. Wouldn't we want to have the latest version of schema in db even though the branch gets deleted? Or are we moving to builds from `main`?

sf-dcp (approved):

damonmcc:
> My understanding is that we currently build published version products using our own branches. Wouldn't we want to have the latest version of schema in db even though the branch gets deleted? Or are we moving to builds from `main`?

true it'd be useful to always have the latest published build in the DB, but we don't yet have anything that relies on or ensures that (e.g. when I look at files in an S3 folder with published versions, can I tell what branch was used to build it? maybe with the commit hash in the file metadata?)

frequently building drafts on main is helpful since, assuming the branch was merged, the build on main would be identical to the draft that was published

dcpy/connectors/edm/build_metadata.py
damonmcc: to maintain our desired import order of `utils -> connectors -> builds`, I tried moving the two functions to better places in [1a1d0dc](https://github.com/NYCPlanning/data-engineering/pull/364/commits/1a1d0dc460d4dc3dfccd497e224bc3a66e8a6b20)

doesn't feel ideal importing from `builds.metadata` but feels like that can't really change without a refresh of what lives in connectors vs utils

sf-dcp (approved):

PR #366, opened 2023-11-15
Refactor pg out of recipe imports

alexrichey:
Refactors sqlalchemy out of recipes and into the postgres client where they belong. Also adds unit tests.

This was spurred from frustrations mocking out sqlalchemy in recipes tests. That was a good indication that all that functionality should be encapsulated in our `postgres` module.

### DIFF
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -7,7 +7,6 @@ import pandas as pd
 from pathlib import Path
 from pydantic import BaseModel
 import shutil
-from sqlalchemy import text, update, Table, MetaData
 from typing import List
 import yaml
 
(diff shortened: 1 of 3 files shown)


dcpy/connectors/edm/recipes.py
```diff
@@ -171,42 +170,41 @@ def import_dataset(
...
             if ds.preprocessor is not None:
                 preproc_mod = importlib.import_module(ds.preprocessor.module)
                 preproc_func = getattr(preproc_mod, ds.preprocessor.function)
                 df = preproc_func(ds.name, df)
-            table_name = ds.import_as or ds.name
-            logger.info(f"Inserting dataframe into {table_name}")
-            pg_client.insert_dataframe(df, table_name)
-
-    with pg_client.engine.begin() as con:
-        con.execute(
```
alexrichey, line 194: FYI, all of this stuff 👇 is moved into the PostgresClient

dcpy/utils/postgres.py
```diff
@@ -153,6 +171,19 @@ def get_table_columns(self, table_name: str) -> list[str]:
         )
         return sorted(column_names["column_name"])
 
+    def add_table_column(
+        self,
+        table_name: str,
+        *,
+        col_name: str,
+        col_type: str,
+        default_value: str = "",
```
alexrichey, line 180: for now, the sole use is strings. But we should revisit this for other types

dcpy/utils/postgres.py
```diff
@@ -153,6 +171,19 @@ def get_table_columns(self, table_name: str) -> list[str]:
...
+    ):
+        q = f"ALTER TABLE {table_name} ADD COLUMN {col_name} {col_type}"
+        if default_value:
```
damonmcc, line 183: should this be `if default_value == ""`? doesn't seem like it'd ever be `None` yet

dcpy/test/connectors/edm/recipes/test_recipes.py
```diff
@@ -69,3 +80,48 @@ def test_serializing_and_deserializing(self):
...
+        )
+        recipes.import_dataset(ds, pg_mock)
+
+        # Verify the values that PostgresClient.insert_dataframe was called with
+        (
+            df_inserted_actual,
+            table_insert_name_actual,
+        ) = pg_mock.insert_dataframe.call_args_list[0][0]
+        assert df_inserted_actual.equals(_mock_preprocessor(ds.name, _test_df))
+        assert table_insert_name_actual == ds.import_as
```
damonmcc, line 127: cool I've never seen `call_args_list` before!

would it be nice to construct `_test_df` with the column `name` already in it or is it nicer to be more dynamic and use `ds.name` to build the "expected" dataframe?

dcpy/utils/postgres.py
alexrichey: Well, this being python, you never really know what you'll get, so I opted for the `truthiness` approach. Happy to change if you feel strongly.

dcpy/test/connectors/edm/recipes/test_recipes.py
alexrichey: Yeah, I haven't used it before either. I've typically just used the `assert_called_with` style methods to check args, but dataframes are weird about equality. 

Re: your question, the point of the test is largely to check that this mock preprocessor that we've bolted onto dcpy actually gets called. Since we have access to that method here, makes sense to just use it to verify outputs. The actual transformation is pretty negligable - just chose it because we do something similar in facdb, IIRC

dcpy/utils/postgres.py
damonmcc: o yea I forgot `if ""` evaluates to `False` !

damonmcc (approved):

Issue #268, opened 2023-10-03
Template DB enhancements


damonmcc:
## motivations

Template DB is intended to be a mock data product for us to test our platform and experiment with approaches we may adopt as standards.

There are a few things that Template DB should be to meet those use cases:
- fast to build
- representative of our current build standards (e.g. a recipe file to declare source data)
- using our platform end-to-end (from loading to distribution)

## ideas after initial build design in #255 
- [x] add a QA page
- [x] create a metadata.yml
- [x] determine the boroughs of green spaces
- [x] deduplicate landmarks records
- [x] export as a shapefile to enable distribution to Open Data

PR #367, opened 2023-11-16
create QA page for Template DB with a map

damonmcc:
related to #357
related to #268 

## notes
- recommend reviewing each commit in sequence
- the purpose of this is to create a new QA page that is relatively simple and might inform future work to improve and generalize out QA pages
- abandoned plans:
  - not adding a ZAP page yet because those build outputs don't yet use our draft/publish approach
  - not adding a KPDB page yet because those build outputs are not in the S3 bucket `edm-publishing`. `connectors.edm.publishing` is pretty hardcoded to only use that bucket

## approach
- expand `dcpy.utils.geospatial` to help create `GeoDataFrame`s and simple interactive maps
- for a product's build outputs that we'd like to QA, use a `dataclass` named `BuildOutput` to structure the details and different forms of the loaded data
- display the entirety of a build output's data, metadata, and minimal summary details like shape and geometry type counts
- use a sequence of buttons to progress through the computationally intensive stages of the page
  1. load data
  2. convert to GeoDataFrames where applicable
  3. generate maps

### before a build version is chosen
Screenshot 2023-12-05 at 1 54 38 PM

## loaded data and metadata
Screenshot 2023-12-05 at 4 28 54 PM

### while converting a DataFrame to a GeoDataFrame
Screenshot 2023-12-05 at 1 55 20 PM

### while generating maps of all GeoDataFrames
Screenshot 2023-12-05 at 1 58 35 PM

### a map of a Template DB build (38,978 points, 2,665 polygons)
Screenshot 2023-12-05 at 1 59 52 PM


python/constraints.txt
```diff
@@ -111,26 +121,35 @@ diagrams==0.23.4
     # via -r python/requirements.in
 diff-cover==8.0.1
     # via sqlfluff
+duckdb==0.9.2
```
fvankrieken, line 124: welcome to the duckdb era

dcpy/utils/geospatial.py
```diff
@@ -12,6 +15,21 @@
 
 from dcpy.utils.logging import logger
 
+GEOMETRY_FORMATS = {
+    "WKT": "EPSG:4326",
+    "WKB": "ESPG:3867",
```
fvankrieken, line 20: Not entirely sure how this is going to be used, but is there a reason to map a format to a specific projection?

dcpy/utils/geospatial.py
```diff
@@ -22,6 +40,72 @@
...
+) -> gpd.GeoDataFrame:
+    data = data.copy()
+    match geometry_format:
+        case "WKT":
+            data[geometry_column] = gpd.GeoSeries.from_wkt(data[geometry_column])
+        case "WKB":
+
+            def _try_wkb(wkb):
+                try:
+                    return pd.Series([shapely.from_wkb(wkb), None])
```
fvankrieken, line 57: this shouldn't be a return here, right? but some sort of variable assignment

dcpy/utils/geospatial.py
```diff
@@ -22,6 +40,72 @@
...
+            )
+        case _:
+            raise NotImplementedError(
+                f"Unsupported geometry format {geometry_format:}. Must be one of the keys in the dict {GEOMETRY_FORMATS:}"
+            )
+    data.drop([geometry_column], axis=1, inplace=True)
+    geo_data = gpd.GeoDataFrame(
+        data,
+        geometry=new_geometry_column,
+        crs=PROJECTIONS["WGS_84_DEG"],
```
fvankrieken, line 72: Just not familiar with gdf constructor - is this assuming that the input is in wgs84? Or converting to

apps/qa/src/components/build_outputs.py
```diff
@@ -0,0 +1,127 @@
...
+
+@dataclass
+class BuildOutput:
+    file_name: str
+    metadata: dict
+    dataframe: pd.DataFrame | None = None
+    geodataframe: gpd.GeoDataFrame | None = None
+    geodataframe_for_display: gpd.GeoDataFrame | None = None
+    map: lmf.Map | None = None
+    qa_warnings: list[str] = field(default_factory=list)
```
alexrichey, line 20: I definitely like using a class to cache the long-running computations. It does seem like the `BuildOutput` is encompassing a little too much though, based on the logic in `generate_geo_data` and `show_build_output`. (I'm an adherent to the Josh Bloch maxim that "classes should strive for immutability")

 Maybe you could start with a minimal set of fields on the `BuildOutput`, then subclass it to have a `GeoBuildOutput` (which would have the  `geodataframe`, `geodataframe_for_display`, and the `map`. Then you could move the display fns into `GeoBuildOutput`.

Since these are just code-style nits, it's not a blocker. Of course I'm happy to pair on it, to scratch my OOP itch 🙂

apps/qa/src/publishing.py
```diff
@@ -2,14 +2,21 @@
 from zipfile import ZipFile
 import streamlit as st
 
+from dcpy.utils import s3
 from dcpy.connectors.edm import publishing
 
+from src.constants import BUCKET_NAME
```
alexrichey, line 8: Can we just use `publishing.BUCKET`?

dcpy/utils/geospatial.py
```diff
@@ -22,6 +40,72 @@
...
+    geometry_column: str,
+    new_geometry_column: str = "geometry",
+) -> gpd.GeoDataFrame:
+    data = data.copy()
+    match geometry_format:
+        case "WKT":
+            data[geometry_column] = gpd.GeoSeries.from_wkt(data[geometry_column])
+        case "WKB":
+
+            def _try_wkb(wkb):
```
alexrichey, line 55: Just curious, why not to use `GeoSeries.from_wkb`?

dcpy/utils/geospatial.py
```diff
@@ -22,6 +40,72 @@
...
+                lambda row: _try_wkb(row[geometry_column]), axis=1
+            )
+        case _:
+            raise NotImplementedError(
+                f"Unsupported geometry format {geometry_format:}. Must be one of the keys in the dict {GEOMETRY_FORMATS:}"
+            )
+    data.drop([geometry_column], axis=1, inplace=True)
+    geo_data = gpd.GeoDataFrame(
+        data,
+        geometry=new_geometry_column,
```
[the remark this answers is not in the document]
sf-dcp: Just curious, why do you need to create a gpd column, `new_geometry_column` before creating a gpd df as opposed to feeding the original column into gpd df?

apps/qa/src/template/template.py
```diff
@@ -0,0 +1,110 @@
...
+    show_build_output,
+    show_map,
+)
+
+
+def template():
+    st.title("QA of Template DB")
+    product_key = sidebar.data_selection("db-template")
+
+    csv_files = [
```
sf-dcp, line 16: It would be nice to use `product_key` to grab data from S3

apps/qa/src/template/template.py
```diff
@@ -0,0 +1,110 @@
...
+        This QA page highlights key scenarios that can indicate potential data issues in a build.
+        """
+    )
+
+    if not product_key:
+        st.warning("No build version selected")
+        return
+
+    if "build_outputs" not in st.session_state:
+        st.session_state["build_outputs"] = {}
```
fvankrieken, line 34: I have a little bit of a gut response to avoid using session_state, though it does simplify code a good amount (and let you render the whole page while things are loading without crazy code). One note is that to generalize, we'll need to specify which product page is being viewed. And we might have to think about memory concerns if we're caching a bunch of big gdfs and maps for multiple products. But that's a problem that we deal with when it comes up 🙂

apps/qa/src/components/build_outputs.py
fvankrieken: I would agree with this, it'd help provide a bit of clarity about what information is stored at which point in the code that's dealing with these bojects

apps/qa/src/template/template.py
damonmcc: yea I've always tried to avoid using session_state as much as I can, but the streamlit docs kept pushing me toward using it for this page's approach

I think it's also why I kept it's use in this new TDB page and not in any components meant to be shared by pages to not impose its use

dcpy/utils/geospatial.py
damonmcc: good call, this should just be a list or enum. and I think I'll rename some things to not imply all coordinate reference systems are proejcted

interesting background on ESPG stuff I just stumbled through:
a European Petroleum Survey Group (EPSG) code is just a unique code that can represent many types of things like coordinate systems, datums, and spheroids ([source](https://support.virtual-surveyor.com/support/solutions/articles/1000261353-what-is-an-epsg-code-)).

we have a few coordinate systems of interest. some are geodesic like [EPSG:4326](https://epsg.io/4326) and have units of degrees. some are projected like [EPSG:4087](https://epsg.io/4087) for the world in meters and [EPSG:2236](https://epsg.io/2263) for NY in meters

dcpy/utils/geospatial.py
damonmcc: unfortunately this is assuming that it's wgs84. to do spatial things later like reprojecting via `GeoDataFrame.to_crs`, it has to have a crs set in the first place, so I figured might as well do it when it's created

would probably be better to make the CRS an argument to `convert_to_geodata` and not assume it here?

apps/qa/src/template/template.py
damonmcc: I think it is, further down on [line 57](https://github.com/NYCPlanning/data-engineering/blob/dm-new-qa-pages/apps/qa/src/template/template.py#L57)

the `csv_files` list here doesn't feel ideal but I imagined we won't always QA all output files to the same degree so tried this approach to declare the sort of "primary" csv files to QA

dcpy/utils/geospatial.py
damonmcc: sorry I shouldn't put a comment in this section explaining why it's so weird

see [4050fc7](https://github.com/NYCPlanning/data-engineering/pull/367/commits/4050fc72f87d76a31b85e30e8d0f50765826702d). the same may be true for `from_wkt` as well so I'll test that

dcpy/utils/geospatial.py
damonmcc: since it's used in a `DataFrame.apply` to create two new columns, I think it has to return rather than assign

dcpy/utils/geospatial.py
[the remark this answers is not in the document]
damonmcc: @sf-dcp don't need to, so I think this is a relic of during my first pass wanting to keep and show both the source column and the converted columns

I think I regret the choice to drop the original column so I should probably change this to keep both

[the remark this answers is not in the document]
damonmcc: see new commits for handling WKT and keeping original geometry column

apps/qa/src/components/build_outputs.py
damonmcc: @alexrichey I do wanna improve this this but KPDB stuff keeps coming up lol down to save it for a follow-up PR?

sf-dcp (approved):

PR #369, opened 2023-11-17
Fix recipes default text value bug

alexrichey:
When I was testing, all the recipe datasets versions were numeric. This broke when it encountered '23v3'.

> ProgrammingError: (psycopg2.errors.SyntaxError) trailing junk after numeric 
literal at or near "23v"
LINE 1: ...dcp_pluto ADD COLUMN data_library_version text DEFAULT 23v3;


damonmcc (approved):

fvankrieken:
Coming in just too late - is this something where we might want the flexibility to have default be numeric? Different type signature and logic based on variable type?

PR #370, opened 2023-11-17
[POC] try using dcpy in data-library actions

damonmcc:
[error](https://github.com/NYCPlanning/data-engineering/actions/runs/6908372987/job/18797511402#step:6:365) during use of dcpy in a data-library script: `ModuleNotFoundError: No module named 'dcpy'`


fvankrieken:
So dcpy will need to be installed - this can happen in the single runner action, or maybe specifically within data-library/library.sh, in the function setup. It's a little tricker when doing it in the docker image publishing, we'll just need to copy dcpy into the docker container as well for install

fvankrieken:
Currently dcpy install in general happens in bash/docker_container_setup.sh

PR #374, opened 2023-11-20
Add Socrata Connector and GH workflow

alexrichey:
This adds a Socrata Connector that will 
- Open a Revision on a dataset
- Push a shapefile
- Push attachments
- Rename/reorder columns
- And, finally, close the revision

Additionally, there's a Github action that will push from the `edm-distributions` bucket to Socrata, taking a dataset name and version.

For review, I'd probably start at the bottom of `publish.py` and work upwards. 

As the comment in Metadata says, I'd checked in some metadata for MIH_TEST. It's in the format that Socrata expects, and we'll need to work backwards to create a DCP base metadata

Adding Casey and Jack for visiblity - no pressure to fisk the code

### DIFF
--- a/dcpy/connectors/socrata/metadata.py
+++ b/dcpy/connectors/socrata/metadata.py
@@ -1,0 +1,148 @@
+"""A first pass at metadata we'd sync to socrata.
+
+Data format:
+`mih_test.metadata` entry is in the format expected by their publishing API.
+`mih_test.columns` is as well, but must be passed to a different endpoint.
+(so while they're both metadata, we have to separate them.)
+
+This should eventually be generated from pure DCP product metadata, as many of
+the fields, especially those at column level, would be quite helpful at all
+stages of the build.
+
+FYI, the mih_test dataset is private, and will remain so.
+"""
+
+datasets = {
+    "mih_test": {
+        "four_four": "p6cc-cxkc",  # A `four_four` is the ID for a socrata dataset
+        "attachments": ["nycmih_metadata.pdf"],
+        "columns": [
+            {
+                "display_name": "the_geom",
+                "datatype_name": "multipolygon",
+                "description": "Geometry type, changed",
+                "field_name": "the_geom",
+            },
+            {
+                "display_name": "Boro",
+                "datatype_name": "text",
+                "description": "The NYC Borough where the Mandatory Inclusionary Housing (MIH) area is mapped. The boro code is a single digit identifier, indicating a particular borough (1 = Manhattan, 2 = Bronx, 3 = Brooklyn, 4 = Queens, 5 = Staten Island).",
+                "field_name": "boro",
+            },
+            {
+                "display_name": "Status",
+                "datatype_name": "text",
+                "description": "Status of the ULURP text amendment application (Adopted = final approval, Pipeline = awaiting approval)",
+                "field_name": "status",
+            },
+            {
+                "display_name": "Project Name",
+                "datatype_name": "text",
+                "description": "Project Name of the Mandatory Inclusionary Housing (MIH) area.",
+                "field_name": "projectnam",
+            },
+            {
+                "display_name": "Date Adopted",
+                "datatype_name": "calendar_date",
+                "description": "The date the ULURP text amendment application was adopted.",
+                "field_name": "dateadopte",
+            },
+            {
+                "display_name": "ZR_ULURP Number",
+                "datatype_name": "text",
+                "description": "The ULURP number of the ULURP text amendment application.",
+                "field_name": "zr_ulurpno",
+            },
+            {
+                "display_name": "ZR_Map",
+                "datatype_name": "text",
+                "description": "The map number referenced in Appendix F of the Zoning Resolution for the mapped Mandatory Inclusionary Housing (MIH) area.",
+                "field_name": "zr_map",
+            },
+            {
+                "display_name": "CD",
+                "datatype_name": "text",
+                "description": "Community District where the Mandatory Inclusionary Housing (MIH) area is mapped.",
+                "field_name": "cd",
+            },
+            {
+                "display_name": "MIH_Option",
+                "datatype_name": "text",
+                "description": "Level of affordability and options applied to the mapped Mandatory Inclusionary Housing (MIH) area.",
+                "field_name": "mih_option",
+            },
+            {
+                "display_name": "Zoning_Map",
+                "datatype_name": "text",
+                "description": "Zoning Map index number where the Mandatory Inclusionary Housing (MIH) area is mapped.",
+                "field_name": "zoning_map",
+            },
+            {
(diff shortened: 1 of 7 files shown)


dcpy/connectors/socrata/publish.py
```diff
@@ -0,0 +1,350 @@
...
+                self.revision_num = self._raw_resource_data["revision_seq"]
+                self.four_four = self._raw_resource_data["fourfour"]
+                self.metadata = self._raw_resource_data["metadata"]
+                self.attachments = self._raw_resource_data["attachments"]
+
+        class RevisionDataSource:
+            def __init__(self, socrata_source):
+                self.socrata_source = socrata_source
+
+                # AR Note: It's not clear to me why/when you would have multiple schemas/output
```
fvankrieken, line 87: Some views on opendata are multiple data sources, could it be a case like that?

https://data.cityofnewyork.us/Housing-Development/Building-Footprints/nqwf-w8eh

One "four_four" (great name for the field) corresponds to two datasets in this case

fvankrieken: Or is this truly for one data source here, not for one "four_four"

fvankrieken:
Bit of a context question - what's the lifecycle of a revision in socrata? Just curious with the logic around open revisions - what led to writing that code, what might cause there to be existing open revisions, and why the need to make sure they're closed.

What other states of revisions are there?

fvankrieken:
From a first pass, this looks great. I'll need to give it another pass after some DayQuil kicks in. 

How have you gone about testing functionality?

dcpy/connectors/socrata/publish.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+def push_shp_from_s3(dataset_name: str, version: str):
+    """Push Shapefile from edm-packaging to Socrata, and sync metadata."""
+    dataset_metadata: dict = datasets[dataset_name]
+    four_four: str = dataset_metadata["four_four"]
+    socrata_metadata = dataset_metadata["metadata"]
+    logger.info(
+        f"Syncing shapefile and metadata for {dataset_name}/{version} to Socrata {four_four}"
+    )
```
damonmcc, line 291: might be nice to have the full `four_four` URL to navigate to either at the start or end of a sync, maybe just the end to see the changes on Open Data

dcpy/connectors/socrata/publish.py
alexrichey: Yeah, it's the latter. A Source is an uploaded dataset (say, an individual shapefile) of which you may have multiple... but I don't quite get why there would be multiple schemas for that one thing.

alexrichey:
> Bit of a context question - what's the lifecycle of a revision in socrata? Just curious with the logic around open revisions - what led to writing that code, what might cause there to be existing open revisions, and why the need to make sure they're closed.
> 
> What other states of revisions are there?

A revision is just either open or closed, and then you apply it, and it becomes read-only, and the GUI won't show that revision anymore.  And you can have multiple open revisions per dataset. So for our purposes, we just open a revision, make modifications, then apply it, and it's gone. Also if you have multiple open revisions, they will warn you in the UI that one is potentially stale. 

Closing the existing revisions was really more of a testing thing, where I wanted a fresh revision to modify and apply, and I didn't want to clutter our Socrata instance with open revisions. It might eventually make sense to remove the call to discard open revisions, but for now it keeps things tidy.

dcpy/connectors/socrata/publish.py
alexrichey: that's a great idea. We can definitely add a link to the dataset in socrata. And maybe if the revision is left hanging, a link to that as well.

alexrichey:
> From a first pass, this looks great. I'll need to give it another pass after some DayQuil kicks in.
> 
> How have you gone about testing functionality?

We have a dummy dataset for Mandatory Inclusionary Housing (MIH_TEST) [here](https://data.cityofnewyork.us/Housing-Development/MIH_Test/p6cc-cxkc) and if you want to poke around the creds are in 1pw, in the `Data Engineering` vault. Feel free to tweak anything, as it won't cause problems.

I think I'm at something like revision 80. I've been pushing various datasets, attachments, and modifying metadata for that dataset. Our Socrata workflow is fairly straightforward, since we just overwrite the data / metadata every time we push. Socrata Revisions have functionality around incremental modifications, ie adding calculated columns, or appending, etc. but since we're note using any of that, our work becomes much easier.

dcpy/connectors/socrata/publish.py
```diff
@@ -0,0 +1,350 @@
...
+    s3.download_folder(
+        DISTRIBUTIONS_BUCKET, f"{dataset_name}/{version}/", download_root_path
+    )
+
+    dataset = Dataset(four_four=four_four)
+    dataset.discard_open_revisions()
+
+    rev = dataset.create_replace_revision()
+
+    data_source = rev.push_shp(path=dataset_path / "shapefile.zip")
```
sf-dcp, line 308: What happens if we successfully push a shapefile but we have a failure anywhere between here and applying the revision below? Does the shapefile get published or is kept as a part of the open revision?

dcpy/connectors/socrata/metadata.py
```diff
@@ -0,0 +1,148 @@
...
+(so while they're both metadata, we have to separate them.)
+
+This should eventually be generated from pure DCP product metadata, as many of
+the fields, especially those at column level, would be quite helpful at all
+stages of the build.
+
+FYI, the mih_test dataset is private, and will remain so.
+"""
+
+datasets = {
```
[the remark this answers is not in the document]
alexrichey: as far as I know, pretty much everything can be public. There are certainly cases where the data can't, but can't think of any cases with the metadata.

dcpy/connectors/socrata/publish.py
alexrichey: Good question - it's the latter. There'll be an open revision that folks can investigate, and close/apply manually

dcpy/connectors/socrata/publish.py
```diff
@@ -0,0 +1,350 @@
...
+            # TODO: diff between name and filename? and what about blob_id?
+            name=attachment_md_raw["filename"],
+            filename=attachment_md_raw["filename"],
+            asset_id=attachment_md_raw["file_id"],
+            blob_id=None,
+        )
+
+
+def push_shp_from_s3(dataset_name: str, version: str):
+    """Push Shapefile from edm-packaging to Socrata, and sync metadata."""
```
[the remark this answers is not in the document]
alexrichey: Good catch - fixed!

sf-dcp (approved):

fvankrieken (dismissed):
Awesome

damonmcc:
happy to approve and test drive this with MIH_test and COLP! seeing some python package merge conflicts though

PR #376, opened 2023-11-20
Fix broken nightly QA runs, zap tests

### LINKED ISSUE #371: Everything is broken in the new nightly build action
https://github.com/NYCPlanning/data-engineering/actions/runs/6918492239/job/18820824380

fvankrieken:
[Latest run](https://github.com/NYCPlanning/data-engineering/actions/runs/6934460801), finished

Resolves #371, mostly. Other than

1. KPDB - still broken on main, failing due to some source data not being present?
2. FacDB - failing due to trying to apply preprocessing to a datasettype that doesn't support it

Commits
1. fix issues around pg connection string
2. fix my issues in testing. First, for the non-pr test case it was failing because of the empty variable. Second, something was always broken for zap. Went back and found my "successful" test run from before, and all I actually successfully did was print out some text.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,7 @@ jobs:
     outputs:
       path_filters: ${{ steps.filter.outputs.changes }}
       matrix: ${{ steps.set-matrix.outputs.matrix }}
+      matrix_raw: ${{ steps.set-matrix.outputs.matrix_raw }}
     steps:
       - uses: actions/checkout@v3
       - uses: dorny/paths-filter@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -147,11 +152,12 @@ jobs:
...
     - name: Pytest
+      env:
+        calls: ${{ toJSON(matrix.calls) }}
       run: |
-        echo '${{ toJson(matrix.calls) }}' \
-          | sed '$ d;1d;s/^[[:space:]]*\"//;s/\",\{0,1\}[[:space:]]*$//'\
-          | while read call; do
-          python3 -m pytest $call
+        echo "$calls" | jq -cr ".[]" | while read call; do
+          IFS=$'\n' arr=( $(xargs -n1 <<<"$call") )
```
fvankrieken, line 159: Couldn't get `readarray` to work as I wanted but `xargs` was simpler @alexrichey

fvankrieken: Hope we never think about that monstrous sed call I initially wrote again. It worked! But so does the tiny `jq` call it's replaced with

products/zap-opendata/pyproject.toml
damonmcc, line 1: is this file needed for the ZAP section of `pytest.json`?
```json
"calls": [
    "-m 'not end_to_end' -s --verbose --verbose --durations=0 --cov .",
    "-m 'end_to_end' -s --verbose --verbose --durations=0 --cov ."
]
```

if it's only because of the `--cov .` parts, probably better to not report ZAP test coverage and drop this file?

fvankrieken:
Reran zap tests now that path filters don't trigger it. Passing!

https://github.com/NYCPlanning/data-engineering/actions/runs/6935230510/job/18864980832

Discovered a new issue with this - still had a bug on a PR that didn't match any of the path changes.

fvankrieken:
> Reran zap tests now that path filters don't trigger it. Passing!
> 
> https://github.com/NYCPlanning/data-engineering/actions/runs/6935230510/job/18864980832
> 
> Discovered a new issue with this - still had a bug on a PR that didn't match any of the path changes. But given that latest run has started running without issue (as opposed to erroring at start of matrix job), seems to be set

Oops nevermind - still broken in case when no changes in path filters

fvankrieken:
> > Reran zap tests now that path filters don't trigger it. Passing!
> > https://github.com/NYCPlanning/data-engineering/actions/runs/6935230510/job/18864980832
> > Discovered a new issue with this - still had a bug on a PR that didn't match any of the path changes. But given that latest run has started running without issue (as opposed to erroring at start of matrix job), seems to be set
> 
> Oops nevermind - still broken in case when no changes in path filters

Fixed

alexrichey (approved):
nice - glad to see you got the bash stuff worked out

PR #379, opened 2023-11-21
QA app docker fixes, GRU page TLC

fvankrieken:
## Commit 1

quick fix for a broken docker build

[Fixed action](https://github.com/NYCPlanning/data-engineering/actions/runs/6947043976)

I feel like I'm starting to tie myself in knots with these python dependencies within the apps. Difficulty of getting dcpy into data-library very tied to this as well. Think we should merge this as is assuming all goes well, but we need to brainstorm a better long term solution - copying around files manually and then having assumptions made in the dockerfile, needing to only build it from a specific script, etc all seems a bit unsustainable.

## Commit 2
Lack of type signature for `dcpy.connectors.s3` led to a mypy-missed issue. Fixed, and also decided to type github workflow runs a bit more rigorously rather than just using loose dicts

### DIFF
--- a/apps/qa/requirements.txt
+++ b/apps/qa/requirements.txt
@@ -1,4 +1,3 @@
-./dcpy
 folium
 geoalchemy2
 mapclassify
--- a/apps/qa/src/gru/helpers.py
+++ b/apps/qa/src/gru/helpers.py
@@ -1,6 +1,4 @@
 import time
-import os
-import boto3
 from datetime import datetime
 import streamlit as st
 import requests
@@ -9,7 +7,7 @@ import re
 from dcpy.utils import s3
 from dcpy.connectors import github
 from dcpy.connectors.edm import recipes
-from .constants import tests
+from .constants import qa_checks
 
 
 def get_source_version(dataset: str) -> dict[str, str]:
@@ -17,11 +15,8 @@ def get_source_version(dataset: str) -> dict[str, str]:
         bucket = "edm-publishing"
         prefix = "gru/dcp_saf/"
         folders = s3.get_subfolders(bucket, prefix)
-        folders = [
-            folder["Prefix"].split("/")[-2]
-            for folder in folders
-            if "latest" not in folder["Prefix"]
-        ]
+        if "latest" in folders:
+            folders.remove("latest")
         latest_version = max(folders)
         timestamp = s3.get_metadata(bucket, f"{prefix}{latest_version}/dcp_saf.zip")[
             "last-modified"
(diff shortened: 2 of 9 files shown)


apps/qa/requirements.txt
```diff
@@ -1,4 +1,3 @@
-./dcpy
```
fvankrieken, line 1: using pyproject.toml instead to not install tests. Figured we should be consistent about how it gets installed in different environments

fvankrieken:
Reverting to draft as I fix a bug in the app, separate from docker stuff

dcpy/connectors/github.py
```diff
@@ -1,25 +1,39 @@
+from __future__ import annotations
```
fvankrieken, line 1: This is so that a `classfunction` can return (typed) its own class. I'll comment below on if we even want that

dcpy/connectors/github.py
```diff
@@ -1,25 +1,39 @@
...
+@dataclass
+class WorkflowRun:
+    name: str
+    status: str
+    conclusion: str
+    timestamp: datetime
+    url: str
+
+    @classmethod
+    def parse_from_github(cls, workflow: dict) -> WorkflowRun:
```
fvankrieken, line 24: For some reason this felt better than defining `__init__`. Though now that I think about it, there's not really a case where we'd create one of these NOT from a github api response... Open to thoughts

apps/qa/src/gru/helpers.py
```diff
@@ -54,11 +49,11 @@ def map_geosupport_version(patched_version: str) -> str:
...
 
-def get_qaqc_runs(geosupport_version: str) -> dict[str, dict]:
-    workflows: dict[str, dict] = {}
-    raw_workflow_runs: list[dict] = []
+def get_qaqc_runs(geosupport_version: str) -> dict[str, github.WorkflowRun]:
+    workflow_runs: dict[str, github.WorkflowRun] = {}
+    raw_workflow_runs: list[github.WorkflowRun] = []
     page = 0
-    while len(workflows) != 7 and (page == 0 or (len(raw_workflow_runs) > 0)):
+    while len(workflow_runs) != 7 and (page == 0 or (len(raw_workflow_runs) > 0)):
```
alexrichey, line 56: Curious what the magical 7 signifies here?

dcpy/connectors/github.py
alexrichey: def ➕ to not not futzing with __init__. Could also just move this constructor out of the class (removing the need to import __future__)

alexrichey (approved):
nice! Had one question about a magic number

dcpy/connectors/github.py
fvankrieken: I personally like static class constructor functions. Just field like `WorkflowRun.parse` reads better than `github.parse_workflow_run`

I'll add the frozen!

apps/qa/src/gru/helpers.py
fvankrieken: Ah. Number of distinct workflows. This function is looking for the most recent run of each, github api returns most recent first so it terminates when they're found. I will make the change though since this number is not that magical, and I have a constant which lists all 7, so I can grab the length pretty easily since that should be determining this

PR #380, opened 2023-11-21
Get Pluto Building

fvankrieken:
- Remove the temp fix for importing qa data
- Look for the right filename for lock file to set version env vars
- fix a bug in qa app - somehow a few months ago while mypy-ing qa, I renamed a function to one that does not exist. 🤷 
- get a longer term solution to the pg_dumps for the pluto qa stuff. Currently the process during a build is
  - pull in qa outputs from last build (sql dumps. And by "last build" I mean the published pluto which has version = bash variable `VERSION_PREV` which is set by a step in the pluto build github action)
  - import them
  - append to them (and replace data if there's already data for the current "version" somehow
  - pg_dump/csv export them, include them in export
Main issue is the pg dumps include the schema, and using the pg_dump cli there's no way to avoid this. Currently added some text processing of the sql file, might make sense to just have this logic rely on csvs instead of sql dumps to really keep things "data only"

Also going to clean up #405

### DIFF
--- a/products/pluto/recipe-minor.yml
+++ b/products/pluto/recipe-minor.yml
@@ -22,3 +22,6 @@ inputs:
       version: latest
     - name: dcp_zoningmapindex
       version: latest
+    - name: dcp_pluto
+      version: latest
+      import_as: previous_pluto
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -61,8 +61,8 @@ jobs:
       - name: Set versions env vars
         working-directory: products/pluto
         run: |
-          echo VERSION=$(yq .version recipe.lock.yml) >> "$GITHUB_ENV" && \
-          echo VERSION_PREV=$(yq '.inputs.datasets.[] | select(.name == "dcp_pluto") | .version' recipe.lock.yml) >> "$GITHUB_ENV"
+          echo VERSION=$(yq .version ${{ inputs.recipe_file }}.lock.yml) >> "$GITHUB_ENV" && \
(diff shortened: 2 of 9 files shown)


fvankrieken:
Also ran into an issue where importing `qaqc_outliers.sql` from last published build had the actual build schema in the pg_dump. Interestingly the other files all had public schema, and they use the exact same machinery around building the tables, dumping them, importing, etc. Going to see if issue stays with this new build or if there was something specifically funky with the last pluto build that caused this

fvankrieken:
Why the hell did I do [this](https://github.com/NYCPlanning/data-engineering/commit/07cad2aacedc7fdc514cabb8c11835dbce738ceb#diff-9e8f391fe30e1dc973b086941eb7fb3ffc68139307807c77ebc33504ad318e57R63)?

fvankrieken:
Simplified this, dropped the work I was doing around versioning and recipes to get this mergable.

products/pluto/recipe-minor.yml
```diff
@@ -22,3 +22,6 @@ inputs:
       version: latest
     - name: dcp_zoningmapindex
       version: latest
+    - name: dcp_pluto
+      version: latest
+      import_as: previous_pluto
```
alexrichey, line 27: Should this be sourced from `VERSION_PREV` in the environment?

alexrichey (dismissed):
Cool - thanks for fixing things up. Had one minor recipe question, but looks good. 

I'm certainly regretting baking the versions into the recipe file itself. Hopefully we can extract that out

fvankrieken:
Yeah I'll address that note. Re baking versions in, I was going to fix that in this draft but wanted to get this merged so pluto works on main, hoping to wrap that and other versioning tweaks up this week

fvankrieken:
Hmm. Currently in pluto, the VERSION_PREV env var comes from the recipe (via lock file), not vice-versa

fvankrieken:
Going to merge as-is for now and address that in my follow-up

Issue #295, opened 2023-10-13
Build dev docker images for use in pr tests


fvankrieken:
Not quite sure that this is 100% the solution we'd want, but just wanted to get some thoughts down and have a spot to brainstorm.

We have an inelegant way of testing changes to our dev/build images at the moment since tests just grab `{image}:latest` from dockerhub, we have to update these images if we want changes reflected in the tests. It would be nicer to build the docker image from the current checked out repository and use that. There are still a couple ways to go about this though, and things to be considered.

Let's say we're just thinking about a single push to an open pr for the time being. We would probably want to
1. check if changes have been made to `docker` or `python` folders in this PR (and bash utils? some of those are used in the docker setup. Maybe those should be isolated)
2. if changes have been made, build (all ?) images in docker folder, and push them somewhere. Theoretically we could push to dockerhub and tag based on pr# or branch name, I would prefer to use [`upload-artefact`](https://github.com/actions/upload-artifact) (and name by pr# or branch name). These are stored by github and expire after 90 days automatically, unlike dockerhub where 1. theyre publicly visible and 2. we would need to clean explicitly. However, not sure if such an image can be easily used in a subsequent job with the syntax `container: {image}` like we currently do. We could also maybe use githubs container registry to have some separation from our "official" images in dockerhub
3. if changes have been made, use newly built/pushed image as container rather than `latest` from dockerhub

Relatively straightforward but still some complexity in implementing. Just 3 for example, would need some interesting code in terms of how you conditionally either get a docker image from dockerhub versus downloading.

There's also then some complexity when we think about having many pushes to a PR - if we're not making changes in each push, we maybe want to be able to re-use the image that was built from our push yesterday to not have tests ballooning in runtime. So maybe at that point it'd be
1. Check if push has changes to `docker` or `python` etc. 
2. If changes have been made, build and push (to artefacts, dockerhub, github container registry, etc)
3. Check if pr has changes to relevant files
4. If 3 has changes (even if 1 did not), get pr-specific image. If not, get dockerhub latest image

fvankrieken:
We also then have the same issue for building on a dev branch when we've made changes. So maybe we
1. on pushes to any branch with changes to relevant files, build docker image and push somewhere. Separate action from build, test, etc
2. on build/test, check if branch has a branch-specific image stored, use it if so, but default then to dockerhub latest

[...]

PR #385, opened 2023-11-27
compile python requirements only weekly

fvankrieken:
The nightly boto updates seem like maybe a bit much - figure we can do this once per week, come in and check it out Monday.

It could be good to address #295 in some way, and have the test for this PR actually use whatever python versions are specified in the action. If we feel like that'd be worth it, happy to work on that this week


fvankrieken:
Actually pending approval, would like to just merge this to get both these commits in main - cherry picked a fix for the test job from my other branch

PR #389, opened 2023-11-27
Ar add parquet recipe support

alexrichey:
I wanted to take Cody for a spin, and adding Parquet support seemed like a fun detour. The unit test was entirely generated by Cody, but everything else took some finessing. 

I added a sample parquet file to [test_nypl_libraries](https://cloud.digitalocean.com/spaces/edm-recipes?path=datasets/test_nypl_libraries/20210122) if you want to test.

### DIFF
--- a/.github/workflows/colp_build.yml
+++ b/.github/workflows/colp_build.yml
@@ -63,7 +63,7 @@ jobs:
       run: ./bash/build_env_setup.sh
 
     - name: Dataloading ...
-      run: python -m dcpy.builds.load
+      run: python -m dcpy.builds.load recipe
 
     - name: Build COLP ...
       run: ./colp.sh build
--- a/dcpy/test/connectors/edm/recipes/test_recipes.py
+++ b/dcpy/test/connectors/edm/recipes/test_recipes.py
@@ -90,9 +90,15 @@ def _mock_preprocessor(name, df):
     return df
 
 
-def _mock_fetch_dataset(ds, _):
+def _mock_fetch_csv(ds, _):
     out_path = TEMP_DATA_PATH / f"{ds.name}.csv"
     _test_df.to_csv(out_path, index=False)
+    return out_path
+
+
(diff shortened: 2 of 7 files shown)


.github/workflows/colp_build.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       run: ./bash/build_env_setup.sh
 
     - name: Dataloading ...
-      run: python -m dcpy.builds.load
+      run: python -m dcpy.builds.load recipe
```
alexrichey, line 66: We need to add these because of a fun `typer` behavior where if there's only one command defined, it just ignores the app command, in this case the `load` in `@app.command("load")`.

dcpy/test/connectors/edm/recipes/test_recipes.py
```diff
@@ -118,10 +118,41 @@ def test_import_dataset(self):
...
+            version="1",
+            import_as="new_table_name",
+            file_type=recipes.DatasetType.parquet,
+            preprocessor=recipes.DataPreprocessor(module="dcpy", function="preproc"),
+        )
+
+        def _mock_fetch_parquet(ds, _):
+            out_path = TEMP_DATA_PATH / f"{ds.name}.parquet"
+            _test_df.to_parquet(out_path, index=False)
+            return out_path
```
damonmcc, line 147: maybe worth moving this to line 98ish to group it with `_mock_fetch_dataset`?

alexrichey: yep, good call. Fixed in latest commit

damonmcc (approved):

.github/workflows/colp_build.yml
fvankrieken: I've come across this a few times and I HATE IT SO MUCH

fvankrieken:
Love. One step closer to moving data library to only spitting out parquet

PR #390, opened 2023-11-27
Build dev docker images/tags

fvankrieken:
I would go commit-by-commit for review

Main goal here - in case where changes result in different docker environment. Desired state (and commits, coincidentally), are:
1. build image and push to dockerhub for the given branch. This happens with pull_request trigger as part of `test` gha. image uses same image name as usual, but publishes with tag "dev-{branch}". Only builds if things in `python` or `docker` folders are edited.
2. be able to use these dev images in builds. This happens via bool input for both the main build workflow and the specific build workflows
3. clean up "stale" image tags - check for active branches, delete all "dev" tags not corresponding to existing branches on github

Commit 4 right now is simply so I can test functionality in 3 more easily. [Here's a run](https://github.com/NYCPlanning/data-engineering/actions/runs/7051310388), I can verify that dev tags existed and then were deleted

Pr tests for this branch are testing commit 1

[Here's a job using a dev tag to build](https://github.com/NYCPlanning/data-engineering/actions/runs/7051225532/job/19193643375#step:2:18)

## Cons to this approach
- slows down testing. Doesn't check in any way if new commits have changed dev image definition since last push to a branch with an open PR. Don't know if there's a huge way around that though
- more complex gha yml

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,11 @@ on:
         type: boolean
         required: true
         default: false
+      dev_image: 
+        description: "Use dev image specific to this branch? (If exists)"
+        type: boolean
+        required: true
+        default: false
   workflow_call:
     inputs:
       dataset_name:
@@ -60,17 +65,24 @@ on:
       recipe_file:
         type: string
         required: true
+      dev_image:
+        type: boolean
+        default: false
 
 jobs:
   health_check:
     name: Health Check
     runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ inputs.dev_image && format('dev-{0}', github.head_ref || github.ref_name) || 'latest' }}
     steps:
       - name: Check inputs
         run: |
+          echo "running with image tag ${{ inputs.dev_image && format('dev-{0}', github.head_ref || github.ref_name) || 'latest' }}"
           echo "Workflow inputs are:"
           echo "${{ toJSON(github.event.inputs) }}"
   template:
+    needs: health_check
     if: inputs.dataset_name == 'template'
     uses: ./.github/workflows/template_build.yml
     secrets: inherit
@@ -79,7 +91,9 @@ jobs:
       logging_level: ${{ inputs.logging_level }}
       build_note: ${{ inputs.build_note }}
       run_export: ${{ inputs.run_export }}
+      image_tag: ${{ needs.health_check.outputs.tag }}
   cbbr:
+    needs: health_check
     if: inputs.dataset_name == 'cbbr'
     uses: ./.github/workflows/cbbr_build.yml
     secrets: inherit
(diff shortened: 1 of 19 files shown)


.github/workflows/python_compile_requirements.yml
```diff
@@ -32,3 +32,4 @@ jobs:
           title: Compile python requirements
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           assignees: fvankrieken, damonmcc, alexrichey, sf-dcp
+          branch: compile_python_reqs
```
fvankrieken, line 35: Ooh the other thing I might add - run nightly build on `push` to this branch. Though ideally we'd need to ensure that dev tags have been published before that happens. Need to think about it

.github/workflows/template_build.yml
```diff
@@ -27,7 +30,7 @@ jobs:
     name: Build
     runs-on: ubuntu-22.04
     container:
-      image: nycplanning/dev:latest
+      image: nycplanning/dev:${{ inputs.image_tag || 'latest' }}
```
fvankrieken, line 33: This was needed in the first commit since it's called by the `test` job. But ended up going with the same approach in commit 2

docker/publish.sh
```diff
@@ -12,40 +15,50 @@ pip3 install requests beautifulsoup4
...
-
+    echo $VERSIONSTRING
     export $(echo "$VERSIONSTRING" | sed 's/#.*//g' | xargs)
     export VERSION=$MAJOR.$MINOR.$PATCH
 
     export BUILD_ARGS="--build-arg RELEASE=$RELEASE --build-arg MAJOR=$MAJOR --build-arg MINOR=$MINOR --build-arg PATCH=$PATCH"
 }
 
 function common {
-    DOCKER_IMAGE_NAME=nycplanning/$1
```
fvankrieken, line 23: This is a function rather than just called because it will run into errors if an invalid image name has been specified. So in the section where we match image name (line 44), an error we define gets thrown rather than something more confusing

.github/workflows/build.yml
```diff
@@ -79,60 +91,79 @@ jobs:
       logging_level: ${{ inputs.logging_level }}
       build_note: ${{ inputs.build_note }}
       run_export: ${{ inputs.run_export }}
+      image_tag: ${{ needs.health_check.outputs.tag }}
```
fvankrieken, line 94: this seemed better than just defining it over and over again, and this is the simplest way to define a re-usable variable within the action as far as I can tell (`env` vars are not accessible when calling other workflows. So couldn't do ${{ env.tag }})

.github/workflows/build.yml
```diff
@@ -45,6 +45,11 @@ on:
         type: boolean
         required: true
         default: false
+      dev_image:
+        description: "(Attempt to) use a dev image specific to this branch?"
```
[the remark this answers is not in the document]
fvankrieken: Yeah that's a bit cleaner

.github/workflows/build.yml
damonmcc: and in the longer-term, the repeated `image_tag: ${{ needs.health_check.outputs.tag }}` lines would be avoided by not having product-specific workflows at all (e.g. `build.yml` calls `dcpy.builds` scripts with some inputs like product name)

.github/workflows/stale_builds.yml
```diff
@@ -17,8 +17,12 @@ jobs:
...
+      DOCKER_USER: ${{ secrets.DOCKER_USER }}
     steps:
       - uses: actions/checkout@v3
-      - name: Delete build artifacts
-        run: |
-          python3 -m dcpy.builds.clean_artifacts
+      #- name: Delete build artifacts
+      #  run: python3 -m dcpy.builds.clean_artifacts schemas
+      - name: Delete docker images
+        run: python3 -m dcpy.builds.clean_artifacts dockerhub_tags
```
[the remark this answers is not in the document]
fvankrieken: Yup - commit 4 is going to be dropped before merging

.github/workflows/python_compile_requirements.yml
damonmcc: (think this is what you mentioned in standup today) happy to save this for a follow-up

fvankrieken: there's probably just a better way to organize these workflows. Something like
- `build` workflow call can take "all" as an input (currently in nightly build it's more duplicating logic in the build workflow)
- docker workflow publishing via workflow call ("test" workflow which builds images doesn't call the docker publish workflow)
- test workflow with just tests (and no docker image) via workflow call

Which then get called by
- nightly build (build all, run tests)
- pr tests (build docker images if needed, run tests)
- bump python packages (build docker images, build all products, run tests)

damonmcc (approved):

damonmcc (approved):

PR #404, opened 2023-12-01
cleanup setting of image tag when no changes present

fvankrieken:
fix bug in implementation in #390 

squasher

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,11 +82,25 @@ jobs:
           echo "matrix_raw=$matrix" >> $GITHUB_OUTPUT
           echo "matrix={\"include\":$(echo $matrix)}" >> $GITHUB_OUTPUT
 
-  build_docker_images:
+  set_docker_image_tag:
+    name: Set docker image tag
     needs: check_changes
     runs-on: ubuntu-22.04
     outputs:
-      tag: dev-${{ github.head_ref || 'latest' }}
+      tag: ${{ steps.tag.outputs.tag }}
+    steps:
+      - name: set image tag
+        id: tag
+        run: |
+          tag="${{ contains(needs.check_changes.outputs.path_filters, 'docker') && format('dev-{0}', github.head_ref || github.ref_name) || 'latest'}}"
+          echo "tag=$tag" >> $GITHUB_OUTPUT
+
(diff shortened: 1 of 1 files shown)


fvankrieken:
Running one last test before finalizing

fvankrieken:
Should be good to go

.github/workflows/test.yml
```diff
@@ -98,17 +112,20 @@ jobs:
           - dev
     steps:
       - uses: actions/checkout@v4
+        if: contains(needs.check_changes.outputs.path_filters, 'docker')
       - name: publish images
         if: contains(needs.check_changes.outputs.path_filters, 'docker')
```
damonmcc, line 117: duplicated lines?

.github/workflows/test.yml
```diff
@@ -98,17 +112,20 @@ jobs:
...
         working-directory: docker
-        run: ./publish.sh "${{ matrix.image }}" "dev-${{ github.head_ref || 'latest' }}"
+        run: ./publish.sh "${{ matrix.image }}" "${{ needs.set_docker_image_tag.outputs.tag }}"
 
   check_formatting:
-    needs: build_docker_images
+    needs: 
+      - set_docker_image_tag
+      - build_docker_images
```
damonmcc, line 124: do jobs like this and below need more than the single job `build_docker_images ` because it might be skipped? 

since it all seems linear and looks like only the steps can be skipped, might be nice if all the `needs:` had a single job listed?

.github/workflows/test.yml
fvankrieken: No, just applying the conditional to the checkout too. Just trying to make it faster if we don't need to build the images.

Can't skip the whole job though because there's not really a way to have other jobs
1. run fine if this job is skipped, AND
2. wait for this job to run if it's not skipped.

As far as I can tell

.github/workflows/test.yml
fvankrieken: To access outputs of any job, they have to be in the "needs"

fvankrieken: What we could do is pass along the outputs though. So outputs of build_docker_images has the outputs of the prior two jobs

PR #413, opened 2023-12-07
Fix Pluto condo flip-flopping, repeat specific builds

fvankrieken:
Think we're good to go.

### Checks

[Passing normal builds](https://github.com/NYCPlanning/data-engineering/actions/runs/7280854986) (with colp failing due to colp-specific bash bug)

[Fixed colp regular build](https://github.com/NYCPlanning/data-engineering/actions/runs/7281693661)

[Successful repeated build](https://github.com/NYCPlanning/data-engineering/actions/runs/7281783483/job/19842970353)

### Original PR description

Have gone in circles on cli/usage and would love some input.

Pretty simple PR really. Commit 1 is the fix for flipping bbls. Commit 2 moves some code around, gets the "plan/recipe" part of builds out of `connectors` and into `builds`.

Commit 3 is the real changes. Adds functionality to
1. provide version via cli input rather than relying on it being checked in
2. add build metadata to outputs (need to include something now to actually also have it included in uploads. It's generated at beginning of build, so need to manually move it into output folder at some point. Not sure we're 100% consistent about output folder syntax so just need to go around and clean that up.
3. provide functionality to repeat a specific build. Slightly different cases - if the former build has `build_metadata.json`, can just use the locked recipe info from there. If not, it takes a "template" recipe and uses the source data versions from a past recipe (but getting file format, preprocessor, data product-level metadata from "template" recipe). If it can't find either, it doesn't attempt to recreate the build.

Main issue is clunkiness around using 3. Logically, this isn't really a new "recipe" - we don't want a "repeat.yml" in each data product, that seems highly redundant. So for now it's a flag that's fed into the CLI. But this also seems a little clunky, especially with displaying this in GHA workflow dispatch - have a box to repeat a specific build, have a box to specify the version of the "new" build being requested, a recipe file path that could be ignored if it's a repeated build... All just a bit messy. One option could be having it be a reserved recipe name? Instead of a separate flag, when "repeat" is fed in as recipe have special logic handled by the cli which interprets that as a command to repeat a build. Keeps the cli/gha interaction just a tad cleaner.

### Additional commits
4 - add some exports in bash
5 - clean up build ghas


dcpy/builds/metadata.py
```diff
@@ -10,3 +29,22 @@ def build_name(name: str | None = None) -> str:
...
+        commit = None
+    if os.environ.get("CI"):
+        run_url = git.action_url()
+    else:
+        run_url = None
+    build_metadata = BuildMetadata(
+        timestamp=timestamp, commit=commit, run_url=run_url, recipe=recipe
+    )
+    with open(output_file, "w", encoding="utf-8") as f:
+        logger.info(f"Writing recipe lockfile to {str(output_file.absolute())}")
```
alexrichey, line 49: copy-pasta 🍝?

dcpy/builds/plan.py
```diff
@@ -179,11 +186,63 @@ def recipe_from_yaml(path: Path) -> Recipe:
...
 
-def plan(recipe_file: Path) -> Path:
-    logger.info("Planning recipe")
-    lock_file = recipe_file.parent / f"{recipe_file.stem}.lock.yml"
+def repeat_recipe_from_source_data_versions(
+    version: str,
+    source_data_versions: pd.DataFrame,
+    template_recipe_path: Path | None = None,
+) -> Recipe:
+    template_recipe_path_ = template_recipe_path or Path("recipe.yml")
```
alexrichey, line 194: Any reason to have `template_recipe_path` be nullable? Thinking Path("recipe.yml") might cause problems depending on where this is invoked.

dcpy/utils/s3.py
```diff
@@ -89,33 +119,26 @@ def list_objects(bucket: str, prefix: str) -> list:
...
-            Key=key,
-            Filename=str(path),
-            Callback=lambda bytes: progress.update(task, advance=bytes),
-        )
+def exists(bucket: str, key: str) -> bool:
+    """Returns true if an object with given bucket and key exists"""
+    try:
+        client().head_object(Bucket=bucket, Key=key)
+        return True
+    except ClientError:
```
alexrichey, line 127: Curious which cases would lead to a ClientError? Do you happen to recall?

alexrichey:
@fvankrieken This all makes sense! I like that the edm connectors now answers more "where" style questions, and `build` handles the "how". I'm still not entirely clear on what belongs where, but this seems totally sensible.

dcpy/utils/s3.py
fvankrieken: If the object doesn't exist!

fvankrieken: Can see if this is a bit too permissive - did some searching and this seemed to be many's recommended solution

fvankrieken:
> @fvankrieken This all makes sense! I like that the edm connectors now answers more "where" style questions, and `build` handles the "how". I'm still not entirely clear on what belongs where, but this seems totally sensible.

Any thoughts on the cli/gha interactivity? Making this not too clunky from a users pov?

dcpy/builds/metadata.py
fvankrieken: Oop

dcpy/builds/plan.py
fvankrieken: Yeah, main reason was that this is that currently, the calls into the cli work down to here. Meaning the `recipe_file` arg makes its way down, so in the case that repeat is specified and no "template" provided, revert to the default.

I think though to your point, two things should happen
1 . use the DEFAULT_RECIPE constant that also feeds into other things
2. have a specific helpful message in case that this isn't found

By our conventions now there should generally be a "recipe.yml" where the code is running, but yeah don't want to turn that assumption into unclear failures

dcpy/builds/plan.py
```diff
@@ -179,11 +186,63 @@ def recipe_from_yaml(path: Path) -> Recipe:
...
+        if recipe_file
+        else Path("recipe.lock.yml")
+    )
+    if recipe_file and not repeat:
+        logger.info("Planning recipe")
+        recipe = plan_recipe(recipe_file, version)
+    else:
+        if version is None:
+            raise Exception("Version must be supplied if repeating a specific build")
+        product_key = publishing.PublishKey("db-pluto", version)
```
damonmcc, line 231: something like `logger.info("Repeating recipe from version {version}")` would be cool, maybe after the exception isn't raied?

damonmcc:
> > @fvankrieken This all makes sense! I like that the edm connectors now answers more "where" style questions, and `build` handles the "how". I'm still not entirely clear on what belongs where, but this seems totally sensible.
> 
> Any thoughts on the cli/gha interactivity? Making this not too clunky from a users pov?

the cli/gha here seems like a good way to handle re-running a past recipe for now. feels like most clunky/messing feelings come from us still having to write logic in gha workflows that can't yet be as simple as "run something with these arguments"

dcpy/builds/plan.py
fvankrieken: Also - other unimplemented thing right now - in these cases, we need to provide product key. See hardcoded db-pluto in this line.

Already an input to the gha. Though here we need the mapping of "product name" to "s3 folder" ("pluto" to "db-pluto" etc. Recipe files include that, so maybe the cleanest way to handle this require ALL repeats to have a base recipe, just to at least get metadata like product name from as a starting point

fvankrieken: That'll also resolve @alexrichey 's comment on nullable template recipe for generating build_metadata

fvankrieken:
Do we want to keep generating source data versions csv for backwards compatibility? I would lean with no. Either way I also need to update publishing util to look for build_metadata, then source data versions

damonmcc:
> Do we want to keep generating source data versions csv for backwards compatibility? I would lean with no. Either way I also need to update publishing util to look for build_metadata json, then source data versions csv

my reflex is that we should keep generating them for now. I don't think we wanna plan on programmatically reading them when we have that same info in a build_metadata json. but I wonder if that csv is an expected attachment to any datasets we distribute?

fvankrieken:
> > Do we want to keep generating source data versions csv for backwards compatibility? I would lean with no. Either way I also need to update publishing util to look for build_metadata json, then source data versions csv
> 
> my reflex is that we should keep generating them for now. I don't think we wanna plan on programmatically reading them when we have that same info in a build_metadata json. but I wonder if that csv is an expected attachment to any datasets we distribute?

That's a good point, I think at least some users (GIS, AD) might like seeing them. Quicker to glance at than a json file

dcpy/builds/plan.py
fvankrieken: Fixed in latest commits

fvankrieken:
Think I'm largely done here - going to run some tests, chiefly
- testing all builds since I've fiddled with gha machinery a bit
- testing a repeat build of a specific published pluto version

products/cpdb/bash/05_export.sh
```diff
@@ -18,6 +18,8 @@ mkdir -p output && (
     csv_export geospatial_check &
     csv_export source_data_versions &
```
fvankrieken, line 19: Do we care about consistency between

```bash
csv_export source_data_versions
```

vs

```bash
cp ../source_data_versions.csv ./
```

Both are perfectly valid in the current setup - `load.py` creates the table, then loads it into the build engine

fvankrieken:
@sf-dcp - this doesn't quite resolve the "previous" issue you've talked about in pluto. Really the recipe logic right now in general works in two cases
- a "new" build (assumes previous is latest)
- recreating a specific build, keeping all data source versions constant.

There's no concept of "redoing" a specific version with updated sources. I.e. building pluto 23v3 with latest available sources, since 23v3 already exists. It'll be a valid new build, but the version will be "wrong" in some sense.

Gonna keep working on logic around versions, finding previous version, etc in #434 after this is merged. Happy to chat more about what I'm thinking there as well if you're interested

products/cpdb/bash/05_export.sh
damonmcc: personally prefer `csv_export`. feels like that keeps it explicit that `source_data_versions` is one of a many exports and if we ever want them all to go to a new directory we'd only have to change that in one place

sf-dcp:
> @sf-dcp - this doesn't quite resolve the "previous" issue you've talked about in pluto. Really the recipe logic right now in general works in two cases
> 
> * a "new" build (assumes previous is latest)
> * recreating a specific build, keeping all data source versions constant.
> 
> There's no concept of "redoing" a specific version with updated sources. I.e. building pluto 23v3 with latest available sources, since 23v3 already exists. It'll be a valid new build, but the version will be "wrong" in some sense.
> 
> Gonna keep working on logic around versions, finding previous version, etc in #434 after this is merged. Happy to chat more about what I'm thinking there as well if you're interested

Okay, will keep in mind!

damonmcc (dismissed):

damonmcc:
kicked off a build of all datasets with this branch [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7290578596)

fvankrieken:
> kicked off a build of all datasets with this branch [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7290578596)

I cancelled it - that would have overwritten outputs that GIS is currently qcing

damonmcc:
> > kicked off a build of all datasets with this branch [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7290578596)
> 
> I cancelled it - that would have overwritten outputs that GIS is currently qcing

sorry should have checked before I started it, couldn't tell if the latest commits had been tested

would it have overwritten them with different data?

fvankrieken:
Yup. Build all currently builds a pluto major, they're reviewing pluto minor from this branch.

I'll branch off of this branch and run all builds if you'd like to see them run in one action

fvankrieken:
Build running [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7290950344)

fvankrieken:
Link to build metadata from latest run for colp

https://nyc3.digitaloceanspaces.com/edm-publishing/db-colp/draft/fvk-pluto-deterministic-builds-2/build_metadata.json

damonmcc (approved):

PR #417, opened 2023-12-11
reorganize qa app code

fvankrieken:
Just wanted to take a quick stab at a more helpful folder structure. Easiest to look at the repo instead of changes

https://github.com/NYCPlanning/data-engineering/tree/main/apps/qa

Changes only really happen within the src folder. In it, we have
- index.py
- pages
   - one folder per page, with code both for streamlit components, more pure python helpers, and the pages themselves within. Any code within any page folder is truly specific to that page (well, mostly)
 - shared (open to other name: code that's shared across different product pages)
   - components: all functions that render streamlit views, and in some cases return values as well (selectboxes, etc)
     - file names by categorization of components
   - utils: python utilities

Certainly not the only way we could do this, but the current setup had more ambiguity

TODO

- [ ] rebase main and resolve issues once @damonmcc 's templatedb pr is merged
- [ ] rename `source_report_utils.py` since it's in a utils folder now


sf-dcp:
1. Is the goal of this issue to make the current structure more clear or change the app to an ideal state? 
One idea is to have each page as an instance of a QA page class and modify as necessary. It seems that some code is repeated across pages: for example, the logic around validating presence of `ProductKey`.

2. Use of `dcpy`/`data-library`: let's say there is a util function we want to use in a QA page, i.e. DO link. One way is to import that function directly to the QA page. From my previous PR, it seems that the preferred way is to 1) create a component and 2) import that component to the QA page. Is this logic true for all cases or only some?

fvankrieken:
> 1. Is the goal of this PR to make the current structure more clear or to change the app to an ideal state?
>    One idea is to have each page as an instance of a QA page class and modify as necessary. It seems that some code is repeated across pages: for example, the logic around validating presence of `ProductKey`.
> 2. Use of `dcpy`/`data-library`: let's say there is a util function we want to use in a QA page, i.e. DO link. One way is to import that function directly to the QA page. From my previous PR, it seems that the preferred way is to 1) create a component and 2) import that component to the QA page. Is this logic true for all cases or only some?

1. bit of both, but more the former I suppose - nudge it towards a more ideal state given how the code is written at this moment
2. I think it depends. I can think of a couple cases where I personally would lean towards different things
  - a dcpy util is needed. It is a one line call that gets fed to something else on the CPDB page. This is specific to CPDB, so would keep it in the CPDB folder, not `shared`. It doesn't clutter or reduce the readability of the CPDB page, so I don't put it in something like `src/pages/cpdb/helpers.py` or `src/pages/cpdb/components.py`
  - similar case, specific to CPDB, but it's a 15 line streamlit component. I would keep it off of the cpdb page for readability and then put it in `src/pages/cpdb/components.py`
  - any component/utility is re-used by many pages -> goes into `shared`

sf-dcp:
@fvankrieken, that makes sense. Lastly, should we move `index.py` to `/pages` dir since it's a page too?

alexrichey (approved):
Cool - seems sensible

fvankrieken:
> @fvankrieken, that makes sense. Lastly, should we move `index.py` to `/pages` dir since it's a page too?

Not the way its set up now - index is really the entry point (or really, the one true page), and there's actually a "home" page within the pages directory.

damonmcc (approved):
no worries if this gets merged before the other QA app pr

sf-dcp (dismissed):

PR #419, opened 2023-12-14
QAQC App: Add list of new & vanished BBLs for PLUTO

### LINKED ISSUE #335: PLUTO QAQC enhancement - output list of all new lots
As part of the PLUTO QAQC process, a request of the GIS team is to output a list of all new lots (lots that are in the new version and not in the previous version).  Having this list readily available will greatly aid in their QAQC process.

sf-dcp:
Fixes #335. The PR consists of 2 commits:
1. Generates a new QAQC table with new and disappeared BBLs during a PLUTO build. 
2. Adds the generated table to the QAQC page. 
    * I re-deployed the qaqc [app](https://de-qaqc.nycplanningdigital.com/?page=PLUTO) on my branch. 
    * Go to PLUTO -> `draft` version -> `sf-qaqc-app-pluto-lots` build and scroll all the way down to see the table.
    
      image

    * If choosing a build where the table is not available (earlier versions), here is the view: 
    
      image
 
## TO DO:
Confirm correctness of the query:
- [x] after Finn merges #413, re-run pluto minor and make sure there are zero records in the table. <-- done
- [x] for pluto major build (23v3), imported `previous_pluto` recipe is 23v3. However, 23v2 or 23v2.1 are expected. <-- #413 addressed this concern by adding an option to **repeat** existing build.

### DIFF
--- a/products/pluto/pluto_build/sql/qaqc_bbl_diffs.sql
+++ b/products/pluto/pluto_build/sql/qaqc_bbl_diffs.sql
@@ -1,0 +1,30 @@
+DROP TABLE IF EXISTS qaqc_bbl_diffs;
+CREATE TABLE qaqc_bbl_diffs (
+    source_version text,
+    bbl text,
+    type text
+);
+
+-- add new bbl records
+INSERT INTO qaqc_bbl_diffs (
+    SELECT
+        :'VERSION' AS source_version,
+        ap.bbl::float::bigint::text AS bbl,
+        'new' AS type
+    FROM archive_pluto AS ap
+    LEFT JOIN previous_pluto AS pp
+        ON ap.bbl::float::bigint = pp.bbl::float::bigint
+    WHERE pp.bbl IS NULL
+);
+
+-- add vanished bbl records
+INSERT INTO qaqc_bbl_diffs (
+    SELECT
+        :'VERSION_PREV' AS source_version,
+        pp.bbl::float::bigint::text AS bbl,
+        'vanished' AS type
+    FROM previous_pluto AS pp
+    LEFT JOIN archive_pluto AS ap
+        ON ap.bbl::float::bigint = pp.bbl::float::bigint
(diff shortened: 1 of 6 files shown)


products/pluto/pluto_build/sql/qaqc_bbl_diffs.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS qaqc_bbl_diffs;
+CREATE TABLE qaqc_bbl_diffs (
+    source_version text,
+    bbl bigint,
```
alexrichey, line 4: It looks like we have `bbl` in a few other places. I'm curious why... is it just to remove leading zeros? cc @fvankrieken @damonmcc

damonmcc: could be, or part of a battle against floats. I think it's an inherited choice rather than one that we made lol

seems like BBLs should be considered strings to me since they're IDs and we'd never want to treat them like numbers

alexrichey (dismissed):
Looks good to me!

products/pluto/pluto_build/sql/qaqc_bbl_diffs.sql
sf-dcp: @damonmcc, I agree. I will update `bbl` type the `qaqc_bbl_diffs` table. However, my understanding is that we will revise `bbl` type to be of text type in all other generated tables during pluto build separately, not in this PR. Lmk if that's correct.

damonmcc (dismissed):

damonmcc (approved):

sf-dcp:
Note to self: when joining tables on `bbl`, need to cast `bbl` to float type first in order to cast to bigint; otherwise, it returns an error. The inserted records has `bbl` casted to text data type.

PR #421, opened 2023-12-15
run tests on push to main

damonmcc:
the dcpy pytest part of our Nightly QA [failed](https://github.com/NYCPlanning/data-engineering/actions/runs/7218198722/job/19667240605#step:6:60) on main last night. seems to be due to the fact that our docker images used in those tests aren't rebuilt & published when we merge chaneges to main

this change should ensure that Nightly QA uses up-to-date docker images

> [!NOTE]
> These changes will not cause product builds to run on merges to main.
> This note format is a new github markdown feature [[docs here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)].

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -1,5 +1,8 @@
 name: Docker Images - Build and Publish
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       build-base:
@@ -30,7 +33,7 @@ env:
 jobs:
   build-base:
     runs-on: ubuntu-22.04
-    if: inputs.build-base
+    if: inputs.build-base || github.event_name == 'push'
     defaults:
       run:
         shell: bash
@@ -43,7 +46,7 @@ jobs:
 
   build-geosupport:
     runs-on: ubuntu-22.04
-    if: inputs.build-geosupport
+    if: inputs.build-geosupport || github.event_name == 'push'
     defaults:
       run:
         shell: bash
@@ -56,7 +59,7 @@ jobs:
 
   dev:
     runs-on: ubuntu-22.04
-    if: inputs.dev
+    if: inputs.dev || github.event_name == 'push'
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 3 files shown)


fvankrieken:
Since the desired behavior is to build and publish docker images on push to main, should we just do that without the extra tests? So push trigger for the build docker image job for main rather than this job

damonmcc:
> Since the desired behavior is to build and publish docker images on push to main, should we just do that without the extra tests? So push trigger for the build docker image job for main rather than this job

I guess the desired behavior is really to run tests on main that reflect the code on main, we just happen to only run them at midnight at the moment. seems like we'd only benefit from running them along with the image push like we do on dev branches

fvankrieken:
That's fine. I think ideally, we don't need to test just after a push to main, since we tested before merging. Any weirdness that sneaks through should be caught by nightly qa. 

It won't work quite as written though. You'll need to 
- make docker build step run on pushes to main (potentially only if there are relevant changes)
- add logic to push to "latest" image tag rather than "dev-main" (convention used for dev images)
The workflow and logic baked into the yml file are already a tad convoluted, so it also might just be simpler to leave this workflow as is. since it's already doing a lot without being our method of publishing production dev images

damonmcc:
ah gotcha, I didn't think about dev-main vs latest

damonmcc:
and now some apparent complexity to adding the push trigger to docker_publish.yml: how to set and use the 4 boolean inputs when the trigger is push?

fvankrieken:
Maybe for each of the jobs
`if: inputs.build-base || github.event_name == 'push'`
?

fvankrieken (approved):

sf-dcp:
> Note
> 
> These changes will not cause product builds to run on merges to main. This note format is a new github markdown feature [[docs here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)].

@damonmcc, stealing the note format!

damonmcc:
may want to follow-up to this with a file filter to prevent rebuilding docker images on _all_ commits to `main`

PR #422, opened 2023-12-15
fix qa app python packages

damonmcc:
although tests on main which use `leafmap` are passing, the QA app can't find the package (see screenshot)

this change fixes that by adding `leafmap` to the list of packages we want the QA to use. that list is a subset of our primary list in [`python/requirements.in`](https://github.com/NYCPlanning/data-engineering/blob/main/python/requirements.in)

tested by delplying QA app from this branch and it [seems to have fixed it](https://de-qaqc.nycplanningdigital.com/?page=Template+DB)

image

### DIFF
--- a/apps/qa/requirements.txt
+++ b/apps/qa/requirements.txt
@@ -1,5 +1,6 @@
 folium
 geoalchemy2
+leafmap
 mapclassify
 matplotlib
 numerize


fvankrieken (approved):

damonmcc:
app deployed after merge [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7225399655)

PR #425, opened 2023-12-18
delete existing files when uploading folder to s3

fvankrieken:
A little late, but resolution to our most infamous bug. I was going to include this in a different branch but figured let's just get this merged.

Would we prefer any sort of safety measures? Or actually checking if a folder exists? Or just this - clean out the prefix in s3 before uploading anything

### DIFF
--- a/dcpy/utils/s3.py
+++ b/dcpy/utils/s3.py
@@ -18,6 +18,7 @@ from rich.progress import (
     TextColumn,
     TimeRemainingColumn,
 )
+from dcpy.utils.logging import logger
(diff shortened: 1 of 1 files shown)


damonmcc (approved):
lgtm! on safety measures, if `client_.delete_object` fails when a folder doesn't exist, then we should check before calling `delete()`

damonmcc:
noting that we used `delete()` locally and it doesn't throw an error or warning when the folder doesn't exist, so no extra safety measures needed

Issue #313, opened 2023-10-23
Implement Remaining Recipes Files


alexrichey:
Scope: Implement recipes files for remaining products (ie all except for Pluto and FacilitiesDB). 

This should be very straightforward. Facilities and Pluto (minor) were the odd cases. For the most part, it should be a matter of copying the requirements out of shell scripts. However, you may also need to check the QA steps, as these often pull in a previous version of the product from recipes to compare against. (these should also go in the recipes file)

I looked through the various products, and they looked straightforward. But we should scope this out, ie add checklists for the products to modify, and investigate any potential oddities. 

cc @fvankrieken from our retro conversation about lagging infra changes, e.g. recipes.

alexrichey:
- [ ] cbbr
- [ ] checkbook
- [x] colp
- [ ] cpdb
- [ ] db-zap-opendata
- [ ] developments
- [x] facilities
- [ ] factfinder
- [ ] knownprojects
- [x] pluto
- [x] template
- [ ] zap-opendata
- [ ] zoningtaxlots

fvankrieken:
I'm working on cpdb as part of fisa enhancements

damonmcc:
I'll do KPDB soon, should be the last one!

PR #434, opened 2023-12-19
Create devdb recipe

fvankrieken:
#313 

Couple main changes

- create devdb recipe - relatively straightforward
- dcpy.utils.versions - different version types as data classes, with logic around parsing, bumping, labels, etc. Definition of "VersionStrategy" as well, which is currently either a "SimpleVersionStrategy" (bump_latest_release or first_of_month at the moment) or some other more complex structure (right now BumpLatestRelease which has a value of int for how much to bump).
- dcpy.build.plan - logic around new version strategy stuff
- dcpy.connectors.edm.publishing - new previous function to lookup previous version

A couple potential big outstanding questions. 
- One is the long comment below. 
- Another is the handling of strings vs Version objects. I go back and forth a bit, but in general outside of versions leave them as strings. Seems a little easier in some ways, and just use the objects for logic like bumping. But maybe we should have the objects themselves in the recipe objects? Probably. Presumably can override how they get serialized (override `__str__` or `__repr__`?)


dcpy/builds/metadata.py
```diff
@@ -19,7 +19,7 @@ class BuildMetadata(BaseModel, extra="forbid"):
 
     def __init__(self, **data):
         if "version" not in data:
-            recipe: Recipe = data["recipe"]
+            recipe = data["recipe"]
```
fvankrieken, line 22: Mypy gave a little warning about this, so just removed it

dcpy/builds/plan.py
```diff
@@ -96,19 +88,24 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
...
+            case None:
+                raise Exception("No version provided")
+            case versions.SimpleVersionStrategy.first_of_month:
+                recipe.version = versions.FirstOfMonth.generate().label
+            case versions.SimpleVersionStrategy.bump_latest_release:
+                recipe.version = versions.bump(
+                    previous_version=publishing.get_latest_version(recipe.product),
+                    bump_type=recipe.version_type,
+                )
+            case versions.BumpLatestRelease() as bump:
```
fvankrieken, line 102: These two different "bump latest release"s are a bit clunky. But basically the desired behavior for me was having both of the following be valid
```yml
version_strategy: bump_latest_release
version_strategy:
  bump_latest_release: 2
```

Since DevDB is labeled with quarters, but we always bump by two quarters.

Maybe this is a slightly silly approach in general. This versioning scheme is maybe more dataset-level than recipe-level, and DevDB really should be "biannual" rather than "every two quarters". We could also specify that it prints out in "quarter" format.

Rationale for current implementation is that it's nice to just parse a version string without additional metadata, from a code perspective. 22Q4 is unambiguously a quarter, and it's nice to not have to parse that to "biannual". However, we're never going to be parsing a version on its own, but rather always for a specific data product. Meaning DevDB 22Q4 is pretty impossible to miss as the second release for DevDB of 2022, and we don't necessarily need to specify `bump_latest_release: 2` to know that the next DevDB release will be 23Q2.

I think I've talked myself into reimplementing this. @damonmcc and @sf-dcp would appreciate some input. Might be a bit clearer if you look down at dcpy.utils.versions as well

fvankrieken: After thinking about it, sort of thinking about leaving as is as a first pass. Need to write a test for `previous` function in publishing. I think we retool as we start to expand dataset-level metadata

fvankrieken: @alexrichey

dcpy/test/utils/test_versions.py
```diff
@@ -17,12 +19,19 @@ def test_parsing_invalid_version(self):
...
-            ["minor", "23v2.1", "23v2.2"],
+        for bumped_part, bump_by, v, v_expected in [
+            ["major", None, "23v2", "23v3"],
+            ["major", None, "23v2.1", "23v3"],
+            ["minor", None, "23v2", "23v2.1"],
+            ["minor", None, "23v2.1", "23v2.2"],
+            [None, 1, "23Q1", "23Q2"],
+            [None, 2, "23Q4", "24Q2"],
+            [None, 7, "23Q2", "25Q1"],
+            [None, 7, "2023-01-01", "2023-08-01"],
```
alexrichey, line 34: I doubt it'd cause any problems, but this behavior feels a little ambiguous. (would naively expect a 7 day increment, but that's not something we ever do, so...)

dcpy/builds/plan.py
alexrichey: yeah, I've also gone back and forth on all of that. Another thought: we could just nix the `version_strategy` in the yml, and put it under the version key. So either of these options would all be valid:
```
version: 2023v1
```

```
version: {bump_latest_release: 2}
```

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -47,3 +50,46 @@ def test_publish(create_bucket, create_temp_filesystem, mock_data_constants):
...
+            )
+
+
+@pytest.mark.parametrize(
+    "product, version, expected",
+    [
+        ("test_product_1", "23v3.1", "23v3"),
+        ("test_product_1", "23v3", "23v2.2"),
+        ("test_product_1", "24v3", "23v3.1"),
+        pytest.param("test_product_1", "23v2", None, marks=pytest.mark.xfail),
```
alexrichey, line 75: neat, hadn't used `pytest.mark.xfail` before

dcpy/utils/versions.py
```diff
@@ -1,41 +1,165 @@
...
+
+    @abstractmethod
+    def bump(
+        self, version_subtype: VersionSubType | None = None, bump_by: int | None = None
+    ) -> Version:
+        raise NotImplementedError("Version is an abstract class")
+
+
+@dataclass(order=True)
+class MajorMinor(Version):
```
alexrichey, line 43: if you wanted to be really fancy, you could also implement a `def __lt__` so that these guys sort correctly in the case of big minor versions, ie so that 23v11 is greater that 23v2. Cherry on top!

alexrichey (approved):
Looks good!

dcpy/utils/versions.py
```diff
@@ -1,41 +1,165 @@
...
-        + parsed.get("major", "")
-        + ("." + parsed["minor"] if "minor" in parsed else "")
-    )
-
-
-def first_of_month() -> str:
-    """Return a string representing the first of the current month."""
-    return date.today().strftime("%Y-%m-01")
+@dataclass
+class RegexSpmMatch:
```
damonmcc, line 11: whats Spm short for?

dcpy/connectors/edm/publishing.py
```diff
@@ -80,6 +79,36 @@ def get_draft_builds(product: str) -> list[str]:
     return sorted(s3.get_subfolders(BUCKET, f"{product}/draft/"), reverse=True)
 
 
+def previous(product: str, version: str | versions.Version):
```
damonmcc, line 82: a return type hint would be nice here

damonmcc (approved):
cool stuff! always love to see `pytest.mark.parametrize` lol

dcpy/utils/versions.py
alexrichey: this is a schmancy Simple Pattern Match class

dcpy/test/utils/test_versions.py
fvankrieken: Yeah that's fair.

For now the only dataset that follows this convention is ztl, which is monthly. Actual dated releases are "20230101". Still not quite ideal - I think we change ztl moving forward to drop the "-01"

dcpy/utils/versions.py
fvankrieken: It actually works as implemented - dataclasses sort by their types, not serialization. And `order=True` means that gt/lt logic goes by the defined order of fields - in this case: year, major, minor

dcpy/utils/versions.py
fvankrieken: Close - ripped off from [here](https://github.com/aronhoff/regex_spm), it's Structural Pattern Matching

dcpy/utils/versions.py
alexrichey: wow! That's awesome

damonmcc:
@fvankrieken 
looks like this PR dropped `developments/bash/01_dataloading.sh` but didn't add a call like `python3 -m dcpy.builds.load`, so DevDB is failing on main [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7393675363/job/20113963933). I can fix it while adding a KPDB recipe.yml?

fvankrieken:
Oops. I would prefer to handle it in my devdb branch since I'm going to be touching a lot of those files. If it'll be ready to be merged tomorrow that's fine though, tomorrow is a crazy meeting day so I might not get to much

PR #437, opened 2023-12-21
try using a generic issue template to track data updates

damonmcc:
## background

we don't currently have a standardized way to track data updates from end-to-end. by track I primarily mean the ability for anyone in GDE to wonder "what's the status of that update?" and find an up-to-date answer

we currently have issue templates for each product, but the bulk of what makes them product-specific are lists and details on source data. those details may be more valuable as documentation stored in other forms (readmes, wiki)

## approach

beyond using a single issue template to track updates, [an issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) seems like an interesting way to create a template and get structured user input when creating the new issue [here](https://github.com/NYCPlanning/data-engineering/issues/new/choose) (how much use input is really needed/useful though if it's just a checklist?)


damonmcc:
gonna merge this since it seems it won't have an effect until the changes are in `main` and this shouldn't disrupt anything

PR #444, opened 2023-12-26
add packaging for data products

### LINKED ISSUE #351: GIS Automations: COLP Data Processing
Automate the transformations that GIS applies to COLP, as outlined [here](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BFE79C434-743F-4847-B56B-D1603963C243%7D&file=Open%20Data%20Update%20Cycle%20Instructions.docx&action=default&mobileredirect=true)

Macro Steps:

- [ ] Pull from publishing folder on Digital Ocean
- [ ] Apply transformations
- [ ] Output to new bucket

Individual Transformations to call out:

- [ ]      (Item 6) "Confirm/Review data schema is the same as previous version 
             No missing geographic values (i.e. Borough, CD, etc) 
             No blank values in feature class, should be “null” instead "

- [ ] "XLSX: Convert csv to xlsx. IMPORTANT:This must be done carefully to avoid corrupting data that must retain leading zeros" 
- [ ] Generate README pdf from [sharepoint](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7B6D29AD78-E128-456A-8D21-99811CCA51BD%7D&file=colp_readme.docx&action=default&mobileredirect=true). Should this reside in (perhaps) markdown in our github repo?
- [ ] (Item 10) "update GDB and SHP metadata, Import metadata from previous version to GDB, Update citation publication date to last day of month (i.e. 6/30/2022) 
- [ ]

damonmcc:
related to #351 

This is a first pass at adding a data update "packaging" feature to `dcpy` for DE to transform files that have passed QA and should be distributed by GIS or by further automation (e.g. push to Socrata).

See related issue [here](https://github.com/NYCPlanning/data-engineering/issues/351#issuecomment-1869740199) for thoughts on the stages of DE product releases.

Although COLP is the intended first test-case, it seemed useful to seek review on this approach before iterating on it and implementing the relatively complex COLP packaging logic. Instead, new tests and Template DB were used to test these changes.

## changes
- refresh relevant diagrams to reflect new concepts of "package" and "distribute" for builds
- make testing of `dcpy.utils` and `dcpy.connectors` more flexible
- add `connectors.edm.packaging` with COLP and TDB as the first examples

## tests
- added `test_packaging.py` and expanded the `mock_data_constants` in `conftest.py` to be used in new tests
- Template DB runs on this branch: [Build](https://github.com/NYCPlanning/data-engineering/actions/runs/7343560522), [Publish](https://github.com/NYCPlanning/data-engineering/actions/runs/7343602062), [Package](https://github.com/NYCPlanning/data-engineering/actions/runs/7351162171)
- COLP packaging failing as expected for two different reasons: requested version not found in publish folder ([run here](https://github.com/NYCPlanning/data-engineering/actions/runs/7341296371/job/19988743880#step:6:75)) and packaging logic is not implemented ([run here](https://github.com/NYCPlanning/data-engineering/actions/runs/7341310115/job/19988780348#step:6:100))

## notes
- new things here like the `PackageMetadata` dataclass are not meant to be permanent approaches to or permanent sources of truth of product metadata. that longer-term work has just started and this is limited to "packaging"
- screenshot of packaged Template DB files in S3 reflecting logic in `packaging.templatedb()` function
  - image


dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+from typing import Callable
+from dataclasses import dataclass
+import typer
+
+from dcpy.utils import s3
+from dcpy.utils.logging import logger
+from dcpy.connectors.edm import publishing
+
+BUCKET = "edm-distributions"
+BUCKET_ACL = s3.string_as_acl("public-read")
```
fvankrieken, line 13: I think you can do
```python
BUCKET_ACL: s3.ACL = "public-read"
```

To not need to pass it to a function

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+
+DOWNLOAD_ROOT_PATH = Path(".library")
+OUTPUT_ROOT_PATH = Path(".output")
+
+
+@dataclass
+class PackageKey(publishing.ProductKey):
```
fvankrieken, line 20: Is this distinct from a PublishKey in any way? Either in usage or intent: a PublishKey is meant to be a unique identifier of a published dataset and version, and it seems like this could just re-use that here

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+import typer
+
+from dcpy.utils import s3
+from dcpy.utils.logging import logger
+from dcpy.connectors.edm import publishing
+
+BUCKET = "edm-distributions"
+BUCKET_ACL = s3.string_as_acl("public-read")
+
+DOWNLOAD_ROOT_PATH = Path(".library")
```
fvankrieken, line 15: Would make this something else, maybe .publish? .package?

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+    @property
+    def path(self) -> str:
+        return f"{self.product}/{self.version}"
+
+
+def colp(publish_key: publishing.PublishKey, package_key: PackageKey):
+    raise NotImplementedError
+
+
+def templatedb(publish_key: publishing.PublishKey, package_key: PackageKey):
```
fvankrieken, line 36: Just for my own understanding - seems that what we're going for with one of these functions is
- download our "official" output in edm-publishing/{dataset}/{version}
- "package" it locally (in the new .output folder). So .library -> .output
- push the packaged version to edm-distributions (not in this call, but logically the next step)

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+@dataclass
+class PackageMetadata:
+    product: str
+    packaged_name: str
+    packaging_function: Callable
+
+
+DATASET_PACKAGE_METADATA = {
+    "db-colp": PackageMetadata("db-colp", "dcp_colp", colp),
+    "db-template": PackageMetadata("db-colp", "dcp_template_db", templatedb),
```
fvankrieken, line 58: Wrong product!

dcpy/connectors/edm/packaging.py
fvankrieken: Or maybe we need a tweak then, if the "product" in a product key is something like "db-pluto" instead of "pluto". I think it SHOULD cover both use cases, even if it doesn't at this moment. And could be handled differently in different places depending on what's needed

damonmcc: I think the difference in folder structure (and so the path property) led me to make PackageKey:
```
class PublishKey(ProductKey):
    ...
    @property
    def path(self) -> str:
        return f"{self.product}/publish/{self.version}"
```
vs
```
class PackageKey(publishing.ProductKey):
    ...
    @property
    def path(self) -> str:
        return f"{self.product}/{self.version}"
```

related to what we were saying earlier thought about publish and package being so similar we may wanna merge them later. I also thought about extracting all ProductKey stuff out of these cli-focused files, but thought adding packing to highlight potential abstraction first would help

dcpy/connectors/edm/packaging.py
damonmcc: good thing nothing uses `PackageMetadata.product! maybe a sign to drop the field until it's useful

dcpy/connectors/edm/packaging.py
fvankrieken: Gotcha. Yeah I think ideally we'd have one key that really is just that - product name and version. And then it could have like "publish_path" and "package_path" properties. But then we do need more metadata that's really product level. Like Pluto with `name: pluto` and `publishing_folder: db-pluto` etc.

But that's a bit of a broader discussion, and not sure if that's blurring lines a bit with how we've implemented things right now. But something for us to sort out - are these keys meant to map to specific entities like folders in s3/specific build outputs? Or just a version of a product, across multiple locations. Arguments for both, and I think keeping them separate like you have for now probably makes sense

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+}
+
+
+def upload(package_key: PackageKey) -> None:
+    local_folder_path = OUTPUT_ROOT_PATH / package_key.path
+    meta = s3.generate_metadata()
```
fvankrieken, line 64: Would we maybe want to use existing metadata? In this case, we're losing the metadata of the s3 files in publishing (losing as in not shuttling them along)

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+        contents_only=True,
+        metadata=meta,
+    )
+
+
+def get_packaged_versions(product: str) -> list[str]:
+    return sorted(s3.get_subfolders(BUCKET, f"{product}/"), reverse=True)
+
+
+def transform_for_packaging(
```
fvankrieken, line 79: reason to not just call the passed function?

dcpy/connectors/edm/packaging.py
```diff
@@ -0,0 +1,150 @@
...
+        output_path / "templatedb.csv",
+        output_path / "templatedb_packaged.csv",
+    )
+
+
+@dataclass
+class PackageMetadata:
+    product: str
+    packaged_name: str
+    packaging_function: Callable
```
fvankrieken, line 53: Can you add types of args and return value to the callable?

dcpy/connectors/edm/packaging.py
damonmcc: so using `s3.get_metadata` when data is downloaded and passing it in into this upload function as an arg?

dcpy/connectors/edm/packaging.py
damonmcc: I went back and forth about that. [testing](https://github.com/NYCPlanning/data-engineering/blob/dm-packaging-colp/dcpy/test/connectors/edm/test_packaging.py#L54-L58) led me to isolate this, otherwise a test would have to call `packaging.package` which has a single arg of `publish_key`

to keep the `test-product` inputs in `test/conftest.py` and not define test things in source code (like `packaging.DATASET_PACKAGE_METADATA`), the benefits of modularity seemed worth the cost of having this simple function

dcpy/connectors/edm/packaging.py
damonmcc: yup! noting though that, like the push, the download isn't in this call either and it's from edm-publishing/{dataset}/**publish**/{version}

and might be nice to have these functions be more structured/class-based, but I didn't wanna invest too much into making this part rigid when the actual packaging transformations are gonna vary so much by product and haven't really been attempted yet

dcpy/connectors/edm/packaging.py
fvankrieken: for 

```python
packaging.transform_for_packaging(
    publish_key,
    package_key,
    mock_data_constants["TEST_PACKAGE_METADATA"].packaging_function,
)
```

can you not do
```python
mock_data_constants["TEST_PACKAGE_METADATA"].packaging_function(publish_key, package_key)
```

docs/diagrams/workflow_product_build.drawio.png
sf-dcp, line 1: Looks nice

dcpy/connectors/edm/packaging.py
fvankrieken: That's roughly what I was getting at. Definitely open for discussion though

fvankrieken: It might make more sense to use `build_metadata.json` as source of truth than s3 file metadata though

fvankrieken: Which we are generating for all products using recipe now

fvankrieken:
Nice. A few ongoing convos but looks great

dcpy/connectors/edm/packaging.py
damonmcc: doesn't look like the entirety of `build_metadata.json` can become s3 file metadata because it has dictionaries as values ([run here](https://github.com/NYCPlanning/data-engineering/actions/runs/7350128126/job/20011262884#step:6:1619)). I confirmed that the nested dicts cause this by running tests locally

also using `generate_metadata()` may be more desirable so that the packaged files reflect the datetime, commit, and action run that actually packaged them. plus `build_metadata.json` will still be a file with that packaged data

fvankrieken: That makes sense. I'm good with that.

With build_metadata.json, we'd def want to drop the recipe, that's what's causing the issue.

fvankrieken: But for now seems we don't necessarily need anything from it

fvankrieken:
I added some small conflicts with my publish tweak sorry!

dcpy/connectors/edm/packaging.py
damonmcc: sounds good! I'll drop/revise the relevant commits c82fe4e and 057e692

.github/workflows/publish.yml
```diff
@@ -74,8 +75,7 @@ jobs:
           latest: ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
         run: |
           python3 -m dcpy.connectors.edm.publishing publish \
-            -p db-${{ inputs.product }} \
-            -b ${{ inputs.draft_build }} \
-            -v ${{ inputs.publishing_version }} \
-            -a ${{ inputs.product_acl }} \
-            $latest
```
fvankrieken, line 81: You dropped this!

fvankrieken: Make sure you include the final backslash too, on the `--acl` line

damonmcc: oops!

fvankrieken (approved):

damonmcc:
seems like a good idea to squash this when I merge it

dcpy/test/conftest.py
```diff
@@ -18,8 +27,95 @@ def aws_credentials():
...
+def create_buckets(aws_credentials):
     """Creates a test S3 bucket."""
     with mock_s3():
-        s3.client().create_bucket(Bucket=TEST_BUCKET_NAME)
+        for bucket in TEST_BUCKETS:
+            s3.client().create_bucket(Bucket=bucket)
         yield  ## the yield within the mock_s3() is key to persisting the mocked session
+
+
+@pytest.fixture(scope="module")
```
alexrichey, line 38: Can we move this to the publishing test module?

dcpy/test/connectors/edm/test_packaging.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+import os
+from dcpy.connectors.edm import publishing, packaging
+
+
+def test_bucket_empty(create_buckets, mock_data_constants):
```
alexrichey, line 6: Generally we want to avoid tests depending on each other, or depending on order of execution. Maybe we move this check into the other tests?

dcpy/test/connectors/edm/test_packaging.py
```diff
@@ -0,0 +1,67 @@
...
+        packaging.get_packaged_versions(
+            product=mock_data_constants["TEST_PRODUCT_NAME"]
+        )
+        == []
+    )
+
+
+def test_package_fails(create_buckets, create_temp_filesystem, mock_data_constants):
+    # packaging should fail when version is not published
+    with pytest.raises(AssertionError):
```
alexrichey, line 20: Should we also assert on the error message, to make sure we're catching the error that we expect?

alexrichey (approved):
Great start! Had a few comments about the tests, but this all seems sensible.

dcpy/test/conftest.py
damonmcc: `mock_data_constants()` is used in both the publishing and the new packaging test modules. it used to be in `test/connectors/edm/conftest.py` but it seemed nice to delete and consolidate that file with `test/conftest.py`

`create_bucket()` already being defined at this level set the precedent, it's used in tests for connectors and utils

PR #447, opened 2023-12-27
add manual schema override for nightly builds

fvankrieken:
Special schema for nightly builds to differentiate it as a special build environment from main. This way, if we're qa'ing something on main over multiple days, it won't get overwritten

[Nightly qa action](https://github.com/NYCPlanning/data-engineering/actions/runs/7341035203)

Pluto running as I write this, but other products have successfully built and exported using "nightly_qa" as schema and draft folder

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,9 @@ on:
       repeat:
         type: boolean
         default: false
+      build_name:
+        type: string
+        required: false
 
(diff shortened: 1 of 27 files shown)


damonmcc (approved):
nice! so a schema would only be called `main` if we manually run a build on `main` and don't specify a build name?

fvankrieken:
Yup!

PR #453, opened 2023-12-28
add option to publish to latest

fvankrieken:
We had never added this, so seemed about time

[action that successfully published to latest](https://github.com/NYCPlanning/data-engineering/actions/runs/7349107438)

This will be squashed

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: publish
         env:
-          latest: ${{ inputs.latest == 'true' && '--latest' || '' }}
+          latest: ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
         run: |
           python3 -m dcpy.connectors.edm.publishing publish \
             -p db-${{ inputs.product }} \
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,6 +29,7 @@ on:
         description: "Publish to 'latest' folder?"
         type: boolean
         default: true
+        required: true
       product_acl:
         description: "Set access control list for output in DO"
         type: choice
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,6 +25,10 @@ on:
         description: "Version of official release"
         type: string
         required: true
+      latest:
+        description: "Publish to 'latest' folder?"
+        type: boolean
+        default: true
       product_acl:
         description: "Set access control list for output in DO"
         type: choice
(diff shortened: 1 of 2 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -194,18 +194,19 @@ def publish(
     acl: s3.ACL,
     publishing_version: str | None = None,
     keep_draft: bool = True,
-    target_bucket: str | None = None,
```
fvankrieken, line 197: This seemed unnecessary, and easy enough to add back in if we ever move publishing.

Or actually, should we keep for kpdb?

.github/workflows/publish.yml
```diff
@@ -65,9 +70,12 @@ jobs:
         run: ./bash/docker_container_setup.sh
 
       - name: publish
+        env:
+          latest: ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
```
fvankrieken, line 74: interestingly this only works with `github.event.inputs.latest` and not `inputs.latest`. Not sure if the latter can truly be treated as a boolean, in general in GHA boolean logic is a little funny

dcpy/connectors/edm/publishing.py
damonmcc: down to drop it. I just checked and KPDB doesn't use publishing.py since we never publish it. it's currently using `s3.upload_folder()`

(future thoughts) seems like we might wanna pick between file-level access via ACL and bucket-level access via multiple similar buckets

fvankrieken: Oh yeah this would be relevant if kpdb drafts were going to `edm-publishing` and then we were publishing to private, but that obviously makes no sense :)

dcpy/connectors/edm/publishing.py
```diff
@@ -194,18 +194,19 @@ def publish(
...
     max_files: int = s3.MAX_FILE_COUNT,
+    latest: bool = False,
```
damonmcc, line 198: just curious why the default here is false but it's true in the github action. maybe so any potential dcpy tests wouldn't overwrite the more human/manual concept of "latest"?

damonmcc (approved):

dcpy/connectors/edm/publishing.py
fvankrieken: Something like that. Seems good by default (if we're manually doing anything with command line, etc) to not push to multiple places. But if we're doing something more official (using gha) we'll generally be wanting to push to latest.

PR #456, opened 2023-12-28
Update lpc_historic_districts, tweak data library workflows

fvankrieken:
lpc_historic_districts uid/four-four changed for some reason, leading to failures of recurring job.

Also added tweaks to create an issue when this job fails. See [failing action here](https://github.com/NYCPlanning/data-engineering/actions/runs/7350292406) and created issue #455

[Succeeding update of lpc_historic_districts](https://github.com/NYCPlanning/data-engineering/actions/runs/7350281797)

Also added ability to use staging image/action for these data library jobs which use the published action. [Passing open data update](https://github.com/NYCPlanning/data-engineering/actions/runs/7350440650/job/20012099702)

### DIFF
--- a/.github/workflows/data_library_docker.yml
+++ b/.github/workflows/data_library_docker.yml
@@ -42,8 +42,8 @@ jobs:
       - name: Log in to Docker Hub
         uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
         with:
-          username: $DOCKER_USERNAME
-          password: $DOCKER_PASSWORD
+          username: ${{ env.DOCKER_USERNAME }}
+          password: ${{ env.DOCKER_PASSWORD }}
 
       - name: Build and push Docker image
         uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
--- a/.github/workflows/data_library_open_data.yml
+++ b/.github/workflows/data_library_open_data.yml
@@ -41,3 +41,20 @@ jobs:
           s3: true
           compress: true
           output_format: pgdump shapefile csv
+
+  create_issue_on_failure:
+    runs-on: ubuntu-22.04
+    needs: dataloading
+    if: ${{ always() && (github.event_name == 'schedule') && contains(needs.dataloading.result, 'failure') }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          sparse-checkout: .github
+      - name: Create issue on failure
+        uses: JasonEtco/create-an-issue@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ACTION: "Open Data Weekly Updates"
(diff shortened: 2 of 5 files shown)


fvankrieken:
Have noticed that dockerhub login isn't working since movement to 1pass. Y'all can hold off on review

fvankrieken:
Okay, good to go!

.github/workflows/data_library_open_data.yml
```diff
@@ -35,9 +40,36 @@ jobs:
...
       - uses: NYCPlanning/action-library-archive@v1.1
+        if: ${{ !(inputs.use_staging_image) }}
         with:
           name: ${{ matrix.dataset }}
           latest: true
           s3: true
           compress: true
           output_format: pgdump shapefile csv
+
+      - uses: NYCPlanning/action-library-archive@v1.2
```
fvankrieken, line 51: Tried to do it dynamically but forgot that github does not let you specify stuff in a `uses` clause with variables. So this is a bit inelegant, but it works. Only real downside other than repeated logic is that both docker images are pulled as part of the action. But only adds about 30 s to runtime

.github/workflows/data_library_quarterly.yml
```diff
@@ -61,6 +65,17 @@ jobs:
...
+        with:
+          name: ${{ matrix.dataset }}
+          version: ${{ github.event.inputs.version }}
+          latest: true
+          s3: true
+          compress: true
+          output_format: pgdump shapefile csv
+
+      - uses: NYCPlanning/action-library-archive@v1.2
+        if: ${{ inputs.use_staging_image }}
```
fvankrieken, line 78: Did not add the issue creation to this one since it is run manually when the quarterly updates are put out

.github/workflows/data_library_open_data.yml
```diff
@@ -35,9 +40,36 @@ jobs:
...
+          name: ${{ matrix.dataset }}
+          latest: true
+          s3: true
+          compress: true
+          output_format: pgdump shapefile csv
+
+  create_issue_on_failure:
+    runs-on: ubuntu-22.04
+    needs: dataloading
+    if: ${{ always() && (github.event_name == 'schedule') && contains(needs.dataloading.result, 'failure') }}
```
fvankrieken, line 63: added the event_name clause after testing. A little tough to test it working explicitly but we use this logic elsewhere

.github/workflows/data_library_docker.yml
```diff
@@ -42,8 +42,8 @@ jobs:
       - name: Log in to Docker Hub
         uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
         with:
-          username: $DOCKER_USERNAME
-          password: $DOCKER_PASSWORD
+          username: ${{ env.DOCKER_USERNAME }}
+          password: ${{ env.DOCKER_PASSWORD }}
```
damonmcc, line 46: lol

damonmcc (approved):

PR #458, opened 2023-12-28
FacDB Enhancement

### LINKED ISSUE #449: FacDB enhancement - add GIS' post processing to the build
## Motivation

After the DE team builds FacDB, GIS performs additional transformations on the data before distributing it downstream. [Here](https://github.com/NYCPlanning/gis-facilities-and-pops/tree/main) is the link to the code Jack wrote for post processing steps.   Snippet of these steps: 
image

The goal is to incorporate transformation logic from the [script](https://github.com/NYCPlanning/gis-facilities-and-pops/blob/main/scripts/facilities_processing.py) to current FacDB build, producing outputs ready for GIS team review. 

## Enhancements
* [ ] Update schema
* [ ] Ensure correct projection

# Notes
* GIS' post-process also includes metadata updates and distribution of outputs. It would be considered as a part of packaging rather than building a product; therefore, not addressing it in this enhancement.

sf-dcp:
Resolves #449.

# Changes:

This PR consists of 2 commits: 

1) Specify spatial projection of `EPSG:2263` as a default value when exporting build to geodabase `.gdb` format.
    * This update will impact all data products that use `fgdb_export` and/or `fgdb_partial_export` fn from `bash/utils.sh` script. 
    * Current products that use the function: **FacDB**, **COLP**, and **PLUTO**. Note, PLUTO has its own fn `fgdb_export_pluto` that is built on top of the utils`fgdb_partial_export` fn.
    * If a certain data product needs a different projection, it can be specified when calling the `fgdb_export` fn. 
   
2) Modify `facdb` table before exporting to various data formats using [this](https://github.com/NYCPlanning/gis-facilities-and-pops/blob/main/templates/_template_facilities_schema.json) metadata:
    * Updated column names, data formats, data max length (example, `VARCHAR(250)` for `FACNAME` column).
    * Replace empty strings (`''` and `' '`) with `NULL` value across the table.
    * These modifications are done via an additional step in the end of the build. The rational to do this in the end rather than in the beginning when setting up expected table schema is because it would impact many downstream sql commands. For example, renamed columns would have to be updated in `SELECT` statements in the QA-related queries.

# Tests:
* ### Projection is set correctly:

Run command to get a summary of a geodabase file including its spatial projection: 

```bash
ogrinfo -al -so facilities.gdb
```

Output when NOT specifying projection (example below is for the latest published FacDB version `2023-08-09`): 

 
   image

Output when specifying projection (example below is a draft [build](https://github.com/NYCPlanning/data-engineering/actions/runs/7415362668) ran from `sf-facdb-gis-enhancement` branch):

   
image

* ### Correct column names and data types:
Using the same CLI gdal command, investigated my geodabase (left) versus the FacDB Bytes geodabase (right):
image
They should be the same. 

Also, checked my geodabase against expected table [schema](https://github.com/NYCPlanning/gis-facilities-and-pops/blob/main/templates/_template_facilities_schema.json) defined by GIS.

# Notes:
* Successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7415362668)
* Outputs from the run in `edm-publishing/db-facilities/draft/sf-facdb-gis-enhancement/`. [DO link](https://cloud.digitalocean.com/spaces/edm-publishing?i=266877&path=db-facilities%2Fdraft%2Fsf-facdb-gis-enhancement%2F).

### DIFF
--- a/bash/utils.sh
+++ b/bash/utils.sh
@@ -265,6 +265,7 @@ function fgdb_export_partial {
             -nlt ${geomtype}\
             -t_srs ${target_projection}\
             "$@"
+        ogrinfo -al -so ${filename}.gdb
         rm -f ${filename}.gdb.zip
     )
 }
--- a/bash/utils.sh
+++ b/bash/utils.sh
@@ -4,6 +4,7 @@ DATE=$(date "+%Y-%m-%d")
 recipes_bucket=edm-recipes
 publishing_bucket=edm-publishing
 recipes_url=${AWS_S3_ENDPOINT}/${recipes_bucket}
+default_srs=EPSG:2263
 UTILS_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
 ROOT_DIR=$(dirname ${UTILS_DIR})
 export PYTHONPATH=$ROOT_DIR
@@ -246,7 +247,8 @@ function fgdb_export_partial {
     local geomtype=${2}
     local nln=${3}
     local table=${4}
-    shift 4
+    local target_projection=${5}
+    shift 5
     if [ -n "${BUILD_ENGINE_SCHEMA}" ]; then
         local schema=${BUILD_ENGINE_SCHEMA}
     else
@@ -261,6 +263,7 @@ function fgdb_export_partial {
             -lco GEOMETRY_NAME=Shape\
             -nln ${nln}\
             -nlt ${geomtype}\
+            -t_srs ${target_projection}\
             "$@"
         rm -f ${filename}.gdb.zip
     )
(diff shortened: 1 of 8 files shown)


bash/utils.sh
```diff
@@ -261,6 +263,7 @@ function fgdb_export_partial {
             -lco GEOMETRY_NAME=Shape\
             -nln ${nln}\
             -nlt ${geomtype}\
+            -t_srs ${target_projection}\
             "$@"
         rm -f ${filename}.gdb.zip
```
damonmcc, line 268: `ogrinfo -al -so ${filename}.gdb` seems really nice to have in product build logs anytime we generate a geodabase, it's a lot of good info for not a lot of lines

maybe running it right before or after line 268 with the `rm -f ${filename}.gdb.zip`?

products/facilities/facdb/sql/_update_facdb.sql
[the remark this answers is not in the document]
sf-dcp: good catch! will update it

bash/utils.sh
sf-dcp: @damonmcc, see output with the geodabase summary [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7424876744/job/20205383668#step:14:24)

damonmcc (approved):

sf-dcp:
Pending review from GIS.

damonmcc (approved):

PR #465, opened 2023-12-29
Move data library into dcpy

fvankrieken:
Very little logic changes. Ended up fiddling with docker images a bit to get things working. I had forgotten what a headache it is to get gdal python bindings working.

In this, I
- moved data-library/library into dcpy/library. This includes taking a few relevant bits from data-library/pyproject.toml and incorporating them in root pyproject.toml, plus updating module names within dcpy.library and some paths for tests (and just moving data-library/tests to dcpy/tests/library)
- updated workflow files to reflect this
- got rid of data library docker image. data library jobs now use `build-base` with dcpy installed like builds do. Needed a couple tweaks

Was going to start refactoring data library as part of this but figure it'd actually be nice to get this merged first, keep things slightly more manageable

[Passing single runner action](https://github.com/NYCPlanning/data-engineering/actions/runs/7359065423)
[Passing opendata scheduled action](https://github.com/NYCPlanning/data-engineering/actions/runs/7359079499)


alexrichey (approved):
very nice

fvankrieken:
Also worth just emphasizing - data library published docker image and action are not going to be maintained after this, since it'll just be part of dcpy. Not going to delete them so Te can continue to enjoy them

I don't think we have actions in other repos that use them, and the ones in this repo have been updated to use dcpy (from checked out repo)

Issue #116, opened 2023-08-03
Improved Automated QA for Recipes and Products


alexrichey:
In building FacDB, I ran into a few issues which were trivial to fix, but nonetheless took time to diagnose. They might provide a nice framework to discuss potential improvements.

Here's a sampling of issues I encountered Building FacDB
- `USECODE`s were corrupted on Bytes: they should have been 4 digit numerics, but were cast to numbers. E.g. "0211" -> 211. 
- - This is an interesting case, because this is data that could benefit from our automated checks, but modifications after `edm-publishing` are out of our control.
- Column were changed or removed in source data (e.g. `borough` in dsny_electronicsdrop`)
- Columns had mixed data after being imported into our python dataframes (`zip_code` in `dsny_fooddrop` was sometimes being read in as a float in pandas)
- `dot_parking` ballooned in size and changed completely. It would have been nice to catch this prior to import to EDM Recipes, rather than noticing the discrepancy in the output, and then having to purge the bad data from S3 (I could easily have forgotten to do that).

What I'm thinking for next steps: Perhaps model out the required/important fields in FacDB for 1) a subset of recipes 2) for the output to edm-publishing. The output of this exercise would be some declarative format about expectations (probably fields with types modeled in yml) which we'd then use to write automations to detect and potentially coerce out-of-spec data into a usable format. Modeling might be a nice way to indicate which columns actually matter at the periphery. E.g. at ingestion time, if we no longer have the `boro` field in `dsny_electronicsdrop` does that actually matter? 

There are some neat libraries (e.g. Cerberus) that we might make use of, though I think it would be nice to get our feet wet before making a decision about them.

Thoughts @fvankrieken , @damonmcc . Would love to hear about your pain points as well. If it'd be easier, we could just huddle and jot down some notes.

damonmcc:
random thoughts:

definitely like the focus on what columns a product needs aka a product's expectations

is a significant amount of this pain caused by the current `data_library` approaches? it doesn't save the raw data and we still expect it do some transformations often via it's python scripts?

I imagine this could go too far, but I think I like the general approach of something failing during a build rather than investing too much in pre-build checks. Definitely still love declaring expectations though!

[...]

PR #467, opened 2024-01-01
use dbt to test source and output data

damonmcc:
## why

related to issue #116

We've talked about testing source and output data for our builds. `dbt` is often used to model, transform, and test data. By only adding the modeling and testing parts, we can use `dbt` for data validation without refactoring product build logic.

Template DB seems like a good product to try this out on.

## changes

- python package changes from adding `dbt-postgres` to `requirements.in` and compiling via github action
  - pydantic must be rolled back to use `dbt-core>1.5` (for now, `pydantic 2.x` compatibility seems to be coming soon)
- add dbt project configuration for Template DB
- add yml files to model the output tables and a single source table of Template DB
- add tests to those models using built-in and `dbt-utils` tests
- run dbt tests at the end of `python3 -m build_scripts.transform`

## notes
- a test passing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7392820649/job/20111939790#step:9:158)
- a test that raising a warning [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7392820649/job/20111939790#step:9:134)


products/template/build_scripts/transform.py
```diff
@@ -1,16 +1,41 @@
...
         PG_CLIENT.engine_uri,
         SQL_QUERY_DIR / f"{step_query}.sql",
     )
 
 
+def _run_dbt_step(run_args=list[str]):
+    dbt_cli_call = f"dbt {' '.join(run_args)}"
+    logger.info(f"Running '{dbt_cli_call}' ...")
+    dbt_runner = dbtRunner()
+    runner_result: dbtRunnerResult = dbt_runner.invoke(run_args)
```
damonmcc, line 21: this is where dbt is called, see docs [here](https://docs.getdbt.com/reference/programmatic-invocations)

products/template/dbt_project.yml
damonmcc, line 1: this file configues the dbt project

products/template/profiles.yml
damonmcc, line 1: this file configures database connections for dbt

products/template/build_scripts/transform.py
sf-dcp: is this command equivalent to `dbt run` CLI command?

products/template/build_scripts/transform.py
```diff
@@ -21,12 +46,21 @@ def intermediate():
     _run_step("historic_landmarks")
 
 
-def outputs():
+def product():
```
sf-dcp, line 49: from my limited research, it seems that it is possible to invoke table builds via one CLI command (`dbt run`). I wonder if there is a similar python way to do that...

products/template/profiles.yml
sf-dcp, line 1: what does this code do? Does it configure the schema where to output tables from `product()` fn?

sf-dcp:
Really exciting stuff!

From watching a couple of dbt tutorials, I saw a couple of interesting things that seem to be useful: 

* Having 1 SQL file per table (model) and ability to reference them via Jinja

  image


For example: during `staging` step, you create a bunch of tables such as `stg_nypl_libraries`. From my understanding, you can create a `stg_nypl_libraries.sql` model with the following content:

```SQL
SELECT
    name AS library_name,
    locality AS borough,
    wkb_geometry
FROM nypl_libraries;
```

Instead of: 

```SQL
-- "staging.sql" file

CREATE TABLE stg_nypl_libraries AS SELECT
    name AS library_name,
    locality AS borough,
    wkb_geometry
FROM nypl_libraries;
```


And they you can reference the table in other queries as `{{ ref('stg_nypl_libraries') }}`.

products/template/build_scripts/transform.py
```diff
@@ -21,12 +46,21 @@ def intermediate():
...
+def test():
+    _run_dbt_step(["deps"])
+    _run_dbt_step(["debug"])
+    _run_dbt_step(["test", "--select", "source:*"])
+    _run_dbt_step(["test", "--select", "templatedb"])
+    _run_dbt_step(["test", "--select", "aggregation"])
 
 
 if __name__ == "__main__":
     staging()
```
sf-dcp, line 63: Not sure if we need `staging()` if all it does is creates copies of source tables. Thoughts?

products/template/build_scripts/transform.py
damonmcc: only if run is passed as the first `run_args`. so  `_run_dbt_step(["debug"])` is equivalent to `dbt debug`

I chose to align with the `_run_step` function above but maybe this shouldn't have the word "run" in the name?

products/template/build_scripts/transform.py
damonmcc: I think it'd be `dbt_runner.invoke(["run"])` and with the wrapper I wrote above it'd be `_run_dbt_step(["run"])`

products/template/profiles.yml
damonmcc: (for some reason the comment I left here is labeled as outdated)

this file is teh connection profile ([docs](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles)). it configures database connections, including which schema to default to (I think any can be used if it's explicitely specified in a query)

products/template/build_scripts/transform.py
damonmcc: true it doesn't do much, but it does a couple of valuable things: selects the subset of columns that are actually useful and renames some to be either more readable or more standardized

see [docs](https://docs.getdbt.com/best-practices/how-we-structure/2-staging#staging-models) on ideas behind the value of using staging models

damonmcc:
> For example: during `staging` step, you create a bunch of tables such as `stg_nypl_libraries`. From my understanding, you can create a `stg_nypl_libraries.sql` model with the following content:
> 
> ```sql
> SELECT
>     name AS library_name,
>     locality AS borough,
>     wkb_geometry
> FROM nypl_libraries;
> ```
> 
> Instead of:
> 
> ```sql
> -- "staging.sql" file
> 
> CREATE TABLE stg_nypl_libraries AS SELECT
>     name AS library_name,
>     locality AS borough,
>     wkb_geometry
> FROM nypl_libraries;
> ```
> 
> And then you can reference the table in other queries as `{{ ref('stg_nypl_libraries') }}`.

yup! I started down that route in an "experiment" PR [here](https://github.com/NYCPlanning/data-engineering/pull/344) at first. but seems nice to start to model and test products that exist and save refactoring transformations for later

products/template/profiles.yml
sf-dcp: I see. +1 to dbt from me for their docs👏

products/template/build_scripts/transform.py
sf-dcp: Got it! I think "run" in the fn name is a bit confusing given that there is a command with that keyword. How about `_execute_dbt_step` or `_perform_dbt_step`?

products/template/build_scripts/transform.py
sf-dcp: I thought you were using `dbt` to build the tables so was confused having multiple steps in the `product` fn. Makes sense

PR #473, opened 2024-01-02
Add parquet as export option

fvankrieken:
So gdal is only available via apt/apt-get as version 3.6.2 (on debian stable at least). So I've gone with a manual install of gdal. Took a bit of work making sure the right dependencies for our workflows are installed.

Went about this by making a nycplanning "base" docker image. This image simply is basically just the python3.11-slim debian image with gdal (and necessary support for python bindings, postgres driver, parquet driver) installed. The manual install takes like 15 min, and I figured when we're doing our dev image builds (during PR tests), it'd be nice to have gdal already set up to save time.

So general workflow around builds
- `base` - never automatically built. Builds via workflow dispatch. Base image `python:3.11-slim`. In general, assume we won't be updating this much
- `build-base`/`build-geosupport`/`dev` - all use `base` as their base image. Are built whenever we push to main, and are built for specific dev branches (and used for tests) when changes to `docker` or `python` folders are present.
- `build-geosupport` - bit of a legacy image at this point. Still uses python:3.11-slim

Main outstanding question for me is how we want to go about parquet files. I think it'd be nice to start exporting parquet whenever we export, and gradually phase out sql? We already have import logic for parquet implemented. It might be useful to add logic to try to default to parquet and fall back to pgdump? Open to thoughts


fvankrieken:
Closing for now - just don't want to put the time into getting gdal 3.8.2 working on our debian stable images

fvankrieken:
Reopening - got this working without too much work on our current base image with a manual gdal install

.devcontainer/docker-compose.yml
```diff
@@ -13,30 +13,3 @@ services:
...
-    # Runs app on the same network as the database container, allows "forwardPorts" in devcontainer.json function.
-    network_mode: service:db
-
-    # Uncomment the next line to use a non-root user for all processes.
-    # user: vscode
-
-    # Use "forwardPorts" in **devcontainer.json** to forward an app port locally.
-    # (Adding the "ports" property to this file will not forward from a Codespace.)
-
-  db:
```
fvankrieken, line 25: Just seemed like this was overkill since we're often using dev environments in edm-data now. Figure it makes more sense to set up postgres locally as-needed and default to no db, free up some resources when developing locally

.github/workflows/python_compile_requirements.yml
```diff
@@ -31,6 +29,6 @@ jobs:
           add-paths: python/
           title: Compile python requirements
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
-          assignees: fvankrieken, damonmcc, alexrichey, sf-dcp
+          assignees: ${{ github.event_name == 'schedule' && 'fvankrieken, damonmcc, alexrichey, sf-dcp' || '' }}
```
fvankrieken, line 32: Something was weird in my dev container so ran this via action. Figured in these cases when the PR is generated we don't need to assign us

docker/build-base/Dockerfile
```diff
@@ -1,5 +1,4 @@
-ARG PYTHON_VERSION
-FROM python:${PYTHON_VERSION}-slim
+FROM nycplanning/base:latest
```
fvankrieken, line 1: Debated if "build-base" was now redundant and if we could just "merge" it with base, but base takes longer to build (15 min - not that long but a bit annoying if you're waiting on that for your pr tests), and since we don't expect to bump gdal or python version often we can treat it a bit more constant. Meanwhile build-base can still have its python versions bumped and then tested in a dev image in PRs

.devcontainer/docker-compose.yml
fvankrieken: Maybe I'll add these back in "docker-compose_db.yml" or something like that so the option is easily there

dcpy/library/cli.py
```diff
@@ -17,7 +17,7 @@
 @app.command()
 def archive(
     path: str = typer.Option(None, "--path", "-f", help="Path to config yml"),
-    output_formats: list[str] = typer.Option(["pgdump"], "--output-format", "-o", help="csv, geojson, shapefile, pgdump and postgres"),
+    output_formats: list[str] = typer.Option(["parquet"], "--output-format", "-o", help="csv, geojson, shapefile, pgdump and postgres"),
```
fvankrieken, line 20: I'll revert this for now

docker/build-base/Dockerfile
damonmcc: definitely like having a lowest-level image to put things that take a long time to build and that we don't often change in PRs

damonmcc:
> Main outstanding question for me is how we want to go about parquet files. I think it'd be nice to start exporting parquet whenever we export, and gradually phase out sql? We already have import logic for parquet implemented. It might be useful to add logic to try to default to parquet and fall back to pgdump? Open to thoughts

I'm in favor of exporting all primary formats when we archive a dataset. that would let us phase out sql later and we won't have to run an archive multiple times in the same day (a current issue with Pluto and FacDB using COLP but in different formats). so maybe parquet, pgdumb, and csv?

seems like action run time and data-library bucket storage would be the only cons?

fvankrieken:
> > Main outstanding question for me is how we want to go about parquet files. I think it'd be nice to start exporting parquet whenever we export, and gradually phase out sql? We already have import logic for parquet implemented. It might be useful to add logic to try to default to parquet and fall back to pgdump? Open to thoughts
> 
> I'm in favor of exporting all primary formats when we archive a dataset. that would let us phase out sql later and we won't have to run an archive multiple times in the same day (a current issue with Pluto and FacDB using COLP but in different formats). so maybe parquet, pgdumb, and csv?
> 
> seems like action run time and data-library bucket storage would be the only cons?

I'm open to it. I still think for this PR, let's not change norms - I'd like to do a bit more testing within pipelines, and keep this PR as "enable" parquet rather than switch to parquet

docker/config.sh
```diff
@@ -16,12 +17,25 @@ function install_mc {
...
 
-function install_gdal_python_reqs {
-    apt-get install -y libgdal-dev \
-        && apt-get -y install --reinstall build-essential
+function install_gdal {
+    wget https://apache.jfrog.io/artifactory/arrow/$(lsb_release --id --short | tr 'A-Z' 'a-z')/apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb
+    sudo apt-get install -y -V ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb
+    sudo apt-get update
+
+    sudo apt-get install -y -V libproj-dev libgeos-dev libarrow-dev libparquet-dev cmake
```
damonmcc, line 25: to confirm, is this to install arrow/parquet things because _we_ wanna use GDAL + parquet or because GDAL 3.8.2 requires these?

fvankrieken: The GDAL parquet driver requires them

fvankrieken: for the other packages here, proj and geos are needed for gdal, cmake is used to run the build/install

dcpy/library/cli.py
fvankrieken: reverted

damonmcc (approved):
🚀

PR #474, opened 2024-01-02
Data library cleanup

fvankrieken:
Lot of little commits but will likely be easier to go by commit

A bunch of workflows, bash commands, etc used docker or published action to invoke data library. I've gone through and updated them to use dcpy instead. Also,
- remove some duplicate templates. ZTL templates were all in two places
- move some templates from product folders to dcpy/library/templates. I left some - these are the cases where there's more actual logic ahead of the data-library call, so doesn't make sense entirely to just have it as a standalone template in library code. Yet

Passing actions
- [DevDB inputs](https://github.com/NYCPlanning/data-engineering/actions/runs/7401684260)
- [PLUTO CAMA](https://github.com/NYCPlanning/data-engineering/actions/runs/7401737250)
- [PLUTO numblgs](https://github.com/NYCPlanning/data-engineering/actions/runs/7401736366)
- [PLUTO pts](https://github.com/NYCPlanning/data-engineering/actions/runs/7401734607)
- [ZAP weekly](https://github.com/NYCPlanning/data-engineering/actions/runs/7403595401)
- [ZAP single](https://github.com/NYCPlanning/data-engineering/actions/runs/7410901664)
- [ZTL dataloading](https://github.com/NYCPlanning/data-engineering/actions/runs/7401817334)


dcpy/library/cli.py
```diff
@@ -36,18 +36,19 @@ def archive(
...
-            push=push,
-            clean=clean,
-            latest=latest,
-            name=name,
-            compress=compress,
-            inplace=inplace,
-            postgres_url=postgres_url,
-            version=version,
-        )
+        for output_format in output_formats:
```
fvankrieken, line 39: A little silly to be doing it this far upstream in the process. But that's what we do now when we specify multiple formats. And since we're going to move away from multiple formats in general, doesn't seem worth the effort to produce multiple output files from the same gdal object (especially since data library is a maze. Initially thought I'd just quickly do that but it's not as simple as one might hope)

fvankrieken: And I'm still preserving this functionality rather than going just to pgdump for these cases 
1. just to not change functionality at the moment
2. for datasets where we historically always have csv outputs, converting those to parquet would be easier than with the pgdumps. Not sure if we will want to for historical recipe datasets but just in case.

fvankrieken:
@damonmcc is there any downside to running zap actions as a test? In terms of versioning with data library or something like that.

fvankrieken:
Added a temp commit to not push anything to s3, links to run zap actions now in description

damonmcc:
> @damonmcc is there any downside to running zap actions as a test? In terms of versioning with data library or edm private or something like that.

other than S3 storage concerns (I'm not sure how much space we have/use) don't think there's any downside to having more than just the weekly versions

.github/workflows/zap_single_visible.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     container:
       image: nycplanning/build-base:latest
     env:
-      AWS_S3_LIBRARY_BUCKET: edm-recipes
+      AWS_S3_BUCKET: edm-recipes
```
fvankrieken, line 26: This is something that we can drop once library is using standard dcpy s3 utils rather than its own

damonmcc: since it's always the same bucket looks like it could be dropped from workflows and set in `library.__init__.py`, but happy to save that for later

dcpy/library/cli.py
```diff
@@ -17,7 +17,7 @@
 @app.command()
 def archive(
     path: str = typer.Option(None, "--path", "-f", help="Path to config yml"),
-    output_format: str = typer.Option("pgdump", "--output-format", "-o", help="csv, geojson, shapefile, pgdump and postgres"),
+    output_formats: list[str] = typer.Option(["pgdump"], "--output-format", "-o", help="csv, geojson, shapefile, pgdump and postgres"),
```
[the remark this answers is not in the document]
damonmcc: looks like it's working. I'm seeing both formats in `edm-recipes/datasets/pluto_input_geocodes/20240103/`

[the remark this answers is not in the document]
fvankrieken: No, typer options expect exactly one text field

damonmcc (approved):

PR #476, opened 2024-01-03
drop archived issue templates

### LINKED ISSUE #451: extract product update details from issue templates
we recently archived all issue templates that were specific to a product. but those markdown files have details on source data that's worth preserving in an accessible place

will likely copy those details to the relevant wiki pages or README files

damonmcc:
resolves #451 

For now, I put the product update details in their README files rather than their wiki pages

Will squash these commits before merging to main


fvankrieken:
Seems good - I would say that we should strive to update these readmes in this go around (so @sf-dcp refresh the facdb readme during this build, I'll refresh the devdb one, @alexrichey pluto)

fvankrieken (approved):

damonmcc:
@fvankrieken moved the data library readme to the right place and dropped most of it

PR #477, opened 2024-01-03
add build_name input to zoningtaxlots_build call

damonmcc:
GIS tried running ZTL dataloading but it fails due to invalid workflow ([example](https://github.com/NYCPlanning/data-engineering/actions/runs/7399164235))

that action calls `zoningtaxlots_build.yml` but doesn't provide [the required input](https://github.com/NYCPlanning/data-engineering/blob/dm-fix-ztl-dataloading/.github/workflows/zoningtaxlots_build.yml#L9-L11) `build_name`

PR #478, opened 2024-01-03
add build_name input to pluto_build call

damonmcc:
follow up to #477 because the action failed at the next step ([run logs](https://github.com/NYCPlanning/data-engineering/actions/runs/7400747383))


jackrosacker (commented):
👍🏼

PR #481, opened 2024-01-03
add KPDB recipe and fix DevDB data loading

### LINKED ISSUE #313: Implement Remaining Recipes Files

damonmcc:
resolves #313 and DevDB breaking after https://github.com/NYCPlanning/data-engineering/pull/434

### DIFF
--- a/dcpy/utils/versions.py
+++ b/dcpy/utils/versions.py
@@ -103,7 +103,7 @@ class FirstOfMonth(Version):
     ) -> FirstOfMonth:
         if version_subtype is not None:
             raise Exception(
-                f"Version subtype {version_subtype} not applicable for Quarterly versions"
+                f"Version subtype {version_subtype} not applicable for FirstOfMonth versions"
             )
         bump_by = bump_by or 1
         new_year = int(self.year + (self.month + bump_by - 1) / 12)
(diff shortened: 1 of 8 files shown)


dcpy/utils/versions.py
```diff
@@ -120,11 +120,47 @@ def generate() -> FirstOfMonth:
                 raise Exception("Version parsing failed")
 
 
+@dataclass(order=True)
```
fvankrieken, line 123: Ooh. This is going to collide with the current "first of month". All our other "Today" ones are formatted "{year}{month}{day}"

Maybe worth biting the bullet and making "monthly" rather than "first of month" for ztl, and format it "{year}-{month}". Or standardizing it as a "major" build every month. Food for thought. Though it'll need to be addressed here in some way so we don't have ambiguous versions.

Or we could leave KPDB for now, which I thought we had said was the plan yesterday?

dcpy/utils/versions.py
```diff
@@ -120,11 +120,47 @@ def generate() -> FirstOfMonth:
...
+        if version_subtype is not None:
+            raise Exception(
+                f"Version subtype {version_subtype} not applicable for Today versions"
+            )
+        bump_by = bump_by or 1
+        new_year = int(self.year + (self.month + bump_by - 1) / 12)
+        new_month = int(
+            self.month + (self.day + bump_by - 1) / 31
+        )  # not every month has 31 days!
+        new_day = ((self.day + bump_by - 1) % 31) + 1
```
fvankrieken, line 145: Probably should just use datetime utilities here

dcpy/utils/versions.py
damonmcc: yea I also hit some other fun stuff [here](https://github.com/NYCPlanning/data-engineering/blob/5cfe42977415bb6fdc2591fead1aa6e37cc16468/dcpy/test/utils/test_versions.py#L38) with date math, bumping date-based versions could probably use a refactor to use the `datetime` library or something

PR #482, opened 2024-01-04
fix devdb dataloading

damonmcc:
fixes DevDB builds breaking after #434 ([example](https://github.com/NYCPlanning/data-engineering/actions/runs/7393675363/job/20113963933))

a DevDB build passing with these changes [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7404016831)

### DIFF
--- a/.github/workflows/developments_build.yml
+++ b/.github/workflows/developments_build.yml
@@ -6,9 +6,18 @@ on:
       image_tag:
         type: string
         required: false
-      build_name: 
+      build_name:
         type: string
         required: true
+      recipe_file:
+        type: string
+        required: true
+      version:
+        type: string
+        required: false
+      repeat:
+        type: boolean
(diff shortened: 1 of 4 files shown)


.github/workflows/developments_build.yml
```diff
@@ -46,8 +55,14 @@ jobs:
         working-directory: ./
         run: ./bash/build_env_setup.sh
 
-      - name: 1. dataloading for EDM builds
-        run: ./devdb.sh dataloading edm && ls -l
+      - name: Set product environment variables
+        run: cat version.env >> "$GITHUB_ENV"
```
fvankrieken, line 59: Would like to change this a bit but that's what I'm already doing in my branch 🙂

fvankrieken (approved):
Thanks 🤦

PR #485, opened 2024-01-04
DevDB Continued Refresh

fvankrieken:
Relatively straightforward. Waiting @damonmcc's version pinning of devdb tests

Commits
1. Move majority of version.env vars into recipe, and grab previous version during planning
2. update readme
3. use parquet for a single dataset
4. more error handling in looking up past versions


dcpy/builds/plan.py
```diff
@@ -107,6 +108,19 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
...
+        recipe.product, recipe.version
+    )
+    if previous_recipe is not None:
+        if recipe.env is None:
+            recipe.env = {}
+        recipe.env["version_prev"] = previous_recipe.label
+
+    env = recipe.env or {}
+    for key in env:
+        os.environ[key] = env[key]
```
fvankrieken, line 123: To be clear, this doesn't update bash/shell env. Just for use within this specific python call

dcpy/connectors/edm/publishing.py
```diff
@@ -104,11 +108,20 @@ def previous(product: str, version: str | versions.Version) -> versions.Version:
...
                 f"{product} - {version} is not published and appears to be 'older' than latest published version. Cannot determine previous."
             )
 
 
+def try_get_previous_version(
+    product: str, version: str | versions.Version
+) -> versions.Version | None:
+    try:
+        return get_previous_version(product, version)
+    except (LookupError, ValueError):  # versions not found, or don't parse correctly
```
alexrichey, line 121: Should we let these Exceptions bubble up? Or is there a legitimate case where there wouldn't be a previous build?

fvankrieken: ValueError potentially. I mainly didn't want to break anybody's code right now, but it's going to be tough to sort out issues if we don't raise errors.

The LookupErrors we have some "valid" cases - new data product, or repeat of original build where there truly is no previous. I'd rather return None and let build logic decide what to do than throw an error here, personally

fvankrieken:
> Nice, though I do wonder if we're conflating template variables and env variables. Wondering if we might either:
> 
> 1. just set the recipe's `env` variables into the environment at the beginning of a build. Might makes sense to have some dcpy functionality to set env vars, for various reasons
> 2. or just call these `vars` instead of `env`, and don't bother putting them in the os.environ
> 
> In any case, I do really like capturing vars in the recipe, esp if we upload the recipe file to `edm-publishing`, just to make it explicit what went into the build.
> 
> Thoughts @damonmcc @sf-dcp ?

Yeah separating env and template vars seems like a good distinction. Especially since env often contains secrets - not that bash env vars are making it into the recipe's `env` field, but just seems good to keep that distinction explicit and make sure nobody makes some unfortunate assumptions

fvankrieken:
> > Nice, though I do wonder if we're conflating template variables and env variables. Wondering if we might either:
> > 
> > 1. just set the recipe's `env` variables into the environment at the beginning of a build. Might makes sense to have some dcpy functionality to set env vars, for various reasons
> > 2. or just call these `vars` instead of `env`, and don't bother putting them in the os.environ
> > 
> > In any case, I do really like capturing vars in the recipe, esp if we upload the recipe file to `edm-publishing`, just to make it explicit what went into the build.
> > Thoughts @damonmcc @sf-dcp ?
> 
> Yeah separating env and template vars seems like a good distinction. Especially since env often contains secrets - not that bash env vars are making it into the recipe's `env` field, but just seems good to keep that distinction explicit and make sure nobody makes some unfortunate assumptions

There are multiple cases where "input" vars still need to be made part of the bash env for the rest of the build, so maybe we should move all of that to template vars, and then export the entire `vars` field, rather than the targeted way it's being done for version and version_prev for devdb in the github action in this pr (and pluto as well, though currently no changes in this PR)

alexrichey:
@fvankrieken I kinda lean towards 
> so maybe we should move all of that to template vars, and then export the entire vars field
Just for consistency across the entire build process. But I don't have a strong opinion, and I'm happy if you want to get this in and we can let it percolate.

dcpy/connectors/edm/publishing.py
alexrichey: Gotcha with the LookupError. I'd be a little worried that catching `ValueError`s might cause some silent issues, but totally get erring on the side of not breaking things.

alexrichey (approved):

fvankrieken:
Going to make that env/var tweak, then re-run devdb build before merging

dcpy/library/templates/dcp_mappluto.yml
```diff
@@ -4,7 +4,7 @@ dataset:
   acl: public-read
   source:
     url:
-      path: https://s-media.nyc.gov/agencies/dcp/assets/files/zip/data-tools/bytes/nyc_mappluto_{{ version }}_unclipped_shp.zip
+      path: https://s-media.nyc.gov/agencies/dcp/assets/files/zip/data-tools/bytes/nyc_mappluto_{{ version | replace(".","_") }}_unclipped_shp.zip
```
fvankrieken, line 7: This is so we can still have mappluto as "23v3.1" in data library but it will fetch the url with "23v3_1"

bash/export_recipe_env.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
fvankrieken, line 1: Unfortunately for now it made sense to do it this way, just for ease of running during builds

fvankrieken: One day we'll get build execution into python

fvankrieken:
[build all on the latest commit of this branch](https://github.com/NYCPlanning/data-engineering/actions/runs/7466765874)

alexrichey (approved):
Nice!

PR #491, opened 2024-01-07
improve TDB build tests

damonmcc:
## why

We been running end-to-end tests for Template DB whenever Template DB is built as a test. They were meant to ensure that certain expectation of the build were being met: certain tables and files exist.

With table testing via `dbt` in place, some of these tests are redundant and can be removed.

## changes

See commits in order for changes. notable chanegs:
- [move dbt calls to more sensible places](https://github.com/NYCPlanning/data-engineering/pull/491/commits/5b8a990c62dcd493daf240f9a8324eb57e0f64d8)
- [model and test intermediate tdb tables](https://github.com/NYCPlanning/data-engineering/pull/491/commits/1eaef43d6dff8e6181d505aa4318a1280b18a635)

## notes
- This is still **not** using dbt to transform data. This is still modeling and testing, something that we could sooner introduce to real product builds.
- Logs [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7442948223/job/20247124058#step:9:100) show the results of testing intermediate tables


products/template/sql/_intermediate_models.yml
```diff
@@ -0,0 +1,53 @@
...
+    columns:
+      - name: library_name
+        tests:
+          - not_null
+      - name: borough
+      - name: wkb_geometry
+    tests:
+      - unique:
+          name: libraries_unique_compound_key
+          column_name: "(library_name || '-' || borough)"
```
fvankrieken, line 15: Just chatted in person, but this syntax is awkward (and as it turns out, doesn't quite work as expected) compared with dbt_utils.unique_combination_of_columns

products/template/tests/test_build.py
```diff
@@ -6,22 +6,6 @@
...
-        "lpc_landmarks",
-    ]
-    actual_tables = PG_CLIENT.get_schema_tables()
-    for table in expected_source_tables:
-        assert table in actual_tables
-
-
 # * test Transform stage
 @pytest.mark.end_to_end()
 def test_transform_staging():
```
fvankrieken, line 11: Would you want dbt to cover this as well?

products/template/sql/_intermediate_models.yml
damonmcc: will drop use of the built-in unique test for compound keys

products/template/tests/test_build.py
damonmcc: true, I just chose to start with the output-end and save these for later. the [`_sources.yml`](https://github.com/NYCPlanning/data-engineering/blob/dm-e-to-e-product-test/products/template/sql/_sources.yml#L11) is also incomplete

happy to cover all tables in this PR if it seems worth it

fvankrieken: Nah totally up to you

damonmcc: I'll hold off on covering all staging and source tables

fvankrieken (approved):

products/template/build_scripts/transform.py
```diff
@@ -37,6 +37,7 @@ def _invoke_dbt(run_args=list[str]):
 
 
 def staging():
+    _invoke_dbt(["test", "--select", "source:*"])
```
sf-dcp, line 40: came here to say I like new fn names 👏

sf-dcp:
@damonmcc, do you know if it's possible to customize dbt logs? I feel like removing `[RUN]` logs and just keeping result ones (`[PASS]` or `[FAIL]`) would make it cleaner.

image

products/template/build_scripts/transform.py
```diff
@@ -37,30 +37,27 @@ def _run_dbt_step(run_args=list[str]):
...
-    _run_dbt_step(["test", "--select", "templatedb"])
-    _run_dbt_step(["test", "--select", "aggregation"])
+    _execute_sql_script("product/templatedb")
+    _execute_sql_script("product/aggregation/templatedb_boroughs")
+    _invoke_dbt(["test", "--select", "product.templatedb"])
+    _invoke_dbt(["test", "--select", "product.aggregation"])
 
 
 if __name__ == "__main__":
+    _invoke_dbt(["deps"])
```
sf-dcp, line 59: How does line 59 work? Trying to understand package management between our env and the `template
/packages.yml` file

products/template/build_scripts/transform.py
```diff
@@ -37,30 +37,27 @@ def _run_dbt_step(run_args=list[str]):
...
 
 
 def intermediate():
-    _run_step("libraries")
-    _run_step("green_spaces")
-    _run_step("historic_landmarks")
+    _execute_sql_script("libraries")
+    _execute_sql_script("green_spaces")
+    _execute_sql_script("historic_landmarks")
+    _invoke_dbt(["test", "--select", "libraries green_spaces historic_landmarks"])
```
sf-dcp, line 48: How does dbt know where to look for models? For `intermediate` step, your data contract is in `sql` dir. For `product`, it's in `product` dir.

products/template/build_scripts/transform.py
damonmcc: this is testing the source tables, which were created during data loading

products/template/build_scripts/transform.py
damonmcc: the [`dbt_project.yml`](https://github.com/NYCPlanning/data-engineering/blob/dm-e-to-e-product-test/products/template/dbt_project.yml#L0-L1) file tells dbt where all model files live via `model-paths: ["sql"]` ([docs](https://docs.getdbt.com/reference/project-configs/model-paths))

and two notes on how dbt knows where to look:
- 100% agree that those intermediate queries and their yml files (they may be called [`properties.yml`](https://docs.getdbt.com/reference/configs-and-properties) files) could be in a named subdirectory to group things together. I just erred on the side of not changing too much in this pass
- there are many many ways to select certain models or tests (node selection syntax  [docs](https://docs.getdbt.com/reference/node-selection/syntax)). in this case, I explicitly listed the model names. so dbt finds and runs all tests related to those model

products/template/build_scripts/transform.py
damonmcc: it parses a project's `packages.yml` file and installs the listed packages in the same directory as that project. so since I've run this locally, I have a folder that's `template/dbt_packages/dbt_utils/` so the project can use `dbt_utils`

this may be more ideal to do during build environment setup, but it has to be specific to the dbt project (the folder `products/template`)

Issue #461, opened 2023-12-29
FacDB 24Q1


damonmcc:
### Product Name

facilities

### Build Version

24Q1

### Status of Update

- [x] source data extracted
- [x] draft build succeeded
- [x] draft build passed QA
- [x] data packaged and distributed

sf-dcp:
Update steps (old issue template)

# FacDB Source Data Updates
Like most of our data products, source data must be updated in data library before FacDB is run. As there are are many source datasets with varied update processes, this issue template should be opened to track progress towards updating all source data

**All source data listed is to be uploaded as .csv files**

## Geosupport
- [ ] Ensure latest Geosupport version is present in production EDM Docker images

## Version Env
- [ ] Update previous version for QCQA in `version.env` file

## Scraped by data library

- [ ] bpl_libraries
Source: Scraped from BPL website
Source url: https://www.bklynlibrary.org/locations/json

- [ ] nypl_libraries
Source: Scrape from NYPL website
Source url: https://www.nypl.org/locations/list

- [ ] uscourts_courts
Source: Court locator for NY state
Source url: http://www.uscourts.gov/court-locator/city/New%20York/state/NY

## Source data from OpenData
To see if a dataset needs to be uploaded, check date last updated in open data against version in data library

- [ ] dca_operatingbusinesses https://data.cityofnewyork.us/Business/Legally-Operating-Businesses/w7w3-xahh

- [ ] dcp_colp https://www1.nyc.gov/site/planning/data-maps/open-data.page#city_facilities

- [ ] dcp_facilities_with_unmapped https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-selfac.page

- [ ] dcla_culturalinstitutions https://data.cityofnewyork.us/Recreation/DCLA-Cultural-Organizations/u35m-9t32

- [ ] dfta_contracts https://data.cityofnewyork.us/Social-Services/DFTA-Contracts/6j6t-3ixh

- [ ] doe_busroutesgarages https://data.cityofnewyork.us/Transportation/Routes/8yac-vygm

- [ ] sca_enrollment_capacity https://data.cityofnewyork.us/Education/Enrollment-Capacity-And-Utilization-Reports-Target/8b9a-pywy

- [ ] dohmh_daycare https://data.cityofnewyork.us/Health/DOHMH-Childcare-Center-Inspections/dsg6-ifza

[...]

PR #495, opened 2024-01-08
FacDB Regular Update

sf-dcp:
Related to #461. 

## Commits: 
1. Update current (24v1) and previous FacDB versions.
2. Revise build template in FacDB `Readme.md` (updated source data links, instructions for data extract).
3. `doe_universalprek` source data changed since its last version. This commit is to update code logic for the new version of the dataset. The changes are related to column names and values in `type` column. 
4. `dfta_contracts` source data changed since its last version. This commit is to update code logic for the new version of the dataset. The changes are related to column names and values in `contract_type` column. The new logic has been discussed and agreed on [here](https://github.com/NYCPlanning/data-engineering/issues/461#issuecomment-1910967112). 
5. Update templates in data-library with new links, description, socrata id
6. Update versions for `dcp_pops` and `data_contracts`  in FacDB `recipe.yml`.


products/facilities/facdb/sql/doe_universalprek.sql
```diff
@@ -2,41 +2,41 @@ DROP TABLE IF EXISTS _doe_universalprek;
...
+    state AS zipcode,
     boro,
-    NULL as borocode,
-    NULL as bin,
-    NULL as bbl,
+    NULL AS borocode,
+    NULL AS bin,
+    NULL AS bbl,
     (CASE
-        WHEN type = 'DOE' THEN 'DOE Universal Pre-K'
```
sf-dcp, line 16: The query relies on the `type` column to produce facdb classifications. The values in the `type` column have changed since the dataset's last version, but the mapping of old to new values appears to be 1-1. 

I kept old values in the query (for example: `WHEN type = <old_val> OR type = <new_val> ...`) in case we would like to build FacDB using an older version of this dataset.

products/facilities/recipe.yml
```diff
@@ -48,7 +48,7 @@ inputs:
       version: "20210409"
     # TODO: revisit this pin
     - name: dcp_pops
-      version: "20230320"
+      version: "20240131"
```
damonmcc, line 51: should this version pinning and the TODO above it be dropped to just use the latest?

damonmcc (approved):

products/facilities/recipe.yml
sf-dcp: Yes, I agree. Curious why it was pinned before. I will make the edit

sf-dcp:
@damonmcc, is it okay to merge today or should we wait for facdb to pass qa?

damonmcc:
@sf-dcp I'd go with merging and keeping the issue open

that way any corrections due to QA could be in a new small PR

PR #506, opened 2024-01-11
fix ZTL QA page

### LINKED ISSUE #270: Fix ZTL QA versions
When changes were made in #109 , ztl "sql" versions had their slashes removed. Old data was not updated. So we need to either 
1. update old data in edm-data (defaultdb, schema dcp_zoningtaxlots)
   - tables named for version
   - version in rows in `qc_...` tables
2. fix code to add back slashes for versions in qa

Will go with option 1 unless you feel strongly otherwise @damonmcc

damonmcc:
resolves #270 

The issue is about the old format of version names in tables in edm-data (`defaultdb.dcp_zoningtaxlots`).

But it doesn't seem like we need to remove the backslashes from old sql table names and records. There doesn't seem to be any code that parses the values and we're planning to do more dynamic generation of QA artifacts based on actual published data in S3.

So this PR just fixes the broken QA page. This doesn't fix the (experimental) source data report in the ZTL page because that requires setting the envar `BUILD_ENGINE_SCHEMA` in the QA droplet (I think via ssh?).

See ZTL page in deployed QA app [here](https://de-qaqc.nycplanningdigital.com/?page=Zoning+Tax+Lots)

## before
image

## after
image

### DIFF
--- a/apps/qa/src/shared/components/sources_report.py
+++ b/apps/qa/src/shared/components/sources_report.py
@@ -1,5 +1,6 @@
 import pandas as pd
 import streamlit as st
+from dcpy.utils import postgres
 from dcpy.connectors.edm import publishing
 from src.shared.utils.source_report import (
     get_source_data_versions_to_compare,
@@ -8,6 +9,8 @@ from src.shared.utils.source_report import (
     compare_source_data_row_count,
     dataframe_style_source_report_results,
 )
+
+from src import QAQC_DB, QAQC_DB_SCHEMA_SOURCE_DATA
 
 
 def sources_report(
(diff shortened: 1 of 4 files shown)


fvankrieken:
Might make sense in general to just make one client at essentially the root level of the app and reference it within functions rather than passing it around. Though I imagine that we currently actually query multiple databases from the qa app given the current disparity of code. Seems good to go as a fix!

fvankrieken (approved):

damonmcc:
sounds good! forgot to drop the commit/revert combo so I'll do that before merging

PR #509, opened 2024-01-12
DevDB 23Q4 continued tweaks

fvankrieken:
Tweaks to get new census data loading, add a space so that sqlite statement is "valid"


.github/workflows/build.yml
```diff
@@ -54,6 +54,10 @@ on:
           - WARNING
           - INFO
           - DEBUG
+      build_name:
+        description: "Build name (defaults to branch name)"
+        type: string
+        required: false
```
fvankrieken, line 60: Figured it's nice to have the option to specify a custom build name

damonmcc: I like it. I think any DB schema that isn't the same as an existing repo branch will get deleted at the end of the week though, just noting

damonmcc (approved):

.github/workflows/build.yml
fvankrieken: Yeah it's a little bit of a downside. But still nice every once in a while to be able to run two builds from the same branch without much hassle

fvankrieken:
Tmp fix with the space in the yml template while awaiting official fix of https://github.com/OSGeo/gdal/issues/9093

PR #510, opened 2024-01-12
improve use of envars in python files

### LINKED ISSUE #288: use of environment variables in `dcpy` during builds
per [this discussion](https://github.com/NYCPlanning/data-engineering/pull/269#discussion_r1349316650) during review of #269

## problems
- due to mixed use of the persistent or a temporary postgres DB, builds use environment variables in different ways
- setting relevant envars locally isn't straightforward
- for running builds, we'd like separation between "config" and "app" logic (see [Twelve-Factor App](https://12factor.net/) as a model)
- constructing/determining the name of a draft build (e.g. `dm-little-branch` or `pr_1000`) during the app logic violates this separation
  - especially if it involves invoking the git CLI

## considerations
- envars set by a python script are not applied to the higher-level environment that called the script. so bash scripts during a build will not reflect any envars set in an earlier python script

## TODO proposed approach
- set an environment variable before the "app" logic of the build begins?
- when python is to-be-run, check if a `.env` exists and use it if it does?

damonmcc:
resolves #288 

These changes are to to standardize how we retrieve envars since we're happy with how they're being set (only set in github action and test setups rather than when we run application code).

Seems more ideal to use `os.environ["ENVAR_NAME"]` rather than `os.environ.get("ENVAR_NAME")` when there's no default value so that our code raises an error when a required envar isn't set.

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -400,11 +400,11 @@ def _cli_wrapper_upload(
 ):
     acl_literal = s3.string_as_acl(acl)
     if not output_path.exists():
-        raise Exception(f"Path {output_path} does not exist")
+        raise FileNotFoundError(f"Path {output_path} does not exist")
     build_name = build or os.environ["BUILD_NAME"]
-    if build_name is None:
-        raise Exception(
-            "Build name must either be supplied via CLI or found in env var 'BUILD_NAME'."
+    if not build_name:
+        raise ValueError(
+            f"Build name supplied via CLI or the env var 'BUILD_NAME' cannot be '{build_name}'."
         )
     logger.info(
         f'Uploading {output_path} to {product}/draft/{build_name} with ACL "{acl}"'
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -401,7 +401,7 @@ def _cli_wrapper_upload(
     acl_literal = s3.string_as_acl(acl)
     if not output_path.exists():
         raise Exception(f"Path {output_path} does not exist")
-    build_name = build or os.environ.get("BUILD_NAME")
+    build_name = build or os.environ["BUILD_NAME"]
     if build_name is None:
         raise Exception(
             "Build name must either be supplied via CLI or found in env var 'BUILD_NAME'."
--- a/dcpy/test/connectors/edm/test_publishing.py
+++ b/dcpy/test/connectors/edm/test_publishing.py
@@ -33,6 +33,8 @@ def test_upload(create_buckets, create_temp_filesystem, mock_data_constants):
 
 
 def test_publish(create_buckets, create_temp_filesystem, mock_data_constants):
+    data_path = mock_data_constants["TEST_DATA_DIR"]
+    publishing.upload(output_path=data_path, draft_key=draft_key, acl=TEST_ACL)
     publishing.publish(
         draft_key=draft_key, acl=TEST_ACL, version=None, keep_draft=False
     )
--- a/dcpy/test/utils/test_s3.py
+++ b/dcpy/test/utils/test_s3.py
@@ -1,5 +1,6 @@
 import pytest
 import os
+from botocore.exceptions import EndpointConnectionError
 from dcpy.utils import s3
 from dcpy.test.conftest import TEST_BUCKET, TEST_BUCKETS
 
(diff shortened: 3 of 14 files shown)


dcpy/utils/s3.py
```diff
@@ -92,9 +92,11 @@ def _progress():
 
 def client() -> S3Client:
     """Returns a client for S3."""
+    aws_s3_endpoint = os.environ["AWS_S3_ENDPOINT"]
+    # when running tests with mock S3 buckets, endpoint_url can be None but not "" or any fake url
+    endpoint_url = None if not aws_s3_endpoint else aws_s3_endpoint
```
alexrichey, line 97: Sorry, approved the PR, but took another look, and I'm not sure about this. Is there a way to accommodate this in the test, rather than changing util code?

damonmcc: yea this was hard to pick an approach for. I'll change it and use a test to show how `botocore` handles the input and throws exceptions

dcpy/test/utils/test_s3.py
```diff
@@ -8,9 +9,38 @@
...
+    with pytest.raises(ValueError, match="Invalid endpoint: testing"):
+        os.environ["AWS_S3_ENDPOINT"] = "testing"
+        s3.client()
+
+    with pytest.raises(ValueError, match="Invalid endpoint: "):
+        os.environ["AWS_S3_ENDPOINT"] = ""
+        s3.client()
+
+
+@pytest.mark.skip(reason="takes ~5 seconds")
```
alexrichey, line 30: Hmm... maybe you can set the connection timeout for s3 to be a very small unit of time?

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -33,6 +33,8 @@ def test_upload(create_buckets, create_temp_filesystem, mock_data_constants):
 
 
 def test_publish(create_buckets, create_temp_filesystem, mock_data_constants):
+    data_path = mock_data_constants["TEST_DATA_DIR"]
+    publishing.upload(output_path=data_path, draft_key=draft_key, acl=TEST_ACL)
     publishing.publish(
         draft_key=draft_key, acl=TEST_ACL, version=None, keep_draft=False
     )
```
damonmcc, line 40: added these lines so that `test_publish` doesn't depend on `test_upload` having been run first

dcpy/test/utils/test_s3.py
damonmcc: i _think_ @fvankrieken set the timeout to 120 seconds [here](https://github.com/NYCPlanning/data-engineering/pull/94/commits/7cdf852835ef6fc4e6971f04b1bb4fd7dda1c2b7) because some files really took longer than the default 60 seconds

this test was mainly just to help me figure out where boto was getting mad about various endpoint url formats. happy to drop it. if someone sets a bad url, boto will raise this descriptive error

dcpy/utils/s3.py
damonmcc: @alexrichey I've revised https://github.com/NYCPlanning/data-engineering/pull/510/commits/8343aadbb6229479fcb53f03d51e23e63e0e0fdb

dcpy/test/utils/test_s3.py
alexrichey: Yeah, I was thinking that in tests you could potentially override the timeout. Happy with whatever route you go though

dcpy/test/utils/test_s3.py
```diff
@@ -8,9 +9,27 @@
...
 
 
+# assert that the default mock credentials work
+def test_client_create(aws_credentials):
+    s3.client()
+
+
+# assert client creation results for invalid client endpoint_url options
+def test_client_create_errors(aws_credentials):
+    with pytest.raises(TypeError, match="str expected, not NoneType"):
```
alexrichey, line 19: wait, does this block actually raise an exception?

dcpy/test/utils/test_s3.py
```diff
@@ -8,9 +9,27 @@
...
+        s3.client()
+
+    with pytest.raises(ValueError, match="Invalid endpoint: "):
+        os.environ["AWS_S3_ENDPOINT"] = ""
+        s3.client()
+
+
 def test_list_buckets(create_buckets):
     """Tests listing s3 buckets."""
-    # buckets = s3.client().list_buckets()["Buckets"][0]["Name"]
```
alexrichey, line 13: ✂️

alexrichey (approved):
Had one question and one nit, but looks good to me!

dcpy/test/utils/test_s3.py
damonmcc: the first `with pytest.raises` block? yea we can't set envars to non-strings

dcpy/connectors/edm/publishing.py
```diff
@@ -401,7 +401,7 @@ def _cli_wrapper_upload(
     acl_literal = s3.string_as_acl(acl)
     if not output_path.exists():
         raise Exception(f"Path {output_path} does not exist")
-    build_name = build or os.environ.get("BUILD_NAME")
+    build_name = build or os.environ["BUILD_NAME"]
     if build_name is None:
         raise Exception(
```
fvankrieken, line 406: ✂️  This will never get raised now

dcpy/test/utils/test_s3.py
fvankrieken: Missing an `s3.client()` on line 21?

fvankrieken: Though having no endpoint is fine by boto3 standards - defaults to standard aws endpoint (we use DO hence our custom one). And something around the default endpoint behavior is how moto mocks it I think (hence the os.environ.pop in conf_test)

fvankrieken: Didn't see Damon's response, that makes sense. Is that worth testing in this case then? If its simply os.environ functionality, not related to client behavior

dcpy/connectors/edm/publishing.py
damonmcc: see [0a04fdd](https://github.com/NYCPlanning/data-engineering/pull/510/commits/0a04fdd846212106e953bc85d5c5503f4d02103f). still seemed worth raising if `BUILD_NAME = ""` since nothing downstream of this seems to

dcpy/test/utils/test_s3.py
damonmcc: see [a9a642f](https://github.com/NYCPlanning/data-engineering/pull/510/commits/a9a642feba597a0fdd1b57537a60c65b8aca37a9)

fvankrieken: That makes sense, let's leave it. Or maybe add a comment as to why it's there?

PR #525, opened 2024-01-17
Add create issue job for all scheduled workflows

fvankrieken:
Seemed low-effort and useful enough to add. Now that nightly_qa should actually be passing nightly, figured for it and all other recurring jobs it'd be useful to create issues so nothing fails too silently.

Happy to remove nightly_qa from the list if we feel like it might get spammy. I hope not though.

Will drop the second commit.

As is, it ran [this job](https://github.com/NYCPlanning/data-engineering/actions/runs/7561798946), creating #531

### DIFF
--- a/.github/workflows/nightly_qa.yml
+++ b/.github/workflows/nightly_qa.yml
@@ -24,3 +24,20 @@ jobs:
       recipe_file: recipe
       dev_image: false
       build_name: ${{ inputs.build_name || 'nightly_qa' }}
+
+  create_issue_on_failure:
+    needs: [test, build]
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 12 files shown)


.github/workflows/nightly_qa.yml
```diff
@@ -11,16 +11,35 @@ on:
...
+    steps:
+      - run: exit 1
+
+  fake_success:
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - run: exit 0
```
damonmcc, line 28: ✂️?

damonmcc (approved):
seems nice! (from IRL chat) it'd be cool to query a github variable to fill in the `ACTION: "..."` lines

PR #533, opened 2024-01-18
Expand parquet support, make default datasettype

fvankrieken:
[Build all](https://github.com/NYCPlanning/data-engineering/actions/runs/7613210139)

Note that this isn't testing what works using parquet imports, just makes sure that current functionality is not broken, in terms of importing via preference (parquet -> pgdump -> csv)

Main changes here are 
- data library - default to generating pgdump (for now, in case things go weird as we try to use parquet) and parquet files for every export
- connectors.edm.recipes and build.plan - if no dataset file type supplied, assume that we want to import parquet, and fall back to pg_dump and then csv if parquet is not found

### DIFF
--- a/.github/workflows/pluto_input_cama.yml
+++ b/.github/workflows/pluto_input_cama.yml
@@ -85,8 +85,7 @@ jobs:
           library archive --name pluto_input_cama_dof \
             --path templates/pluto_input_cama_dof.yml \
             --version ${{ steps.processing.outputs.version }} \
-            --output-format pgdump --output-format csv \
-            --latest --s3 --compress
+            --latest --s3
 
       - name: Clean CAMA
         run: ./pluto clean cama
--- a/dcpy/builds/load.py
+++ b/dcpy/builds/load.py
@@ -23,10 +23,6 @@ def import_dataset(
     pg_client: postgres.PostgresClient,
 ):
     """Import a recipe to local data library folder and build engine."""
-    ds_table_name = ds.import_as or ds.name
-    logger.info(
-        f"Importing {ds.name} into {pg_client.database}.{pg_client.schema}.{ds_table_name}"
-    )
 
     if ds.version == "latest" or ds.version is None:
         raise Exception(f"Cannot import a dataset without a resolved version: {ds}")
@@ -41,7 +37,10 @@ def import_dataset(
         preproc_func = None
 
     recipes.import_dataset(
-        ds.dataset, pg_client, preprocessor=preproc_func, import_as=ds.import_as
+        ds.dataset,
+        pg_client,
+        preprocessor=preproc_func,
+        import_as=ds.import_as,
     )
 
 
--- a/dcpy/builds/plan.py
+++ b/dcpy/builds/plan.py
@@ -40,8 +40,8 @@ class InputDataset(BaseModel, use_enum_values=True, extra="forbid"):
 
     @property
     def dataset(self):
-        if self.version is None or self.file_type is None:
-            raise Exception(f"Dataset {self.name} requires both version and file_type")
+        if self.version is None:
+            raise Exception(f"Dataset {self.name} requires version")
 
         return recipes.Dataset(
             name=self.name, version=self.version, file_type=self.file_type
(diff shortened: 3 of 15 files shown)


dcpy/builds/plan.py
```diff
@@ -176,14 +176,10 @@ def get_source_data_versions(recipe: Recipe):
 
 
 def _apply_recipe_defaults(recipe: Recipe):
-    recipe.inputs.dataset_defaults = (
-        recipe.inputs.dataset_defaults
-        or InputDatasetDefaults(file_type=recipes.DatasetType.pg_dump)
```
fvankrieken, line 181: The logic around default preference is now handled in connectors/edm/recipes.py

dcpy/builds/load.py
```diff
@@ -23,10 +23,6 @@ def import_dataset(
     pg_client: postgres.PostgresClient,
 ):
     """Import a recipe to local data library folder and build engine."""
-    ds_table_name = ds.import_as or ds.name
-    logger.info(
-        f"Importing {ds.name} into {pg_client.database}.{pg_client.schema}.{ds_table_name}"
```
fvankrieken, line 28: This also gets logged in the called recipes.import_dataset, meaning this was printed twice each time before

dcpy/connectors/edm/recipes.py
```diff
@@ -159,7 +160,8 @@ def import_dataset(
         }
         df.rename(columns=columns, inplace=True)
         pg_client.insert_dataframe(df, ds_table_name)
-        pg_client.add_pk(ds_table_name)
+        if id_column is None:
+            pg_client.add_pk(ds_table_name)
```
fvankrieken, line 164: Should maybe do this differently. This is letting you specify what column in the input dataset is the id column. In the case of this dataset, there is a column called `id` that seems to be serial. Since the dataset has an id column, none is added.

However, this column isn't then specified as a primary key/unique index. So I could add that. However, that might be making assumptions about data and validation in a spot that we don't really want it (at time of import). And in that case, maybe here it would make more sense to specify a different label for a newly-generated id column rather than the label for the existing one in the dataset.

Thoughts?

.github/workflows/pluto_input_cama.yml
```diff
@@ -85,8 +85,7 @@ jobs:
           library archive --name pluto_input_cama_dof \
             --path templates/pluto_input_cama_dof.yml \
             --version ${{ steps.processing.outputs.version }} \
-            --output-format pgdump --output-format csv \
-            --latest --s3 --compress
+            --latest --s3
```
damonmcc, line 88: any consequence of dropping `--compress`?

dcpy/connectors/edm/recipes.py
```diff
@@ -130,7 +155,13 @@ def import_dataset(
         )
         if preprocessor is not None:
             df = preprocessor(ds.name, df)
+        columns = {
+            column: column.replace("_", "_").replace("'", "_") for column in df.columns
+        }
+        df.rename(columns=columns, inplace=True)
```
damonmcc, line 161: seems worth noting why this is now needed

fvankrieken: Oops, there's also a typo there. Replacing underscore with another underscore.

dcpy/connectors/edm/recipes.py
damonmcc: hmmm yea I feel like since the only effect of specifying `id_column` is that a PK isn't generated we should just generate a new PK every time

that way, any builds that expect it will have it and there'd be no risk of being mislead by specifying a column that hasn't been checked. in the future when we refactor build logic we can always investigate, test, and use a source ID column and rely less of an expected `ogc_fid`-ish column generated at archival

damonmcc: noting we're gonna try always adding `ogc_fid` on import when the source dataset isn't a pgdump

dcpy/connectors/edm/recipes.py
```diff
@@ -109,6 +130,9 @@ def import_dataset(
     preprocessor: Callable[[str, pd.DataFrame], pd.DataFrame] | None = None,
 ):
     """Import a recipe to local data library folder and build engine."""
+    if ds.file_type is None:
+        ds.assign_file_type([DatasetType.parquet, DatasetType.pg_dump, DatasetType.csv])
```
fvankrieken, line 134: This is where the "ordering" of dataset preference is defined

fvankrieken: Not sure why I put ordering in quotes

dcpy/connectors/edm/recipes.py
```diff
@@ -130,7 +154,14 @@ def import_dataset(
...
         if preprocessor is not None:
             df = preprocessor(ds.name, df)
+
+        # make column names more sql-friendly
+        columns = {
+            column: column.replace("-", "_").replace("'", "_") for column in df.columns
+        }
+        df.rename(columns=columns, inplace=True)
         pg_client.insert_dataframe(df, ds_table_name)
+        pg_client.add_pk(ds_table_name, "ogc_fid")
```
fvankrieken, line 164: Figured it made more sense on the recipe import side to specify "ogc_fid" as id column rather than in the pg util, keep the code in utils a bit more pure

fvankrieken:
@damonmcc [latest facdb passing dataloading](https://github.com/NYCPlanning/data-engineering/actions/runs/7627384326/job/20784475237). All others passed dataloading as well

damonmcc (approved):

Issue #523, opened 2024-01-16
Handling "ogc_fid" column


fvankrieken:
When gdal creates a pg_dump, it creates a serial pk column "ogc_fid", which was not present in the original source dataset. These are not generated for csv/parquet outputs.

Because unique, simple ids are useful, many product builds have logic that uses these ogc_fid columns, and therefore will break when naively importing parquet files. So we need a fix. I propose we have our parquet/csv to_sql methods handle this. Can have an optional `id` arg which can specify the name of the id column if present, and if not have it create a serial pk column. We could simply keep it as "ogc_fid", or just call it "id". "ogc_fid" is nice in that it seems unlikely that we'd ever have a collision, I could imagine we have a dataset that has an "id" column that's not just a serial db pk.

fvankrieken:
This was implemented, presumably we'll have to phase out these columns in some way post overhaul, but I don't think this really needs to be scoped right now

PR #534, opened 2024-01-18
DevDB - Geocode applications during build

### LINKED ISSUE #536: DevDB - geocode job applications during build
The DevDB datasync action currently does a few things

- call data library step for three socrata datasets
  - one hpd dataset
  - dob permits issued
  - dob bis job applications (legacy system that seemingly still gets some use - dataset is regularly updated)
- geocode
  - for two hpd datasets, import datasets from edm-recipes, geocode and the archive them with data library. The second one is one that we have only manually loaded once
  - for applications data, another recipes dataset is imported - dob now applications. This is the newer system for filing applications with DOB. It is available on socrata, but we get a custom version via ftp from dob (with more columns than what's available on socrata). This is appended to the bis applications, geocoded, and exported as one dataset (dob_geocode_results)

Logic in DevDB build then imports ALL of these datasets - 2 hpd, 2 job applications - both ungeocoded and geocoded. Geocoded results are then joined based on runtime-generated ids of the input datasets line up with the geocode results. Which they should for  This means that if datasync runs, then dob_now data is loaded manually, and then devdb is built, thousands of records will join erroneously.

fvankrieken:
Resolves #536. Related to #523 as well

A few main changes
- removal of geocoding applications during datasync action. Datasync action still geocodes hpd data which is archived as first step of job
- call to geocode_dob.py during build. Instead of being aimed at a more raw concatenation of bis and now data, it's aimed at _init_devdb, which effectively has the same cleaning done on it upstream that was being done in the sql query in this file before
- little bit of sql cleanup

I tested this by making an intermediate table - persisted the DRAFT table defined in cte in _geo.sql - from both the archived geocode results and by using the new geocoded _init table, and ensuring they were completely identical (`select * from a EXCEPT select * from b`, and vice-versa)


dcpy/utils/geocode.py
fvankrieken, line 1: This isn't going to actually be part of this PR, but late yesterday just wanted to start working on moving geosupport to more usable, typed classes. Would love that to be part of python-geosupport repo at some point

fvankrieken:
[Passing build](https://github.com/NYCPlanning/data-engineering/actions/runs/7575918809)

.github/workflows/developments_datasync.yml
```diff
@@ -5,17 +5,14 @@ on:
     - cron: "0 0 * * MON"
   workflow_dispatch:
     inputs:
-      version:
-        description: "Would you like to specify a version for dob_jobapplications?"
```
alexrichey, line 9: Such polite phrasing. A little sad to see it go

products/developments/python/geocode_dob.py
```diff
@@ -35,15 +29,15 @@ def geocode(input):
...
-                borough=borough,
+                street_name=input["address_street"],
+                house_number=input["address_numbr"],
+                borough=input["boro"],
                 mode="tpad",
             )
             geo = parse_output(geo)
             geo.update(
                 dict(
-                    uid=uid,
```
alexrichey, line 46: Does changing these keys from uid to id change anything downstream?

fvankrieken: Yup - just used within the build as a unique row identifier for _init_devdb and downstream tables. They don't make it to export though. You'll see the same replacement in the sql files

alexrichey (approved):
Awesome

PR #535, opened 2024-01-18
[experiment] expand repeat build feature

damonmcc:
This allows us to repeat builds that have a `metadata.yml` rather than `build_metadata.json` or `source_data_versions.csv`

see builds using this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-facdb-repeat-23v2)

### update

likely will abandon this endeavor. for all source data that was a pg_dump (list [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/facilities/recipe.yml#L12-L39)) the versions used in FacDB 2023-08-09 were not documented in `metadata.yml`. so recreating that build would take more investigating and guesswork than seems worth it right now

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,14 @@ on:
           - pluto
           - ztl
           - all
+      repeat:
+        description: "Repeat existing build?"
+        type: boolean
+        default: false
+      version:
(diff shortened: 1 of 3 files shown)


fvankrieken:
There's a metadata.yml??

PR #540, opened 2024-01-23
Use new nycoc_checkbook as source of checkbook data for cpdb

fvankrieken:
https://github.com/NYCPlanning/data-engineering/actions/runs/7629076331

Outputs look good! Hadn't dug into if the "cpdb_capital_spending" was very transformed, but seems like it's working properly. Spot checked some projects and verified that they agree with older checkbook data from other source


products/cpdb/recipe.yml
```diff
@@ -23,7 +23,8 @@ inputs:
     - name: doitt_buildingfootprints
 
 # Projects
-    - name: cpdb_capital_spending
+    - name: nycoc_checkbook
+      version: 2010-2023
```
fvankrieken, line 27: pinned for now - we should perhaps think a bit more about how we actually want to version this, when we trigger the job, etc

damonmcc (approved):
love a little pr!

Issue #471, opened 2024-01-02
PFF 2024


damonmcc:
### Product Name

pff

### Build Version

2024

### Status of Update

- [x] source data extracted
- [x] draft build succeeded
- [x] draft build passed QA
- [x] data packaged and distributed

fvankrieken:
Update from pop consists of
- 18-22 ACS data. Needs to be ingested roughly analogously to past "manual updates"
- data dictionary for 18-22 ACS data. Needed to regenerate metadata.json files
- "Updated asian subgroups" excel file - seems not needed by us, has mappings of new demographics columns. Think it's essentially a subset of data dictionary
- ACS0610 updated social tab - exactly what it sounds like. 1 of 4 datasets of 0610 acs with updated data
- "ForDelivery_DDHCA2020_forPFF.xlsx" - still trying to figure out specifically what's needed here. Potentially just new decennial data, replacing last year's dhc decennial data?

fvankrieken:
Todo

- [ ] ingest new acs data, get pipeline working
- [ ] update 0610 data, get exported
- [ ] figure out ddhca data

fvankrieken:
Distribution doesn't quite exist yet

PFF has passed internal QA, is awaiting feedback/enhancement requests from AE (which I have offered as they're trying to simplify logic on their end)

fvankrieken:
Going to leave open until it's known that the outputs are properly ingested by AE by their current migration scripts

damonmcc:
AE will show Population team the data in the app so they can QA it

fvankrieken:
AE confirming data is good to go after review with Pop

PR #541, opened 2024-01-23
PFF 24Q1

### LINKED ISSUE #539: PFF source data complications
Mainly getting some thoughts down around stuff going into #471. This could be part of #499.

## Motivation

PFF has presented some unique cases around updating source datasets. Currently, the PFF workflow largely consists of pivoting tabular data from excel sheets and concatenating/dumping into a csv. Relatively straightforward. However, there's a bit of complexity in terms of trying to map some of the entities into our current architecture. 

Mainly, this update is focused on ACS data. ACS has data for year 2010, then gets updates lagging slightly. So last year, we published 2021 data, this year we're updating with 2022. On the AE side of things, the desired end result is comparison of the latest data to the 2010 data. A year ago, these files were checked directly into the repo for simplicity. I archived them in data library as raw excel files just to move it a bit more to our desired state. So in DO, we have
- acs
  - 2010
    - acs.xlsx
  - 2021
    - acs.xlsx
  - 2022
    - acs.xlsx
However, this year, they also have some updated 2010 data. Specifically for one of the four sheets in the 2010 acs.xlsx. So first issue arises that "2010" is not a unique enough "version" of this dataset, since it can be updated. Second, it's a bit awkward to be updating just part of it - in our current system, the easiest way to do this would be to download the xl, replace the sheet, re-load. On the side of PFF/building, we need acs 2010 data, but the "latest" version of it, meaning "acs_2010" maybe needs to be its own dataset.

## Possible implementations

Thinking about our plans to archive raw source data, one possibility that might make sense is mapping one raw source dataset to multiple "recipe" datasets. These files really do have 4 datasets (social, economic, etc - the sheets within the xl), and in the current case, we really want to update just one of those. So assuming we have raw source data in `edm-inbox` for now, we could have something like

```mermaid
flowchart TB
  subgraph edm-inbox
    acs_2010/acs_2010_20230101.xlsx
    acs_2010/acs_2010_20240101_social_update.xlsx
  end
  data_library
  acs_2010/acs_2010_20230101.xlsx-->data_library
  data_library_2
  acs_2010/acs_2010_20240101_social_update.xlsx-->data_library_2
  subgraph edm-recipes
    acs_2010/20230101/acs_2010.parquet
    acs_2010/20240101/acs_2010.parquet
  end
  data_library-->acs_2010/20230101/acs_2010.parquet
  acs_2010/20230101/acs_2010.parquet-->data_library_2
  data_library_2-->acs_2010/20240101/acs_2010.parquet
```

Then at build time, we simply look for latest acs_2010. This is maybe a tad clunky - I don't love updating a dataset with itself in general, and only conditionally. Not quite sure what this would look like in terms of a library template yml/python script, needing to sometimes just grab raw data only, sometimes append, etc.

It might be nicer to go with

```mermaid
flowchart TB
  subgraph edm-inbox
    acs_2010/acs_2010_20230101.xlsx
    acs_2010_social/acs_2010_20240101_social_update.xlsx
  end
  data_library
  acs_2010/acs_2010_20230101.xlsx-->data_library
  data_library_2
  acs_2010_social/acs_2010_20240101_social_update.xlsx-->data_library_2
  subgraph edm-recipes
    acs_2010_social/20230101/acs_2010_social.parquet
    acs_2010_economic/20230101/acs_2010_economic.parquet
    acs_2010_social/20240101/acs_2010_social.parquet
  end
  data_library-->acs_2010_social/20230101/acs_2010_social.parquet
  data_library-->acs_2010_economic/20230101/acs_2010_economic.parquet
  data_library_2-->acs_2010_social/20240101/acs_2010_social.parquet
```

Then at build time, acs_2010_social and acs_2010_economic are both inputs, and latest can be grabbed separately for each.
This would then need a template yml for each of the sub-datasets of course, and there still is some conditionality - maybe just name of the excel file and the sheet to look at. But those two things are relatively easily parameterizable. It's maybe a tad inelegant in our current framework that after loading a "complete" acs dataset to edm-inbox, you'd have to run multiple data library jobs, each opening this bulky excel file. Maybe more ideally, there's one data library job that is triggered by uploading a dataset (or other trigger for non-manual files) that can then process and spit out multiple recipe datasets, but that's certainly way outside our infrastructure at the moment, and we'd have to think about how that gets configured - 1 yml file per recipe dataset, and something separate for the raw data processing job? Seems like better to go with something that aligns just a tad better with our current system for now

fvankrieken:
Part of #471, resolves #539 (at least in part)

Want to get this merged before getting into trying to use more of a "recipe"-style approach, as this works for the moment. Touched some data library stuff, commit-by-commit is def easiest

### Context - what is this python file?

Referring to `products/factfinder/pipelines/acs_manual_update.py`

This python file essentially is the whole "build" script for PFF acs data - one of the subdatasets of PFF. It basically takes a bunch of tables from excel inputs and pivots them

### Commits
1. rename default data library path for local loading. Just seemed easiest at this point - "library" no longer a top level folder from where library is being run, and we already have the root ".library" gitignored folder. Happy to change, just seems to make sense.
   - though now that I think about this, this will maybe just cause a headache for merge conflicts with others right now. Happy to drop (@sf-dcp lemme know)
2. self-explanatory. Wanted it for column names in generating metadata files
3. Also self-explanatory
4. logic around streamlining logic a bit for new manual acs updates. Added a function to generate metadata from excel data dictionary input, which currently is not in S3 or checked in to the repo, nor is this function called in the actual "pipeline". I included more info than before because I figured it couldn't hurt. This is a bit ugly, need to think of how I actually want to do this longer term. But something quick and dirty for now to get things building.
5. split up one of the source datasets. There are sort of two entities in the acs updates, the reference 2010 dataset, of which its four "sub"-datasets can be updated individually, and the latest ACS. These are handled just slightly differently overall, so I only changed logic around importing the 2010 one - splitting it up into 4 datasets, one per excel sheet, and getting it in data library as parquet files. There's some clunkiness, it's pretty manual to load them, but it's an improvement. There are some data library tweaks here to make it possible to re-use scripts, define what args are sent in.
   - might be worth just biting the bullet and doing the same for "acs" - as in, have two acs datasets in recipes, "dcp_pop_acs2010" and just "dcp_pop_acs" which is then versioned by year - I don't see a scenario where we update dataset "vintages" other than the 2010 one. And then for each of those two have each of the subcategories. Main rationale to not right now is that the extra manual steps might add room for manual error

As a follow-up, I want to actually use a recipe, but that's going to involved expanding recipe import logic to handle more python-based pipelines, so wanted to get this in main before working on that separately


products/factfinder/pipelines/acs_manual_update.py
```diff
@@ -111,11 +113,33 @@ def transform_all_dataframes(year):
     return combined_df
 
 
-def filter_by_metadata(df, year):
```
fvankrieken, line 114: @damonmcc do you know why this was being done before? Do they only want a subset of fields in AE? I'll follow up with Tim/Tyler too

fvankrieken:
If someone is already glancing at this, just updated the description a bit

products/factfinder/pipelines/acs_manual_update.py
fvankrieken: The new acs data is <10% larger than the last one, so this isn't hugely consequential in terms of number of fields

fvankrieken: Given that we're now generating metadata files from data dictionaries, I think it's fine not to. Though maybe should have some validation that metadata and df fields actually align?

damonmcc: yea filtering via metadata seems like overkill

damonmcc (approved):

PR #542, opened 2024-01-23
Data library - expand pydantic use beyond validation

fvankrieken:
Feel like I'm rearranging deck chairs on the titanic here slightly, but wanted a bit more ability to specify in a somewhat type-safe way more information in our library yml templates

Started on commit 3 and then decided that was far too much rearranging. So reverted for now so I could come back if I want, but will probably leave that. Leaving as food for thought but will rebase before merging.

Easiest to view the 2 commits separately just because of all the file changes in commit 1. But you can see there that the main changes to the functionality are
- no duplicated "name" field (name of dataset, name of destination - before, there was validation that these were the same). For script sources, assume script has same name as dataset
- no need for `version: {{version}}`
- for script entries, make subsection freeform dictionary. Most require a path field, but not all, so this seemed simplest. That means if a template requires a script but has no args, it should be denoted as `script: {}`. 

Commit 2, we have a few things going on
- use pydantic object instead of freeform dict throughout logic in ingest, config. Clean up logic to remove redundant calls
- Abolish config.yml in export - we already have config.json, and going in line with our decision for build_metadata exported in json, seemed to make sense to just keep json if we're just keeping one

### DIFF
--- a/dcpy/library/config.py
+++ b/dcpy/library/config.py
@@ -75,7 +75,7 @@ class Config:
             version = self.version_socrata(_config.source.socrata.uid)
         else:
             # backwards compatibility before templates were simplified
-            if _config.version == r"{{version}}":
+            if _config.version and _config.version.replace(" ", "") == r"{{version}}":
                 _config.version = None
             version = self.version or _config.version or self.version_today
         config = self.parsed_rendered_template(version=version)
--- a/dcpy/library/config.py
+++ b/dcpy/library/config.py
@@ -1,21 +1,15 @@
-import importlib
-import json
 from datetime import datetime
-import requests
-import yaml
-from jinja2 import Template
-import os
-from typing import Optional, Tuple
 from functools import cached_property
+from jinja2 import Template
+import importlib
+import os
+import pytz
+import requests
+import subprocess
+import yaml
 
-from .utils import format_url, get_execution_details
-from .validator import Validator
-
-
-# Custom dumper created for list indentation
-class Dumper(yaml.Dumper):
-    def increase_indent(self, flow=False, indentless=False):
-        return super(Dumper, self).increase_indent(flow, False)
+from .utils import format_url
+from .validator import Validator, Dataset
 
 
 class Config:
@@ -25,7 +19,7 @@ class Config:
     file to pass into the Ingestor
     """
 
-    def __init__(self, path: str, version: Optional[str] = None):
+    def __init__(self, path: str, version: str | None = None):
         self.path = path
         self.version = version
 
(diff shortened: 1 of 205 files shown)


dcpy/library/config.py
```diff
@@ -36,21 +30,16 @@ def unparsed_unrendered_template(self) -> str:
...
 
     @property
-    def parsed_unrendered_template(self) -> dict:
+    def parsed_unrendered_template(self) -> Dataset:
         """parsing unrendered template into a dictionary"""
-        return yaml.safe_load(self.unparsed_unrendered_template)
+        return Dataset(**yaml.safe_load(self.unparsed_unrendered_template)["dataset"])
 
-    def parsed_rendered_template(self, **kwargs) -> dict:
+    def parsed_rendered_template(self, version: str) -> Dataset:
```
fvankrieken, line 37: We never use other kwargs here. Debated getting rid of this entirely, but it's convenient when path requires version specified

dcpy/library/config.py
```diff
@@ -79,113 +64,86 @@ def version_today(self) -> str:
...
 
-    @property
-    def compute_yml(self) -> str:
-        return yaml.dump(
-            self.compute,
-            Dumper=Dumper,
-            default_flow_style=False,
-            sort_keys=False,
-            indent=2,
+def get_execution_details() -> dict[str, str]:
```
fvankrieken, line 117: Decided it made more sense to just have this here instead of utils. No logic change.

Debated making it structured, but it's just generated and dumped so left as dict for now

dcpy/library/validator.py
```diff
@@ -38,37 +38,35 @@ class Socrata(BaseModel):
...
-    dependents: Optional[List[str]] = None
-
-    class Config:
-        extra = Extra.allow
+    info: str | None = None
+    url: str | None = None
+    dependents: List[str] | None = None
 
 
 class Dataset(BaseModel):
```
alexrichey, line 62: Should we add `extra="forbid"` to the class, to disallow extra fields in the source yml?

dcpy/library/config.py
```diff
@@ -79,113 +64,87 @@ def version_today(self) -> str:
...
-                "geometry": geometry,
-            }
-
-        path = config["dataset"]["source"]["url"]["path"]
-        subpath = config["dataset"]["source"]["url"]["subpath"]
-        config["dataset"]["source"]["url"]["gdalpath"] = format_url(path, subpath)
-        config["execution_details"] = get_execution_details()
+                os.system(f'curl -o {path} "{url}"')
+            else:
+                raise Exception("Impossible")
```
alexrichey, line 99: Is it?

alexrichey (approved):
Great - this all makes sense. It's nice that we've got unit tests for this. I'm curious if you did any other testing?

dcpy/library/config.py
fvankrieken, line 99: @alexrichey saw you commented so going to respond - pydantic does validate that string belongs to one of the literals listed, and this is to appease mypy. I did mean to at least return a sensible error because this as written is slightly confusing.

fvankrieken:
> Great - this all makes sense. It's nice that we've got unit tests for this. I'm curious if you did any other testing?

I added a unit test to validate all template files with pydantic, going to do a bit of spot testing of datasets

dcpy/library/config.py
alexrichey: Thanks - yeah, I commented and then saw that we'd defined these in a literal, and assumed it was a `mypy` fix.

fvankrieken:
Alright, ingested all datasets (that don't require specific versions or local files) up to loading into gdal, and then spot tested a script and socrata source

fvankrieken:
@alexrichey - realized I also needed to update template files not in dcpy. I did, and also made sure the new code is backwards compatible (and added a test for it)

dcpy/library/utils.py
```diff
@@ -49,36 +47,6 @@ def format_url(path: str, subpath: str) -> str:
...
-            .decode()
-        )
-        return {
-            "type": "manual",
-            "user": git_user,
-            "timestamp": timestamp,
-        }
+def get_all_templates():
+    """Get names of all dataset templates included in dcpy"""
+    return [file.stem for file in TEMPLATE_DIR.glob("*")]
```
damonmcc, line 52: doesn't seem to be used, maybe was gonna be part of a new test?

PR #543, opened 2024-01-23
Colp packaging

alexrichey:
This is just a first, very rough, step to validate csv files against our metadata. Currently validation will just return a list of exceptions, e.g. for COLP:
```
[['COLUMN_MISMATCH', "Invalid column(s) found in source data: {'bOROUGH'}."],
 ['COLUMN_MISMATCH', "Column(s) missing from source data: {'BOROUGH'}."],
 ['INVALID_DATA',
  'Column BBL contains 17428 invalid records, for example: 3035880036.00000000'],
 ['INVALID_DATA',
  'Column MAPBBL contains 17428 invalid records, for example: 3035880036.00000000'],
 ['INVALID_DATA', 'Column CATEGORY has 2 empty values'],
 ['INVALID_DATA', 'Column EXPANDCAT has 54 empty values']]
```

This could certainly benefit from unit tests, and there's some deduplication that can happen in the validation function. But this works against COLP for now, and there are higher priority tasks. 

I'd certainly appreciate feedback on the metadata format itself. Any comments about field names, formats, etc. would be great to incorporate before we add metadata for other products. Also, we'd originally put some product specific metadata in `dcpy` but I'm not convinced it should live there, so for now I just put the metadata with the COLP product itself.

The next step is to add a `distributions` key to the COLP metadata, and use that update COLP on Socrata. I'll be working on that next.

Also, you can read commit-by-commit to get a sense of exactly what I changed in the metadata when I ported it from PDF to yml. First commit is a faithful representation, then following that are my modifications.

### DIFF
--- a/dcpy/metadata/validate.py
+++ b/dcpy/metadata/validate.py
@@ -1,5 +1,6 @@
 import pandas as pd
 from pathlib import Path
+import shapely
 import yaml
 
 
@@ -100,8 +101,25 @@ def validate_csv(csv_path: Path, metadata_path: Path):
                         ]
                     )
 
-            case "geometry":
-                pass
+            case "wkb":
+
+                def is_wkb_valid(g):
+                    try:
+                        shapely.wkb.loads(g)
+                        return True
+                    except Exception:
+                        return False
+
+                invalids = csv_df[~csv_df[col_name].apply(is_wkb_valid)]
+                if not invalids.empty:
+                    sample = invalids.iloc[0][col_name]
+                    errors.append(
+                        [
+                            "INVALID_DATA",
+                            f"The {col_name} column contains {len(invalids)} invalid wkb value(s), for example: {sample}",
+                        ]
+                    )
+
             case "text":
                 pass
             case _:
--- a/dcpy/metadata/validate.py
+++ b/dcpy/metadata/validate.py
@@ -1,0 +1,140 @@
+import pandas as pd
+from pathlib import Path
+import yaml
+
+
+def validate_csv(csv_path: Path, metadata_path: Path):
+    """Validate a csv against a metadata file."""
+    csv_df = pd.read_csv(csv_path, dtype=str)
+    csv_df_non_null = csv_df.fillna("")
+
+    metadata = yaml.safe_load(open(metadata_path, "r"))
+
+    errors = []
+
+    # Find mismatched columns
+    # Assumes that if `appears_in` isn't set, every filetype will contain the column
+    metadata_column_names = {
+        x["name"]
+        for x in metadata["columns"]
+        if not x.get("appears_in") or "csv" in x.get("appears_in")
+    }
+    csv_headers = set(csv_df.columns)
+
+    extras_in_csv = csv_headers.difference(metadata_column_names)
+    if extras_in_csv:
+        errors.append(
+            [
+                "COLUMN_MISMATCH",
+                f"Invalid column(s) found in source data: {extras_in_csv}.",
+            ]
+        )
+
+    not_found_in_csv = metadata_column_names.difference(csv_headers)
+    if not_found_in_csv:
+        errors.append(
+            [
+                "COLUMN_MISMATCH",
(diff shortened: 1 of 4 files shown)


dcpy/utils/s3.py
```diff
@@ -231,13 +231,13 @@ def download_folder(
...
     *,
-    include_prefix_in_export: bool = False,
+    include_prefix_in_export: bool = True,
 ) -> None:
     """Given bucket, prefix filter, and export path, download contents of folder from s3 recursively"""
     if prefix[-1] != "/":
         raise NotADirectoryError("prefix must be a folder path, ending with '/'")
     for obj in list_objects(bucket, prefix):
-        key = obj["Key"].replace(prefix, "") if include_prefix_in_export else obj["Key"]
+        key = obj["Key"] if include_prefix_in_export else obj["Key"].replace(prefix, "")
```
alexrichey, line 240: 🐛

dcpy/utils/s3.py
```diff
@@ -231,13 +231,13 @@ def download_folder(
     prefix: str,
     export_path: Path,
     *,
-    include_prefix_in_export: bool = False,
+    include_prefix_in_export: bool = True,
```
alexrichey, line 234: Swapping the default, so that nothing changes for existing callers. (No previous callers specified this kwarg)

products/colp/metadata.yml
```diff
@@ -0,0 +1,282 @@
...
+    display_name: BBL
+    description: >-
+      10-digit identifier for a tax lot, consisting of the borough code followed
+      by the tax block followed by the tax lot. The borough code is one numeric
+      digit. The tax block is one to five numeric digits, preceded with leading
+      zeros when the block is less than five digits. The tax lot is one to four
+      digits and is preceded with leading zeros when the lot is less than four
+      digits. For condominiums, this is usually the unit BBL.
+    data_type: bbl
+    data_readme_type: double
```
alexrichey, line 61: thoughts about this name? This is for cases where we have a concrete type that we can validate (e.g. a BBL), but what appears in the README currently is a less specific type, like `double`.

damonmcc: slight preference for `readme_data_type` but no big deal

if we aren't validating READMEs (not suggesting we do), do we need this field? maybe near future for generating READMEs?

products/colp/metadata.yml
damonmcc, line 1: nice! this would be used to validate tabular data, push to distributions like Socrata, and generate metadata files like a README or PDF?

and seems like it assumes there's a single primary table, which I kinda like since Socrata forces us to treat a product that way. I imagine we'll wanna validate other output tables/files but that seems like a lower priority

dcpy/metadata/validate.py
```diff
@@ -0,0 +1,158 @@
...
+    # Validate Data in Columns
+    for col in metadata["columns"]:
+        col_name = col["name"]
+        if col_name in not_found_in_csv:
+            continue
+
+        match col["data_type"]:
+            case "bbl":
+                with_invalid_bbl = csv_df[~csv_df["BBL"].str.match("^\d{10}$")]
+                if not with_invalid_bbl.empty:
```
damonmcc, line 52: could `with_invalid_bbl` be `invalids` like the other data types?

damonmcc (approved):
noted some things but LGTM!

noting that it's related to #351 for tracking

products/colp/metadata.yml
alexrichey: @damonmcc cool, yeah I agree. Will change in the followup PR

products/colp/metadata.yml
alexrichey: @damonmcc yep, all of those things

dcpy/metadata/validate.py
alexrichey: Sure could. Will change in followup

fvankrieken:
Sorry, this slipped from my attention today, but this looks great. Excited about where this is all going!

PR #546, opened 2024-01-24
improve FacDB source data archiving

damonmcc:
related to #461

## motivations

The Open Data page for `dycd_afterschoolprograms` is gone and the data seemed to always include more than just afterschool programs. DYCD's has a new dataset to replace it.

## changes

- use `edm-recipes/inbox` to stage and archive `usnps_parks`
- add `dycd_service_sites` to data library
- in FacDB replace use of `dycd_afterschoolprograms` with `dycd_service_sites`

## relevant action runs
- archiving of `usnps_parks` [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7659315759/job/20874175109) and sucessful re-run of FacDB on main [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7650122006)
- archiving of new `dycd_service_sites` [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7648080368) and run of FacDB on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7673219155/job/20915377146)
- run of FacDB named `dm-archive-for-facdb-better` [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7674394015) after using only the latest year's records from `dycd_service_sites`

## notes

- `dycd_afterschoolprograms` is no longer on Open Data
- `dycd_service_sites` is from Open Data [here](https://data.cityofnewyork.us/Social-Services/DYCD-Program-Sites/ebkm-iyma/about_data) and has a `FiscalYear` column with values of 2020, 2021, 2022. It's effectively the same as `dycd_afterschoolprograms` but appends every year of data
- `dycd_afterschoolprograms` vs DYCD Program Sites
  - 1,599 -> 9,392 rows
    - Program Sites has 2,875 rows in fiscal year 2022
  - 17 -> 30 columns
  - relevant category columns seem to be `program_type` -> `Service Category`
    - 18 -> 18 distinct values
    - values are so similar, the datasets seems to represent the same entities
  - `dycd_afterschoolprograms` seems to always have had more than just afterschool programs in it
- the FacDB pipeline code to pre-process DYCD data was investigated and adapted for the new data. the old address data was  new-line-separated values in a single column. see screenshots of jupyter notebook cells below
  - old data
    - image
  - new data
    - Screenshot 2024-01-26 at 4 26 41 PM


.github/workflows/build.yml
```diff
@@ -21,35 +21,31 @@ on:
...
-      build_note:
-        description: "A short note about this build (optional)"
-        type: string
-        required: false
       logging_level:
         description: "Logging level"
         type: choice
         required: true
-        default: WARNING
+        default: INFO
```
damonmcc, line 48: the default logging level in our python logging utils [here](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/utils/logging.py#L16) is `INFO`

fvankrieken:
Should have read a bit more - seems like significant enough changes to warrant a new dataset in edm-recipes

.github/workflows/build.yml
```diff
@@ -21,35 +21,31 @@ on:
...
+        description: "Recipe filename to use"
         type: choice
         options:
           - recipe
           - recipe-minor
         required: true
         default: recipe
       version:
-        description: "Version of dataset being built"
+        description: "Version of dataset being built (overrides version defined in recipe)"
```
fvankrieken, line 32: A note - there's no override for socrata sources, for those it always looks up "rowsUpdatedAt". Do you think we should have that manual override be possible?

dcpy/library/templates/dycd_service_sites.yml
```diff
@@ -0,0 +1,37 @@
+dataset:
+  name: &name dycd_service_sites
+  version: "{{ version }}"
```
fvankrieken, line 3: Don't need this line anymore.

Actually a few things with latest main are not needed
- no version if not actually specified
- no &name in `name` field
- no name field in destination (assumes its the `name` of the dataset. Before, there was validation that would throw an error if they didn't match, so I just got rid of it entirely)

products/facilities/facdb/datasets.yml
```diff
@@ -156,9 +156,9 @@ datasets:
     scripts:
       - dpr_parksproperties.sql
 
-  - name: dycd_afterschoolprograms
+  - name: dycd_service_sites
     scripts:
-      -  dycd_afterschoolprograms.sql
+      -  dycd_service_sites.sql
```
fvankrieken, line 161: From whatever happens with this file, would it be easy to scrap and just use the recipe instead? Just curious

products/facilities/facdb/pipelines.py
```diff
@@ -364,19 +364,14 @@ def dsny_electronicsdrop(df: pd.DataFrame):
...
-        lambda x: str(x).replace("nan", "").split("\n")[-1]
-    )
-    df["spatial"] = df.spatial.apply(lambda x: x.replace("(", "").replace(")", ""))
-    df["longitude"] = df.spatial.apply(lambda x: x.split(",")[-1])
-    df["latitude"] = df.spatial.apply(lambda x: x.split(",")[0])
+def dycd_service_sites(df: pd.DataFrame):
+    def _nan_to_string(value: object) -> str:
+        return str(value).replace("nan", "")
+
+    df["address"] = df["street_address"].apply(_nan_to_string)
```
fvankrieken, line 371: Some specific odd cases accounted for before - zipcode last 5 characters of first line for example. Did you do any validation that the imported dataset was good to go?

.github/workflows/build.yml
damonmcc: your talkin about setting a version when archiving source data? this is just the version of the product build so I'm not sure this action could/should do anything to versions of sources

I think it'd be nice to have the option to override. rowsUpdatedAt is a great default for socrata sources. but if data changes >once a day and we don't wanna overwrite an archive or start using timestamps in versions, an override in data_library_single.yml would be nice

fvankrieken: Oh oops. That is what I was referring to

products/facilities/facdb/pipelines.py
fvankrieken: Ah again should have read a little more closely. Cool

fvankrieken (approved):

products/facilities/facdb/datasets.yml
damonmcc: seems like FacDB has a sql script like this for every source. it standardizes column names, does some categorizing, and makes a geometry column from lat long columns. and at the end it does `CALL append_to_facdb_base('_dycd_service_sites');`

so scrapping this one would probably scrapping and combining em all

fvankrieken: Gotcha. Or logic to just call script based on dataset name without using this file?

damonmcc: oh did you mean scrapping the `datesets.yml` file? def agree this file isn't the best way to do this!

fvankrieken: Precisely!

damonmcc:
noting that I forgot to filter this new source data to only have the latest year when it's used in FacDB

would rather do that during the build that during the archive since multiple years of data might be useful in a different data product/analysis

PR #550, opened 2024-01-25
bug - use of unrendered template within config.compute

fvankrieken:
And now second commit - update existing open nightly qa issues on failure instead of opening new ones


damonmcc (approved):

fvankrieken:
Cleanup of #542

PR #552, opened 2024-01-26
Recipes for non-postgres destinations, tweaks for PFF DDHCA

fvankrieken:
Continuing on #541, making changes for #471

## Big Picture
A couple things here, but main changes are
- adding the concept of different destinations for imported datasets in builds. Typically, we have assumed that they're being loaded to a postgres instance being used as a build engine. Now, they can be
  - pg table (as before)
  - pd df (in memory)
  - file (still using xlsxs here for the moment sadly)
  where in python, the result of a load keeps track of the loaded datasets and their destinations
- using this logic within PFF for both ACS and Decennial manual updates

## Context
Pff is basically pivoting data from a bunch of inputs. In this case we have 
- ACS inputs - essentially 8 datasets, 1 table per "domain" per year, where we compare 2010 and the "latest" year (in this case, 2022). These generate 2 outputs, one csv per year
- Decennial inputs - 2 files for essentially 16 datasets, 1 table per domain per year, but just for last two decennial census years, 2010 and 2020, but also we have two separate files, the DHC and the "Detailed DHC-A" aka DDHCA.

Inputs are all very wide - one row per geography with many columns for statistics based on various variables. The outputs are in long format, with variable/geog as row, and 5 outputs (various stats of the same variable) as columns

## Notes for review
I went sort of two directions slightly with refactoring the acs/decennial manual update scripts using recipes. Decennial loads everything in memory at once, concatenates, then exports to csv. ACS exports to csv sequentially - thought mainly that it made sense to not keep a bunch of lingering dfs in memory if needed. Curious on people's thoughts, I should likely just pick one of those. Performance isn't a huge issue so maybe just the Decennial approach for readability.

There's room for reducing code a bit in terms of slightly redundant code in acs/decennial, but on that front would like to get this merged and get outputs to AE sooner rather than later. But still feedback welcome


dcpy/connectors/edm/recipes.py
```diff
@@ -60,7 +62,8 @@ def assign_file_type(self, file_type_preferences: list[DatasetType]) -> DatasetT
             raise FileNotFoundError(
                 f"No datasets of types {file_type_preferences} found in {self.s3_folder}"
             )
-        self.file_type = file_type
+        if mutate:
+            self.file_type = file_type
```
fvankrieken, line 72: Wanted this because in the case of something like reading a dataframe rather than flushing out a recipe, it makes sense to just get the file type rather than mutate an object

dcpy/builds/load.py
```diff
@@ -58,17 +102,65 @@ def load_source_data(
...
-    pg_client = postgres.PostgresClient(schema=build_name)
-    setup_build_environments(pg_client)
-
-    pg_client.create_table_from_csv(
-        recipe_lock_path.parent / "source_data_versions.csv"
+    if plan.InputDatasetDestination.postgres.value in [
+        dataset.destination for dataset in recipe.inputs.datasets
+    ]:
+        pg_client = postgres.PostgresClient(schema=build_name)
+        setup_build_environments(pg_client)
```
alexrichey, line 109: It looks like `setup_build_environments` also deletes the s3 draft folder for the product. Do we need to do that with other destinations? (In any case, I'm more in favor of just breaking that function up into its s3 and pg component parts)

fvankrieken: Ooh that's a good catch. Yes, I'll break up the logic in there

dcpy/builds/plan.py
```diff
@@ -26,18 +26,25 @@ class RecipeInputsVersionStrategy(str, Enum):
     copy_latest_release = "copy_latest_release"
 
 
-class DataPreprocessor(BaseModel, use_enum_values=True, extra="forbid"):
+class DataPreprocessor(BaseModel, extra="forbid"):
```
alexrichey, line 29: I can't quite recall why `use_enum_values=True` was used, but it might have been that otherwise, when we deserialize the recipe into a yaml file, it these values aren't output as strings, but python classes. In any case, curious if this was causing problems?

dcpy/connectors/edm/publishing.py
```diff
@@ -201,10 +201,11 @@ def legacy_upload(
         prefix = Path(publishing_folder)
     else:
         prefix = Path(publishing_folder) / s3_subpath
-    version_folder = prefix / version
-    key = version_folder / output.name
+    key = prefix / version
     meta = generate_metadata()
     if output.is_dir():
+        print(output)
```
alexrichey, line 207: ✂️✂️

dcpy/connectors/edm/recipes.py
```diff
@@ -47,7 +53,9 @@ def s3_folder(self) -> str:
     def s3_key(self) -> str:
         return f"{self.s3_folder}/{self.file_name}"
 
-    def assign_file_type(self, file_type_preferences: list[DatasetType]) -> DatasetType:
+    def assign_file_type(
```
alexrichey, line 56: This object's mutability feels a little bit awkward... I wonder if makes more sense for this dataset object to simply list the available file types that it finds on s3, instead of specifying one. (so we'd change `file_type` to be `file_types`, or a method called `get_file_types`). Then we can leave the assigning for the `plan.plan_recipe` function

dcpy/builds/plan.py
fvankrieken: Not quite sure why/how I took this one out to be honest. This class doesn't have any enums so it isn't useful, but it certainly wasn't doing any harm.

dcpy/connectors/edm/recipes.py
fvankrieken: That makes sense, I'll try something out

dcpy/builds/load.py
fvankrieken: Do we actually want to be deleting s3 artifacts here? Functionality in `edm-publishing` deletes folder when loading to s3. If my build errors out in the middle of a build but prior to loading, I'd prefer the previous build's output don't vanish. If you agree, I'm just going to scrap that bit of code

dcpy/utils/s3.py
```diff
@@ -106,7 +106,6 @@ def client() -> S3Client:
 def list_objects(bucket: str, prefix: str) -> list:
     """Lists all objects with given prefix within bucket"""
     objects: list = []
-    prefix = _folderize(prefix)
```
fvankrieken, line 109: I sort of get what I was going for here, but we have utils for things within a folder, list_objects really should be a prefix

dcpy/builds/load.py
alexrichey: Good question... It does seem really misplaced. Scrapping it seems good to me

fvankrieken:
To this point

> I went sort of two directions slightly with refactoring the acs/decennial manual update scripts using recipes. Decennial loads everything in memory at once, concatenates, then exports to csv. ACS exports to csv sequentially - thought mainly that it made sense to not keep a bunch of lingering dfs in memory if needed. Curious on people's thoughts, I should likely just pick one of those. Performance isn't a huge issue so maybe just the Decennial approach for readability.

Decennial runs out of memory when running locally, so maybe points to doing things incrementally. [In CI runs fine](https://github.com/NYCPlanning/data-engineering/actions/runs/7901535629)

dcpy/builds/plan.py
```diff
@@ -26,18 +26,25 @@ class RecipeInputsVersionStrategy(str, Enum):
...
+
+
 class InputDataset(BaseModel, use_enum_values=True, extra="forbid"):
     name: str
     version: str | None = None
     file_type: recipes.DatasetType | None = None
     version_env_var: str | None = None
     import_as: str | None = None
     preprocessor: DataPreprocessor | None = None
+    destination: InputDatasetDestination | None = None
```
alexrichey, line 47: All the nullability here makes me think we probably need to either 
1. implement a custom deserializer, or
3. have separate classes to differentiate between what we've serialized from yaml, and what we've actually resolved. So maybe this class should be called a InputDatasetRaw where everything is nullable, then we have another class which is stricter. That would allow us to remove the assertions you've added, I think?

Playing around with how this could work, and it could look something like this: 
```
class Strict(BaseModel):
    name: str
    version: str

class Lax(Strict):
    version: str | None = None # Only override this field, to loosen the constraint. 
    
l = Lax(name="hi", version="there") # used for deserializing
s = Strict() # used for resolved. Throws exception bc missing name and version
```
The inheritance direction feels backwards somehow. My gut says you'd want to subclass to make something _more_ specific (ie remove nullability) rather than open it up. But such is life in OOP-land.

alexrichey (approved):
Looks good! Had one thought about future direction for deserializing using Pydantic, but that's a longer term thing to play around with.

dcpy/builds/plan.py
fvankrieken: My gut is that if we went with separate classes, might be best to avoid the inheritance route, and just have them be fully distinct. So you'd only have a strict when expecting a strict, only have a lax when expecting a lax. Issue then is we need to go up the chain so to speak, have "lax" recipes and "computed"/strict recipes as separate classes or something like that. Not necessarily a bad thing, but a bit boilerplate-y. But might be worth it for these sorts of situations.

Isn't that a bit orthogonal from the custom deserializer though? Not entirely sure how you mean we would use it to get around the `| None`s

alexrichey: Re: the custom deserializer, I was thinking that a large part of why we allow nullability for these class attributes is that we omit values in the serialized data (ie the yaml files) then immediately compute them after we deserialize. e.g. Setting the filetype of a dataset object. So if we were able to hook into the deserializer, we could avoid those types of situations potentially. It's probably my less preferred option though.

fvankrieken: Ah I see what you mean. Yeah I'd prefer to not handle that sort of thing during deserialization, that seems like it'd get messy in a hurry

alexrichey:
> @alexrichey any specific thoughts on [this](https://github.com/NYCPlanning/data-engineering/pull/552#issuecomment-1944070251)? As in general logic flow in [decennial](https://github.com/NYCPlanning/data-engineering/blob/fvk-pff/products/factfinder/pipelines/acs_manual_update.py) vs [acs](https://github.com/NYCPlanning/data-engineering/blob/fvk-pff/products/factfinder/pipelines/acs_manual_update.py) (linking files rather than commits so the logic is easier to follow) with regards to keeping more in memory (load all, transform all, export all) vs doing things sequentially (load one, transform one, export one, repeat for all inputs). And the general code structure decisions made in those files

Sounds like the only option is to continually write to a file, or a series of files, right? Seems sensible, unless I'm missing something

PR #555, opened 2024-01-28
fix QA app deploy and page for Template DB

damonmcc:
## motivations

All QA pages have a warning to use `st.query_params` in place of experimental functions ([docs](https://docs.streamlit.io/library/api-reference/utilities/st.query_params)).

QA deploy action started to fail silently due to `no space left on device` when pulling the docker image ([example](https://github.com/NYCPlanning/data-engineering/actions/runs/7687967415/job/20948646427#step:4:57)). There were ~54 images in the DO droplet's storage (`docker info` [logs](https://github.com/NYCPlanning/data-engineering/actions/runs/7688144034/job/20949055232#step:4:33) and `docker images` [logs](https://github.com/NYCPlanning/data-engineering/actions/runs/7688275454/job/20949323912#step:4:76))

Template DB QA page fails to show show S3 file metadata and fails to show GeoDataFrames due to a MemoryError when converting the geometry column to a string. That conversion is needed to display the values using `st.dataframe`.

## actions runs

- Template DB QA page [here](https://de-qaqc.nycplanningdigital.com/?page=Template+DB)
- all runs of the QA App deploy action [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/qa_deploy.yml?query=branch%3Adm-qa-page-template-db)
- successful fail on error [example](https://github.com/NYCPlanning/data-engineering/actions/runs/7688221358/job/20949216878#step:4:105)
- successful deploy [example](https://github.com/NYCPlanning/data-engineering/actions/runs/7688447071/job/20949708589)
### old deprecation warning

Screenshot 2024-01-28 at 11 16 18 AM

## now showing GeoDataFrame properly

Screenshot 2024-01-28 at 11 19 53 PM

### DIFF
--- a/dcpy/utils/s3.py
+++ b/dcpy/utils/s3.py
@@ -16,7 +16,7 @@ from rich.progress import (
 )
 import typer
 from typing import Any, Literal, TYPE_CHECKING, cast, get_args
-from dataclasses import dataclass
+from pydantic import BaseModel
 
 from dcpy.utils import git
 from dcpy.utils.logging import logger
@@ -38,8 +38,7 @@ ACL = Literal[
 MAX_FILE_COUNT = 50
 
 
-@dataclass
-class Metadata:
+class Metadata(BaseModel):
     last_modified: datetime
     content_length: int
     content_type: str
--- a/dcpy/utils/s3.py
+++ b/dcpy/utils/s3.py
@@ -45,7 +45,7 @@ class Metadata(BaseModel):
     custom: dict[str, Any]
 
 
-def generate_metadata():
+def generate_metadata() -> dict[str, str]:
     metadata = {
         "date_created": datetime.now(pytz.timezone("America/New_York")).strftime(
             "%Y-%m-%dT%H:%M:%S%z"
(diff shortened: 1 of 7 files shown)


dcpy/utils/s3.py
```diff
@@ -38,15 +38,14 @@
 MAX_FILE_COUNT = 50
 
 
-@dataclass
-class Metadata:
+class Metadata(BaseModel):
```
fvankrieken, line 41: Why the change from dataclass to pydantic basemodel?

fvankrieken:
Ooh good digging on the deploy failing

fvankrieken (approved):
One very minor question but lgtm!

dcpy/utils/s3.py
damonmcc: I was trying to figure out how to go from Metadata to dictionary and for some reason things like `dataclass.asdict()`  wasn't working

so I changed it to a BaseModel to use`.model_dump()`. seems like we're using BaseModel in a lot of other places too

fvankrieken: Yeah I at times wonder if it'd just be cleaner to just use pydantic, even if it's overkill for some of our cases.

fvankrieken: muddying the water there is also a `dataclasses` submodule of pydantic

damonmcc:
> Ooh good digging on the deploy failing

I was so confused when the pages weren't changing after deploying

PR #557, opened 2024-01-29
fix failing Nightly QA tests and builds

### LINKED ISSUE #554: Scheduled Action Failure - ⚙️ Nightly QA
[Failed action](https://github.com/NYCPlanning/data-engineering/actions/runs/7691432145)

damonmcc:
resolves #554

## changes
- for all Template DB source tables, test their key and geometry columns
  - These tests error [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7700891161/job/20985754617#step:10:109) (rather than pass or fail) because expected column names can't be found. This happens before any product sql queries are run (e.g. staging.sql) and seems like a preferable way for a build to react to the problem.
- return pgdump to its former glory as the preferred source data file format

## relevant builds on this branch
- all products [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7701094151)
- Template DB in tests [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7701912670/job/20989058328?pr=557)

### TDB source versions before changes
Screenshot 2024-01-29 at 1 50 43 PM

### TDB source versions after changes
Screenshot 2024-01-29 at 1 58 03 PM

### DIFF
--- a/dcpy/builds/plan.py
+++ b/dcpy/builds/plan.py
@@ -15,8 +15,8 @@ from dcpy.connectors.edm import recipes, publishing
 DEFAULT_RECIPE = "recipe.yml"
 LIBRARY_DEFAULT_PATH = recipes.LIBRARY_DEFAULT_PATH
 RECIPE_FILE_TYPE_PREFERENCE = [
-    recipes.DatasetType.parquet,
     recipes.DatasetType.pg_dump,
+    recipes.DatasetType.parquet,
     recipes.DatasetType.csv,
 ]
 
--- a/products/template/sql/_sources.yml
+++ b/products/template/sql/_sources.yml
@@ -4,6 +4,45 @@ sources:
   - name: tdb_sources
     schema: "{{ env_var('BUILD_ENGINE_SCHEMA') }}"
     tables:
+      - name: nypl_libraries
+        columns:
+          - name: name
+            tests:
+              - unique
+              - not_null
+          - name: wkb_geometry
+            tests:
+              - not_null
+
+      - name: bpl_libraries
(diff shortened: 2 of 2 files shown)


dcpy/builds/plan.py
```diff
@@ -15,8 +15,8 @@
 DEFAULT_RECIPE = "recipe.yml"
 LIBRARY_DEFAULT_PATH = recipes.LIBRARY_DEFAULT_PATH
 RECIPE_FILE_TYPE_PREFERENCE = [
-    recipes.DatasetType.parquet,
     recipes.DatasetType.pg_dump,
+    recipes.DatasetType.parquet,
```
fvankrieken, line 19: No :(

products/template/sql/_sources.yml
damonmcc, line 1: these tests effectively ensure that the listed columns exist in the loaded source data tables

other than testing that all values in the key column are not null and unique, this tests that all values in the column `wkb_geometry` are not null. when there are exceptions, the tests for that source reflect that

an alternative is to use the dbt contract feature ([docs](https://docs.getdbt.com/reference/resource-configs/contract)), but that requires listing every column and it's data type. that'd be a lot to do here when we know `wkb_geometry` is the issue and may not be ideal when we often only care about a fraction of all columns

dcpy/builds/plan.py
damonmcc: lol ahead of your time!

sf-dcp (approved):

sf-dcp:
@damonmcc made me merge.

PR #559, opened 2024-01-29
include source data file type in build metadata

damonmcc:
related to #544 

When we plan a recipe, we resolve the versions and file types of source data. While investigating source data changes, it's helpful to know what file type was used for each.

The failing test is not due to these changes and was the motivation for them.

## screenshots
Screenshot 2024-01-29 at 12 49 29 PM

### DIFF
--- a/dcpy/builds/plan.py
+++ b/dcpy/builds/plan.py
@@ -14,6 +14,11 @@ from dcpy.connectors.edm import recipes, publishing
 
 DEFAULT_RECIPE = "recipe.yml"
 LIBRARY_DEFAULT_PATH = recipes.LIBRARY_DEFAULT_PATH
+RECIPE_FILE_TYPE_PREFERENCE = [
+    recipes.DatasetType.parquet,
+    recipes.DatasetType.pg_dump,
+    recipes.DatasetType.csv,
+]
 
 
 class RecipeInputsVersionStrategy(str, Enum):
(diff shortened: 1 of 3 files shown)


dcpy/builds/plan.py
```diff
@@ -165,13 +170,19 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
         if ds.version == "latest":
             ds.version = recipes.get_config(ds.name, "latest")["dataset"]["version"]
 
+    # Determine the recipe file type
+    for ds in recipe.inputs.datasets:
+        if ds.dataset.file_type is None:
+            ds.file_type = ds.dataset.assign_file_type(RECIPE_FILE_TYPE_PREFERENCE)
+        ds.file_type = ds.dataset.file_type
+
```
fvankrieken, line 178: This is much more sensible than when I was doing it, thank you for this

dcpy/connectors/edm/recipes.py
```diff
@@ -130,12 +130,9 @@ def import_dataset(
     preprocessor: Callable[[str, pd.DataFrame], pd.DataFrame] | None = None,
 ):
     """Import a recipe to local data library folder and build engine."""
-    if ds.file_type is None:
```
fvankrieken, line 133: Let's maybe throw an error in this case now?

fvankrieken (approved):
One little potential tweak but it's minor, lgtm

dcpy/connectors/edm/recipes.py
damonmcc: true, I added it to `import_dataset` where we already throw an error if version is unresolved

fvankrieken (approved):

PR #560, opened 2024-01-29
Source data qa

fvankrieken:


.devcontainer/docker-compose.yml
sf-dcp, line 1: did you deploy the app locally?

fvankrieken: What do you mean by that?

dcpy/connectors/edm/recipes.py
```diff
@@ -248,3 +305,62 @@ def purge_recipe_cache():
...
+    metadata = [
+        get_metadata(dataset, v).to_row()
+        for v in versions
+        #    if v not in ["testor", "2023_executive_points", "2017_adopted_points", "2017_adopted_polygons"]
+        #    and not (dataset == "doitt_buildingfootprints" and v == "20230414")
+        #    and not (dataset == "test_nypl_libraries")
+    ]
+    metadata_df = pd.DataFrame.from_records(metadata)
+    metadata_df["event_source"] = "scrape"
+    pg_client.insert_dataframe(
```
fvankrieken, line 348: As you can see, went a very lightweight approach for now of "migrations" - just insert df, append to existing table. So pandas.to_sql creates table if it doesn't exist

dcpy/utils/logging.py
sf-dcp, line 1: Looks nice. I'm not sure about the code placement though. I would think that logging file would consist of logging module stuff. What about a separate file `metadata.py`?

fvankrieken: Yeah I'm with you on that

apps/qa/bash/docker_build_and_publish.sh
sf-dcp, line 1: Could you give an example when you would need a recipe? Also, curious if there is a risk if/when recipes in container falling out of sync from the product dir ones.

fvankrieken: That's where the qa app is getting the list of source datasets per data product for this "source data" page

dcpy/utils/logging.py
fvankrieken: It's been moved now

apps/qa/bash/docker_build_and_publish.sh
```diff
@@ -9,6 +9,12 @@ set_error_traps
 
 cp $ROOT_DIR/pyproject.toml $PROJECT_DIR
 cp $ROOT_DIR/python/constraints.txt $PROJECT_DIR
+
+mkdir $PROJECT_DIR/products
+for dataset in "cbbr" "colp" "cpdb" "developments" "facilities" "pluto" "zoningtaxlots"; do
```
fvankrieken, line 14: Per @sf-dcp 's question, this is a bit hacky. I'll add a comment in the code explaining why it's currently like this

sf-dcp:
Looking at the `metadata_logging` table in postgres, there are values `NULL`, `Manual`, `could not parse`. Do you know in what case(s) we get one of these values?

fvankrieken:
> Looking at the `metadata_logging` table in postgres, there are values `NULL`, `Manual`, `could not parse`. Do you know in what case(s) we get one of these values?

- NULL are before any metadata like this was included in the config.json/yml files in edm-recipes. 
- Manual means run locally, if there's no name after it that means `git` command line call did not return a username. 
- Could not parse is when something went wrong generating this info - not sure if maybe this happened when the docker image was still getting used

apps/qa/bash/docker_build_and_publish.sh
damonmcc: another way that comes to mind is moving the product recipe.yml files into dcpy, but that doesn't seem worth the trouble yet

.devcontainer/docker-compose.yml
sf-dcp: Sorry, I did see you redeploying the app on your brunch. I'm just confused by this change in the dev container.

fvankrieken: Gotcha. It allows you to view the app if you're running it from within the dev container!

fvankrieken: It used to be included in this file, confusingly in the `db` section. Hence why the app worked locally in the past. But when I dropped the db a little while back, this port-forwarding was dropped too, so this is just re-adding it back in

dcpy/test/utils/test_metadata.py
damonmcc, line 1: ❤️

dcpy/connectors/edm/recipes.py
```diff
@@ -173,3 +290,80 @@ def purge_recipe_cache():
...
+        logger.warn(f"Could not delete from table: {e}")
+
+    ## scrape metadata, insert into table
+    versions = get_all_versions(dataset)
+    metadata = [
+        get_metadata(dataset, v).to_row()
+        for v in versions
+        #    if v not in ["testor", "2023_executive_points", "2017_adopted_points", "2017_adopted_polygons"]
+        #    and not (dataset == "doitt_buildingfootprints" and v == "20230414")
+        #    and not (dataset == "test_nypl_libraries")
```
damonmcc, line 336: ✂️?

fvankrieken:
Per discussion with @sf-dcp , I should write some tests for connectors.edm.recipes as part of this PR

dcpy/connectors/edm/recipes.py
damonmcc: or trying to keep this around for later resolution of the datasets that didn't work?

fvankrieken:
Issues might come up from multiple loading of multiple file formats - need to check how I'm logging in commit 6

damonmcc:
> Issues might come up from multiple loading of multiple file formats - need to check how I'm logging in commit 6

maybe a filename column? so `(name, version, filename, timestamp, runner, event_source)`

fvankrieken:
> > Issues might come up from multiple loading of multiple file formats - need to check how I'm logging in commit 6
> 
> maybe a filename column? so `(name, version, filename, timestamp, runner, event_source)`

I don't think we want to see those as separate events though, I'd prefer that one gha run of data library is one row in this table

sf-dcp:
> > > Issues might come up from multiple loading of multiple file formats - need to check how I'm logging in commit 6
> > 
> > 
> > maybe a filename column? so `(name, version, filename, timestamp, runner, event_source)`
> 
> I don't think we want to see those as separate events though, I'd prefer that one gha run of data library is one row in this table

what about a binary col for each data format? or a column containing a list of data formats

damonmcc:
I like a list over binary columns. so a column of type array?

dcpy/library/archive.py
```diff
@@ -119,19 +119,21 @@ def __call__(
...
-                    # from the version of the file currently getting added to latest
+                diff_version = [
+                    k
+                    for k in keys_in_latest
+                    if self.s3.info(k)["Metadata"].get("version", "") != version
+                ]
+
+                # Remove keys from the latest directory that have versions different
+                # from the version of the file currently getting added to latest
+                if len(diff_version) > 0:
```
fvankrieken, line 123: I feel like this hasn't been working? Haven't we had differing versions in latest folders

fvankrieken:
This has been open/tinkered with long enough that a fresh full review would be good. Commit-by-commit will certainly make the most sense

dcpy/connectors/edm/recipes.py
```diff
@@ -271,3 +324,73 @@ def purge_recipe_cache():
...
+
+    logged_metadata = get_logged_metadata([dataset])
+    logged_versions = {m["version"] for _, m in logged_metadata.iterrows()}
+
+    ## scrape metadata, insert into table
+    versions = get_all_versions(dataset)
+    metadata = [
+        get_archival_metadata(dataset, v).to_row()
+        for v in versions
+        if v not in logged_versions
```
fvankrieken, line 389: Rationale for this being if we archive v 20240101 on 20240102, and then re-archive it on 20240103, edm-recipes only gives us the latest. But both events should now be logged when actually running `library` (see logging commit). If events are logged by data library and already in the table, they don't need to be re-scraped

Theoretically, after merging this and scraping all the data once, we won't need to do it ever again (but we know that's not true)

dcpy/utils/s3.py
```diff
@@ -354,7 +352,7 @@ def get_filenames(bucket: str, prefix: str) -> set[str]:
 def get_subfolders(bucket: str, prefix: str, index=1) -> list[str]:
     prefix = _folderize(prefix)
     prefix_path = Path(prefix)
-    subfolders = set()
+    subfolders: set[str] = set()
```
alexrichey, line 355: surprised we're not getting a dangerous default here, since sets are mutable 🤷‍♂️

alexrichey (approved):
good stuff!

PR #562, opened 2024-01-29
Ar colp socrata md

alexrichey:
This wraps up the initial phase of distribution, namely validating datasets and distributing to socrata. Adds pydantic models, metadata for templatedb, and an initial unit test for column calculations for various datasets and destinations.


alexrichey:
@damonmcc Would love some input on naming in the metadata. It might be useful to do a quick review / demo tomorrow morning beforefand?

products/colp/metadata.yml
```diff
@@ -8,19 +8,51 @@ description: |
   Questions or feedback on errors within the City Owned and Leased Properties can be directed to DCAS Land Use Planning at (212) 386-0622 or RESPlanning311@dcas.nyc.gov.
 
 tags: [COLP, IPIS, city owned, leased, property, Department of Citywide Administrative Services, DCAS, Department of City Planning, DCP, New York City, NYC]
+each_row_is_a: City Owned or Leased Property
+
+distributions:
```
damonmcc, line 13: is there a distribution that the csv will be going to?

products/colp/metadata.yml
```diff
@@ -33,22 +65,39 @@ columns:
       - ["3", Brooklyn]
       - ["4", Queens]
       - ["5", Staten Island]
+    format_overrides:
+      csv:
+        name: BOROUGH
+      socrata:
+        display_name: BOROUGH
```
damonmcc, line 72: these seem like a mix of format and distribution overrides. if it were format only I think it'd be `csv:` and `shapefile:`

definitely seems reasonable to be thinking about destinations though so I'm not sure which would be a better way to frame these

products/colp/metadata.yml
damonmcc: bytes, but we aren't implementing distribution to it yet

products/colp/metadata.yml
damonmcc: per pairing: reworking structure of overrides

alexrichey:
@damonmcc I'm fixing one or two minor issues (e.g. adding faker to the requirements) but this is otherwise good to go

damonmcc (approved):
LGTM but the new tests aren't running because the python package `faker` isn't found. it may just need to be added to the list `dependencies = [...]` in `pyproject.toml`

dcpy/metadata/models.py
```diff
@@ -0,0 +1,127 @@
...
+    description: str
+    data_type: str
+
+    # Optional
+    data_source: str | None = None
+    example: Any | None = None
+    is_nullable: bool = True
+    is_primary_key: bool = False
+    readme_data_type: str | None = None
+    values: list[tuple] = []
```
fvankrieken, line 54: Is mutable default args an issue in pydantic models?

fvankrieken: Ignore me, it's fine (thank goodness)

alexrichey:
@damonmcc Yeah, having a bit of trouble re(setting) up my dev containers to recompile the requirements file. But after that, should be all fixed

fvankrieken:
> @damonmcc Yeah, having a bit of trouble re(setting) up my dev containers to recompile the requirements file. But after that, should be all fixed

Per @damonmcc 's comment, you should still add it to the dependencies for dcpy in pyproject.toml

dcpy/metadata/models.py
alexrichey: Thankfully not!

Issue #547, opened 2024-01-24
KPDB aggregations have duplicates and ambiguous columns


damonmcc:
see GIS comment [here](https://github.com/NYCPlanning/edm-overview/issues/1009#issuecomment-1908636591)

damonmcc:
have added new boolean columns named `has_project_phasing` and `has_future_units` to help identify records that should be use for housing projections

also added an ouput table/csv called `summary_record_phasing` to show counts of records with future units grouped by the record sources and status that inform the "Phasing Assumptions" that went into the design of KPDB

damonmcc:
had added columns like `units_net_in_nta` and `within_5_yearts_in_nta` so that we calculate phased units in aggregations. if a record is split between multiple geographies, the phased units must be calculated for each geography

damonmcc:
shared latest outputs with GIS today

PR #565, opened 2024-01-30
start testing KPDB builds

### LINKED ISSUE #518: model KPDB output tables
## motivation

KPDB output tables are relatively complex and often have duplicate values. In the main output table it's hard to tell if/which duplicates are either valid or erroneous and should be corrected. Aggregation tables seems to be duplicating the duplicates in the main table. There are many types of "entities" in KPDB which makes it hard to talk about the data (Clusters, Projects, DOB Jobs, etc.).

The [KPDB wiki](https://github.com/NYCPlanning/data-engineering/wiki/Product:-KPDB) is out-of-date and isn't a data contract that new builds can be easily checked against

## ideas

We recently added `dbt` to model and test Template DB tables (example [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/template/sql/product/_product_models.yml)). The same could be done for KPDB to describe and test the most importable build outputs.

damonmcc:
resolves #518, related to #547

This adds data tests for the primary KPDB table using `dbt`. It also starts to test some of the aggregation tables. These could be considered loose data contracts ([docs](https://docs.getdbt.com/docs/collaborate/govern/model-contracts)).

Many of the filename and directory changes are because `dbt` expects the configuration yml files to be in subdirectories ([docs](https://docs.getdbt.com/reference/model-properties)) and expects to find sql filenames matching the names of tables defined in config files.

I'll follow up on this with more tests to investigate and fix issues with KPDB data found by GIS.

## changes
- use latest KPDB corrections
- ensure aggregations use final KPDB table
- move and consolidate aggregation-related scripts
- create primary KPDB table before aggregations
- add and use dbt to test high priority KPDB tables

## results of changes

builds using this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-kpdb-for-gis)

latest example of testing primary KPDB table [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7744371614/job/21117967250#step:10:61) and aggregation tables [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7745995963/job/21123203139#step:12:58)


products/knownprojects/bash/01_dataloading.sh
```diff
@@ -6,7 +6,7 @@ max_bg_procs 5
 if [ -n "${BUILD_ENGINE_SCHEMA}" ]; then
     echo "Dropping and creating build schema '$BUILD_ENGINE_SCHEMA'"
     run_sql_command "DROP SCHEMA IF EXISTS ${BUILD_ENGINE_SCHEMA} CASCADE;"
-    # run_sql_command "VACUUM (FULL);"
+    run_sql_command "VACUUM (ANALYZE);"
```
damonmcc, line 9: build time was gradually increasing when building frequently and I think this is helping keep it low and consistent

products/knownprojects/kpdb.sh
damonmcc, line 1: this isn't used anymore, we just call each build stage in the github workflow file

pyproject.toml
```diff
@@ -100,6 +101,7 @@ dependencies = [
     "beautifulsoup4",
     "boto3",
     "botocore",
+    "dbt-postgres",
```
damonmcc, line 104: this is needed because the KPDB build uses the `nycplanning/build-base` image and runs `python3 -m pip install . -c ./python/constraints.txt`. therefore this list of packages are the only ones installed

this wasn't needed to use dbt in Template DB beacuse it uses the `nycplanning/dev` image which has all packges listed in requirements.txt

products/knownprojects/kpdb.sh
damonmcc, line 1: this isn't used anymore, we just call each build stage in the github workflow file

pyproject.toml
fvankrieken: build-base has its own requirements file in `docker/build-base`, this spot is for dependencies for dcpy. This also should be added to the main requirements.in if it's not!

products/knownprojects/sql/aggregate/sca/boundaries_school_subdistricts.sql
```diff
@@ -338,7 +338,7 @@ FROM (
         b.proportion_in_subdist_1 AS proportion_in_subdist,
         round(a.units_net * b.proportion_in_subdist_1) AS units_net_in_subdist
     FROM
-        _kpdb AS a
+        kpdb AS a
```
fvankrieken, line 341: Huh. so seems like these aggregate tables were using not the final table before as their input?

fvankrieken:
This looks great! Few small notes, main thing is getting dbt-postgres in `build-base/requirements.txt` instead of dcpy's dependencies, but other than that lgtm

products/knownprojects/profiles.yml
```diff
@@ -0,0 +1,14 @@
...
+
+dcp-de-postgres:
+  target: dev
+  outputs:
+    dev:
+      type: postgres
+      host: "{{ env_var('BUILD_ENGINE_HOST') }}"
+      user: "{{ env_var('BUILD_ENGINE_USER') }}"
+      password: "{{ env_var('BUILD_ENGINE_PASSWORD') }}"
+      port: "{{ env_var('BUILD_ENGINE_PORT') | as_number }}"
```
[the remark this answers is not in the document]
damonmcc: i like newlines too, will do

and this is needed because dbt can't use the full constructed DB URL. if you meant the weird `as_number` for the port variable, without that this resolves to `port: "1234"` which fails rather than `port: 1234` ([docs](https://docs.getdbt.com/reference/dbt-jinja-functions/profiles-yml-context))

products/knownprojects/sql/product/_product_models.yml
```diff
@@ -0,0 +1,53 @@
...
+    columns:
+      - name: project_id
+      - name: source
+      - name: record_id
+        tests:
+          - not_null
+          - unique:
+              config:
+                severity: error
+                error_if: ">250"
```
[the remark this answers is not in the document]
damonmcc: just that the Housing team's latest round of QA was ok with the current amount of duplicate records (labor intensive to keep fixing)

but these thresholds are definitely the types of things we'd wanna review with and have set by product owners/SMEs as we use these more

fvankrieken (approved):

PR #568, opened 2024-02-02
improve KPDB units phasing and aggregations

### LINKED ISSUE #547: KPDB aggregations have duplicates and ambiguous columns

damonmcc:
resolves #547 

## vocab

- "units": housing units
- "gross units": units associated with the record
- "net units": units that haven't been accounted for by another related record
- "phased units": net units that have completion predictions spread out over specific time periods (see [Phasing Assumptions](https://github.com/NYCPlanning/data-engineering/wiki/Product:-KPDB#phasing-assumptions))

## significant changes

- add new columns related to a record's units: `has_future_units`, `has_project_phasing`, `future_units_without_phasing`
- fix +/-1 rounding error in units phasing calculations
- generate a deduplicated version of the primary output table for use in aggregation queries
  - choosing to create a new table rather then deduplicate the primary one. these duplicates are used by the Housing team during QA to inform corrections
- add phased unit calculations to aggregations
- generate actual aggregations of future units which have one row per geographic boundary
  - rather than just the "longform" tables which have one row per portion of a KPDB record in a geographic boundary

## action runs

All builds using this branch are [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-kpdb-debug)


fvankrieken:
Per the discussion today, what really are net units again?

.github/workflows/test_helper.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           # TODO: Gradually expand this to cover more products
           sqlfluff lint products/template/
           sqlfluff lint products/cpdb/sql/
-          sqlfluff lint products/knownprojects/
+          sqlfluff lint products/knownprojects/ --templater=jinja
```
fvankrieken, line 31: Would you want to wait till I get dbt templater merged, and you can use that? Or I can try to switch it over in my branch. Either way!

products/knownprojects/sql/aggregate/_aggregate_longform_models.yml
```diff
@@ -0,0 +1,54 @@
+version: 2
+
+models:
+  - name: longform_nta_output
+    description: "Project records split by intersecting NTAs"
```
fvankrieken, line 5: Are these accurate descriptions? I'm not quite sure what this means

products/knownprojects/sql/aggregate/longform_cd_output.sql
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************************************************************************************************************************
-Sources kpdb - finalized version of KPDB build
+Sources kpdb_deduplicated - finalized version of KPDB build with no duplicatre record_id values
```
fvankrieken, line 2: Nitty nit but there are many `duplicatre`s in these comments (similar vibe to socrate actually)

damonmcc: and `survery`! my middle finger is too trigger happy

table descriptions in sql files might be non-ideal anyway if they can be described in a yml. will fix

.github/workflows/test_helper.yml
damonmcc: I'm down to wait

damonmcc:
@fvankrieken 
> Per the discussion today, what really are net units again?

good call! I just revised the PR description.

I believe the type of record determines how we would describe net units ( like "completed" vs "permitted for" construction) I'll revise the "future units" concept in this PR to pull back from my assumption that everything without phasing can be described as having future units

products/knownprojects/sql/aggregate/_aggregate_longform_models.yml
damonmcc: yup their accurate. I'll revise them for clarity, so maybe something like `Project records with a row for each NTA it intersects with`

expand/explode/unnest are relevant concepts but an unnest isn't used in the queries and they don't seem to work well in a table description for non-SQL ppl

fvankrieken: Ah gotcha. Thanks, that description is clearer!

PR #574, opened 2024-02-07
improve KPDB testing and dbt test schemas

damonmcc:
related to #547 

These changes have been pulled out of PR #568 to simplify it and let it be more about fixing data issues

The motivations to  delete a schema that was created by running dbt tests are
- dbt provides no way to delete it when new tests are run, it must be manually deleted. incremental building and testing is usually how it's used so I'm not shocked
- If we don't delete them, tables with inaccurate test results and with names of since-deleted tests will exists and be confusing
- if we don't consider these schemas during use of `delete_stale_schemas()`, we'll inadvertently delete them while keeping their build schemas. so we'd lose test results for builds we're trying to keep around

## relevant action runs
- logs of a KPDB build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7817924478/job/21326972771#step:8:23) where it would delete the tests schema during data loading


sf-dcp:
From high level, it sounds like you previously had dbt tests creating a schema with test result tables in db.

* This PR addresses rebuilding on the same branch, correct? In other words, the problem you are addressing exists when you previously built a schema with the same name (i.e. branch name). 
* Does dbt append to existing tables instead of overwriting them completely? In your second bullet point, you are mentioning incorrect results and test names - can you show an example? 
* Not sure I understand your last bullet point. 

@damonmcc

damonmcc:
@sf-dcp 

> From high level, it sounds like you previously had dbt tests creating a schema with test result tables in db.

yup! a dbt test always creates a table in a schema ([docs](https://docs.getdbt.com/docs/build/data-tests#:~:text=Test%20result%20tables%20are%20created%20in%20a%20schema)). In our case, those schemas are `BUILD_SCHEMA__tests` due to configurations like [this](https://github.com/NYCPlanning/data-engineering/blob/main/products/knownprojects/dbt_project.yml#L9).

> * This PR addresses rebuilding on the same branch, correct? In other words, the problem you are addressing exists when you previously built a schema with the same name (i.e. branch name).

yes, as well as nightly QA runs and Template DB test builds in PRs

> * Does dbt append to existing tables instead of overwriting them completely? In your second bullet point, you are mentioning incorrect results and test names - can you show an example?

It overwrites them completely. So the incorrectness I'm refereeing to is really caused by tables that were created by tests that have been renamed or deleted.

So if we write and run `test_primary_key`, we'll have a table named `test_primary_key`. And then if we rename it to `test_compound_key` and run, we'll have a new table called `test_compound_key`.

Now we have two tables. One of them, `test_primary_key`, is an artifact of code that has changed and so the table shouldn't exist anymore since it doesn't reflect the current state of a product's tests.

Or in another example, a test that was failing was called `test_important_fact`. We renamed it and it's passing, but the table `test_important_fact` still exists and has records in it because it had been failing. If we looked in the DB, it'd seem like something's wrong with a build even though no tests failed during it.

> * Not sure I understand your last bullet point.

We currently delete "stale" DB schemas using [this](https://github.com/NYCPlanning/data-engineering/actions/workflows/stale_builds.yml) github action. It currently deletes all schemas that don't match the names of branches in the repo. So it's gonna delete schemas like `kpdb.in_progress__tests` even when we have branch `in_progress` still open and want to keep all relevant schemas.

products/knownprojects/bash/03_aggregate.sh
```diff
@@ -28,10 +28,10 @@ echo "SCA sggregations are complete"
 ## Do other aggregation
 # Aggregate KPDB projects to Community Districts
```
sf-dcp, line 29: It seems like the code below creates different level aggregations at census tract, community district, and nta. If that's true, I would revise the comment above and the echo cmd below

products/knownprojects/dbt_project.yml
```diff
@@ -7,3 +7,6 @@ model-paths: ["sql"]
 tests:
   +store_failures: true
   schema: "_tests"
+
+vars:
+  max_duplicate_projects: 250
```
sf-dcp, line 12: just curious, does this variable represent unique count of records that have dupes or the total count of records? For example, I have 2 unique id records; each id has 4 more dupes. So the total is 10 records with 2 unique ids...

products/knownprojects/sql/aggregate/_aggregate_models.yml
```diff
@@ -32,5 +32,39 @@ models:
             - cdta
           config:
             severity: error
-            error_if: ">250"
+            error_if: ">{{ var('max_duplicate_projects') }}"
```
sf-dcp, line 35: when a condition for error is true, does it stop the pipeline?

products/knownprojects/bash/03_aggregate.sh
damonmcc: revised!

products/knownprojects/dbt_project.yml
damonmcc: I believe it's unique count of records that have dupes

and I think the variable name should be changed to reflect it's used to test "project records" rather than "projects". I'll change that

damonmcc: I shouldn't put a screenshot of the data here but an example of the outputs is

| unique_field | n_records |
|--------|--------|
| a_123 | 2 |
| b_123 | 50 |
| c_123 | 3 | 

and this would contribute 3 toward that threshold of 250 duplicates. a recent run of this [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7835217216/job/21380150443#step:10:45) showed 65 and the relevant table has 65 rows like those above

products/knownprojects/sql/aggregate/_aggregate_models.yml
damonmcc: yup, but we can always set any or all tests to warn instead of error if we don't think they should stop a build

dcpy/builds/metadata.py
```diff
@@ -39,6 +39,12 @@ def build_name(name: str | None = None) -> str:
     return name.lower().replace("-", "_")
 
 
+def build_tests_name(build_name: str) -> str:
+    if os.environ.get("BUILD_ENGINE_SCHEMA"):
+        return os.environ["BUILD_ENGINE_SCHEMA_TESTS"]
```
sf-dcp, line 44: Is there a case where we have `BUILD_ENGINE_SCHEMA` but not `BUILD_ENGINE_SCHEMA_TESTS` env var? If so, we should probably check for both env variable in the if statement or handle the exception?

products/knownprojects/dbt_project.yml
sf-dcp: Nice!!! Thanks for the example!

sf-dcp:
> @sf-dcp
> 
> > From high level, it sounds like you previously had dbt tests creating a schema with test result tables in db.
> 
> yup! a dbt test always creates a table in a schema ([docs](https://docs.getdbt.com/docs/build/data-tests#:~:text=Test%20result%20tables%20are%20created%20in%20a%20schema)). if the schema doesn't exists, it's created. In our case, those schemas are `BUILD_SCHEMA__tests` due to configurations like [this](https://github.com/NYCPlanning/data-engineering/blob/main/products/knownprojects/dbt_project.yml#L9).
> 
> > * This PR addresses rebuilding on the same branch, correct? In other words, the problem you are addressing exists when you previously built a schema with the same name (i.e. branch name).
> 
> yes, as well as nightly QA runs and Template DB test builds in PRs
> 
> > * Does dbt append to existing tables instead of overwriting them completely? In your second bullet point, you are mentioning incorrect results and test names - can you show an example?
> 
> It overwrites them completely. So the incorrectness I'm refereeing to is really caused by tables that were created by tests that have been renamed or deleted.
> 
> So if we write and run `test_primary_key`, we'll have a table named `test_primary_key`. And then if we rename it to `test_compound_key` and run, we'll have a new table called `test_compound_key`.
> 
> Now we have two tables. One of them, `test_primary_key`, is an artifact of code that has changed and so the table shouldn't exist anymore since it doesn't reflect the current state of a product's tests.
> 
> Or in another example, a test that was failing was called `test_important_fact`. We renamed it and it's passing, but the table `test_important_fact` still exists and has records in it because it had been failing. If we looked in the DB, it'd seem like something's wrong with a build even though no tests failed during it.
> 
> > * Not sure I understand your last bullet point.
> 
> We currently delete "stale" DB schemas using [this](https://github.com/NYCPlanning/data-engineering/actions/workflows/stale_builds.yml) github action. It currently deletes all schemas that don't match the names of branches in the repo. So it's gonna delete schemas like `kpdb.in_progress__tests` even when we have branch `in_progress` still open and want to keep all relevant schemas.

Thank you for clarifying. I wanted to make sure I had a full context before my review. Besides a few minor questions/comments across the PR, all looks good!

dcpy/builds/metadata.py
damonmcc: oops that's a typo! both lines should be `BUILD_ENGINE_SCHEMA_TESTS`

Issue #102, opened 2023-08-01
CPDB enhancements


damonmcc:
[Project Statement](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/doc2.aspx?sourcedoc=%7Bad134f75-e828-4efe-a7ea-0dcb0bb01c1f%7D&action=edit)

damonmcc:
awaiting review of draft [Project Statement](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/doc2.aspx?sourcedoc=%7Bad134f75-e828-4efe-a7ea-0dcb0bb01c1f%7D&action=edit) to then finish and share with Capital Planning team for confirmation

damonmcc:
awaiting Capital Planning feedback on project statement

damonmcc:
from KW/Capital Planning on 9/12:

some questions and comments we’re hoping to cover:

- CPDB FISA commitments/liquidations data improvements
- In which tables will the new columns be included?
- Projects, commitments
- Will new columns be replacing any columns, or are we just adding new columns?
- How does this change
  - Citycost
  - Totalcost
  - CCexempt
  - CCnonexempt
  - Totalcityplannedcomit
  - Totalplannedcommit
- Will EDM still be including CheckbookNYC information?
- Should naming convention include underscores to improve readability?
- How are column naming conventions influenced by GIS column name character limits?
- How should we let DE know what the descriptions in the data dictionary should be?
- Can you please also include NYCHA developments shapefiles in the mapping process? Currently projects at NYCHA developments are not mapped against the FacDB or any other spatial reference.

 

Over the longer term, we’d also like to improve our understanding of fuzzy string matching to ultimately work with agencies to improve their naming conventions and increase the % of projects that are mapped

- What is the hierarchy of shapefiles that are used for fuzzy string matching?
- How easy is it to update the CPDB build to include new shapefiles for mapping purposes?
- Can we update the algorithm use sponsor agency for agency ID to improve fuzzy string matching (ie. NYPL as sponsor agency assumes all projects are linked to libraries)
- Are parks capital projects compared not only to their capital project shapefile but to FacDB as well?
- How can we support outreach on Capital Planning’s end to improve naming conventions and increase our mapping?

 Helpful links:

[...]

PR #585, opened 2024-02-14
add cpdb_dcpattributes exports back in

fvankrieken:
At some point in #102, I decided we didn't need these. We do for qa purposes, and it does no harm to keep them in the internal distribution. I've added them back

### DIFF
--- a/products/cpdb/sql/_create_export.sql
+++ b/products/cpdb/sql/_create_export.sql
@@ -1,13 +1,19 @@
+DROP TABLE IF EXISTS cpdb_dcpattributes_pts;
+SELECT * INTO cpdb_dcpattributes_pts
+FROM cpdb_dcpattributes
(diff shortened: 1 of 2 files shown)


products/cpdb/sql/_create_export.sql
```diff
@@ -1,13 +1,19 @@
...
+
+DROP TABLE IF EXISTS cpdb_dcpattributes_poly;
+SELECT * INTO cpdb_dcpattributes_poly
+FROM cpdb_dcpattributes
+WHERE ST_GEOMETRYTYPE(geom) = 'ST_MultiPolygon';
+
 DROP TABLE IF EXISTS cpdb_projects_pts;
 SELECT * INTO cpdb_projects_pts
 FROM cpdb_projects_shp
-WHERE
```
fvankrieken, line 4: cpdb build is based on one ccp, so this clause does nothing

damonmcc (approved):

PR #588, opened 2024-02-15
Fix missing push to latest in publish

fvankrieken:
A troublesome bug - we need to make sure this hasn't led to GIS publishing older datasets


alexrichey (approved):
So this would be potentially a problem when we just default to `latest` in recipes, right?

Issue #575, opened 2024-02-08
Add CEQR Type II source data to data library and implement initial dbt script in `ceqr-survey`


damonmcc:
Source Data Review doc [here](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BBAE15377-87A6-4188-9637-36A89AC1DF98%7D&file=CEQR_Type_II_Data_Source_Review.xlsx&action=default&mobileredirect=true)

for datasets where Status = "approved", we can start adding recipes and extracting data

- [x] Zoning
- [x] E-Designation
- [x] Air
- [x] Noise

fvankrieken:
For simplicity, seems to make sense to limit this issue to the datasets needed for the demo - categories zoning, air quality, and noise

For datasets that correspond to CEQR App, we've elected to archive to data library the outputs that we've produced in `edm-publishing/ceqr-app-data`. Longer term it'd be a lot nicer to archive raw data and process as part of extract code, but that'll need to wait till we really get into #484 

In addition to getting these datasets in `edm-recipes`, they should each have a corresponding script in `products/ceqr/models/intermediate/`. This script should pull in the dataset (imported via dcpy recipe/load logic) using dbt's `source` function, then format the table so that it can be easily joined to pluto bbls to find violations.
- table should have 3 columns: 
  - "variable" - corresponds to dataset (see "Variable" column in data source excel)
  - "id" - some unique identifier of row within its dataset
  - "geom" - the geometry of the dataset buffered as specified in "Data Processing" column in data source excel

fvankrieken:
Zoning category issues are a bit of a special case - since pluto is being imported, we should be able to simply apply flags based on the data present without any spatial logic

fvankrieken:
Just commenting to remind myself to edit this and make a little mermaid diagram for proposed data flow/processing steps in the product build

Issue #486, opened 2024-01-05
Green Fast Track support


AmandaDoyle:
## Project Description

**DE will extract source data, do geospatial calculations, and produce outputs needed for the new Green Fast Track survey GIS is making.**

### Friday 2/23

For a demo of the POC, DE and GIS will focus on a subset of CEQR categories: Zoning, E-designation, Air, Noise.

- [x] https://github.com/NYCPlanning/data-engineering/issues/576
- [x] https://github.com/NYCPlanning/data-engineering/issues/575
- [x] Add build logic
  - https://github.com/NYCPlanning/data-engineering/issues/609
  - https://github.com/NYCPlanning/data-engineering/issues/607
  - https://github.com/NYCPlanning/data-engineering/issues/608
- [x] Export FileGeodatabase for use in survey
  - https://github.com/NYCPlanning/data-engineering/issues/590
- [x] Show the Green Fast Track POC

### Wednesday 3/20

Planning Support will host a forum with applicant reps for feedback on Eligibility Tool. The goal is to confirm it would work for their workflow.

- [x] By 3/6 share survey language with Planning Support
- [x] NYCPlanning/edm-overview#1029
- [x] On 3/11 Review Tool with Planning Support and EARD
- [x] By 3/14 discuss outline of Tool demo and get outputs for pilot projects
- [x] Demo at Applicant Forum for ~10 minutes

### Friday 4/19

Improve GFT tool by incorporating remaining variables and critical feedback.

- [x] #708
- [x] #741

### June 3 public launch
- [x] https://github.com/NYCPlanning/edm-overview/issues/1033
- [x] #814
- [x] #861
- [x] #872
- [x] #816
- [x] #664
- [x] #893
- [x] confirm GFT build aligns with Data Source Review doc
- [x] https://github.com/NYCPlanning/edm-overview/issues/1044
- [x] https://github.com/NYCPlanning/edm-overview/issues/1012
- [x] https://github.com/NYCPlanning/edm-overview/issues/1038

### after public launch
- [x] #887
- [ ] add source data documentation to GFT app

## Background

The [City Environmental Quality Review](https://www.nyc.gov/site/planning/applicants/environmental-review-process.page) (CEQR) process identifies and assesses the potential environmental impacts of land use actions that are proposed by public or private applicants.

[...]

PR #592, opened 2024-02-16
Add dcp_lion to library, elevated railways to `ceqr_survey`

fvankrieken:
Part of #486, #575 

Resulting table
image

A bit interesting that things like airtrain are included. There are 56 distinct segments, so might be worth making sure that all of them are appropriate. See row_type description in [lion metadata](https://s-media.nyc.gov/agencies/dcp/assets/files/pdf/data-tools/bytes/lion_metadata.pdf) (page 32)

### DIFF
--- a/dcpy/library/ingest.py
+++ b/dcpy/library/ingest.py
@@ -81,10 +81,19 @@ def translator(func):
             fields=dataset.destination.fields,
         )
 
-        layerName = srcDS.GetLayer(0).GetName()
+        if srcDS.GetLayerCount() > 1:
+            if dataset.source.layer_name:
+                src_layer = dataset.source.layer_name
+            else:
+                raise Exception(
+                    "Multiple layers in source dataset found, must specify one in recipe."
+                )
+        else:
+            src_layer = srcDS.GetLayer(0).GetName()
+
         sql = dataset.destination.sql
         if sql:
-            sql = sql.replace("@filename", layerName)
+            sql = sql.replace("@filename", src_layer)
 
         # Create postgres database schema and table version if needed
         if output_format == "PostgreSQL":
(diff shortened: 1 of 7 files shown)


dcpy/library/ingest.py
```diff
@@ -81,10 +81,19 @@ def wrapper(self: Ingestor, *args, **kwargs) -> Tuple[list[str], str, str]:
...
-        layerName = srcDS.GetLayer(0).GetName()
+        if srcDS.GetLayerCount() > 1:
+            if dataset.source.layer_name:
+                src_layer = dataset.source.layer_name
+            else:
+                raise Exception(
+                    "Multiple layers in source dataset found, must specify one in recipe."
+                )
+        else:
+            src_layer = srcDS.GetLayer(0).GetName()
```
fvankrieken, line 92: I feel like this is a decent check, in general our outputs assume one layer so we shouldn't be allowing multiple layers going through. Might raise errors if some of our input datasets do have multiple layers and by some convention we're just getting one of them. Would appreciate thoughts

fvankrieken:
Hmm. Archival of lion [just failed](https://github.com/NYCPlanning/data-engineering/actions/runs/7936606882/job/21672252919). I got this error locally, then it resolved, so I assumed it was something that I had inadvertantly fixed, but seems that it might be happening inconsistently? Should poke around a bit before merging

dcpy/library/ingest.py
damonmcc: yea since we're always trying to output one table, seems good to error if multiple are found and none specified

damonmcc (approved):
LGTM but not sure if that parquet issue is resolved

fvankrieken:
Put off investigation lion parquet issue, wrote up in #600

PR #597, opened 2024-02-19
Fix recipe destination default

fvankrieken:
Get nightly qa passing again

Fixes issues causing #591 and failure in #595 

[Nightly qa run for this branch](https://github.com/NYCPlanning/data-engineering/actions/runs/7964084937)

### DIFF
--- a/dcpy/builds/plan.py
+++ b/dcpy/builds/plan.py
@@ -3,7 +3,7 @@ from enum import Enum
 import os
 import pandas as pd
 from pathlib import Path
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 import typer
 from typing import List
 import yaml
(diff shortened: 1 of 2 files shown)


dcpy/builds/plan.py
```diff
@@ -63,7 +63,9 @@ def dataset(self):
 class InputDatasetDefaults(BaseModel, use_enum_values=True):
     file_type: recipes.DatasetType | None = None
     preprocessor: DataPreprocessor | None = None
-    destination: InputDatasetDestination | None = None
+    destination: InputDatasetDestination = Field(
+        default=InputDatasetDestination.postgres, validate_default=True
```
fvankrieken, line 67: The fun dance of pydantic's `use_enum_values`
- If you just set default value to `InputDatasetDestination.postgres`, builds fail because conversion of enum to enum value happens at validation, and normally default values are not validated
- If you set default value to `InputDatasetDestination.postgres.value`, builds work just fine but `mypy` correctly sees that you're doing something that's not type safe

So here we are

fvankrieken: I think we might want to just write a custom `BaseModel` - just pydantic's basemodel, but with `extra="forbid"` and a custom deserializer so that we don't have to use `use_enum_values` and can still use the enums themselves within python while deserializing properly. I don't understand why there isn't a built-in implementation for that

damonmcc (approved):

PR #598, opened 2024-02-19
Add CATS, State Facility, and Title V models to `ceqr_survey`

sf-dcp:
Related to #575.

### What: 
Create air-quality intermediate tables aka "models" for: 
* cats permits
* state facility permits
* title V facility permits

All three models have the following resulting columns: `variable`, `id`, and `geom` (which is **buffered** geometry). The three tables have the same logic (except different distance buffer).

### Commits:
1) Add sources of these datasets to dbt project, specifying required columns and tests. <--- 🚨 NEED FEEDBACK on validation tests here.

2) Create staging models of these datasets: select needed columns, add custom `variable` column, transform geometry to `2263` projection, and filter if needed (applicable to cats permits).

3) Create intermediate tables with buffered geometry in 2 steps: 
    * spatially join with pluto to add tax lot geometry
    * create `geom` buffered column by applying buffer to tax lot geom column or to the original point geom column. The latter happens if original point wasn't mapped to a tax lot.
 
<--- 🚨 NEED FEEDBACK on the spatial queries.

### Successful run: 
* [here](https://github.com/NYCPlanning/data-engineering/actions/runs/7980369506/job/21789858966)
    * it fails a test for `dcm_arterial_highways` source model which is outside of this PR. 

### Notes: 
* All 3 source datasets have same column "geom" and it's in hex-encoded WKB format (ex: `0101000020E6100000BE1589096A7852C0657094BC3A534440`). Casting to geometry sql data type converts the column to `Point`.

### DIFF
--- a/products/ceqr_survey/models/_sources.yml
+++ b/products/ceqr_survey/models/_sources.yml
@@ -100,6 +100,32 @@ sources:
           - name: facility_name
             tests:
               - dbt_utils.at_least_one
-          - name: wkb_geometry
+          - name: geom
+            tests:
+              - not_null
+
+      - name: nysdec_state_facility_permits
+        columns:
+          - name: permit_id
+            tests:
+              - unique
+              - not_null
+          - name: facility_name
             tests:
               - dbt_utils.at_least_one
+          - name: geom
+            tests:
+              - not_null
+
+      - name: dep_cats_permits
+        columns:
+          - name: applicationid
+            tests:
+              - unique
+              - not_null
+          - name: status
+            tests:
+              - not_null
+          - name: geom
+            tests:
+              - not_null
--- a/products/ceqr_survey/models/intermediate/int__dep_cats_permits.sql
+++ b/products/ceqr_survey/models/intermediate/int__dep_cats_permits.sql
@@ -32,10 +32,7 @@ final AS (
     SELECT
         variable,
         id,
-        (CASE
-            WHEN bbl_geom IS NULL THEN ST_BUFFER(permit_geom, 400)
-            ELSE ST_BUFFER(bbl_geom, 400)
-        END) AS geom
+        ST_BUFFER(COALESCE(bbl_geom, permit_geom), 400) AS geom
     FROM cats_permits_with_pluto
 )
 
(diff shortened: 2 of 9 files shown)


sf-dcp:
Sample cats permits with 400-feet buffer (including tax lots & point geoms): 

image

products/ceqr_survey/models/_sources.yml
```diff
@@ -89,12 +89,37 @@ sources:
         columns:
           - name: permit_id
             tests:
-              - unique
+              - unique # TODO: need to confirm this assumption
```
fvankrieken, line 92: Confirmed?

damonmcc: agree we can consider our testing of it as confirmation

damonmcc:
@sf-dcp I fixed the highway bug in #606 if it'd be nice to see a full build on this branch

products/ceqr_survey/models/staging/stg__pluto.sql
```diff
@@ -6,11 +6,11 @@ WITH mappluto_wi AS (
 
 final AS (
     SELECT
-        CAST(bbl AS text) AS bbl,
         zonedist1,
         zonedist2,
         zonedist3,
         zonedist4,
+        CAST(bbl AS text) AS bbl,
```
fvankrieken, line 13: Can we move this back? I assume this was from `sqlfluff fix` being run not from the home directory?

products/ceqr_survey/models/intermediate/int__dep_cats_permits.sql
```diff
@@ -0,0 +1,42 @@
...
+-- create buffer of 400 feet around tax lot (bbl_geom column). 
+-- If tax lot is null, create buffer around point (geom column)
+final AS (
+    SELECT
+        variable,
+        id,
+        (CASE
+            WHEN bbl_geom IS NULL THEN ST_BUFFER(permit_geom, 400)
+            ELSE ST_BUFFER(bbl_geom, 400)
+        END) AS geom
```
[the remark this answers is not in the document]
sf-dcp: Thanks, will revise!

damonmcc (approved):

PR #603, opened 2024-02-20
fix pluto projection in CEQR Survey

damonmcc:
this makes subsequent spatial operations with PLUTO simpler since other source data is projected in 2263


sf-dcp (approved):

sf-dcp:
❤️ tiny PRs

damonmcc:
those force pushes were to fix sql formatting

PR #619, opened 2024-02-24
improve Template DB

### LINKED ISSUE #268: Template DB enhancements

damonmcc:
resolves #268 

End-to-end test failing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8101460990/job/22141825321?pr=619#step:7:60) due to exported metadata files and new shapefiles, passing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8101704638/job/22142602432#step:7:28)

I imagine we'd like to have any new "make shapefiles" code to be in our utils. But for the urgency of using this in the ongoing GIS automation/metadata work, I kept it in the Template DB codebase for now.

### DIFF
--- a/products/template/build_scripts/export.py
+++ b/products/template/build_scripts/export.py
@@ -1,5 +1,8 @@
+import os
 import shutil
+import zipfile
 from dcpy.connectors.edm import publishing
+from dcpy.utils import geospatial
 from dcpy.utils.logging import logger
 
 from . import PRODUCT_S3_NAME, BUILD_NAME, OUTPUT_DIR, PG_CLIENT
@@ -9,11 +12,17 @@ METADATA_FILES = [
     "build_metadata.json",
 ]
 BUILD_TABLES = {
-    "templatedb": ["csv"],
+    "templatedb": [
+        "csv",
+        "shapefile_points",
+        "shapefile_polygons",
+    ],
 }
 
 
 def export():
+    if OUTPUT_DIR.exists():
+        shutil.rmtree(OUTPUT_DIR)
     OUTPUT_DIR.mkdir(parents=True)
     # export metadata files
     for filename in METADATA_FILES:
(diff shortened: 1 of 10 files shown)


products/template/metadata.yml
```diff
@@ -40,8 +40,15 @@ dataset_package:
...
       type: csv
-    - name: primary_shapefile
-      filename: template.shp.zip
+    - name: shapefile_points
+      filename: templatedb_points.shp.zip
+      type: shapefile
+      overrides:
+        columns:
+          wkb_geometry:
+            name: the_geom
```
damonmcc, line 49: noting that things like this column name override aren't being tested. that's what the GIS automation/metadata work is for

products/template/build_scripts/export.py
```diff
@@ -24,11 +33,40 @@ def export():
...
-                    f"Exporting table\n\t{table_name}\n\tas a {file_type} to\n\t{OUTPUT_DIR}"
-                )
                 data.to_csv(file_path.with_suffix(".csv"), index=False)
+            elif "shapefile" in file_type:
+                shapefile_directory = OUTPUT_DIR
+                geodata = geospatial.convert_to_geodata(
+                    data,
+                    geometry_column="wkb_geometry",
+                    geometry_format=geospatial.GeometryFormat.wkb,
+                    crs=geospatial.GeometryCRS.wgs_84_deg,
```
alexrichey, line 47: question: do we normally export our package shapefiles in WGS 84?

products/template/sql/green_spaces.sql
sf-dcp, line 1: tiiiiny nit: ✂️ line 1?

products/template/sql/product/_product_models.yml
```diff
@@ -13,24 +13,11 @@ models:
       - name: wkb_geometry
         tests:
           - not_null:
-              where: "bbl is not null"
+              where: bbl is not null
```
sf-dcp, line 16: what does "where" tag do?

products/template/build_scripts/export.py
damonmcc: good question. the COLP data on Open Data [here](https://data.cityofnewyork.us/City-Government/City-Owned-and-Leased-Property-COLP-/fn4k-qyk2/data_preview) looks to be WGS 48 (looking at the numbers in the lat, long, and geom columns)

but COLP in our DB looks to be state plane and I don't see any reprojecting in the export code. so maybe Open Data reprojects after we upload?

products/template/sql/product/_product_models.yml
damonmcc: it limits a test to records where the condition is true ([docs](https://docs.getdbt.com/reference/resource-configs/where))

sf-dcp: niceeee

sf-dcp (approved):
Good job!

products/template/metadata.yml
damonmcc: ideally column names are consistent across file types, but the rationale for the renaming here is:
- shapefiles have a character limit for column names, so we either end up with unreadable auto-truncated column names or we design with the flexibility of setting them ourselves
- legacy reasons: we often want to improve column names in our "internal" tables without disrupting downstream destinations like Open Data
- since Template DB is being used to test our pipelines, this column renaming is really just to mimic an unfortunate aspect of real data products

products/template/build_scripts/export.py
sf-dcp: Perhaps when we abstract to `utils`, we want to be able to explicitly specify a desired projection

sf-dcp:
@damonmcc, when we export data to a specific geom type like here (points and polygons), would it fail silently if it can't translate the geom?

damonmcc:
@sf-dcp I think since the code to get specific geom types is selecting rows of a GeoDataFrame, it isn't translating geoms so couldn't fail for that reason

the code below is pretty much "where geoemtry type is MultiPolygon"
```python
geodata = geodata.loc[geodata.geom_type == "MultiPolygon"]
```

PR #624, opened 2024-02-26
Add dbt test for checking projection

sf-dcp:
## Motivation
Currently our source data has different projections and it takes a bit of work to confirm a geometry projection for a given table. One way is to check target projection in a dataset template in `data-library`. Another way is to use gdal CLI to get the projection. It would be nice to have a test that would check a geometry column to be in state plane (`EPSG:2263`). 

## Approach
You can write custom schema tests in dbt. I wrote my own test that checks a column's projection, and it fails if there is at least 1 record with a projection other than `2263`.

## Commits
1. Create the test in `macros` directory
2. Add the test to 2 tables: `dcp_mappluto_wi` and `int__dep_cats_permits_buffer`. These tablas have different projections, allowing us to check correct behavior from the test. 

#### Successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8055140004/job/22001301833). From this run, you can see that:
* `dcp_mappluto_wi` gets a [warning](https://github.com/NYCPlanning/data-engineering/actions/runs/8055140004/job/22001301833#step:7:75) as expected. 
* `int__dep_cats_permits_buffer ` passes the [test](https://github.com/NYCPlanning/data-engineering/actions/runs/8055140004/job/22001301833#step:7:294).

## TODO
* [ ] After this pr is reviewed, delete the test from pluto table. It's not needed here because pluto geom column gets casted to correct projection in staging.

### DIFF
--- a/products/ceqr_survey/macros/test_is_2263_srs.sql
+++ b/products/ceqr_survey/macros/test_is_2263_srs.sql
@@ -1,0 +1,22 @@
+-- tests column values to have 2263 projection (state plane)
+
+{% test is_2263_srs(model, column_name) %}
+
+WITH validation AS (
+    SELECT
+        {{ column_name }} AS geom_column
+    FROM {{ model }}
+),
+
+validation_errors AS (
+    SELECT
+        geom_column
+    FROM validation
+    WHERE ST_SRID(geom_column) != 2263
+
+)
+
+SELECT *
+FROM validation_errors
+
+{% endtest %}
--- a/products/ceqr_survey/models/_sources.yml
+++ b/products/ceqr_survey/models/_sources.yml
@@ -91,7 +91,7 @@ sources:
           - name: wkb_geometry
             tests:
               - not_null
-              - is_2263_srs:
+              - is_epsg_2263:
                   config:
                     severity: warn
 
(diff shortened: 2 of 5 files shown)


products/ceqr_survey/models/_sources.yml
```diff
@@ -91,6 +91,9 @@ sources:
           - name: wkb_geometry
             tests:
               - not_null
+              - is_2263_srs:
+                  config:
+                    severity: warn
```
sf-dcp, line 96: I set severity to `warn` level here because otherwise it stops the build (which we probably want in other cases like after building staging tables). This test is run on the source data before we do anything to it.

damonmcc (approved):
nice!

and for that todo you noted, it doesn't seem like a bad idea to keep testing source data. the test is the proof that we should reproject (which we already do)

products/ceqr_survey/macros/test_is_2263_srs.sql
fvankrieken, line 1: Would prefer a name without "srs", maybe `test_is_epsg_2263`. While it is a spatial reference system, "SRS" is a term more used in gdal specifically, more common language choice for these seem to be "coordinate system"

fvankrieken:
This is awesome!

sf-dcp:
Per Finn's suggestion, renamed the test.

Re-ran the build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8083267134/job/22085974400). 

Decided to keep the test under PLUTO per Damon's rec

fvankrieken (approved):

PR #625, opened 2024-02-27
GHA maintenance

### LINKED ISSUE #622: Stagger nightly build and weekly compile python requirements
See #621, failed test in #620. These jobs running at the same time seem to be throttling db resources, leading to both failing

fvankrieken:
1. closes #622 , move python compile job to noon on saturdays
2. fix logic around docker images. Logic was right in assigning tag, but in case of "changes since last push to pr", missed case of a rebase from main which changes python/docker stuff. In this case, but when there are NO pr changes, nothing should happen because they should still be using latest. So just need to check if there are BOTH push and pr changes before building dev images

### DIFF
--- a/.github/workflows/python_compile_requirements.yml
+++ b/.github/workflows/python_compile_requirements.yml
@@ -1,7 +1,7 @@
 name: Compile python requirements
 on:
   schedule: 
-  - cron: 0 5 * * SUN
+  - cron: 0 17 * * SAT
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/python_compile_requirements.yml
```diff
@@ -1,7 +1,7 @@
 name: Compile python requirements
 on:
   schedule: 
-  - cron: 0 5 * * SUN
+  - cron: 0 17 * * SAT
```
fvankrieken, line 4: These are UTC if that wasn't clear! So moved up 12 hr from Sunday 0:00 to Sat 12:00

damonmcc: (not for this PR) I wonder if we can put time zone in these

damonmcc (approved):

damonmcc (approved):

PR #629, opened 2024-02-28
rename ceqr_survey to green_fast_track

fvankrieken:
Do give it a look over for the actual renamings within files

Building [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8087386444)


products/green_fast_track/models/product/_product_models.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 
 models:
-  - name: ceqr_survey_bbls
+  - name: greenbbls
```
damonmcc, line 4: should be `green_fast_track_bbls`?

fvankrieken: oops ty, not sure how that happened

damonmcc (approved):

Issue #632, opened 2024-02-29
Data Library - Processing Scaffolding


fvankrieken:
Per https://github.com/NYCPlanning/data-engineering/issues/484#issuecomment-1969720450, set up for now at least a scaffold of code to take yml template and run custom script (if called for), geoprocessing steps, etc.

sf-dcp:
From [review](https://github.com/NYCPlanning/data-engineering/issues/633#issuecomment-1973433345) of custom scripts, the following set of templates seem to be representative of our custom scripts ingestion: 

* `nysed_nonpublicenrollment` <-- data read from a specific tab in an excel file
* `dcp_pad` <-- extract zipfile from an API
* `bbl_libraries` <-- extract json from an API
* `doe_pepmeetingurls` <-- scrapes data from a website with requests & beautiful soup. This recipe represents datasets that can't be simply pulled, and we need to account for the logic with a custom script extraction.
* `fisa_capital_commitments` <-- a csv with custom delimiter & no headers. Similar datasets also require custom encoding when reading from a csv. 

cc: @fvankrieken

Issue #631, opened 2024-02-29
Data Library - Convert raw data to parquet


fvankrieken:
This should take result of #499 and convert it to parquet. This will obviously depend on file type, as defined in `source` section of yml template. For shapefiles, geodatabases, etc. it might still make sense to use gdal, though geopandas is an option as well. In my opinion, code simplicity is most important, so even if gpd is less performant (not sure it is, just speculating), I would prefer it to gdal. 

This should output a parquet file locally that can feed into the preprocessing steps of the extraction

fvankrieken:
Edge cases:
- geodatabase with multiple layers where we only need one of them

sf-dcp:
Adding on requirements for this task: 

* Reading in the following formats:
  * csv
  * geodabase (with 1 or multiple layers)
  * shapefile
  * zipped version of any of the formats above
  * json
  * geojson
  * excel?
  
* Correct geometry in output (parquet stores geometry in wkb format)
* Unchanged column names in output
* Unchanged data in output

sf-dcp:
Update: implemented 1st iteration of `convert_raw_to_parquet` fn in #658. The fn currently converts `csv`, `geodatabase`, and `shapefile` formats to parquet. 

`TODO`:  
- [ ] implement json
- [ ] geojson
- [ ] zipped format
- [ ] for csv files, implement point geom made out of Longitude & Latitude columns
- [ ] geoparquet (as an export format)
- [ ] name geometry column as `geom` in output parquet file
- [ ] break up `transform_to_parquet` fn into 2 parts: 1) read raw data into a `geopandas` df and 2) save the df to parquet
- [ ] simplify tests (perhaps by using complete, not partial, fake templates)

fvankrieken:
For raw json, it'd make sense to have an optional field that takes input in form of jq query to parse a slightly nested structure into something tabular

sf-dcp:
All data formats have been implemented.

Issue #499, opened 2024-01-09
Data Library - Archive Raw Data


fvankrieken:
There's a bit of this which could overlap with some of #498, mainly in that some of this will end up in some metadata changes. But I think that there are some specific metadata tweaks that align well with this change, while #498 is meant to be a bit more comprehensive and expansive, as opposed to a tweak of what we currently have.

I see a few components of this
- Actual archival of raw source data. This involves some decisions. Do we want to store this data in a separate bucket? Different parent folder in recipes, mimicking structure? So recipes/raw/{dataset_name}/{version} and recipes/datasets/{dataset_name}/{version}. Or we could simply put them in the same folders. I don't love that - sometimes I imagine we'll need to get version information from a dataset itself possibly, and I don't like in general running into cases where we have a dataset/version folder that has raw but not processed data for some reason. 
  - this also gets into "version" vs "date" issue, and maybe we need to think about this a little bit.
  - separate buckets is convenient in that we could use the exact same utils for pushing/pulling, just replacing bucket name. But I guess we could do the same thing with parent folder. Maybe we go to two new buckets for this rather than putting them into recipes, and get us out of the "recipe" naming ambiguity (edm-recipes vs recipe.yml)
- Breaking up of extract logic a bit. These are sort of mini ETLs in and of themselves, so thinking of the true "extract" part (or first "load" part?) as dumping raw data into s3, then being able to "transform" (clean/preprocess/reproject/etc) this and output a parquet file. This also means updating logic around script sources. We want to move towards a model where when we have custom script "sources", we really have "path" or "socrata" etc sources, and then apply some sort of operators to the dataset rather than the current model which views a python script as a "source" in and of itself

This maybe should be holstered until we fully move to parquet, just for cleanliness. Work doesn't need to wait though, just fully operationalizing

fvankrieken:
Rescoping this issue a bit per discussion on 2/28.

For now, this python code should live in `dcpy/extract` (maybe in file called `ingest.py` if we don't feel like it's a loaded term). The logic in this file should
1. Take in a yml template (from library), specifically the `source` section
[...]

PR #636, opened 2024-02-29
Scaffold for dcpy.extract

fvankrieken:
#632 

@sf-dcp identified 8 datasets largely representative of the breadth of types of datasets we ingest. We're going to be working with those as we move forward. We're including the "old" template as a subfield for reference while we develop

For now separated out yml/pydantic fields of "archive raw" metadata and "transform to parquet" metadata, so that they can be added separately by me and @sf-dcp as we work on #499 and #631, respectively. These will probably be merged to something like "source" at some point. Or maybe not!

Added one test for now - test that all template files valid.

This will be squashed


dcpy/extract/ingest.py
```diff
@@ -0,0 +1,47 @@
+from pathlib import Path
+from pydantic import BaseModel
+
+from dcpy.connectors.edm import recipes
+from . import PARQUET_PATH
+
+
+class ImportDefinition(BaseModel):
```
fvankrieken, line 8: Rationale for this being it's nice to have a class that is strictly a definition (with optional types) separate from the actual "computed" config (which can then have non-optional types, maybe slightly different structure)

dcpy/extract/templates/bpl_libraries.yml
```diff
@@ -0,0 +1,39 @@
+name: bpl_libraries
+acl: public-read
+source:
+  url: https://www.bklynlibrary.org/locations/json
+
+library_dataset:
```
fvankrieken, line 6: Added a comment in the class definition, but figured easiest thing for now is to include the entire "old" template as a subfield, so we can add/delete freely in the actual new template but have the old ones here for reference

dcpy/connectors/edm/recipes.py
```diff
@@ -192,6 +192,14 @@ def to_row(self) -> dict:
         }
 
 
+def archive(config: Config, file_path: Path):
+    ### this could either take a config object and path to a file, and archive them both in a folder together,
+    ### or take path to folder that takes dumped config as well as file output and loads whole folder
+    ### likely the latter in case of multiple output formats. Not necessarily something we want to be doing long term
+    ### but for flexibility for now it might make sense
+    raise NotImplemented
```
sf-dcp, line 200: Is this function responsible for creating desired output formats (i.e. csv, pgdump) or would it be a separate fn? If it will happen here, then the latter option in your comment shouldn't be a concern

fvankrieken: I would say it should happen prior to this, but open for debate

sf-dcp: I noticed that we are archiving raw data in `ingest.py`. What do you think about moving this function to `ingest.py`?

dcpy/extract/ingest.py
```diff
@@ -0,0 +1,33 @@
+from dcpy.connectors.edm import recipes
+from . import PARQUET_PATH, models
+
+
+# doesn't strictly need to be recipe config, but at least config-like object
+def archive_raw_dataset(dataset: str, version: str | None = None) -> recipes.Config:
+    """Given dataset and potentially a semantic version, ingest it from its source and archive it in its
+    raw format, with metadata, in edm-recipes/raw_data"""
```
sf-dcp, line 8: `edm-recipes/{dataset}/{version}/raw_data` or `edm-recipes/raw_data`?

dcpy/connectors/edm/recipes.py
fvankrieken: Maybe. My gut is have it here, just because there are a few things going on in extract
- grab raw data
- push raw data to s3 (edm-recipes)
- transform to parquet
- process
- push to s3 (edm-recipes)

I think the interaction with s3 portions should live within connectors.edm.recipes, and not use our s3 utils directly. We're definitely blurring the line a little bit here, and there's an argument to unify a bit more code. But to me, we're keeping `extract` logic as the logic around pulling data, archiving it, and getting it ready for our pipelines, not dealing with the actual storage/representation of datasets that we've archived. So that logic should stay in recipes

dcpy/extract/ingest.py
fvankrieken: the latter is what we're going with for now

dcpy/connectors/edm/recipes.py
fvankrieken: With the rationale being that our extract code doesn't need to know about the details of where we're storing things, our edm recipes connector logic can take care of that

dcpy/extract/ingest.py
```diff
@@ -0,0 +1,33 @@
...
+    import_definition = models.read_import_definition(dataset)
+
+    ## logic to grab dataset based on ImportDefinition, dump locally
+
+    ## logic to "compute" recipes.Config object, similar to Config.compute in library
```
sf-dcp, line 13: I'm leaning towards computing `Config` object separately and pass it into `archive_raw_dataset` because:
1) should make testing of `Config` object and archival of raw data easier as these functions will have 1 task to perform. 
2) `archive_raw_data` will be more readable.
3) It feels confusing getting a `Config` object as a result of a fn named `archive_raw_data`. Otherwise, I would rename the fn something like `archived_raw_data_config`

Let me know your thoughts

fvankrieken: That makes sense to me!

fvankrieken: See commit 4. Other thing in general that I would LOVE thoughts on is naming of things. I feel like I'm going in circles and using synonyms to mean different things and am just too deep in it at this point to know what makes sense

dcpy/extract/templates/bpl_libraries.yml
sf-dcp: Nice solution! I didn't realize we are using "production" templates to build new data library :) Questions:
* It looks like our original templates have `dataset` not `library_dataset` object. Does new object name work with our current data library infra?
* `transform_to_parquet_metadata` <--- what type of data you are thinking of here?

dcpy/test/extract/test_models.py
```diff
@@ -0,0 +1,13 @@
+import yaml
```
sf-dcp: rename file to `test_config`?

dcpy/extract/ingest.py
sf-dcp: Okay, will keep in mind to share feedback about naming! And I agree, "template" makes sense since we use it when we specifically refer THE templates lol

dcpy/connectors/edm/recipes.py
sf-dcp: makes sense

sf-dcp:
Finished my review - all looks good! Nice job. Have a couple of clarifying questions + need to rename the test. Otherwise ready to approve

sf-dcp (approved):

dcpy/test/extract/test_models.py
fvankrieken: Done!

dcpy/extract/templates/bpl_libraries.yml
fvankrieken: 1st point - It doesn't. these are not really meant to be used by anything in code, just to live in the files so that we can reference them. As in, as we're building these new ones, make sure that we're not forgetting anything. Once we're done with this project, we can remove this field. If we really wanted to point data library at these for some reason, it would work with very minimal tweaks.

For your second point, not sure - just wanted to have a field that is self contained for #631 , so that additions can be made without collisions from work on #499. Presumably file type might be needed - for a txt within a zip for dcp pad for example. But just wanted a dedicated place for anything needed in implementation of #631. Can merge these or drop the latter later if it's unnecessary.

PR #640, opened 2024-03-01
revise Green Fast Tract

damonmcc:
related to https://github.com/NYCPlanning/edm-overview/issues/1029

These are changes to the Green Fast Track categories that we've already implemented. These came from discussions with Planning Support and GIS

- export all points, not just points that didn't join to a lot
- export a list of all values that contribute to each flag (e.g. CATS Permit IDs as `abc, xyz`, E-Des as `E-215, E-655`)
- add columns and exports for a new "Unpermitted Industrial Sources" Air Quality variable to flag nearby lots with manufacturing land use
- build in stages to make build logs more readable and useful

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-gft-update)

## notes
- tried to build using the latest Vents data (data library run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8164322120)), but that shapefile only has 1 row so I pinned the Vents version here to the last valid version. following-up with GIS to remake the data
- some lots are near >50 CATS Permits. will follow-up with GIS to determine how to handle those (maybe change value to `10+ values` or something)

## screenshots
Screenshot 2024-03-06 at 9 46 19 AM

### DIFF
--- a/products/green_fast_track/models/product/green_fast_track_bbls.sql
+++ b/products/green_fast_track/models/product/green_fast_track_bbls.sql
@@ -28,17 +28,12 @@ flags_ranked AS (
 flags_wide AS (
     SELECT
         {% for row in variables -%}
-            MAX(
-                CASE
-                    WHEN
-                        variable_type = '{{ row["variable_type"] }}'
-                        THEN variable_id
-                END
+            ARRAY_TO_STRING(
+                ARRAY_AGG(variable_id) FILTER (WHERE variable_type = '{{ row["variable_type"] }}'), ', '
             ) AS "{{ row['label'] }}",
         {% endfor %}
         bbl
     FROM flags_ranked
-    WHERE row_number = 1
     GROUP BY bbl
 )
 
--- a/products/green_fast_track/models/product/green_fast_track_bbls.sql
+++ b/products/green_fast_track/models/product/green_fast_track_bbls.sql
@@ -19,6 +19,7 @@ flags_ranked AS (
         bbl,
         variable_type,
         variable_id,
+        distance,
         ROW_NUMBER()
             OVER (PARTITION BY bbl, variable_type ORDER BY distance)
         AS row_number
@@ -28,8 +29,12 @@ flags_ranked AS (
 flags_wide AS (
     SELECT
         {% for row in variables -%}
+            /* construct a comma-separated list of values ordered by distance and name */
             ARRAY_TO_STRING(
-                ARRAY_AGG(variable_id) FILTER (WHERE variable_type = '{{ row["variable_type"] }}'), ', '
+                ARRAY_AGG(variable_id ORDER BY distance ASC, variable_id ASC) FILTER (
+                    WHERE variable_type = '{{ row["variable_type"] }}'
+                ),
+                ', '
             ) AS "{{ row['label'] }}",
         {% endfor %}
         bbl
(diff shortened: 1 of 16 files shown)


fvankrieken:
Would you want to move buffers to `intermediate/buffers` folder and flags to `intermediate/flags` as part of this pr? As I think we'd settled on last week for desired organization

products/green_fast_track/models/product/green_fast_track_bbls.sql
```diff
@@ -28,17 +28,12 @@ flags_ranked AS (
...
     SELECT
         {% for row in variables -%}
-            MAX(
-                CASE
-                    WHEN
-                        variable_type = '{{ row["variable_type"] }}'
-                        THEN variable_id
-                END
+            ARRAY_TO_STRING(
+                ARRAY_AGG(variable_id) FILTER (WHERE variable_type = '{{ row["variable_type"] }}'), ', '
```
fvankrieken, line 32: Do we want any ordering here? Either by distance(/`row_number`), or alphabetical, etc

products/green_fast_track/recipe.yml
```diff
@@ -10,6 +10,7 @@ inputs:
     - name: nysdec_state_facility_permits
     - name: nysdec_title_v_facility_permits
     - name: dcp_air_quality_vent_towers
+      version: 20240201
```
fvankrieken, line 13: Why the pin?

fvankrieken: Answered in the PR description!

products/green_fast_track/models/product/green_fast_track_bbls.sql
damonmcc: I'll look at some values but I think they're already ordered during the CTE above because of `OVER (PARTITION BY bbl, variable_type ORDER BY distance)`

and distance seems like a good choice

fvankrieken:
You said "where requested, export values for all adjacent features (e.g. CATS Permit IDs as "abc, xyz")", but seems like we're doing array_agg for all. Are there cases where they just want one? Would be fairly straightforward to add a case statement, do either the old logic or your new logic based on variable name. Or more specifically, add "aggregate_method" to the `variables` table, and base which type of aggregation we do on that

products/green_fast_track/models/product/green_fast_track_bbls.sql
fvankrieken: Or just by row_number - order already computed, and in that case we're just enforcing what the postgres compiler hopefully is already doing

damonmcc:
@fvankrieken good call on doing "nearest only" for some. since Planning Support's reasoning for "nearest only" seemed to be "we don't _need_ all values". I erred on the side of "give all values" for this first pass so that we can see if the assumption that "all values" will be overwhelming is actually true for most projects

from the meeting notes in the [survey issue](https://github.com/NYCPlanning/edm-overview/issues/1012), looks like Arterial Highways and Shadows (not implemented yet) are the only variables where "nearest only" might be desirable

products/green_fast_track/models/product/green_fast_track_bbls.sql
damonmcc: looks like it was doing furthest -> nearest! and even though there's lots of ties (same lot with >1 CATs Permit), it was still deterministic

I added `order by distance asc, variable_id asc`. the `variable_id` part is to resolve distance ties and flags that where distance is null

damonmcc:
@fvankrieken
> Would you want to move buffers to `intermediate/buffers` folder and flags to `intermediate/flags` as part of this pr? As I think we'd settled on last week for desired organization

would rather do that in a follow-up and focus on changes GIS needs

fvankrieken (approved):

PR #643, opened 2024-03-03
check dbt projects in PR tests

damonmcc:
We discussed and would like to start applying [Code Conventions](https://github.com/NYCPlanning/data-engineering/wiki/Code-Conventions) for our dbt projects.

Although we use sqlfluff to lint sql files, it's limited to the formatting of queries and can't check things that involve compiled dbt sql files and properties yaml files.

[`dbt-checkpoint`](https://github.com/dbt-checkpoint/dbt-checkpoint) is a set of pre-commit hooks that can be used locally and in CI to check that our dbt files adhere to the rules we choose to enforce ([dbt blog](https://docs.getdbt.com/blog/enforcing-rules-pre-commit-dbt)).

example of new tests failing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8127931481/job/22213314373#step:7:56) due to some models not being listed in a properties yaml file (no `_staging_models.yml`) and then passing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8127953377/job/22213358437#step:7:56) after the file is added with all model names in it

### DIFF
--- a/.github/workflows/test_helper.yml
+++ b/.github/workflows/test_helper.yml
@@ -98,9 +98,6 @@ jobs:
           ./bash/docker_container_setup.sh
           ./bash/build_env_setup.sh
 
-      - name: Temp install pre-commit
-        run: python3 -m pip install pre-commit
-
       - name: Check dbt files
         run: |
           export BUILD_ENGINE_DB=db-green-fast-track
--- a/.github/workflows/test_helper.yml
+++ b/.github/workflows/test_helper.yml
@@ -52,7 +52,7 @@ jobs:
           sqlfluff lint products/knownprojects/ --templater=jinja
           sqlfluff lint products/pluto/pluto_build/sql/
           sqlfluff lint products/zoningtaxlots/sql/
-          export BUILD_ENGINE_DB=db-ceqr-survey
+          export BUILD_ENGINE_DB=db-green-fast-track
           sqlfluff lint products/green_fast_track/models --templater=dbt
 
   mypy:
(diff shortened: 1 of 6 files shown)


.github/workflows/test_helper.yml
```diff
@@ -69,6 +69,46 @@ jobs:
...
+          BUILD_ENGINE_USER: "op://Data Engineering/EDM_DATA/username"
+          BUILD_ENGINE_PASSWORD: "op://Data Engineering/EDM_DATA/password"
+          BUILD_ENGINE_PORT: "op://Data Engineering/EDM_DATA/port"
+
+      - name: Setup build environment
+        run: |
+          ./bash/docker_container_setup.sh
+          ./bash/build_env_setup.sh
+
+      - name: Temp install pre-commit
```
fvankrieken, line 101: Would we want to just add this to `requirements.in` and the `dev` image?

.github/workflows/test_helper.yml
```diff
@@ -69,6 +69,46 @@ jobs:
...
+        run: python3 -m pip install pre-commit
```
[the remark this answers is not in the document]
fvankrieken: commented at about the same time - let's just include it in this one?

.github/workflows/test_helper.yml
```diff
@@ -69,6 +69,46 @@ jobs:
...
+
+      - name: Check dbt files
+        run: |
+          export BUILD_ENGINE_DB=db-green-fast-track
+          dbt deps --profiles-dir products/green_fast_track --project-dir products/green_fast_track
+          dbt seed --profiles-dir products/green_fast_track --project-dir products/green_fast_track
+          dbt docs generate --profiles-dir products/green_fast_track --project-dir products/green_fast_track
```
damonmcc, line 109: these all have to be run before the checks because the checks parse the artifacts of compiling the sql queries: `target/manifest.json` and `target/catalog.json`

.github/workflows/test_helper.yml
```diff
@@ -69,6 +69,46 @@ jobs:
...
+          pre-commit run --all-files
```
damonmcc, line 110: `--all-files` means that this checks all files, not just files changed in a given PR. this is consistent with all of our other "linting"

Issue #396, opened 2023-11-28
Accurate versioning for datasets from edm-publishing/datasets


fvankrieken:
Many data library datasets are pulled from edm-publishing. Generally, they look in `edm-publishing/datasets/{dataset}/staging`, grab the file, and use the current date (date of data library execution) as version. This isn't accurate, as we can be grabbing the same dataset over and over again and labelling it as "new" versions.

Case in point - dof_shoreline. It hasn't been updated since 2022
<img width="676" alt="image" src="https://github.com/NYCPlanning/data-engineering/assets/9454672/2149366b-0f15-445f-bec6-ed3627625a2a">

But it loads frequently (strangely so - need to glance into this) in `edm-recipes`
<img width="674" alt="image" src="https://github.com/NYCPlanning/data-engineering/assets/9454672/98b73fdb-8771-429d-82fe-d9767daee281">

Meaning we're generating a ton of duplicate data, and potentially using "different" versions of this dataset in our product builds that are all actually using the same data source

PR #654, opened 2024-03-05
dcpy.extract - archive raw data

fvankrieken:
#499 

Also addressing #396 

There are a couple big steps to happen here, largely squished into commit 2

Commit 1 is simpler - functions to grab GIS datasets from `edm-publishing` bucket. Since we currently assume "staging" folder is latest, using that as a proxy - use that to get last updated timestamp, look for that version and confirm that an actual version from that date exists (rather than blindly trusting that the last updated reflects the accurate version).

Commit 2 is the meat of this. 

Starting with the types, in `extract.metadata.py` we have the `Template` object. This refers to the template.yml files we store in library/extract, and is meant to be a literal reading of them (does not get "computed" as in data library). The other big ones are in edm/recipes.py. These types are meant to be accessed/used by the recipes api, so I wanted them to live here. For now, have both a "LibraryConfig" type (refers to config.json structure currently used in dcpy.library) and "ExtractConfig". Towards the end of this project we likely want to align these a bit more (or at least just figure out how things are going to be backwards compatible) but for now, while we develop, I'm just keeping them entirely separate. This ExtractConfig is the "computed" version of the template.

What goes into computing it? I don't call it "computing" anymore, because I wanted the data flow to be a bit more intuitive. We have a couple main steps from reading a template to archiving raw data.
- generate runtime timestamp
- read in template, passing version as jinja template options. This does mean that having {{version}} in the yml is only supported when version is supplied explicitly. Added 4th commit where this its validated that vars in template are supplied.
- get version if not supplied. For socrata or GIS dataset, look it up. Otherwise, use generated timestamp to create "today" version
- download file (if not a local file)
- archive file, with config that includes version, timestamp, and filename of raw file. This is stored in `edm-recipes/raw_datasets/{dataset}/{timestamp.isoformat()}

Commit 3 - tests. I don't test the "full" call, but each of the components of it (each of the above steps)

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -406,6 +406,8 @@ def _assert_gis_dataset_exists(name: str, version: str):
 
 
 def get_latest_gis_dataset_version(dataset_name: str) -> str:
+    """Get latest version of GIS-published dataset in edm-publishing/datasets
+    from metadata of dataset in 'staging' folder"""
     ## TODO
     ## might make more sense to just list folders, grab "max" version
     ## by that logic, should likely try to parse str version to dcpy version object
@@ -418,6 +420,10 @@ def get_latest_gis_dataset_version(dataset_name: str) -> str:
 
 
 def download_gis_dataset(dataset_name: str, version: str, target_folder: Path):
+    """Download GIS-published dataset from edm-publishing/datasets.
+    Capitalizes supplied version when looking in s3 due to current conventions.
+    Only quarterly (24a/24A) datasets currently use format other than just numeric datestrings
+    """
     ## TODO - assumes versions are numeric or geosupport (which we use "24a" vs gis "24A")
     version = version.upper()
     assert target_folder.is_dir(), f"Target folder '{target_folder}' is not a directory"
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -104,6 +104,10 @@ class ExtractConfig(BaseModel, extra="forbid"):
                             raise Exception(f"Unknown org {s}")
 
         class EdmPublishingGisDataset(BaseModel, extra="forbid"):
+            """Dataset published by GIS in edm-publishing/datasets"""
+
+            # Some datasets here will phased out if we eventually get data
+            # directly from GR or other sources
             type: Literal["edm_publishing_gis_dataset"]
             name: str
 
--- a/dcpy/extract/metadata.py
+++ b/dcpy/extract/metadata.py
@@ -32,6 +32,7 @@ class Template(
 
 
 def get_jinja_vars(s: str) -> set[str]:
+    """Get all variables expected in a jinja template string"""
     env = jinja2.Environment()
     parsed_content = env.parse(s)
     return meta.find_undeclared_variables(parsed_content)
@@ -40,6 +41,8 @@ def get_jinja_vars(s: str) -> set[str]:
 def read_template(
     dataset: str, version: str | None = None, template_dir: Path = TEMPLATE_DIR
 ) -> Template:
+    """Given dataset name, read yml template in template_dir of given dataset
+    and insert version as jinja var if provided."""
     file = template_dir / f"{dataset}.yml"
     logger.info(f"Reading template from {file}")
     with open(file, "r") as f:
@@ -64,6 +67,9 @@ def read_template(
 
 
 def get_version(template: Template, timestamp: datetime) -> str:
+    """Given parsed dataset template, determine version.
+    If version's source has no custom logic, returns formatted date
+    from provided datetime"""
     match template.source:
         case recipes.ExtractConfig.Source.Socrata() as socrata:
             return extract_socrata.get_version(socrata)
@@ -76,6 +82,8 @@ def get_version(template: Template, timestamp: datetime) -> str:
 def get_config(
     template: Template, version: str, timestamp: datetime, file_name: str
 ) -> recipes.ExtractConfig:
+    """Simple wrapper to produce a recipes ExtractConfig from a parsed template
+    and other computed values"""
     return recipes.ExtractConfig(
         name=template.name,
         version=version,


dcpy/connectors/edm/publishing.py
```diff
@@ -55,6 +55,13 @@ def path(self) -> str:
         return f"{self.product}/draft/{self.build}"
 
 
+@dataclass
+class GisDataset:
```
fvankrieken, line 59: Can name these something else - meant to refer to datasets in `edm-publishing/datasets`

dcpy/connectors/edm/recipes.py
```diff
@@ -26,6 +30,76 @@
...
+    def s3_path(self):
+        return Path(RAW_FOLDER) / self.name / self.timestamp.isoformat()
+
+
+class DatasetKey(BaseModel, extra="forbid"):
+    name: str
+    version: str
+
+
+class ExtractConfig(BaseModel, extra="forbid"):
```
fvankrieken, line 48: This is largely ported from dcpy/extract/config.py. If these types are going to be in config files in edm-recipes, wanted them here. Much like "libraryconfig" below. For now, differentiating them by dcpy module name

fvankrieken:
First pre-review comment from @sf-dcp - add docstring to "config" types

fvankrieken:
Tests failing due to weird bpl_libraries api failure. Hopefully resolves on its own

dcpy/extract/ingest.py
```diff
@@ -1,30 +1,87 @@
...
+            )
+        case recipes.ExtractConfig.Source.Api() as api:
+            return _download_file_web(api.endpoint, f"{template.name}.{api.format}")
+        case recipes.ExtractConfig.Source.Socrata() as socrata:
+            return _download_file_web(
+                utils.Socrata.get_url(socrata), f"{template.name}.{socrata.extension}"
+            )
+
+
+def archive_raw_dataset(dataset: str, version: str | None):
```
fvankrieken, line 55: This needs a better name, it shouldn't have the same name as the recipes function which does the actual archiving. Something that better represents the scope of this - get data from source, determine metadata, archive, and return config/metadata. Please help

dcpy/connectors/edm/recipes.py
alexrichey: This feels like part of a broader conversation, and doesn't have to be addressed as part of this, but I do wonder if there's a better spot for this. I suppose I've been conceiving of `edm.connectors` as a set of utilities to interact with collections of items on s3, without necessary having a knowledge about those items. `connectors.edm.publishing` being a good example. 

Curious if anything except for `dcpy.extract` use the Source classes? The Socrata orgs in particular feel misplaced.

dcpy/connectors/edm/publishing.py
```diff
@@ -392,6 +392,41 @@ def get_data_directory_url(product_key: ProductKey) -> str:
...
+    if not s3.exists(BUCKET, _gis_dataset_path(name, version)):
+        print(_gis_dataset_path(name, version))
+        print(s3.list_objects(BUCKET, _gis_dataset_path(name, version)))
+        print(s3.exists(BUCKET, _gis_dataset_path(name, version)))
+        raise FileNotFoundError(f"GIS dataset {name} has no version {version}")
+
+
+def get_latest_gis_dataset_version(dataset_name: str) -> str:
+    ## TODO
+    ## might make more sense to just list folders, grab "max" version
```
alexrichey, line 410: 👍

dcpy/extract/ingest_scripts/doe_pepmeetingurls.py
alexrichey, line 1: Just ideating here, but is this actually a... connector to DOE?

dcpy/extract/utils.py
alexrichey, line 1: Should these go in the existing Socrata connector?

dcpy/extract/utils.py
```diff
@@ -0,0 +1,36 @@
+from datetime import datetime
+import requests
+
+from dcpy.connectors.edm import recipes
+
+
+class Socrata:
```
alexrichey, line 7: Should this go in the existing Socrata connector?

dcpy/test/extract/resources/ingest_script.py
```diff
@@ -0,0 +1,7 @@
+from dcpy.extract import TMP_DIR
+
+
+def runner():
+    path = TMP_DIR / "t.txt"
```
alexrichey, line 5: Should we use an actual temp dir for tests, distinct from what we'd actually use? (e.g. for local testing?)

dcpy/test/extract/test_utils.py
```diff
@@ -0,0 +1,18 @@
+from dcpy.connectors.edm import recipes
+from dcpy.extract import utils
+
+
+class TestSocrata:
```
alexrichey, line 5: This makes a callout to Socrata? I don't think it'll be problematic, but I suppose it's conventional to separate out these potentially flaky tests.

dcpy/extract/metadata.py
```diff
@@ -28,12 +30,34 @@ class Template(
...
+    vars = get_jinja_vars(template_string)
+    if version is None and len(vars) > 0:
+        if vars == {"version"}:
+            raise Exception(
+                "Version must be supplied explicitly to be rendered in template"
+            )
+        else:
+            raise Exception(f"Unsupported jinja vars found in template: {vars}")
+    else:
+        if not (len(vars) == 0 or vars == {"version"}):
```
alexrichey, line 55: I personally find not-or conditions hard to digest at first. Vs. `if len(vars) > 0 and vars != {"version"}`, but that maybe just be a me thing

dcpy/extract/metadata.py
```diff
@@ -28,12 +30,34 @@ class Template(
...
+def read_template(
+    dataset: str, version: str | None = None, template_dir: Path = TEMPLATE_DIR
+) -> Template:
+    file = template_dir / f"{dataset}.yml"
     logger.info(f"Reading template from {file}")
     with open(file, "r") as f:
         template_string = f.read()
-    template_string = jinja2.Template(template_string).render(**kwargs)
+    vars = get_jinja_vars(template_string)
+    if version is None and len(vars) > 0:
```
alexrichey, line 47: is empty string a valid version? (ie should we just do `if not version and ...)`

alexrichey (approved):
Lots of good stuff! Provoked a lot of big questions, so would love to talk about some of the bigger picture. Of course, it's tied in heavily to dcpy structure as well, which isn't exactly going to be figured out tomorrow. But the broad strokes make sense, so I'm happy if you want to get this in and then iterate.

dcpy/connectors/edm/recipes.py
fvankrieken: Yeah. We talked a bit about this in person obviously, but just to recap - these types by domain certainly belong more in dcpy.extract, while connectors.edm.recipes more just moves them around. But it's still nice for type safety for recipes to at least know what they are.

Would probably make sense to move out `models` of different domains to a separate module in dcpy, so that they can be organized more logically by domain but still available to code throughout dcpy without recipes connector and extract code being mutually dependent on one another

dcpy/extract/ingest_scripts/doe_pepmeetingurls.py
fvankrieken: Hmm

dcpy/extract/metadata.py
fvankrieken: I'm with you

dcpy/extract/utils.py
fvankrieken: I think so. Was debating waiting for a follow-up but think that makes enough sense to do now

dcpy/connectors/edm/recipes.py
```diff
@@ -370,6 +455,49 @@ def log_metadata(config: Config):
...
+        return {
+            "name": self.name,
+            "version": self.version,
+            "timestamp": self.timestamp,
+            "runner": self.runner,
+        }
+
+
+def get_archival_metadata(name: str, version: str | None = None) -> ArchivalMetadata:
+    logger.info(f"looking up metadata for {name}/{version}")
```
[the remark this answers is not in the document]
fvankrieken: Addressed [here](https://github.com/NYCPlanning/data-engineering/pull/654/commits/989b49f1f8b43a31116b85bceda5e5457240545d)

dcpy/test/extract/resources/ingest_script.py
fvankrieken: Addressed [here](https://github.com/NYCPlanning/data-engineering/pull/654/commits/989b49f1f8b43a31116b85bceda5e5457240545d)

dcpy/extract/metadata.py
fvankrieken: Addressed [here](https://github.com/NYCPlanning/data-engineering/pull/654/commits/989b49f1f8b43a31116b85bceda5e5457240545d)

dcpy/extract/metadata.py
fvankrieken: Addressed [here](https://github.com/NYCPlanning/data-engineering/pull/654/commits/989b49f1f8b43a31116b85bceda5e5457240545d)

dcpy/extract/ingest_scripts/doe_pepmeetingurls.py
fvankrieken: I think this one I'm going to kick down the curb until we flush out the rest of extract logic. But probably

fvankrieken: I guess just about every data source is coming from a connector of sorts. Yeah this makes sense

alexrichey: Yeah, that was sort of a shower thought, imagining a world where these are all connectors, and everything under `dcpy.connectors.*` can return a generic Dataset object that can be operated on (ie. geocoded, dumped into a db, pushed to recipes etc.) I don't know how at-odds that is with current thoughts are dcpy.extract though

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -103,3 +120,25 @@ def test_previous(
...
+        publishing._assert_gis_dataset_exists(TEST_GIS_DATASET, v_fail)
+
+
+def test_get_latest_gis_dataset_version(create_buckets, add_gis_datasets):
+    today = datetime.now().strftime("%Y%m%d")
+    assert today == publishing.get_latest_gis_dataset_version(TEST_GIS_DATASET)
+
+
+def test_download_gis_dataset(
+    create_buckets, add_gis_datasets, create_temp_filesystem: Path
```
sashafilippova, line 139: create_temp_filesystem <--- nice to know you can yield a data structure!

dcpy/connectors/edm/publishing.py
```diff
@@ -392,6 +392,41 @@ def get_data_directory_url(product_key: ProductKey) -> str:
...
+    ## by that logic, should likely try to parse str version to dcpy version object
+    ## so we don't have strange cases of an unexpected subfolder (that starts with an _)
+    ## this would let us not have to supply versions for geosupport datasets ("24a", etc)
+    metadata = s3.get_metadata(BUCKET, _gis_dataset_path(dataset_name, "staging"))
+    version = metadata.last_modified.strftime("%Y%m%d")
+    _assert_gis_dataset_exists(dataset_name, version)
+    return version
```
sashafilippova, line 417: Just to clarify: is `staging` dir here equivalent to `latest` in recipes?

fvankrieken: Essentially. They update the staging folder before it's moved to "production", so it's where we can get it soonest

dcpy/extract/ingest.py
```diff
@@ -1,30 +1,87 @@
...
 
-    ## logic to push local dataset and "computed" config to raw archive in s3
 
-    ## return Config, from which s3path should be computed
+def download_file_from_source(template: metadata.Template, version: str) -> Path:
+    """From parsed config template and version, download raw data from source
+    and return path of saved local file."""
+    match template.source:
+        ## Non reqeust-based methods
+        case recipes.ExtractConfig.Source.LocalFile() as local_file:
```
sashafilippova, line 33: curious, do we compare instances or classes here? For example, it looks like we instantiating `recipes....LocalFile()`. Also, does `local_file` refer to `template.source` or  `recipes....LocalFile()`? Don't understand

dcpy/extract/ingest.py
sashafilippova: copilot recs for naming: 
* `get_data_archive_and_return_config`
* `retrieve_raw_data_and_get_metadata`

dcpy/extract/metadata.py
```diff
@@ -0,0 +1,61 @@
...
+):
+    """Definition of a dataset for ingestion/processing/archiving in edm-recipes"""
+
+    name: str
+    acl: recipes.ValidAclValues
+
+    ## these two fields might merge to "source" or something equivalent at some point
+    ## for now, they are distinct so that they can be worked on separately
+    ## when implemented, "None" should not be valid type
+    source: recipes.ExtractConfig.Source.Options
```
sashafilippova, line 27: Could `Source.Options` live outside of `ExtractConfig`? Or is there a reason why it's packed under `ExtractConfig`? It feels confusing referencing `ExtractConfig` in `Template` because they seem to be independent entities...

fvankrieken: It was more thinking of the `ExtractConfig` as the "complete" object, but I think you're right. Ties in a bit to some of Alex's points around where these types live too. I think I wanted to not have too complex a web of interacting entities, less spread out, but per both your points its resulted in some types being in places that don't really match their domain

dcpy/extract/ingest.py
fvankrieken: It compares the type. I agree that the syntax is a little funky, but that's python for you. `local_file` is `template.source`, but the typer knows that it is a LocalFile (with a path)

dcpy/test/extract/test_extract_metadata.py
```diff
@@ -0,0 +1,30 @@
...
+        val = metadata.Template(**s)
+        assert val
+
+
+def test_jinja_templating():
+    raw_template = metadata.read_template("dcp_atomicpolygons")
+    assert isinstance(raw_template.source, recipes.ExtractConfig.Source.FileDownload)
+    assert (
+        raw_template.source.url
+        == "https://s-media.nyc.gov/agencies/dcp/assets/files/zip/data-tools/bytes/nyap_.zip"
```
sashafilippova: is it safe to assume that `dcp_polygons` won't change their url?

dcpy/test/extract/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    ## https://data.cityofnewyork.us/Social-Services/Outcome-of-Preventive-Cases-Closed-By-Borough-And-/q663-gvx6/about_data
+    ## this dataset has "no plan for updates"
+    ## it may eventually be deleted but for now is useful as a psuedo-constant
+    test_set = recipes.ExtractConfig.Source.Socrata(
+        type="socrata",
+        org=recipes.ExtractConfig.Source.Socrata.Org.nyc,
+        uid="q663-gvx6",
```
sashafilippova: Not sure what dataset this is but assuming it's not likely to change over time (uid and version)

dcpy/extract/metadata.py
sashafilippova: Got it, thank you for clarification!

sf-dcp (approved):
Exciting stuff! Does make more sense overall compared to the previous pr

dcpy/extract/ingest.py
fvankrieken: So the distinction being `template.source` is of type `LocalFile | Socrata | ...`, so any attributes of `LocalFile` cannot be accessed. But once we match it, `local_file` is essentially the equivalent of `cast(LocalFile(), template.source)` so it's type-safe to access it's class attributes as a `LocalFile`

dcpy/test/extract/test_utils.py
fvankrieken: It is not - per the comment a few lines up, there are "no plan for updates" for this dataset. Have tweaked this slightly in the latest commit (and moved this test to `socrata` connector)

fvankrieken:
> Exciting stuff! Does make more sense overall compared to the previous pr

And hopefully these PRs keep trending in that direction!

PR #655, opened 2024-03-06
Add converter from socrata -> DCP metadata

alexrichey:
This enables us to generate metadata in our format from the published dataset on Socrata. It's meant to be a little rough - I'm thinking GIS could run this from a Github action, and use the resulting metadata as a starter. For now, I'll be using it to generate metadata, just from the command line.


dcpy/connectors/socrata/metadata.py
```diff
@@ -0,0 +1,146 @@
...
+        None,
+        "--output-path",
+        "-o",
+        help="Path to export the metadata to",
+    ),
+):
+    logger.info(f"fetching metadata for {four_four}")
+    md = metadata.make_dcp_metadata(
+        pub.Dataset(four_four=four_four).fetch_metadata()
+    ).model_dump()
```
fvankrieken, line 117: Just talked about this in person, but might be able to specify `exclue_none` here? No worries if you'd prefer to be more explicit below anyways

alexrichey: @fvankrieken yeah, let me investigate that. I'd also like to look into that for recipes, so that the `recipe.lock.yml` files aren't so bloated.

alexrichey: @fvankrieken  implemented

dcpy/connectors/socrata/metadata.py
```diff
@@ -0,0 +1,129 @@
...
+        "name": c["fieldName"],
+        "display_name": c["name"],
+        "description": c["description"],
+        "data_type": dcp_type,
+        "example": sample,
+    }
+    if c["renderTypeName"] == "multipolygon":
+        dcp_col["readme_data_type"] = "geometry"
+
+    ENTIRELY_ARBITRARY_MAX_STANDARDIZED_VALS = 10
```
fvankrieken, line 46: Great name

dcpy/connectors/socrata/metadata.py
```diff
@@ -0,0 +1,129 @@
...
+    if (
+        samples
+        and len(samples) < ENTIRELY_ARBITRARY_MAX_STANDARDIZED_VALS
+        and len(samples) == int(c["cachedContents"].get("cardinality"))  # type: ignore
```
fvankrieken, line 50: Not that everything needs to be 100% type safe, in particular when working with json/dict responses from apis, but just curious on your thoughts between pydantic types and typeddict for these objects, having done a bit of work with them

dcpy/connectors/socrata/metadata.py
```diff
@@ -0,0 +1,129 @@
...
+    ):
+        dcp_col["values"] = [
+            [s["item"], "<FILL_IN_THE_VALUE_DESCRIPTION!>"] for s in samples
+        ]  # type: ignore
+    return models.Column(**dcp_col)  # type: ignore
+
+
+def make_dcp_metadata(socrata_md):
```
fvankrieken, line 58: type signature? Code is relatively self-evident but it still helps coming in to this code a bit unfamiliar, even if it's just `dict`

dcpy/connectors/socrata/publish.py
```diff
@@ -69,6 +69,17 @@ class Attachment(TypedDict):
...
+        class Column(TypedDict):
+            """Column fields returned from the Socrata API."""
+
+            name: str  # ex: Project Name
+            fieldName: str  # ex: projectnam
+            description: str
+            renderTypeName: str
+
+            # cached attributes of a dataset, e.g. top 10 most common values,
+            cachedContents: NotRequired[dict]
```
fvankrieken, line 81: just curious, do you know what this refers to in terms of "cached"? Like, why is something cached in socrata? When do we see cached contents in a response vs not?

alexrichey: That's a great question. For MIH, the only column without `cachedContents` is `the_geom`. So maybe certain types aren't cacheable.

dcpy/connectors/socrata/metadata.py
alexrichey: That's a good question... In this case, I was using typeddict more for documentation purposes than anything, but it might not be a bad idea to just default to one or the other.

dcpy/connectors/socrata/metadata.py
alexrichey: @fvankrieken added types

fvankrieken (approved):

PR #656, opened 2024-03-06
revise Green Fast Track

damonmcc:
related to https://github.com/NYCPlanning/edm-overview/issues/1029

Most of the apparent changes are due to reorganizing/renaming sql files to align more with our new [dbt Code Conventions](https://github.com/NYCPlanning/data-engineering/wiki/Code-Conventions#dbt).

Could do more model renaming to more strictly adhere to the proposed conventions, but would rather do that in a follow-up PR so it can be focused on and not hold up these changes to data requested by GIS.

The more significant changes here are
- use the latest Vents source data from GIS
- as requested by GIS, change exported source data to union all buffers (e.g. a single geometry for all elevated railway buffers)

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-gft-sources)


products/green_fast_track/models/intermediate/int__spatial_flags.sql
damonmcc, line 1: build [before](https://github.com/NYCPlanning/data-engineering/actions/runs/8192787339/job/22405101455#step:7:278) and [after](https://github.com/NYCPlanning/data-engineering/actions/runs/8193380942/job/22407019864#step:7:278)

I think this saves 2 minutes

.pre-commit-config.yaml
damonmcc, line 1: these changes make the local use of `pre-commit run --all-files` easier by eliminating the need to run the following commands first:
```bash
dbt clean --no-clean-project-files-only --profiles-dir products/green_fast_track --project-dir products/green_fast_track
dbt docs generate --profiles-dir products/green_fast_track --project-dir products/green_fast_track
```

.github/workflows/test_helper.yml
```diff
@@ -74,7 +74,7 @@ jobs:
     runs-on: ubuntu-22.04
     container: nycplanning/dev:${{ inputs.image_tag }}
     env:
-      BUILD_NAME: ${{ github.head_ref || github.ref_name }}
+      BUILD_NAME: ${{ github.head_ref || github.ref_name }}__pr_tests
```
damonmcc, line 77: explanation from relevant commit:

> use a more distinct schema for PR-based dbt formatting tests
> 
> the pre-commit tests require a DB connection and running `dbt seed`. if they use the simple branch name schema, there are collisions with builds

.pre-commit-config.yaml
sf-dcp: can you explain why do we want to generate dbt docs as a part of running dbt tests? And are they accessible anywhere after running the tests?

alexrichey:
Might be nice to automate creating a lineage graph...

.pre-commit-config.yaml
damonmcc: sure! (and I should add the first part of this to our wiki)

for many of the linting tests we do/would want to run using `dbt-checkpoint` require the existence of `manifest.json` and `catalog.json`. those files are artifacts of compiling the dbt project. the project can be compiled by generating docs or running all the queries. since this linting is most valuable to run _before_ building, generating docs is a recommended approach to enable this linting ([docs](https://github.com/dbt-checkpoint/dbt-checkpoint?tab=readme-ov-file#setup))

we don't have access to these docs since they're generated during tests on every branch and nothing is done with the generated files. to @alexrichey's note of `Might be nice to automate creating a lineage graph`, these docs could be served up in some way to allow us to inspect the graph of a dev branch. and the docs generated by `main` would be a source of truth for our data products!

PR #658, opened 2024-03-07
Data library: convert raw to parquet

### LINKED ISSUE #631: Data Library - Convert raw data to parquet

sf-dcp:
Related to #631.

### Background 
As described in the issue, we would like to explore `GDAL` alternatives for translating data formats into parquet. I did research on available tools/packages, and I found that 1) most of them, if not all, are built on top of GDAL and 2) only `pandas`+`geopandas` and `GDAL` are able to work with a variety of data formats (both spatial and non-spatial data) as opposed to a small set of formats. Btw, `geopandas` uses Fiona which in turns is an api to GDAL. 

The goal of this PR is a POC to replace `GDAL` with `geopandas` for a format translation. And the reason why we are exploring `GDAL` alternatives is because we are looking for better code readability. 

### Commits

1) With the new data library implementation, templates are read and "computed" in the `ConfigExtract` object. This commits expands `ConfigExtract` with ` transform_to_parquet_metadata` section via `ToParquetMeta` class. The idea is that this section includes all necessary info for translating raw data into parquet. 

2) Create the `transform_to_parquet` fn. Note, I implemented conversion of only 3 input data formats: `csv`, `geodabase`, and `shapefile`. If we decide to move forward with this PR approach, need to add zipped formats, json, geojson, and excel. Some details below: 
* Text data formats can be spatial or non-spatial. If it doesn't have a geometry, the fn reads a `csv` into regular pandas df and dump into parquet. The process is a bit more complicated if a `csv` does have spatial data:     
     * read it into regular pandas df, 
     * convert column containing geometry into spatial, 
     * convert df into geopandas df, and 
     * dump into parquet.
  
* With geospatial data formats the process is straightforward. Read data into geopandas and save into parquet. Boom.

* There is an "interesting" (read annoying) quirk with dtypes and `NaN` value in pandas which impact translation of **text** data into parquet. It doesn't impact data formats that can be read into `geopandas` directly. I outline them in the PR comments below.
 
3) Create fake data for tests and the test itself. It tests `csv`, `shapefile`, and `geodabase` formats. The logic is a bit convoluted around creating test `ConfigExtract` objects. Each test is run on 1 data format at a time. 

4 & 5) Fix other tests that use `ConfigExtract` or `Template` objects. 

### 🚨 Looking for feedback 
* General input about replacing `GDAL` with `pandas` + `geopandas`
* Structure of `ToParquetMeta` class. Perhaps it needs to be more strict what attributes are expected depending on type
* A better way to organize test objects and artifacts like test yaml templates. 
* Function/class/variable naming

Sorry for the long PR description. Lmk if it's easier for me to do a walk-through! 

**Edit (as of 3/25):** 
This is first iteration of the `transform_to_parquet` fn. `TODO` items for next PR:
- [ ] implement json
- [ ] geojson
- [ ] zipped format
- [ ] for csv files, implement point geom made out of Longitude & Latitude columns
- [ ] geoparquet (as an export format)
- [ ] name geometry column as `geom` in output parquet file
- [ ] break up `transform_to_parquet` fn into 2 parts: 1) read raw data into a `geopandas` df and 2) save the df to parquet


sf-dcp:
### Promised `pandas` quirks description

* gpd dataframe MUST have at lease 1 geom column. It will throw an error if it's not present. It can read only 1 geom column into gpd dataframe. Other geom columns will be held as a string or a wkt. 

* When saving to parquet, gpd df will only convert "active" geom column into wkb

* Classification of data types in `pandas`:
     * column contains only integers (without missing values) --> integer column
     * column contains numbers with decimals (with or without missing values) --> float column 
     * column contains mix of numbers and non-numeric values (strings) --> object column
     * column with integers and missing values (`NaN`) --> is automatically converted into a column with a floating-point type (float64) to accommodate the `NaN` values.

dcpy/extract/ingest.py
```diff
@@ -84,10 +90,92 @@ def extract_and_archive_raw_dataset(dataset: str, version: str | None):
...
+    """
+
+    # create new dir for raw data and output parquet file
+    if TMP_DIR.is_dir():
+        shutil.rmtree(TMP_DIR)
+    TMP_DIR.mkdir()
+
+    if local_data_path:
+        # TODO: fix logic for cases when it's a directory, not a file
+        assert local_data_path.is_file(), "Local path should be a valid file"
```
damonmcc, line 122: if this raises an error when `local_data_path` is a directory, that seems like a good thing

dcpy/test/extract/resources/transform_to_parquet_template.yml
damonmcc, line 1: looks like the formatting of these might be wrong because it makes `encoding` the top-level key and everything else it's value. I think the tests are working because they iterate through these and then use the value of `format` to get a file

seems like these could something like
```yaml
- config:
  encoding:
  format: csv
  delimiter:
  geometry:
    geom_column: wkt
    layer:
    crs: EPSG:4326
```

dcpy/test/extract/conftest.py
damonmcc, line 1: yay fake data!

dcpy/test/extract/resources/transform_to_parquet_template.yml
fvankrieken: There's no top-level key here - it's an array element that's a dict, so encoding, format, delimiter, geometry are all at the same level

fvankrieken: I will say though, if these fields are optional, it's a bit cleaner in my opinion to just not have `encoding`, `delimiter`, etc rather than have value-less keys. But let's maybe brainstorm a bit about these types before making any changes

damonmcc (approved):
lgtm!! just left a couple comments

and I tried to think of other names for `transform_to_parquet_metadata` and `ToParquetMeta` but couldn't think of any that were worth holding up these (great) changes for

dcpy/extract/ingest.py
```diff
@@ -84,10 +90,92 @@ def extract_and_archive_raw_dataset(dataset: str, version: str | None):
...
+        )
+        logger.info(f"Downloaded raw data from s3 to {local_data_path}")
+
+    data_load_config = config.transform_to_parquet_metadata
+
+    match data_load_config.format:
+        case "shapefile" | "geodabase":
+            gdf = gpd.read_file(
+                local_data_path,
+                geometry=data_load_config.geometry.geom_column,
```
fvankrieken, line 140: Are these fields allowed to be null? For gpd.read_file's signature, I mean. 

For shapefile and geodatabase (typo there btw on line 137), is that a relevant field? Or since it's a special field, can gpd just tell without it being specified?

dcpy/extract/ingest.py
```diff
@@ -84,10 +90,92 @@ def extract_and_archive_raw_dataset(dataset: str, version: str | None):
...
+            )
+
+            if (
+                not data_load_config.geometry.crs
+                and not data_load_config.geometry.geom_column
+            ):
+                gdf = df
+
+            else:
+                geom_column = data_load_config.geometry.geom_column
```
fvankrieken, line 160: Would be good to have validation that this column exists in the df

fvankrieken:
Looks great. I'd like to spend a little more time looking at the tests closely, and then I think you and I should get together to chat about cleaning up the ToParquetMeta class (and validation around it), but excited to see this where it is

dcpy/extract/ingest.py
sf-dcp: I was thinking about a zipped shape file... But it's probably considered as a file, not a folder

dcpy/test/extract/resources/transform_to_parquet_template.yml
sf-dcp: I agree. I was also considering whether we need to have classes for each data format with a specific set of attributes versus having it like this. For example, `delimiter` is a csv attribute only while `layer` is relevant to geodabase files.

dcpy/extract/ingest.py
sf-dcp: are you thinking to have validation in code here or in a data contract?

fvankrieken: To me that sort of validation belongs here - if we're explicitly providing geospatial information as part of this conversion, I'd prefer a helpful error at time of conversion.

As written now, I think we'd simply get a pandas error. Which is relatively helpful, but my gut response is that it'd be helpful to see something like "specified geometry column does not exist in test.csv" or something like that

dcpy/test/extract/resources/transform_to_parquet_template.yml
fvankrieken: Yeah my instinct is to go that way, just because validation then simply happens at time of reading in from yml

dcpy/extract/ingest.py
sf-dcp: Update: `gdb` format is considered a directory. My test didn't catch the assert statement because I misspelled geodatabase as geodabase 🥲 Socrate moment

dcpy/extract/ingest.py
sf-dcp: Thanks for catching the misspelling! I realized I never ran a test on "geodabase" format because of it. I will fix it. 

Regarding your questions, I would say no for geospatial data formats because it successfully converts the geom column to wkb format in parquet (which is specific to geom columns): 

Screenshot 2024-03-19 at 11 30 37 AM

Having said that, `geom_column` is relevant for cases when we have multiple geometries

fvankrieken: Do geodatabases and shapefiles support multiple geometries? For a single layer in a geodatabase.

sf-dcp: Now that I'm thinking about it, they don't support it. It's more relevant to parquet/geoparquet, and text data formats. Should I remove the `geom_column` from the fn signature?

dcpy/test/extract/resources/transform_to_parquet_template.yml
sf-dcp: Per chat with Finn, new structure will look like this and will be enforced with pydentic classes by data format: 

```yaml
- format: csv
  delimeter: # optional
  encoding: # optional
  geometry: # optional
    geom_column: # required for geospatial csvs. str (if geometry object) or dict of x and y columns (lon and lat)
      x:
      y:
    crs: # required for geospatial csvs

- format: excel
  tab: # required
  encoding: # optional
  geometry: # optional
    geom_column: # required for geospatial csvs. str or dict of x and y columns
      x:
      y:
    crs: # required for geospatial csvs

- format: shapefile
  encoding: # optional
  crs: # required

- format: geodatabase
  encoding: # optional
  crs: # required
  layer: # optional
```

fvankrieken:
Did you run an actual dataset? Go from source -> archive raw -> parquet

sf-dcp:
> Looks great! In a way the last test is a bit odd, we're sort of having a lot of code to make sure that we're simply not losing/mutating data when we export to parquet. It might make sense to break the main function up into two functions: input dataset to dataframe, then dataframe to parquet. Then there wouldn't be the need to recreate a lot of the function in the test itself.

That makes sense! I will add it to my todo list for next PR. I was also thinking of creating complete fake templates for our `test_ingest.py` instead of having a list of items in a yaml file.

fvankrieken:
Yeah testing for all of this probably also warrants some brainstorming. After the code reorg!

sf-dcp:
> Did you run an actual dataset? Go from source -> archive raw -> parquet

I relied on the tests. When I tried running `run.py`, I was getting `ImportError` because of relative paths. How do you run it?

fvankrieken:
`python3 -m dcpy.extract.run archive_raw bpl_libraries`

Issue #665, opened 2024-03-08
update columns in KPDB aggregation tables


damonmcc:
Make changes to KPDB align with [Schema Doc](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7B93462D17-6D0B-485B-BB44-53CCFAF1338C%7D&file=field_descriptions_20240304.xlsx&action=default&mobileredirect=true) from GIS. That doc lists the fields used in an in-development ESRI app to show past and future housing development.

That app expects KPDB data aggregated by NTA and CDTA with the following columns for each aggregation type:

## column details from Schema Doc

| column name | description |
|--------|--------|
| within_5_years | # of units that should be completed within the next 5 years |
| from_5_to_10_years | # of units that should be completed within the next 5-10 years |
| after_10_years | # of units that should be completed sometime after 10 years from this year |
| future_phased_units_total | Total units included in three kpdb phasing buckets; Aggregate of within_5_years, from_5_to_10_years, after_10_years  |
| future_units_without_phasing | Uncompleted units without phasing |
| kpdb_completed_units | Units that are likely included in HousingDB--but are definitely not future units. Perhaps pending Cert of Occ? |
| units_net | The number of future units that should be counted as part of the project after removing completed DOB units and deduplication |
| units_gross | The proposed number of units associated with a project in geography as reported in the source data |
| records_in_cdta | How many projects not necessarly in app |
| senior_housing | Count of projects in geography where project is at least partially senior housing |
| nycha | Count of projects in geography where project is at least partially a NYCHA project |
| classb | Count of projects in geography where project is at least partially class b |

damonmcc:
only columns left are `senior_housing`, `nycha`, and `classb` but they aren't urgent due to data quality concerns

PR #666, opened 2024-03-08
revise KPDB aggregations

damonmcc:
related to #665. this doesn't resolve it because this doesn't add `senior_housing`, `nycha`, and `classb` to aggregation outputs

see [commit messages](https://github.com/NYCPlanning/data-engineering/pull/666/commits) for all changes

successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9878675634)

### screenshots of new columns in aggregation outputs

Screenshot 2024-07-10 at 2 10 35 PM

Screenshot 2024-07-10 at 2 09 41 PM

Screenshot 2024-07-10 at 2 10 10 PM

Screenshot 2024-07-10 at 2 10 59 PM


products/knownprojects/sql/create_kpdb.sql
```diff
@@ -32,44 +33,55 @@ WITH
...
 ),
 
 future_units_details AS (
     SELECT
         *,
+        NULLIF(units_net, 0) IS NOT NULL AND completed_units = 0 AS has_future_units,
         CASE
             WHEN
-                has_future_units AND NOT has_project_phasing
+                NOT has_project_phasing AND completed_units = 0
```
fvankrieken, line 52: just clarifying this is intended as `(NOT has_project_phasing) AND completed_units = 0` since that is the way postgres interprets it

fvankrieken: It seems like it just checking. Also - what's the reason to use logic on this unit value rather than its status?

products/knownprojects/sql/aggregate/future_units_by_cd.sql
```diff
@@ -23,7 +23,9 @@ future_units_with_cd AS (
         units_gross,
```
fvankrieken, line 23: these are real ripe to be built with dbt

fvankrieken: (but that should wait till doing kpdb entirely - much prefer only the intermediate of testing with dbt, building entirely outside of it)

products/knownprojects/sql/create_kpdb.sql
damonmcc: yes to ur first comment, think it's worth adding the parentheses?

fvankrieken: Nope!

damonmcc: logic on the phased unit values seems more accurate and easier to understand than the status

the status values do determine how units get allocated to the the different phases (`within_5_years`, `from_5_to_10_years`, `after_10_years`). I added the table `summary_record_phasing` to show record counts by `source + status`

Screenshot 2024-07-10 at 2 52 53 PM

fvankrieken (approved):

PR #671, opened 2024-03-11
Ar add product metadata

alexrichey:
Adds product metadata for E-des, Facilities, and a few others. It's probably easiest to go commit by commit.

Commits 1 + 2: Raw extracts of metadata from Socrata for the datasets. (you can probably skip these. They're unaltered outputs from a script)
Commits 3 and onward: cleaning up those files. We can automate some of this cleanup later. But there's some really funky formatting that was probably manually done in Socrata by whoever initially set up the dataset. 

This is a great occasion to critique anything in the metadata structure that could be improved. A few things spring to mind:
- Column types, especially numeric ones (ie, should we ever be distinguishing between floats and doubles? Are BBLS numeric or just text?)
- Terminology around dataset packages. (see [related discussion](https://github.com/NYCPlanning/data-engineering/discussions/793#discussioncomment-9217232))

Also, apologies in advance for the extra commits around e-des. Renaming that file made it difficult to maintain a clean commit history. 

Also, important to note, this metadata will have to be tweaked when we go to actually automate the push of these datasets. And before we actually publish anything, we need to have GIS involved in reviewing the metadata itself. But this is just a first pass at generating product metadata from scraped Socrata datasets.


products/facilities/metadata.yaml
```diff
@@ -65,14 +45,14 @@ dataset_package:
       omit_columns: []
       columns: {}
   attachments:
-  - facilities_23v2_metadata.zip
+  - facilities_metadata.zip
```
alexrichey: It's still an unanswered question how we deal with changing attachment or file names. Maybe there needs to be some templating. For now, I think it's best to just change the filename in our package.

products/facilities/metadata.yml
```diff
@@ -0,0 +1,303 @@
...
+  - [seats, ""]
+  - [beds, ""]
+  - [routes, ""]
+- name: optype
+  display_name: optype
+  description: Indicates whether the operating entity is public or non-public.
+  data_type: text
+  example: Non-public
+  values:
+  - [Non-public, ""]
```
alexrichey, line 236: These picklist values are generated by the scraper when it sees fewer than (IIRC) 10 or so values in the column metadata. The format of the picklist values is a two element list, with the raw value, and the human readable variety. In the cases where the raw value matches the description, I've just left them blank. 

A different case might be boro code, where it's a number, but the description is the boro name itself. e.g. "1" -> "Manhanttan"

damonmcc (approved):

Issue #672, opened 2024-03-11
archive GFT Natural Resources data


damonmcc:
Archived
- [x] State Regulated Freshwater Wetlands and Checkzone
  - [x] checkzone specifically 
- [x] Tidal Wetlands
- [x] Priority Waterbodies
- [x] Significant Natural Communities
- [x] Beaches
- [x] National Wetlands Data

GFT Logic written
- [ ] State Regulated Freshwater Wetlands and Checkzone
- [ ] Tidal Wetlands
- [ ] Priority Waterbodies
- [ ] Significant Natural Communities
- [ ] Beaches
- [ ] National Wetlands Data

other
- [x] `variables.csv` updated

fvankrieken:
@croswell81 @jackrosacker - state-regulated freshwater wetlands comes with a buffer, the described "check zone" described [here](https://services6.arcgis.com/DZHaqZm9cxOD4CWM/arcgis/rest/services/State_Regulated_Freshwater_Wetlands/FeatureServer). It is "a minimum of 100 feet, but has been extended for a limited number of particularly sensitive wetlands". In our excel file, we state a buffer of 100 feet - should I just go with 100 feet, or use the supplied check zones?

jackrosacker:
The data source sheet may need to be edited to consider the wetland and check zone sources as two different datasets and questions. Alex M. confirmed that the check zone question has distinct implications and would need to be asked regardless of the answer to the wetland+ question. 

The preliminary language we have for these is:
- Does your project site contain a [Natural Resource]?
and
- Is your project site within a wetland check zone?

We still need to vet and finalize the language and data categorization here, but from what's been communicated to us it sounds like either a project **contains** a wetland, and/or the project is **within** a wetland check zone. I'd tentatively say that we wouldn't need to calculate 100' buffers, but would just look for a spatial relationship between each lot and the existing check zone polygons.

Matt may have some different info on this - curious to check in later today.

croswell81:
I need to clean-up the natural resources and open space sections of this document. Might be helpful for us to check-in and discuss what's left so we are all on the same page of what is final and what might need some clarification from either Planning Support or EARD.
@fvankrieken @jackrosacker

fvankrieken:
I think it'd be good to chat at some point today. There's also this comment in the natural resources section
> SS wants to touch base with Parks and see if they can share their dataset to avoid duplication
[...]

PR #676, opened 2024-03-12
ESRI Arcgis Feature Server connector

fvankrieken:
Add a new connector for [ESRI Arcgis Feature Service](https://developers.arcgis.com/rest/services-reference/enterprise/feature-service.htm) and use it in library. Keeping this as only part of #672 so that others can use this connector

[Working action](https://github.com/NYCPlanning/data-engineering/actions/runs/8266756531)

### DIFF
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -14,6 +14,7 @@ import yaml
 
 from dcpy.utils import s3, postgres, metadata
 from dcpy.utils.logging import logger
+from dcpy.connectors.esri.arcgis_feature_service import FeatureServer
 
 
 BUCKET = "edm-recipes"
@@ -176,6 +177,7 @@ class LibraryConfig(BaseModel):
         url: Url | None = None
         script: Script | None = None
         socrata: Socrata | None = None
+        arcgis_feature_server: FeatureServer | None = None
         layer_name: str | None = None
         geometry: GeometryType | None = None
         options: list[str] | None = None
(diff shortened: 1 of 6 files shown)


dcpy/connectors/esri/arcgis_feature_server.py
```diff
@@ -0,0 +1,52 @@
...
+from pydantic import BaseModel
+from typing import cast
+
+
+class Server(StrEnum):
+    nys_clearinghouse = "nys_clearinghouse"
+
+    @property
+    def id(self):
+        match self:
```
fvankrieken, line 12: This is silly at the moment, I'll just make a dict and non-class function. Will probably play nicer with pydantic anyways

docker/config.sh
```diff
@@ -2,7 +2,7 @@
 
 # Logic in this file is used within builds of docker containers
 
-COMMON_APT_PACKAGES="curl zip unzip git wget ca-certificates lsb-release build-essential sudo postgresql-client-15 libpq-dev jq locales"
+COMMON_APT_PACKAGES="curl zip unzip git wget ca-certificates lsb-release build-essential sudo postgresql-client-15 libpq-dev jq locales libkrb5-dev"
```
fvankrieken, line 5: Needed to install arcgis python package. Even if we don't go that route, might leave, it's a pretty insignificant package to include

fvankrieken:
Going to incorporate this in extract in my extract PR follow-up to #654

alexrichey (approved):
Cool - looking forward to giving it a spin

damonmcc:
nice! and looks like no need for an esri python package?

fvankrieken:
Not for now at least. `arcgis` is actually pretty nice to use, but this was easy enough to implement (thanks for some points in the right direction @jackrosacker), and `arcgis` had a ton of dependencies, so seemed tidy just to keep things lighter-weight for the time being

PR #685, opened 2024-03-14
Fvk gft natural resources

fvankrieken:
First half of #672

### DIFF
--- a/dcpy/connectors/esri/arcgis_feature_service.py
+++ b/dcpy/connectors/esri/arcgis_feature_service.py
@@ -57,7 +57,7 @@ def query_dataset(dataset: FeatureServer, params: dict) -> dict:
 
 
 def get_dataset(dataset: FeatureServer, crs: int) -> dict:
-    CHUNK_SIZE = 2000
+    CHUNK_SIZE = 100
     params = {"where": "1=1", "outFields": "*", "outSr": crs, "f": "geojson"}
 
     # there is a limit of 2000 features on the server, unless we limit to objectIds only
--- a/dcpy/library/templates/dcp_beaches.yml
+++ b/dcpy/library/templates/dcp_beaches.yml
@@ -1,0 +1,23 @@
+dataset:
+  name: dcp_beaches
+  acl: public-read
+  source:
+    url:
+      path: s3://edm-recipes/inbox/dcp_beaches/nyc_beaches.zip
+      subpath: nyc_beaches.shp
+    geometry:
+      SRS: EPSG:2263
+      type: MULTIPOLYGON
+
+  destination:
+    geometry:
+      SRS: EPSG:2263
+      type: MULTIPOLYGON
+    options:
+      - "GEOMETRY=AS_WKT"
+
+  info:
+    description: |
+      ## DCP NYC Beaches
+      GIS team-produced dataset that merges DPR nyc beaches dataset with manually
+      created features from NPS national beaches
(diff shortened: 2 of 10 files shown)


dcpy/connectors/esri/arcgis_feature_service.py
```diff
@@ -57,7 +57,7 @@ def query_dataset(dataset: FeatureServer, params: dict) -> dict:
 
 
 def get_dataset(dataset: FeatureServer, crs: int) -> dict:
-    CHUNK_SIZE = 2000
+    CHUNK_SIZE = 100
```
fvankrieken, line 60: @alexrichey any thoughts on if maybe this should just be an override for specific problem datasets? Gut is yes

fvankrieken: But also simple to just keep it low since performance isn't really an issue

alexrichey: Interesting... so does this resolve the issue with the sporadic failures? If so, I'd probably just go the simplest route, and just hardcode like you've done here.

fvankrieken: It did. NYS GIS Clearinghouse seemed to error out on any given query at exactly 1 minute

dcpy/library/templates/dcp_beaches.yml
```diff
@@ -0,0 +1,23 @@
+dataset:
+  name: dcp_beaches
+  acl: public-read
+  source:
+    url:
+      path: s3://edm-recipes/inbox/dcp_beaches/nyc_beaches.zip
```
alexrichey, line 6: grim thought: I hope we never need to introduce versions for this dataset

fvankrieken: 🙁

dcpy/library/templates/nysdec_freshwater_wetlands_checkzones.yml
```diff
@@ -0,0 +1,25 @@
+dataset:
+  name: nysdec_freshwater_wetlands
+  acl: public-read
+  source:
+    arcgis_feature_server:
+      server: nys_clearinghouse
+      name: State_Regulated_Freshwater_Wetlands
+      layer: 1
+    geometry:
+      SRS: EPSG:2263
```
alexrichey, line 10: [This](https://services6.arcgis.com/DZHaqZm9cxOD4CWM/ArcGIS/rest/services/State_Regulated_Freshwater_Wetlands/FeatureServer/1) says `Spatial Reference: 26918 (26918)`. Does that matter?

fvankrieken: I believe that's meant for the extent. Currently, this geometry gets fed into the call to arcgis connector, and queries for the data specifically in the given SRS

fvankrieken: The REST service supports that. So to answer your question - not entirely, though it's maybe not great that there's this uncertainty about where transformation happens

dcpy/connectors/esri/arcgis_feature_service.py
alexrichey: ah... interesting. Might make sense down the road to add some retry/backoff logic, but this seems great for now

alexrichey (approved):

PR #698, opened 2024-03-18
Macro to clip to geometry

fvankrieken:
Bunch of tidying commits so it'll be simpler going by commit

1-4 are all a bit of housekeeping really, both in my templates that I created and in our utils

5 includes boro boundaries, merges them to one geom, and adds the macro to clip to geometry

Moved my natural resources-specific code to a new branch

### DIFF
--- a/dcpy/builds/load.py
+++ b/dcpy/builds/load.py
@@ -63,6 +63,8 @@ def import_dataset(
     else:
         preproc_func = None
 
+    if pg_client and (not ds.destination):
+        ds.destination = plan.InputDatasetDestination.postgres
     assert ds.destination, f"Dataset destination not resolved for dataset {ds.name}"
     match ds.destination:
         case plan.InputDatasetDestination.postgres:
--- a/products/green_fast_track/macros/clip_to_geom.sql
+++ b/products/green_fast_track/macros/clip_to_geom.sql
@@ -1,0 +1,9 @@
+{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom") %}
+
+    SELECT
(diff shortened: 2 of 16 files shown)


dcpy/builds/load.py
```diff
@@ -63,6 +63,8 @@ def import_dataset(
     else:
         preproc_func = None
 
+    if pg_client and (not ds.destination):
+        ds.destination = plan.InputDatasetDestination.postgres
```
fvankrieken, line 67: Really this is to make the cli command to import a single dataset simpler. Doesn't make a ton of sense to import a dataset to anything other than pg database

products/green_fast_track/macros/clip_to_geom.sql
```diff
@@ -0,0 +1,9 @@
+{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom") %}
+
+    SELECT
+        {{ dbt_utils.star(from=left, except=[left_by]) }},
+        ST_INTERSECTION({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }}) AS geom
+    FROM {{ left }}
```
fvankrieken, line 6: might be worth adding ability to get some sort of identifier from `right`. For now, stg__nyc_boundary only has one row so it's not relevant. But you could imagine that if you want to clip to boroughs, each row returned should have a boro_id attached

damonmcc: definitely would be a nice feature, but I think my brain processes the singular geom in `clip_to_geom` as involving a one row in the right table anyway

damonmcc (approved):

PR #699, opened 2024-03-19
gft dbt code organization

fvankrieken:
Follow up to #656, renaming models to align with our [dbt Code Conventions](https://github.com/NYCPlanning/data-engineering/wiki/Code-Conventions#dbt)

[Passing build](https://github.com/NYCPlanning/data-engineering/actions/runs/8348664722/job/22851121957)


fvankrieken:
There are some open questions for us still around conventions - when and where we want to be testing. We have some duplicative tests for tables that are then just gettting unioned, and then we're running the same tests again. But for now, very happy to be overly prescriptive and pare down later as we'd like

fvankrieken:
Also somehow I dropped the boroboundaries while rebasing in #698 , so that is here

damonmcc (approved):

PR #703, opened 2024-03-22
GFT: Add Shadows models

sf-dcp:


products/green_fast_track/macros/clip_to_geom.sql
sf-dcp, line 1: I added an additional param for crs. Primarily want to make sure both models always have the same crs before they get spatially joined

products/green_fast_track/seeds/nyc_parks_properties_categories.csv
sf-dcp, line 1: This seed is used in 2 places: 
* on line `148` in `models`/`_sources.yml` to test that there are no categories outside of this list.
* in the `models/staging/stg__nyc_parks_properties.sql` to filter data by `typecategory` column where `allowed=TRUE`

products/green_fast_track/models/staging/stg__nys_parks_properties.sql
```diff
@@ -0,0 +1,17 @@
...
+    FROM {{ source('recipe_sources', 'nysparks_parks_polygons') }}
+),
+
+final AS (
+    SELECT
+        'nys_parks_properties' AS variable_type,
+        objectid AS variable_id,
+        st_transform(wkb_geometry, 2263) AS raw_geom
+    FROM source
+    WHERE upper(county) IN ('BRONX', 'KINGS', 'QUEENS', 'RICHMOND')
```
damonmcc, line 14: missing `MANHATTAN`, I'm seeing two records with that county value

products/green_fast_track/seeds/nyc_parks_properties_categories.csv
fvankrieken: Nice. I like the check to ensure that the list doesn't fall out of date. Would be a good check to live in extract code at some point!

products/green_fast_track/seeds/variables.csv
```diff
@@ -12,3 +12,9 @@ arterial_highways,Arterial Highways,Noise,16,"75 foot buffer"
 elevated_railways,Elevated Railways,Noise,17,"1,500 foot buffer"
 airport,Airport,Noise,18,"Intersects 65db contour"
 e_designations_hazmat,(E) Designations - HazMat,HazMat,HZ1,"HazMat Code = 1"
+nyc_parks_properties,NYC Parks properties,Shadow/Open Space,23,"75 foot buffer. Specific list of allowed typecategory values"
+pops,POPS,Shadow/Open Space,23,"75 foot buffer around BBL lot"
+waterfront_access_wpaa,Waterfront Public Access Areas,Shadow/Open Space,23,"75 foot buffer"
+waterfront_access_pow,Publicly Owned Waterfront,Shadow/Open Space,23,"75 foot buffer"
+nys_parks_properties,NYS Parks Property,Shadow/Open Space,23,"75 foot buffer"
+us_parks_properties,Federal Parks property,Shadow/Open Space,23,"75 foot buffer"
```
fvankrieken, line 20: Don't have to do it now, but probably worth adding a check to ensure that all these variables are represented in "all flags" table, and that "all flags" doesn't have any that aren't present here. I need to wrap up my PR so I can include that

products/green_fast_track/macros/clip_to_geom.sql
```diff
@@ -1,8 +1,11 @@
-{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom") %}
+{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom", crs=2263) %}
 
     SELECT
         {{ dbt_utils.star(from=left, except=[left_by]) }},
-        ST_INTERSECTION({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }}) AS geom
+        ST_INTERSECTION(
+            ST_TRANSFORM({{ left }}.{{ left_by }}, {{ crs }}),
+            ST_TRANSFORM({{ right }}.{{ right_by }}, {{ crs }})
```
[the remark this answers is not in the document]
sf-dcp: Per stand up this morning, the rational to modify the macro is because it didn't work with a CTE for a stage table (usparks). And I couldn't use it directly on source because it was in a different projection. Finn and I will check in to figure out an approach.

products/green_fast_track/models/staging/stg__nys_parks_properties.sql
sf-dcp: In the GDE notes for this variable in the excel sheet, it is the following:
> Filter down to NYC, Use County field and select NYC Counties: Bronx, Kings, Queens, Richmond (none Manhattan - New York)

I understood it as not to include Manhattan. Does that mean the opposite?

@damonmcc

damonmcc: oh sorry I didn't see that! seems odd to exclude Manhattan but the excel sheet says to, so never mind my comment!

products/green_fast_track/macros/clip_to_geom.sql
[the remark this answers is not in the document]
sf-dcp: Would this piece of code live in the macro?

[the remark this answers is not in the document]
sf-dcp: I'm not sure how to implement this...

sf-dcp:
Used Finn's macro in `usparks` staging model. 

New successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8437277933/job/23106730367)

damonmcc (approved):

PR #704, opened 2024-03-22
GFT - Natural Resources

### LINKED ISSUE #672: archive GFT Natural Resources data

fvankrieken:
closes #672 

Will be easiest to go commit by commit

### More maintenance-type stuff

1. unrelated issue, make it easier to use dcpy from cli to import single dataset
2. simple tweak to `clip_to_geom` macro
3. rename elevated railways to "exposed railways" as per the guiding excel sheet
4. limit usfws_wetlands to just 4 watersheds that touch NYC. Original dataset was gigabites (all of NY state), so this makes it a little easier to use

### Meat of new logic

5. for all existing natural resources (from #685 ), add to gft up to int_buffers__all. Since these datasets are not buffered, I currently have `buffer` as null in their staging/buffer tables, then coalesce at int_buffers__all. **Do we feel like this is unsafe?** Or unintuitive. It probably just makes more sense to do this upstream, with the individual tables, I did it this way just to keep those tables more specific to their source data.
    also, this commit does not add expectations to `_sources.yml`, that comes in another commit
6. add three additional datasets, from library to int_buffers__all
7. add expectations for all _sources, both original ones and those added in commit 6
8. get data into the wide table. This adds a custom column which aggregates all natural resources
9. remove constraint that variables in seed should have survey question - I think these are in flux, plus not set for some of them at the moment
10. HEXAGONS. This got runtime for this query from 36 minutes down to 2ish. This is all mainly around the single monstrous multipolygon of natural_resource_shadows, which combines all natural resources (and then buffers them, as they're not buffered to start). It might have been simpler to just leave them as individual buffered polygons and see if that was somewhat performant, but this was fun. And also, it's nice to treat all variables the same in this case, and not need to aggregate that column later


products/green_fast_track/macros/clip_to_geom.sql
```diff
@@ -18,8 +18,13 @@ from "left" table and the resulting geom column named "geom"
...
-        ST_INTERSECTION({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }}) AS geom
+        CASE
+            WHEN ST_COVEREDBY({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }})
+                THEN {{ left }}.{{ left_by }}
+            ELSE ST_INTERSECTION({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }})
+        END
+         AS geom
     FROM {{ left }}
-    INNER JOIN {{ right }} ON ST_INTERSECTS({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }})
+    INNER JOIN {{ right }} ON ST_RELATE({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }}, 'T********')
```
fvankrieken, line 28: It seems we shouldn't be using st_intersects in general if we are looking for overlapping polygons. If they share a line border, that line is returns as the intersection

https://gis.stackexchange.com/questions/273993/st-relate-intersect-ignore-shared-boundary

fvankrieken:
Hex runtime comparison:

[36 min with no "tiling"](https://github.com/NYCPlanning/data-engineering/actions/runs/8634108888/job/23668765055#step:7:624)

[10 minutes with tiling](https://github.com/NYCPlanning/data-engineering/actions/runs/8635781231/job/23674325011#step:7:622)

[3 minutes with tiling and no distance calculated for this specific variable](https://github.com/NYCPlanning/data-engineering/actions/runs/8635973978/job/23674934317#step:7:622)

products/green_fast_track/macros/clip_to_geom.sql
fvankrieken: The perf part of this commit is the upper part though - only calculate intersection if need by, because st_coveredby is much faster than st_intersection

See https://postgis.net/documentation/tips/tip_intersection_faster/

products/green_fast_track/seeds/_seeds.yml
```diff
@@ -15,8 +15,6 @@ seeds:
         tests:
           - not_null
       - name: survey_question
-        tests:
-          - not_null
```
fvankrieken, line 19: Columns seem a tad in flux, and we don't use them from this variable column so made them optional for now

fvankrieken:
Getting pilot projects expecations sorted out, but code is ready for review.

Not quite sure if it makes senes but leaving out export for now just because I've been tinkering up till this point for a while, and would just like to get merged. Export should be straightforward, maybe would make sense to just include here but I'd still prefer to leave as is

fvankrieken:
[Latest build still running](https://github.com/NYCPlanning/data-engineering/actions/runs/8648580996)

damonmcc:
agree that adding new sources to the exported FGDB is a pain and worth doing after this

just to note, one insight I've seen come from the export part is unexpected geometry types

damonmcc:
> Since these datasets are not buffered, I currently have buffer as null in their staging/buffer tables, then coalesce at int_buffers__all. Do we feel like this is unsafe?

It doesn't seem unsafe, maybe isn't consistent with something like the airports that aren't actually buffered either. But down to go with this as is and revisit how we handle non-buffered buffers (call them all `spatial` or something else that's inclusive of buffered and non-buffered source data)

products/green_fast_track/macros/clip_to_geom.sql
alexrichey: Good to know! I wonder if (outside of this macro), 
```
            WHEN ST_COVEREDBY(...)
                THEN orig.geom
            ELSE ST_INTERSECTION(...)
```
warrants a macro of it's own

fvankrieken:
[Build now passing](https://github.com/NYCPlanning/data-engineering/actions/runs/8649284307/job/23715152720), added some helpful comments in a final rebase. Good to go

products/green_fast_track/macros/clip_to_geom.sql
fvankrieken: Or just a postgres function so we don't need to use jinja. given that this is in the official docs, I'm surprised they don't just have an out of the box function that does this

products/green_fast_track/models/staging/stg__dcp_beaches.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    'beaches' AS variable_type,
+    agency || '-' || name AS variable_id,
+    st_union(wkb_geometry) AS raw_geom,
+    NULL AS buffer
```
alexrichey, line 5: Should we be null'ing out the buffer, in general? When I don't have buffer specified, I've just been setting `raw_geom` and `buffer` to the same value

damonmcc (approved):

alexrichey (approved):
Awesome. Esp love the tiling

sf-dcp:
I probably won't get to PR review until tomorrow, and I don't want to hold you back. Feel free to merge without my input

fvankrieken:
> I probably won't get to PR review until tomorrow, and I don't want to hold you back. Feel free to merge without my input

Sounds good - I added a few helpful comments in your honor!

PR #709, opened 2024-03-25
test GFT outputs using "pilot projects"

### LINKED ISSUE #708: use GFT pilot project to test data pipeline
Potential Green Fast Track pilot projects have been identified by Planning Support to be used in training materials.

To ensure we generated expected outputs, we could use these sets of lots as the focus of an integration test for our GFT builds.

We could use any lots for this type of test. But using pilot project lots would mean that their outputs will be accurate while DCP uses them in training materials and we're likely to get quick feedback on any questions we have about expected outputs.

damonmcc:
resolves #708 

We don't have a list of pilot projects from Planning Support that we can confidently commit to this public repo, so I made a list of fake pilot projects.

To test expected outputs for pilot projects, I made these changes:
- add a new seed csv to list the BBLs of pilot projects and their expected outputs (copied from the primary product table)
- add a new product table called `green_fast_track_pilot_projects` which is a subset of the primary product table
- use `dbt_utils.equality` to test that the new product table is equal to the seed file

new test passing in a build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8460159638/job/23177838103#step:7:481)


products/green_fast_track/seeds/expected_green_fast_track_pilot_projects.csv
damonmcc, line 1: this csv is much more readable in the file view [here](https://github.com/NYCPlanning/data-engineering/blob/dm-gft-pilot-projects/products/green_fast_track/seeds/expected_green_fast_track_pilot_projects.csv)

I chose to use DCP's offices and a project named `Beach Airport Lots`. `Beach Airport Lots` is two adjacent lots that are near JFK airport and Rockaway beach. only one of them intersects `stg__panynj_airports`

products/green_fast_track/seeds/_seeds.yml
```diff
@@ -32,3 +33,36 @@ seeds:
           - not_null
           - accepted_values:
               values: ["TRUE", "FALSE"]
+
+  - name: expected_green_fast_track_pilot_projects
+    description: Mock pilot projects and their expected outputs
+    config:
+      # empty columns default to integer type
```
damonmcc, line 40: I'd rather not have to declare these column types (other than `bbl` and `geom`), but I couldn't find any other way

csv values that properly end up as `value | NULL | value` in the table that I tried are `value,,value`, `value,"",value`, and `value,NULL,value`. but in all cases the inferred column type is integer when all values in the column are `NULL`. when that happens, the equality test fails because the "actual" data's column is type text

products/green_fast_track/bash/build.sh
```diff
@@ -7,7 +7,7 @@ dbt deps
 dbt debug
 
 echo "Build seed tables"
-dbt build --select config.materialized:seed
+dbt build --select config.materialized:seed --indirect-selection=cautious --full-refresh
```
sf-dcp, line 10: what do these 2 arguments in the end do?

products/green_fast_track/seeds/_seeds.yml
```diff
@@ -32,3 +33,36 @@ seeds:
...
+        tests:
+          - not_null
+      - name: bbl
+        tests:
+          - not_null
+          - unique
+      - name: geom
+        tests:
+          - not_null
+          - is_epsg_2263
```
sf-dcp, line 68: `is_epsg_2263` <-- 🎉

products/green_fast_track/seeds/_seeds.yml
```diff
@@ -32,3 +33,36 @@ seeds:
...
+        Arterial Highways: text
+        (E) Designations - HazMat: text
+        Special Coastal Risk Districts: text
+        NYC Parks properties: text
+        POPS: text
+        Waterfront Public Access Areas: text
+        Publicly Owned Waterfront: text
+        NYS Parks Property: text
+        Federal Parks property: text
+        geom: geometry(Geometry, 2263)
```
sf-dcp, line 56: wait, I thought dbt doesn't let you check geom data types. Is that not true?

products/green_fast_track/seeds/_seeds.yml
sf-dcp: Makes sense, good to know

sf-dcp:
Everything looks great! Ready to approve. The only little thing I'm not sure about is the names of the seed and the resulting model. If it's a pure test now and in the future, I'm leaning towards renaming `test_...` or something similar

products/green_fast_track/seeds/_seeds.yml
damonmcc: this is setting a column type ([docs](https://docs.getdbt.com/reference/resource-configs/column_types)) not testing it

this column is used in the `dbt_utils.equality` test, but that isn't the same as the data contracts we tried that were checking column types and couldn't handle the geometry type

maybe if we try a data contract using something like what I did here with `geometry(Geometry, 2263)` to be more explicit with the type it would work??

products/green_fast_track/bash/build.sh
damonmcc: `--full-refresh` ensures the entire table is rebuilt, not just values in existing columns. without it, this fails when there are new or different columns ([docs](https://docs.getdbt.com/faqs/Seeds/full-refresh-seed))

`--indirect-selection=cautious` ensures that when we say "test the seed tables" during `dbt build`, we don't run _all_ tests that depend on those tables ([docs](https://docs.getdbt.com/reference/node-selection/test-selection-examples?indirect-selection-mode=cautious#indirect-selection-examples)). in this case, the new test `green_fast_track_pilot_projects_equality` was being run before the table it's defined for exists)

damonmcc:
@sf-dcp good call, I renamed it from `expected_green_fast_track_pilot_projects` to `test_expected_pilot_projects` to also make it shorter (it's not like we have `green_fast_track` in every other seed file name)

sf-dcp (approved):
👏

sf-dcp:
> @sf-dcp good call, I renamed it from `expected_green_fast_track_pilot_projects` to `test_expected_pilot_projects` to also make it shorter (it's not like we have `green_fast_track` in every other seed file name)
> 
> maybe `test_pilot_projects_expected` would be better? it seems worth keeping the word "expected" somewhere in the name

`test_expected_pilot_projects` sounds good!

PR #711, opened 2024-03-25
Ar gft historic buffers

alexrichey:
Latest successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8730593563). All builds from this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Aar-gft-historic-buffers).

Builds [before](https://github.com/NYCPlanning/data-engineering/actions/runs/8712020901/job/23897401244#step:7:842) and [after](https://github.com/NYCPlanning/data-engineering/actions/runs/8713409327) updating contract for main table.

Will add exporting of source data in a follow-up.

---

Still TODO
- [x] nysshpo_register_historic_places datasource has exactly 1k records?!
- [x] null geoms in all_buffers
- [x] Compound Key Error -> Dupes in my data
- [x] certain queries take a very long time

### Historic Buildings (`int_buffers__nysshpo_historic_buildings`)

For this dataset, we ideally take a Point and match via spatial join to PLUTO. However, there are some near hits. As Below:
image
(adjacent building is 'Church of God')

Overall there are 2k records where we don't match. (of 100k total) It's possible that buffering the point by a few feet BEFORE matching to pluto might increase the number of matches. 

### nysshpo_register_historic_places
Same situation as above, but only 23 (of 179) results where we don't directly match a PLUTO BBL. There are a handful where, potentially, buffering the point by a foot or two before matching to PLUTO would yield a BBL match. E.g. `98NR01367 - Jackson Heights Historic District`
image

Also, regarding this comment from the spreadsheet, 

> Use [National Regiser Number] field for name value. May also want to include [Resource Name]

Is the following convention ok: `[National Regiser Number] - [Resource Name]` (e.g. 98NR01367 - Jackson Heights Historic District) works

### Problems with `Archeological Buffers` dataset
When we filter this dataset down to just those polygons that overlap with the NYC boundary, there are about 50. There are two problems. 
1) There's no actual name identifer for what these are, and we get various ID'ish columns, but I'm sure that they constitute meaningful IDs. (e.g. objectid, orig_fid). So if we do use this dataset, what should we use for the `variable_id`?
2) One polygons covers most of Manhattan, and another covers most of Staten Island. 
image
image


alexrichey:
## Notes by Dataset (Ongoing)
#### Archaelogical Buffers
image

products/green_fast_track/models/intermediate/buffers/int_buffers__lpc_landmarks.sql
fvankrieken, line 1: Per in-person convo with @damonmcc, a few things need to happen

1. st_within back to st_contains
2. maybe just group by lm_name and not first by bbl. If we were passing on bbl info, this would make sense and in general makes sense in context of having a single bbl geom with all appropriate lpc_landmarks, but say we have two rows for lpc landmark A, contained in bbls 1 and 2. lpc landmark B is also contained in lot 2. If we first group by bbl, then we end up with a row of lm_names "A, B" for lot 2 and "A" for lot 1. If we then have a project bbl that intersects both of these, when we aggregate the lm_names for the wide table, we'll have "A, A, B" (or "A, B, A"). So it seems to make sense to only aggregate by lm_name to deduplicate, with unioned polygons of multiple bbls for single lm_names

damonmcc: if the Landmark points and PLUTO lots looks like this:
image

the rows of `int_buffers__lpc_landmarks` should look like this:

| variable_id | bbl |
|--------|--------|
| A | 1 |
| B | 1, 2 |
| C | 2 |

products/green_fast_track/models/_sources.yml
```diff
@@ -121,6 +151,54 @@ sources:
...
+        columns:
+          - name: usnnum
+            tests:
+              - not_null
+          - name: usnname
+            tests:
+              - not_null
+          - name: eligibilitydesc
+            tests:
+            - accepted_values:
```
sf-dcp, line 162: 👀 `accepted_values`

products/green_fast_track/models/_sources.yml
```diff
@@ -121,6 +151,62 @@ sources:
...
+        columns:
+          - name: usnnum
+            tests:
+              - not_null
+          - name: usnname
+          - name: eligibilitydesc
+            tests:
+            - accepted_values:
+                values: ['Eligible','Listed', 'Not Eligible', 'Not Eligible - Demolished', 'Undetermined']
+          - name: wkb_geometry
```
sf-dcp, line 164: perhaps I missed it somewhere... Can `wkb_geometry` ben null?

products/green_fast_track/seeds/test_expected_pilot_projects.csv
sf-dcp, line 1: Does this look right?

image

sf-dcp: and this: 

image

sf-dcp:
Reviewed the PR - awaiting last `int_buffers__lpc_landmarks` to be fixed. 

A small nit, perhaps for a different PR: name models (stage/intermediate) after `variable_type`. For example, `int_buffers__lpc_landmarks` doesn't jump as `nyc_historic_buildings`.

products/green_fast_track/seeds/test_expected_pilot_projects.csv
damonmcc: it is logically right, since just one big polygon for that flag. but the string used could be better. I'll looks for other similar flags to emulate

products/green_fast_track/models/_sources.yml
damonmcc: yup very handy! and DRY so long as the strings aren't needed anywhere else or more useful as structured info like  `nyc_parks_properties_categories.csv`

damonmcc:
> Reviewed the PR - awaiting last `int_buffers__lpc_landmarks` to be fixed.
> 
> A small nit, perhaps for a different PR: name models (stage/intermediate) after `variable_type`. For example, `int_buffers__lpc_landmarks` doesn't jump as `nyc_historic_buildings`.

it's definitely hard to quickly see the connection between source data and variables in a lot of models (not always 1-to-1?). would love to improve the model names and start some docs soon

products/green_fast_track/models/_sources.yml
damonmcc: good call. I don't see any nulls in the source data, will add the test

products/green_fast_track/models/intermediate/buffers/int_buffers__lpc_landmarks.sql
damonmcc: sample records in final table after successful build:

Screenshot 2024-04-18 at 10 20 05 AM

damonmcc:
@fvankrieken @sf-dcp
made some changes, ready for re-review!

sf-dcp (approved):
LGTM!

dcpy/library/templates/lpc_historic_district_areas.yml
```diff
@@ -15,7 +15,7 @@ dataset:
 
   destination:
     geometry:
-      SRS: EPSG:4326
+      SRS: EPSG:2263
```
alexrichey, line 18: @damonmcc Apologies for leaving this hanging. I don't believe I reran data library after making this change, so I think that next time we run data library with this dataset, this will break what we've got below in the dbt model. 

I this case, I'll just revert this change to keep `EPSG:4326`. cc'ing @fvankrieken and @sf-dcp because it's perhaps a consideration for `extract`/`ingest` for downstream breaking changes like this, and how to deal with them. Supposing we did change the projection, what's the best way to deal with that? Maybe just creating `lpc_historic_district_areas_v2.yml`? Sort of a poor man's versioning scheme 🤷‍♂️

damonmcc: I guess it'd better to not revert this change and just change the dbt model when it breaks, since we'd rather archive to `2263` and not have to reproject during the build

alexrichey: Looks like it's also used in EDDE, so it's break things there. And even though it's probably trivial to fix there, I think generally best to have consistent projections in data we've archived. Super trivial in this case since we've only ever archived this dataset twice... but also a good opportunity to do things the "right" way with breaking changes

alexrichey (commented):
Thanks so much for kicking this over the finish line while I was out! It mirrors pretty closely the changes I had locally. I'm unable to actually submit an "Approve" on the PR (which makes sense) so this is as close as I can get: ✅

PR #712, opened 2024-03-25
add optional left_columns arg

fvankrieken:
To specify columns, can do so like

```sql
WITH a AS (
    ...
),
clipped_to_nyc AS (
    {{ clip_to_geom(left=a, left_by="wkb_geometry", left_columns=["variable_type", "variable_id"]) }}
)
SELECT * FROM clipped_to_nyc
```

### DIFF
--- a/products/green_fast_track/macros/clip_to_geom.sql
+++ b/products/green_fast_track/macros/clip_to_geom.sql
@@ -5,7 +5,7 @@
             {{ dbt_utils.star(from=left, except=[left_by]) }},
         {% else %}
             {% for column in left_columns %}
-                left.{{ column }},
+                {{ left }}.{{ column }},
             {% endfor %}
         {% endif %}
         ST_INTERSECTION({{ left }}.{{ left_by }}, {{ right }}.{{ right_by }}) AS geom
--- a/products/green_fast_track/macros/clip_to_geom.sql
+++ b/products/green_fast_track/macros/clip_to_geom.sql
@@ -1,7 +1,13 @@
-{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom") %}
+{% macro clip_to_geom(left, right=ref("stg__nyc_boundary"), left_by="geom", right_by="geom", left_columns=[]) %}
 
     SELECT
-        {{ dbt_utils.star(from=left, except=[left_by]) }},
+        {% if left_columns == [] %}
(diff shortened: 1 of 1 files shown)


sf-dcp:
nice! Have you given it a try?

fvankrieken:
I did use it.

Nothing from the right table is selected, and that's why there's the `{{ dbt_utils.star(from=left, except=[left_by]) }}` when no columns are supplied, so that the left geom column is not selected/duplicated

fvankrieken:
You know, I _compiled_ a model using it, not ran. Let me run it with a CTE

sf-dcp:
> I did use it.
> 
> Nothing from the right table is selected, and that's why there's the `{{ dbt_utils.star(from=left, except=[left_by]) }}` when no columns are supplied, so that the left geom column is not selected/duplicated

Got it.  In the second part of the logic below, it seems like we will be selecting `left_by` column, no?

```jinja
        {% else %}
            {% for column in left_columns %}
                left.{{ column }},
            {% endfor %}
```

fvankrieken:
That's iterating through the argument explicitly provided for selecting by the user, so I'm just going to trust people to select what they need in that case

sf-dcp (approved):
Thanks for taking a stab at it!

fvankrieken:
Caught a little bug, works now

sf-dcp:
@fvankrieken, I tried to use the macro in my code like this and I'm getting a syntax error: 

```sql

WITH reprojected AS ( ...
),

clipped_to_nyc AS (
    {{ clip_to_geom(left=reprojected, left_by="geom", left_columns=["objectid"]) }}
),

```

Here is the compiled query:

```sql 
reprojected AS (...
),

clipped_to_nyc AS (
    SELECT
         .objectid,
         ST_INTERSECTION(.raw_geom, "db-green-fast-track"."sf_gft_shadows"."stg__nyc_boundary".geom) AS geom
    FROM 
    INNER JOIN "db-green-fast-track"."sf_gft_shadows"."stg__nyc_boundary" 
    ON ST_INTERSECTS(.raw_geom, "db-green-fast-track"."sf_gft_shadows"."stg__nyc_boundary".geom)


),
```

So the compiled query doesn't know the CTE name. Do you know what I'm doing wrong? You said it worked on your CTE

sf-dcp:
Nevermind, solved the issue! The CTE name should be in quotes when calling the macro

PR #718, opened 2024-03-27
Fvk restructure extract

fvankrieken:
- Rename extract to ingest
- move formerly extract/ingest.py to ingest/extract.py
- move transformation logic (parquet, eventually transform steps) to ingest/transform.py
- above changes reflected in models and tests as well


sf-dcp (approved):
Looks good!

Issue #460, opened 2023-12-29
CPDB 2024 Preliminary


damonmcc:
### Product Name

cpdb

### Build Version

24_preliminary

### Status of Update

- [ ] source data extracted
- [ ] draft build succeeded
- [ ] draft build passed QA
- [ ] data packaged and distributed

fvankrieken:
All external agency partners contacted about source data

fvankrieken:
# Update source data
Read more about who to reach out to and how to update a specific dataset [here](https://github.com/NYCPlanning/db-cpdb/wiki/Maintenance)
## Received from agency partner and manually loaded into data libraries
- [ ] Capital Commitment Plan: fisa_capitalcommitments
- [ ] ddc_capitalprojects_infrastructure
- [ ] ddc_capitalprojects_publicbuildings
- [ ] dot_projects_bridges
- [ ] edc_capitalprojects - new data needs to be appended onto existing source dataset
- [ ] edc_capitalprojects_ferry - new data is often not received and this dataset is not updated with each release because nothing has changed.

## Automatically loaded into data libraries
Before initiating a build of cpdb, verify the latest version in DigitalOcean matches the version in NYC OpenData

# Projects
- [ ] [cpdb_capital_spending](https://github.com/NYCPlanning/db-cpdb/actions/workflows/spending.yml) - updated with action
- [ ] dot_projects_intersections https://data.cityofnewyork.us/Transportation/Street-and-Highway-Capital-Reconstruction-Projects/97nd-ff3i
- [ ] dot_projects_streets https://data.cityofnewyork.us/Transportation/Street-and-Highway-Capital-Reconstruction-Projects/jvk9-k4re
- [ ] dpr_capitalprojects https://www.nycgovparks.org/bigapps
- [ ] dpr_parksproperties https://data.cityofnewyork.us/Recreation/Parks-Properties/enfh-gkve
- [ ] dcp_cpdb_agencyverified (does not get updated)

# Building and lot-level info
- [ ] dcp_mappluto_wi https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-pluto-mappluto.page
- [ ] dcp_facilities https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-selfac.page
- [ ] doitt_buildingfootprints https://data.cityofnewyork.us/Housing-Development/Building-Footprints/nqwf-w8eh

# Spatial boundaries
[...]

PR #719, opened 2024-03-27
CPDB 23prelim

fvankrieken:
Tweaks for #460

### DIFF
--- a/dcpy/library/script/fisa_dailybudget.py
+++ b/dcpy/library/script/fisa_dailybudget.py
@@ -2,6 +2,7 @@ from pathlib import Path
 import pandas as pd
 import re
 
+from dcpy.connectors.edm import recipes
 from . import df_to_tempfile
 from .scriptor import ScriptorInterface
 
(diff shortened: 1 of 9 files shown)


dcpy/library/script/fisa_dailybudget.py
```diff
@@ -72,6 +79,17 @@ def ingest(self) -> pd.DataFrame:
         df = self._dedupe(df)
         return df.sort_values(by="TBL_LAST_DT")
 
+    def ingest(self) -> pd.DataFrame:
```
damonmcc, line 82: since this one uses `previous_version`, should this be `ingest_historical`?

fvankrieken: This is meant to be just the standard ingest. Ingest historical was meant if you had the entire historical archive of files to process. Maybe not the clearest lingo

fvankrieken: I can add some helpful comments/docstrings

damonmcc: gotcha. and looks like `ingest_historical` isn't called in `runner`, so someone would have to edit this file to use `ingest_historical`?

fvankrieken: And by that I mean delete the now-defunct function

damonmcc (approved):

PR #729, opened 2024-04-01
Data library: convert raw to parquet v2.0

sf-dcp:
This is a continuation of #658 PR converting raw data to parquet. Related to #631 issue. 

It's probably easiest to review the PR commit by commit. A big chunk of changed files here is revised test data used in tests.

### Major changes: 

1) Originally, the `to_parquet` function was responsible for reading a local dataset into a pandas or geopandas dataframe (pandas df if data isn't geospatial) and then output to a parquet/geoparquet file. 

    In this PR, we are refactoring this function into 2 functions: reading data into a dataframe & outputting to parquet. First reason is the need to expand the function to work with more and more input data formats --> more code. And second reason is that a code snippet reading local data into a dataframe was repeated in tests. Thus, the need to refactor.

2) Implement a zipped input data format in `to_parquet` in 2 steps: unzip a file & use existing code to process an unzipped file. I added tests for zipped csv, zipped shapefile, and zipped geodatabase. 

3) Implement a csv format that has longitude and latitude columns instead of one geometry column. Test is also present.

4) Name geometry column as `geom` in output parquet file.

5) Due to added data formats, expand test code, generating fake data. Note, test code is becoming messy, and we will refactor it in a separate PR as it will touch tests outside of `to_parquet` function.

### Side note
`to_parquet` fn converts *geospatial* input data into a `geoparquet` data format: no need to explicitly specify a file extension to be `.geoparquet`. This is because `geopandas` df automatically becomes `geoparquet`.

### `TODO` for next PR:  
- [ ] json
- [ ] geojson
- [ ] excel
- [ ] simplify tests (perhaps by using complete, not partial, fake templates) 
    From a convo with Finn, we can simplify tests by using a subset of the `config` object as a function input instead of the entire thing.


dcpy/test/lifecycle/ingest/test_transform.py
sf-dcp, line 1: Note to myself: Not happy how current tests are set up

dcpy/utils/data.py
sf-dcp, line 1: didn't come up with a better name for this file... Currently, using `read_data_to_df` in `lifecycle.ingest.transform` and its test

sf-dcp:
Example of what test data looks like: 

* `.csv` format only:
    image

* `.csv`, `.shp`, `.gdb`, zipped `.csv`/`.shp`/`.gdb`
    image

dcpy/utils/data.py
```diff
@@ -0,0 +1,156 @@
...
+    Config,
+)
+
+import zipfile
+
+EXTACTED_FILES_DIR = Path("extracted_files")
+
+
+def read_data_to_df(
+    config: Config, local_data_path: Path
```
fvankrieken, line 18: If this function is going to take a Config (or any model from `lifecycle`), I don't think it should be in utils. Sort of goes with our discussion today. I know you wanted to limit scope, so for now, maybe it could just take a file.Format and a filename

fvankrieken: Oops you're already doing that. Seems like you can just pass in the file format. For the errors that refer to the raw file name, you can do `local_data_path.stem`

dcpy/utils/data.py
```diff
@@ -0,0 +1,156 @@
...
+
+from dcpy.utils.logging import logger
+from dcpy.models import file
+from dcpy.models.lifecycle.ingest import (
+    Config,
+)
+
+import zipfile
+
+EXTACTED_FILES_DIR = Path("extracted_files")
```
fvankrieken, line 14: Maybe `tmp/extracted_files`? Just so it's gitignored/hidden slightly more.

Also, typo in var name!

dcpy/utils/data.py
fvankrieken: Seems fine! Certainly at least for now

fvankrieken:
A couple small notes, but other than that this looks great

dcpy/utils/data.py
fvankrieken: oops not stem. `local_data_path.name`

sf-dcp: This function will be taking in a model from `lifecycle` - where would you recommend placing this fn in such case? `dcpy.lifeclycle.ingest.tranform`? 

This `data.py` file also has `unzip_file` fn below. Lmk if it makes sense to keep in utils, `data.py` file.

sf-dcp: * Regarding passing in the file format: if it's okay, I would like to hold off this change until next PR because my test also uses this fn. The reorganization of the test code, creating input objects for the `test_to_parquet` fn, will be a bit messy.

* Regarding raw file name, I don't need-need it in this function per se. The only places I'm referring to it are assert messages below: 
    image

   I can def replace it with the Path.name attribute - ty for the suggestion! The only thing is when an input path is zipped csv, the assert message will say "... doesn't exist in `file.csv.zip`" which is fine.

fvankrieken: If you switched to a `file.Format` object as the arg instead of config, it should just be a one-line difference in the tests, no? There'd be room to make things cleaner by not goofing the rest of the template like we'd talked about but for now you could just edit this line

https://github.com/NYCPlanning/data-engineering/pull/729/files#diff-888d97f8d2b095ef22f1837f5039b12008ef5ce31d310b5833051877bcd63eccR96

from this
```
raw_df = data.read_data_to_df(config=config, local_data_path=file_path)
```

to this
```
raw_df = data.read_data_to_df(config=config.format, local_data_path=file_path)
```

I think?

fvankrieken: And then if it's using a more generic type (like `file.Format`, which lives outside of `lifecycle.ingest`) rather than an object very specific to our ingestion code, I think it's fine for this code to keep living in a more pure util file like it is now

sf-dcp: gotcha! This makes sense. I thought I would need to fix test config objects.

sf-dcp:
@fvankrieken, I added 3 commits based on your recs

fvankrieken (approved):

PR #738, opened 2024-04-04
enforce a contract on `green_fast_track_bbls`

damonmcc:
yay! like when creating a seed file with a geometry column [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/green_fast_track/seeds/_seeds.yml#L56), we have to use the PostGIS column types to declare geometry columns in dbt yaml files.

### when columns are missing
Screenshot 2024-04-04 at 12 17 28 PM

### when passing
Screenshot 2024-04-04 at 12 19 38 PM


products/green_fast_track/models/product/_product_models.yml
```diff
@@ -79,7 +77,38 @@ models:
...
+        tests: [not_null]
+      - name: '"Publicly Owned Waterfront Flag"'
+        data_type: string
+      - name: '"Waterfront Public Access Areas"'
+        data_type: string
+        tests: [not_null]
+      - name: '"Waterfront Public Access Areas Flag"'
+        data_type: string
       - name: geom
+        data_type: geometry(Geometry, 2263)
```
sf-dcp, line 111: Nice!!! Do you know whether it checks projection at a column or value level? I'm wondering if it pulls the crs value from geom columns table in Postgres...

sf-dcp (approved):
LGTM!

fvankrieken (approved):

products/green_fast_track/models/product/_product_models.yml
damonmcc: pretty sure it's the column level. I just added a snippet from the docs the PR description that seems to indicate it isn't value level

I imagine value-level contract enforcement would add a lot of time to building a model

damonmcc:
oops I shouldn't have added the Shadow's columns here. gonna drop that change so that builds on `main` don't fail before https://github.com/NYCPlanning/data-engineering/pull/727 is done

products/green_fast_track/models/product/_product_models.yml
sf-dcp: Got it. Do you know if it raises an error if the projection value is zero at the column level?

damonmcc: in this case it would because the data type of `geometry(Geometry, 2263)` is expecting a projection of 2263

PR #746, opened 2024-04-07
fix GFT lot zoning logic

### LINKED ISSUE #741: revise Zoning logic

damonmcc:
resolves #741

### DIFF
--- a/products/green_fast_track/models/unit_tests.yml
+++ b/products/green_fast_track/models/unit_tests.yml
@@ -1,0 +1,11 @@
+unit_tests:
+  - name: test_logic__int_flags__zoning
+    description: "Check the int_flags__zoning logic captures lots with many zoning values."
+    model: int_flags__zoning
+    given:
+      - input: ref('stg__pluto')
+        rows:
+          - {bbl: 1005500023_ish, zonedist1: C1-7, zonedist2: R6, zonedist3: R10, zonedist4: NULL}
+    expect:
+      rows:
+        - {bbl: 1005500023_ish, variable_type: zoning_districts, variable_id: C or M}
--- a/python/requirements.in
+++ b/python/requirements.in
@@ -5,7 +5,7 @@ boto3-stubs[s3]>=1.28.1
 botocore-stubs
 cached-property==1.5.2
 contextily==1.3.0
-dbt-postgres
+dbt-postgres==1.8.0b2
 diagrams>=0.23.3
 faker==23.1.0
 folium==0.14.0


docs/diagrams/logic_gft_lot_zoning.drawio.png
damonmcc, line 1: confirmed with GIS and Planning Support that this lot-level logic is accurate

products/green_fast_track/models/product/_product_models.yml
```diff
@@ -189,10 +189,10 @@ models:
       - name: geom
         data_type: geometry(Geometry, 2263)
 
-  - name: green_fast_track_pilot_projects
+  - name: test__actual__pilot_projects
```
alexrichey, line 192: super nit: should just be one underscore after actual, right?

products/green_fast_track/seeds/test__expected__flags_zoning.csv
alexrichey, line 1: Love how declarative this testing ends up being

products/green_fast_track/models/product/_product_models.yml
damonmcc: I was looking to mimic our conventions with other model prefixes like `int_flags__`, so `test_actual__` would probably be more aligned than doubles before and after

and paraphrasing from dbt docs on "why double underscores at all": double underscores help visually distinguish separate parts of multi-word names

alexrichey: `test_actual__` and `test_expected__` seem good. (Wouldn't nitpick this usually, but when I go copy-pasta, just want to make sure I don't grab `test__expected_pilot_projects`)

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
alexrichey, line 1: Just curious what you're using this for? Just to confirm my understanding... say if we have (trimming down to just two zoning districts, for a mini-example)

bbl = 1
zonedist1 = a
zonedist2 = b
----------
bbl = 2
zonedist1 = b  (swapped from above example)
zonedist2 = a (swapped from above example)

confirming that'd end with a table like this, right?

z1 | z2 | zoning_combo_count
-----------------------------
a  | b  | 1
b  | a  | 1

alexrichey (approved):
Great stuff! Just had the question to make sure I understand that test query, but otherwise 👍

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
damonmcc: that's correct, this is useful for exactly that. I felt pretty in-the-dark about the combinations and their frequencies without this query

we may decide we care more about the unordered "list" of zoning district values independent of which columns they came from, but I wanted to stay as close to the source data as I could for this first pass. and it was informative to see that the type of flip-flop you described does appear in PLUTO

alexrichey: Awesome... had never seen the distinct rollup over a set of columns. Good to know it exists

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
```diff
@@ -0,0 +1,28 @@
...
+
+grouup_by_rollup AS (
+    SELECT
+        zonedist1,
+        zonedist2,
+        zonedist3,
+        zonedist4,
+        count(*) AS zoning_combo_count
+    FROM pluto
+    GROUP BY DISTINCT ROLLUP (zonedist1, zonedist2, zonedist3, zonedist4) -- noqa: PRS
```
fvankrieken, line 18: Two little things
- "grouup"
- what does "distinct" do in this case? I would imagine that's handled by the "group by" but not super familiar with rollup

fvankrieken: But from looking at docs it seems redundant

products/green_fast_track/models/product/test_actual__pilot_projects.sql
fvankrieken, line 1: Should we maybe have a separate `test` folder? Either in, or maybe prefereably outside of `product` folder? Would sort of make sense as a top-level models folder next to stg, int, prod

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
damonmcc: lol will fix the typo

[looks like](https://www.postgresql.org/message-id/bf3805a8-d7d1-ae61-fece-761b7ff41ecc@postgresfriends.org) the `distinct` is useful when doing multiple rollups in a single query? I just ran this without it and I don't see a difference in the results so I'll drop it. it's probably an artifact of my confusion and caution when learning to use rollup

products/green_fast_track/models/product/test_actual__pilot_projects.sql
damonmcc: yea at first I put `test_actual__pilot_projects` and `test_actual__flags_zoning` in `intermediate/flags/test` and `product/test`, which seemed like overkill

since a top-level `tests` folder is the default path for [singular tests](https://docs.getdbt.com/reference/project-configs/test-paths#default), maybe `models/test`?

but the intent here was that `dbt build --select intermediate` would build and test `test_actual__flags_zoning`, so I don't think that'd happen if it were moved to a separate folder (without some additional explicit CLI calls)

fvankrieken: Yeah sorry I meant within models!

Hmm, that's a good point

docs/diagrams/logic_gft_lot_zoning.drawio.png
sf-dcp: Very useful diagram! I would also add it to gft readme

sf-dcp (approved):
LGTM! 

Just wanted to say that having both the drawio visual the `test__expected_flags_zoning` table on your branch made the review easy to understand the the zoning logic! 

Also, excited to have analysis queries in `dbt`! Will make validation easier.

PR #750, opened 2024-04-08
clean up python reqs, dcpy install

fvankrieken:
Had an odd failed job [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8602897170/job/23573973198)

Couple things going on. For some reason, during dcpy install in container, it's uninstalling and reinstalling some python packages. On top of this, python isn't finding those packages when trying to use them.

I thought there might be some weird dependency issue causing the need/desire from pip to reinstall them, so I tried to clean up the requirements. This doesn't seem to have made a change, so I'm trying to force `--no-deps` when installing dcpy in the build image. Going to run a test build to see if this resolves issue

[Issue resolved now](https://github.com/NYCPlanning/data-engineering/actions/runs/8603394140/job/23575198124)

### DIFF
--- a/bash/docker_container_setup.sh
+++ b/bash/docker_container_setup.sh
@@ -11,6 +11,6 @@ else
     export workspace=/__w/data-engineering/data-engineering
 fi 
 
-python3 -m pip install $option . -c ./python/constraints.txt
+python3 -m pip install $option . -c ./python/constraints.txt --no-deps
 
 git config --global --add safe.directory $workspace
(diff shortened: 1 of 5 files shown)


damonmcc:
was it because that failed build was run on a branch that's [one commit behind main](https://github.com/NYCPlanning/data-engineering/compare/fvk-gft-natural-resources-flags...main)? and that one commit has changes to packages?

seems like nightly builds on main haven't had this issue, but definitely like any amount of package/install cleanup! just wondering why other builds have worked

fvankrieken:
Ah maybe you're right. I still think it's nice to keep --no-deps, just make sure that nothing is actually being installed, just making the link to the dcpy folder

damonmcc (approved):

PR #751, opened 2024-04-09
SqlFluff DevDB

fvankrieken:
I did bugger some stuff up with my dcpy install thing yesterday - build images don't necessarily have all requirements of dcpy. I've added some code here to install all dcpy reqs as part of docker image builds. A bit worried how the frankenstein docker image management code continues to grow, but this makes things a little better at least for the time being. 

devdb build failed initially due to above problem (see [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8619159417)), but [passing now](https://github.com/NYCPlanning/data-engineering/actions/runs/8619807505)

Will add rules around capitalization in follow up at some point to not mess with people's gft code


products/developments/sql/_bis.sql
```diff
@@ -2,7 +2,7 @@
 DESCRIPTION:
 	Initial field mapping and prelimilary data cleaning for BIS job applications data
 
-INPUTS:
```
fvankrieken, line 5: Sort of funny that sqlfluff also hunts down trailing whitespace in comments. Fine by me!

pyproject.toml
```diff
@@ -127,20 +127,19 @@ dependencies = [
     "beautifulsoup4",
     "boto3",
     "faker",
-    "GDAL",
```
fvankrieken, line 106: gdal is required for dcpy, but it's just too messy with the annoying versioning issues around alignment of binary version and python version to have it listed here

pyproject.toml
```diff
@@ -43,12 +43,28 @@ CONDITION = "WHERE a.geom IS NOT NULL"
...
+source_column="bct2020"
+output_column="bct2020"
+additional_column_mappings='[("geoid", "centract2020")]'
+group_by='["boro", "bct2020", "centract2020"]'
+join_table="dcp_ct2020"
+geom_join_column="boroct2020"
+years="['2010', '2011']"
+decade="2020"
+geom="CD"
+CAPTURE_DATE="2024-01-01"
```
damonmcc, line 66: (for later) I wonder how else we can handle these to prevent conflicting values with same name. maybe config files in subdirectories?

or maybe the actual value doesn't matter when linting??

damonmcc (approved):
nice!

pyproject.toml
fvankrieken: Yeah the actual value doesn't matter... UNLESS the type is different, for jinja. Luckily we don't have that issue right now, but it's certainly a possibility down the road

PR #757, opened 2024-04-11
Fvk dbt dcpy utils

fvankrieken:


dcpy/lifecycle/builds/dbt.py
```diff
@@ -0,0 +1,103 @@
...
+    source["tables"].append({"name": dataset})
+    return source
+
+
+class RecipeEditor:
+    def __init__(self, path):
+        self.path = path
+
+    def __enter__(self):
+        plan.recipe_from_yaml(self.path)
```
fvankrieken, line 26: Just to validate recipe. Loads it in yaml to preserve comments, order, etc

dcpy/lifecycle/builds/dbt.py
```diff
@@ -0,0 +1,103 @@
...
+        with self.path.open() as f:
+            self.recipe = yaml.load(f)
+        return self.recipe
+
+    def __exit__(self, *args):
+        Recipe(**self.recipe)
```
fvankrieken, line 32: Since this object doesn't know what you're doing with the recipe, good to just validate it before dumping

dcpy/lifecycle/builds/dbt.py
```diff
@@ -0,0 +1,103 @@
...
+
+from dcpy.utils.logging import logger
+from dcpy.models.lifecycle.builds import InputDataset, Recipe
+from dcpy.lifecycle.builds import plan
+
+DEFAULT_RECIPE = "recipe.yml"
+
+app = typer.Typer(add_completion=False)
+
+yaml = YAML()
```
fvankrieken, line 13: Am I annoyed that this package requires that you instantiate a YAML object to read/write yaml from files? Yes I am

dcpy/lifecycle/builds/dbt.py
```diff
@@ -0,0 +1,107 @@
...
+    Optionally creates a stub sql file in staging folder and adds source to _staging_models.yml
+    """
+    with RecipeEditor(project_dir / recipe_path) as recipe:
+        recipe["inputs"]["datasets"].append(
+            InputDataset(name=dataset).model_dump(
+                exclude_defaults=True, exclude_none=True
+            )
+        )
+
+    sources_path = project_dir / "models" / "_sources.yml"
```
sf-dcp, line 65: not sure if this sources path will remain true. Perhaps make this a global variable for visibility?

sf-dcp:
Excited to see dbt `utils` :) 2 nits: 
* would be nice to validate whether a dataset is already present in sources/staging yaml files
* a lil test to confirm it works

sf-dcp:
> > Excited to see dbt `utils` :) 2 nits:
> > 
> > * would be nice to validate whether a dataset is already present in sources/staging yaml files before adding it to the list. If it, probably want to abort it to avoid duplication?
> > * a lil test to confirm the functionality works
> 
> I'm gonna push back on both fronts
> 
> 1. with it being a sort of silly little cli util at the moment, I'd prefer we keep the behavior simple and predictable - just add something to these files - and put the onus on the user to use appropriately
> 2. again with it currently being a silly little cli call, I'd rather not implement a test for what could be a moving target

Let's at least add a `TODO` docstring/comment that tests are needed. That way, we set expectations about usage of this code in the future: if someone attempts to use it, they are aware to test it extensively. Given how much we are adding/changing, I think there is a risk of accumulating a significant chunk of untested codebase, and having some sort of a todo note would be helpful.

sf-dcp (approved):

PR #763, opened 2024-04-12
Scaffold for prerprocessing steps

fvankrieken:
#632 

In essence, this pr is really about the framework - the classes in `dcpy.models.lifecycle.ingest` and the  two `validate` functions in `dcpy.lifecycle.ingest.transform`. The idea is that processing steps can be defined in a yml template like so

```yml
name: bpl_libraries
...
processing_steps:
- name: simple_transformation
- name: drop_columns
  args:
    columns: ['col1', 'col2']
```

For now, functions take and return a dataframe, and any other needed arguments. We're obviously not set on this, and maybe for now just to be completely agnostic, each preprocessing step should just read from and write to a parquet file. But for a quick framework, this seemed easiest. I mainly wanted to focus on validating yml input to python function calls.

Would love input on the structure of where preprocessing functions live. Another module would probably be cleaner than this rather than this weird class, but this worked for now


dcpy/utils/duckdb.py
fvankrieken, line 1: Not used in this PR, but figure we'll need this again at some point

dcpy/lifecycle/ingest/transform.py
```diff
@@ -109,3 +111,158 @@ def to_parquet(config: Config, local_data_path: Path | None = None):
 
     gdf.to_parquet(PARQUET_PATH, index=False)
     logger.info(f"✅ Converted raw data to parquet file and saved as {PARQUET_PATH}")
+
+
+class Preprocessors:
```
fvankrieken, line 116: These functions are very half-baked. I'd like to just leave them for now, and iterate on them while trying to actually get this working for our template datasets

dcpy/lifecycle/ingest/transform.py
```diff
@@ -109,3 +111,156 @@ def to_parquet(config: Config, local_data_path: Path | None = None):
...
+            f"""
+            COPY (
+                SELECT * FROM 's3://edm-recipes/datasets/{dataset}/{version}/{dataset}.parquet'
+                UNION ALL
+                SELECT * FROM '{file}'
+            ) TO '{file}' (FORMAT PARQUET)"""
+        )
+
+
+def validate_function_args(
```
sf-dcp, line 172: Perhaps validation functions like this one should live in `lifecycle`/`ingest`/`metadata`?

sf-dcp:
My thoughts about location of preprocessing functions... These functions (like one dropping columns) potentially can be used in python pipelines too , and `utils` appears to be the right place for utils shared across modules in `dcpy`. And then you can import these functions into your `Preprocessors` class. For the class itself, it seems to fit with other classes in `models` module. 

Otherwise, the framework around preprocessing steps looks good to me!

fvankrieken:
> My thoughts about location of preprocessing functions... These functions (like one dropping columns) potentially can be used in python pipelines too , and `utils` appears to be the right place for utils shared across modules in `dcpy`. And then you can import these functions into your `Preprocessors` class. For the class itself, it seems to fit with other classes in `models` module.
> 
> Otherwise, the framework around preprocessing steps looks good to me!

Yeah that's going to be a funny thing moving forwards, and I think is going to be fairly dependent on what utils we actually create to meet the needs of these datasets. For some, I don't think we need a util - dropping columns on a pd dataframe is already a one-liner, so a wrapper around it in our utils would be a bit odd. But could definitely see a world where we end up putting utils there that get referenced here.

dcpy/lifecycle/ingest/transform.py
fvankrieken: Potentially. This is a sort of question I'd like to kick down the curb - this function is fairly generic at the moment and theoretically could live outside of `lifecycle` even. Though counterpoint, can't think of an example where we'll need to be validating parsed function arguments anywhere outside of lifecycle.

`lifecycle.ingest.metadata` could also use a rename. I feel like it really is about parsing and validation, or generating configs for these datasets, so something a little more specific to that. Which then still goes in line with your suggestion of moving this there.

So long way of saying sounds good. I do want to rename "metadata" file though. Any thoughts? `config` sort of works, though still a bit generic, would prefer something that's an umbrella for parsing and validation but drawing a blank at the moment

sf-dcp: Yeah, I agree. It would be nice to have milestone check-ins to evaluate code architecture, and refactor if needed. Versus doing a major refactor in the end. 

Regarding renaming, `config` or `validate` sounds good

fvankrieken: Done

fvankrieken: And also broke something in the process of course

fvankrieken: It's because we use `config` as a var name all over the place.

Given that the other file names are verbs here, I'm going to go with `configure`

sf-dcp (approved):

PR #771, opened 2024-04-17
Refactor extract slightly and fix flaky tests

fvankrieken:
Extract code felt a little messy with what objects were being passed where, what order of operations was in the "main" function, so just tidied things up a little.

Expanded tests just slightly and more importantly, am mocking api responses. Since they're pretty much all request-based at the moment, thought it made sense to mock `requests.get`, with mocked responses for specific endpoints (so if we add more, they fail by default)


dcpy/test/lifecycle/ingest/test_extract.py
```diff
@@ -1,36 +1,30 @@
...
+from dcpy.lifecycle.ingest import extract, configure
+
 from . import RESOURCES
 
 
 def test_download_file(create_buckets, create_temp_filesystem: Path):
+    """Tests both configure.get_filename and extract.download_file_from_source"""
     with open(RESOURCES / "sources.yml") as f:
-        sources = yaml.safe_load(f)
+        sources = TypeAdapter(list[Source]).validate_python(yaml.safe_load(f))
```
fvankrieken, line 16: @sf-dcp a useful function - this is how I read in https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/test/lifecycle/ingest/resources/sources.yml directly as a list of `ingest.Source` objects

dcpy/connectors/socrata/extract.py
```diff
@@ -37,7 +39,10 @@ def get_version(source: Source):
...
     This is used as a proxy for a 'version' of the dataset."""
     url = get_metadata_url(source)
-    print(url)
     resp = requests.get(url)
     resp.raise_for_status()
     return _get_version_from_resp(resp.json())
+
+
+def download(source: Source, filename: str, dir: Path):
+    web.download_file(get_download_url(source), filename, dir)
```
fvankrieken, line 48: Small wrapper but just thought it made since, since eventually we might move away from using requests for socrata downloads (towards say one of their actual api)

dcpy/lifecycle/ingest/extract.py
```diff
@@ -19,62 +17,64 @@
...
+) -> None:
+    """
+    From parsed config template and version, download raw data from source to provided path
+    """
+    path = dir / filename
+    match source:
         ## Non reqeust-based methods
         case LocalFileSource():
-            return template.source.path
+            pass
```
fvankrieken, line 30: Thought it felt a little silly to pass in the full path and also return it. Though actually, with a default arg for dir it's a bit odd. Need to think about that for a moment

dcpy/test/lifecycle/ingest/resources/tempate.yml
fvankrieken, line 1: RIP almost empty "tempate" file that I added and never used

dcpy/test/lifecycle/ingest/resources/test_sources/bpl_libraries.json
fvankrieken, line 1: These files are all dummy responses for content of `requests.get`.

dcpy/lifecycle/ingest/scripts/doe_pepmeetingurls.py
```diff
@@ -109,10 +107,3 @@ def ingest() -> pd.DataFrame:
         proposals1.append(g1)
     df = pd.DataFrame(proposals1)
     return df
-
-
-def runner(dir=TMP_DIR) -> Path:
-    df = ingest()
-    path = dir / "doe_pepmeetingurls.csv"
-    df.to_csv(path, index=False)
-    return path
```
fvankrieken, line 118: Decided it made more sense for the "script" source to just return data, not convert it to a file. This could maybe change over time, this isn't really working or tested yet. Want the rest of things to settle out (and figure out how many real "script" sources we'll have) before refining. Per a comment from @alexrichey a while back, these maybe will just become their own connectors. We'll see

dcpy/test/lifecycle/ingest/test_configure.py
```diff
@@ -55,7 +70,64 @@ def test_read_template():
     )
 
 
+@mock.patch("requests.get", side_effect=mock_request_get)
+def test_get_version(mock_request_get, create_buckets):
```
damonmcc, line 74: with the order-of-args problem you showed me, does it help having the value of `side_effect` be the same as the "could be made up" first arg? I don't remember what you had `side_effect` set to when we were looking at this locally

dcpy/lifecycle/ingest/extract.py
damonmcc: returning `None` seems nice. there's no default for `filename` and in either case of a provided or default `dir`, those two values would be known by whatever calls this function. so returning the path doesn't seem worth it

dcpy/connectors/socrata/extract.py
```diff
@@ -1,7 +1,10 @@
 from datetime import datetime
+from pathlib import Path
 import requests
 
+from dcpy.connectors import web
 from dcpy.models.connectors.socrata import Source
+from dcpy.connectors import web
```
damonmcc, line 7: ✂️

fvankrieken: Done!

fvankrieken:
Just told @damonmcc - realize there's one more silly thing about organization of extract code right now. Going to add one more commit just moving some logic around

dcpy/lifecycle/ingest/run.py
```diff
@@ -1,10 +1,19 @@
...
+from . import configure, extract, transform, TMP_DIR
 
 
 def run(dataset: str, version: str | None = None):
-    config = extract.extract_and_archive_raw_dataset(dataset, version)
+    config = configure.get_config(dataset, version)
+
+    # download dataset
+    extract.download_file_from_source(
+        config.source, config.raw_filename, config.version
```
fvankrieken, line 12: Should probably clean this up a bit, a little silly to be passing three subfields of the same object

dcpy/lifecycle/ingest/extract.py
```diff
@@ -19,62 +17,64 @@
...
             module = importlib.import_module(
-                f"dcpy.connectors.{template.source.connector}.{template.source.function}"
+                f"dcpy.connectors.{source.connector}.{source.function}"
             )
-            logger.info(f"Running custom ingestion script {template.name}.py")
-            return module.runner(dir)
+            logger.info(f"Running custom ingestion script {source.function}.py")
+            df = module.extract()
+            assert isinstance(df, DataFrame)
+            df.to_parquet(path)
```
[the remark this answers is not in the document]
damonmcc: (in-person) needed for type checking

damonmcc (approved):

dcpy/lifecycle/ingest/configure.py
```diff
@@ -51,28 +61,58 @@ def read_template(
...
+                )
             return timestamp.strftime("%Y%m%d")
 
 
-def get_config(
-    template: Template, version: str, timestamp: datetime, file_name: str
-) -> Config:
-    """Simple wrapper to produce a recipes ExtractConfig from a parsed template
-    and other computed values"""
+def get_filename(source: Source, ds_name: str) -> str:
```
sf-dcp, line 83: is the output of the fn equivalent to `raw_dataset` name attribute in Config object?

fvankrieken: Yes - or more specifically, this is what sets the raw_dataset name attribute

PR #774, opened 2024-04-18
Expand ingest preprocessing for script source coverage

### LINKED ISSUE #633: Data Library - Move "scripts" logic to processing steps
Once #632 is complete. This has two main parts, so far
1. abstract out any code in these scriptors that are generic and can be applied to multiple datasets
2. add "custom" script processing steps for relevant datasets. Maybe 1 to start, if 1 "script" source dataset has already been chosen and prototyped in #499

fvankrieken:
Resolves #633
follows up on #763

Bit of context first

## Motivation/Scope

#763 gave a very basic framework for defining preprocessing steps in yml and validating/running them in python. Pandas was chosen as the tool for transformations here because of ease of data cleaning relative to sql, though I could be convinced to go a different route. The current framework is this `Preprocessor` class in `ingest.transform` that defines the valid preprocessing functions to call. 

For this PR, I wanted to take most/all of the library datasets that have custom "script" sources (which in reality have very normal sources much like our other datasets - api calls, local files, etc), but actually rely on their "source" script to perform some sort of preprocessing. 

The first 6 commits here are more generic ingest stuff, code cleanup, etc - then after that, each commit is adding one or multiple templates from library to ingest, and adding the appropriate yml and new preprocessors as needed to correctly spit out what's needed. Going commit-by-commit will definitely be easiest, and I think I've sorted out most of the cases where I might be overwriting changes in future commits.

Most of the logic is concerned with `dcpy.lifecycle.ingest.transform` - the bottom function reads in the initial parquet file as a dataframe, then applies preprocessing steps as defined in yml input. There are some that are a little more "extract"-focused as well.

Happy to sit down and chat about anything if it'd be helpful

### General
1. Bit of a bug - for consistency, simplest to just move a local file into the tmp/staging/etc dir that external files get downloaded to
2. Also bug - `Path` objects don't serialize nicely by default as it turns out
3. Cleanup of logic around where file outputs are in extract. Currently for simplicity/flexibility, each big step is dumping out a parquet file. There was too much assumed logic around where these files were. I move the default location setting logic to be at runtime (in `run.py`), and the other functions require it as an arg. Just with testing, running it myself, this seemed to make more sense. **would love thoughts around this, as it maybe is still a tad clunky**
4. Bunch of little tweaks to `transform` - mainly around what arguments are geting shuttled along, taking paths as inputs rather than returning as outputs. Fairly in line with commit 3
5. Two things, which are tough to disentangle because I made this one commit instead of 2
   - moving some validation code out of `ingest.configure` where it doesn't belong
   - a custom recursive [`isinstance`](https://github.com/NYCPlanning/data-engineering/pull/774/commits/d891b806658eaadfc89f870441db5a691e11e054#diff-565b79993fcd991ac240d9f5353170e4b9973afa548773f47b00f789d3f4fb59R6) function. isinstance doesn't work for parameterized generics, and there are a few that would be nice to check - literals, types of lists/dicts, etc. Needed custom line for Union too, because `isinstance` does work on them, but wouldn't work if one of the types is a parameterized generic
6. There's a test that validates all yml templates in `ingest`. That ideally should also validate all processing steps defined. Moved this since we're getting a little out of the configure logic

### dataset-specific

Links here will be to python script in `library/script` generally

7. add template for [dcp_sfspd](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/dcp_sfpsd.py) and filter_rows
8. [dob_cofos](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/dob_cofos.py). This needed use of computed version in jinja template, renaming of columns, setting value of columns, and getting my first pass at `append_prev` to actually work. Additionally added the generic pd.Series function. Essentially, allow users to specify a string like "map" or "str.split" that is a callable series function (`df["column"].str.split(...)`), and kw arguments to said function. Also adds logic to validate calls to this function. Due to weirdness around what inspect returns for these functions, I needed to rework the validation function a bit. But all tests passing. For these functions, type cannot be validated but at least the arguments (presence of, whether there missing) is validated
9. [dpr_capitalprojects](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/dpr_capitalprojects.py), [nypl_libraries](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/nypl_libraries.py), [bpl_libraries](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/bpl_libraries.py). This doesn't have any added preprocessing logic, but adds a basic implementation for reading json source to df. The columns thing feels a little hacky but I was having errors trying to dump the "raw" dataframe to parquet. Which we don't technically need but is still nice just as a troubleshooting help.
10. dob_now [permits](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/dob_now_permits.py) and [applications](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/dob_now_applications.py). Also did not need scripting logic, but did need an s3 source. This maybe gets replaced with local file upload? If we're archiving raw data, not sure we need to load to private s3 before archiving raw in s3
11. dob bis applications and permits. These didn't have script sources but were our only two datasets with `sql` specified in gdal options, so that logic needs to be recreated. Tweaks rename_columns and adds a function to clean column names
12. fisa [commitments](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/fisa_capitalcommitments.py) and detailed [budget data](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/script/fisa_dailybudget.py). Needed ability to specify column names when reading in raw data, as well as deduplication logic for the detailed budget data since it appends to prev a la cofos


dcpy/connectors/web.py
```diff
@@ -6,8 +6,11 @@
 
 def download_file(url: str, path: Path) -> None:
     """Simple wrapper to download a file using requests.get."""
+    default_headers = {
+        "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64)",
+    }
     logger.info(f"downloading {url} to {path}")
-    response = requests.get(url)
+    response = requests.get(url, headers=default_headers)
```
fvankrieken, line 13: dpr_capitalprojects api returns error without this

fvankrieken:
TODO before review

- print statement in cofos commit
- combine/clean order of commits in terms of pd series and cofos
- combine, document better json commits (9 and 15)
- tests for pd series function, validation

fvankrieken:
I need to add tests for commit 8 (pd Series logic), but I don't anticipate needing to change non-test code as I've done a lot of use-case testing. Given that, would appreciate input and review!

dcpy/lifecycle/ingest/configure.py
```diff
@@ -107,6 +107,7 @@ def get_config(dataset: str, version: str | None = None) -> Config:
     template = read_template(dataset, version=version)
     filename = get_filename(template.source, template.name)
     version = version or get_version(template.source, timestamp)
+    template = read_template(dataset, version=version)
```
alexrichey, line 110: duped from line 107?

dcpy/utils/validation.py
```diff
@@ -0,0 +1,97 @@
...
+        return (
+            isinstance(obj, dict)
+            and all([_isinstance(key, keytype) for key in obj])
+            and all([_isinstance(obj[key], valtype) for key in obj])
+        )
+    else:
+        return isinstance(obj, cls)
+
+
+# This maybe belongs in a more pure util section of code
```
alexrichey, line 34: mission accomplished ✂️

dcpy/utils/validation.py
```diff
@@ -0,0 +1,97 @@
...
+    ignore_args = ignore_args or []
+    sig = inspect.signature(function)
+
+    pos = [
+        a
+        for a in sig.parameters.values()
+        if a.kind == a.VAR_POSITIONAL or a.kind == a.POSITIONAL_ONLY
+    ]
+    if pos:
+        raise TypeError(
```
alexrichey, line 57: Is this inconsistent with the `raise_error` arg? In the spirit of "a function should do one thing only" maybe it's best to not raise errors and just return the violations?

fvankrieken: Maybe, though this is a different error - the function itself being validated is not a format that allows validation with this function, rather than some sort of error with the arguments (which is what is supposed to be validated here)

dcpy/test/utils/test_validation.py
```diff
@@ -0,0 +1,116 @@
+import pytest
+from dcpy.utils import validation
+
+
+def test_validate_function_args():
```
alexrichey, line 5: Great variety of test cases! For readability, would love it if you could split these into maybe one test per validated function? And would it be possible to just put the dummy data into a list of `[signature, shouldValidate, reason]` and then iterate through them?

dcpy/utils/validation.py
alexrichey, line 1: Should we name this something like `introspect`, or `py_functions` (or `inexplicably_missing_from_std_lib` 🙂)

fvankrieken: Yeah I think something along those lines would make sense

dcpy/utils/validation.py
alexrichey: Gotcha - I do think that's somewhat unexpected behavior for a function called `validate_function_args`. My gut is that supplying args vs kwargs here isn't so different as to warrant such a change in behavior (ie, raising error vs return errors) but I get the distinction

fvankrieken: Yeah mainly I was thinking of my use case - I can play around and see if I can just make it slightly more flexible, because I agree that on the face of it, it doesn't necessarily warrant different behavior

alexrichey (approved):
Very cool to see this in action, specifically the templates with all the processing steps. Had a few nits, but otherwise all good. I was wondering what's the expected way to handle operations that are more complicated than we can encapsulate in the declarative processing_steps? Ie what if there's some really complicated filtering? (or does that use-case not really exist in our recipe sources?)

fvankrieken:
I have currently managed to cover our use cases (I think!), but I also think it's perfectly fine to have to have a specific preprocessing function called `process_cofos` or something like that if need be

alexrichey:
> I have currently managed to cover our use cases (I think!), but I also think it's perfectly fine to have to have a specific preprocessing function called `process_cofos` or something like that if need be

Gotcha - where does that type of thing live?

dcpy/utils/data.py
```diff
@@ -118,6 +119,17 @@ def read_data_to_df(
...
+            with open(local_data_path) as f:
+                json_str = json.load(f)
+            if data_format.normalize:
+                gdf = pd.json_normalize(json_str, **data_format.normalize)
+            else:
+                gdf = pd.read_json(json_str)
+            # Feels a tad hacky to have this here instead of in preprocessing
+            # But json columns were causing parquet issues
+            if data_format.columns:
+                gdf = gdf[data_format.columns]
```
damonmcc, line 133: curious what the parquet issues are

does this mean that if a template doesn't have a `columns` list then `read_data_to_df` might cause parquet issues?

fvankrieken: This was specifically conversion of a json column - and seems to be a pandas/pyarrow serialization issue. Hence why this is only defined in the json source format. I only needed it for this specific dataset

fvankrieken:
> > I have currently managed to cover our use cases (I think!), but I also think it's perfectly fine to have to have a specific preprocessing function called `process_cofos` or something like that if need be
> 
> Gotcha - where does that type of thing live?

Could be in the preprocessor object along with the other functions. Not the most elegant for now

damonmcc (approved):

PR #777, opened 2024-04-19
Get publishing version from build metadata

fvankrieken:
Pretty self explanatory - get version from build_metadata.json instead of from version.txt.

Figure we'll still keep the version.txt, it's nice to have. But also have it be consistent across products now

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -9,6 +9,7 @@ from io import BytesIO
 from zipfile import ZipFile
 import typer
 from typing import Callable, TypeVar
+import yaml
 
 from dcpy.utils import s3, git, versions
 from dcpy.utils.logging import logger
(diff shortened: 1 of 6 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -18,18 +19,24 @@
...
+
 def get_version(product_key: ProductKey) -> str:
-    """Given product key, gets version
-    Assumes existence of version.txt in output folder
-    """
-    return s3.get_file_as_text(BUCKET, f"{product_key.path}/version.txt").splitlines()[
-        0
-    ]
+    """Given product key, gets version"""
+    return get_build_metadata(product_key).version
```
fvankrieken, line 39: Very small wrapper now, but still nice to have as its own function given that we often want this, and then code dealing with it doesn't need to know about `BuildMetadata` class at all

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -50,6 +50,7 @@ def test_upload(create_buckets, create_temp_filesystem, mock_data_constants):
 
 
 def test_publish(create_buckets, create_temp_filesystem, mock_data_constants):
+    """Tests publish function as well as get_version and get_latest_version"""
```
fvankrieken, line 53: Expand the rest of this function to see how get_version and get_latest_version is tested here. Not the most unit of tests given that it's also testing publish, but it's still fairly complete in terms of what we'd want tested

damonmcc (approved):

damonmcc:
noting that this was related to #776

PR #786, opened 2024-04-22
Ingest: Clean up & Simplify Tests

sf-dcp:
Related to #631.

### Motivation
Clean up `to_parquet` fn and its associated tests. 

### Changes
1) Previously,  `to_parquet` fn took in a `recipes.Config` object as its input and then parsed it inside. Given that the Config object is evolving (the structure of Config object may change, i.e. new attributes) and we just need its `file.Format` object, I replaced `recipes.Config` with `file.Format`. This simplifies tests dramatically because we don't need to worry about creating a whole fake `recipes.Config` object to test `to_parquet`. 

2) I removed the part downloading data from s3 from `to_parquet` fn. We already download data in `run.py` before we call `to_parquet`, so the data is already available locally. This change makes  `to_parquet` more atomic. 

3) Update test resources. Previously, we were creating fake `recipes.Config` objects in a test file, and the code started becoming lengthy because of a wide variety of data formats (test cases). So make it more simple, I create custom objects in `transform_to_parquet_template.yml` where each object, i.e. test case, consists of `file.Format` and test file name. I feel it's easier to maintain these tests long term if we have all info about test cases in one place, the yaml file.

### DIFF
--- a/dcpy/lifecycle/ingest/transform.py
+++ b/dcpy/lifecycle/ingest/transform.py
@@ -14,12 +14,17 @@ from dcpy.models import file
 from dcpy.models.lifecycle.ingest import FunctionCall
 from dcpy.utils.logging import logger
 from dcpy.connectors.edm import recipes
-from . import TMP_DIR, PARQUET_PATH, configure
+from . import TMP_DIR, configure
 
 OUTPUT_GEOM_COLUMN = "geom"
 
 
-def to_parquet(file_format_config: file.Format, local_data_path: Path):
+def to_parquet(
+    file_format_config: file.Format,
+    local_data_path: Path,
+    dir: Path = TMP_DIR,
+    output_filename: str = "tmp.parquet",
+) -> None:
     """
     Transforms raw data into a parquet file format and saves it locally.
 
@@ -33,16 +38,17 @@ def to_parquet(file_format_config: file.Format, local_data_path: Path):
     Parameters:
         file_format_config (file.Format): Config object containing geometry info.
         local_data_path (Path): Path to the local data file.
+        dir (Path): Directory to use for output file.
+        output_filename (str): Output file name.
 
     Raises:
         AssertionError: `local_data_path` does not point to a valid file or directory.
         AssertionError: If `geom_column` is present in yaml template but not in the dataset.
     """
 
-    # create new dir for raw data and output parquet file
-    if TMP_DIR.is_dir():
-        shutil.rmtree(TMP_DIR)
-    TMP_DIR.mkdir()
+    # create new dir for output parquet file if doesn't exist
+    dir.mkdir(parents=True, exist_ok=True)
+    output_file_path = dir / output_filename
 
     assert (
         local_data_path.is_file() or local_data_path.is_dir()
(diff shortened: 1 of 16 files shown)


dcpy/test/lifecycle/ingest/test_transform.py
```diff
@@ -1,103 +1,69 @@
...
     transform,
     PARQUET_PATH,
 )
 
 from dcpy.utils import data
-from . import RESOURCES, TEST_DATA_DIR, FAKE_VERSION
+from . import RESOURCES, TEST_DATA_DIR
+
+
+class TestConfig(BaseModel):
```
sf-dcp, line 18: I created this pydantic class to validate objects from `transform_to_parquet_template.yml`. I expect this yaml file to grow with more test cases, and it would be nice to validate them. I'm not sure about the location of this class, however. Should it be moved to `models` module or stay here?

dcpy/lifecycle/ingest/transform.py
```diff
@@ -45,22 +44,12 @@ def to_parquet(config: Config, local_data_path: Path | None = None):
         shutil.rmtree(TMP_DIR)
     TMP_DIR.mkdir()
```
fvankrieken, line 45: Maybe cut these three lines as well? If the tmp dir is more being handled by `run` at this point

dcpy/test/lifecycle/ingest/test_transform.py
fvankrieken: Love this! I would keep it here for sure.

fvankrieken:
Love it! Have one comment about a few lines that maybe should be cut but other than that this is great, everything is much cleaner with these tweaks

dcpy/test/lifecycle/ingest/test_transform.py
```diff
@@ -1,103 +1,69 @@
...
+    """
+    Test pydentic class used to validate input yaml file
+    """
+
+    file_name: str
+    to_parquet_config: Format
```
damonmcc, line 24: (just an idea, not a request to change asap)

I wonder if better names for these might be `file_config: FileConfig`? since it seems like the models included in `models.file.Format` are used as configs, and one of their attributes is named `format`

dcpy/lifecycle/ingest/transform.py
sf-dcp: I actually did cut these lines at first. But when I ran the tests with pytest, I got an error that the folder doesn't exist. Would you recommend to move this code snippet under the tests setup?

dcpy/test/lifecycle/ingest/test_transform.py
fvankrieken: I would prefer not calling something else a config. I think "File Format" is pretty intuitive (and a commonly used term). But maybe call the attribute of these different formats `type` rather than `format`

dcpy/lifecycle/ingest/transform.py
fvankrieken: I think that'd make sense

fvankrieken: We're in a slightly funky place right now where there's some semi-assumed folder/file structure that could use some clarifying, but this PR is nice and isolated so that can be more taken care of in #773 which is more meant for overall refactoring (and has some overlap with this PR).

Like ideally, test code doesn't need to "know" about some default temporary directory ideally. But that's better than getting errors here or having code that could wipe out the downloaded raw file before it's transformed to parquet

sf-dcp: I see. What about modifying this snippet as: "check this dir exists. if it does, move on. If it doesn't, create it for me." and keeping it inside of the `transorm.py`? I wonder if there could be a case where you want to use the function alone, outside of `run.py`

fvankrieken: And yeah, seems fine to make sure that the output folder exists and make it if it doesn't. I just still think ideally that's coming from some sort of function arg (with default perhaps) rather than a hard-coded constant

sf-dcp: I think I've seen that in your PRs, may steal it for this fn

PR #789, opened 2024-04-23
revise GFT lot zoning logic

### LINKED ISSUE #741: revise Zoning logic

damonmcc:
resolves #741


products/green_fast_track/models/product/green_fast_track_bbls.sql
```diff
@@ -55,11 +55,20 @@ flags_wide AS (
...
     {% for row in variables -%}
-        CASE
-            WHEN f."{{ row['label'] }}" IS NULL THEN 'No'
-            ELSE 'Yes'
-        END AS "{{ row['label'] }} Flag",
-        f."{{ row['label'] }}",
+        {% if row['variable_type'] == 'zoning_districts' %}
+            /* zoning_districts has a Category column */
+            f."{{ row['label'] }}" AS "{{ row['label'] }} Category",
+            array_to_string(ARRAY[pluto.zonedist1, pluto.zonedist2, pluto.zonedist3, pluto.zonedist4],
```
fvankrieken, line 61: Does this include nulls in the array?

damonmcc: oh true! this approach is definitely ugly and not DRY

maybe better if I make sure the list generated during `int_flags__zoning` is accessible here?

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
```diff
@@ -2,50 +2,93 @@ WITH pluto AS (
...
+district_categories AS (
+    SELECT * FROM {{ ref('zoning_district_categories') }}
+),
+
+districts_exploded AS (
     SELECT
         bbl,
-        UNNEST(ARRAY[zonedist1, zonedist2, zonedist3, zonedist4]) AS zd
+        UNNEST(ARRAY[zonedist1, zonedist2, zonedist3, zonedist4]) AS zd,
+        COALESCE(zonedist1, zonedist2, zonedist3, zonedist4) IS null AS zd_all_null
```
fvankrieken, line 17: Should be able to just check zonedist1 - pluto has logic to ensure that you never have zonedist{n+1} is not null if zonedist{n} is null

products/green_fast_track/models/product/green_fast_track_bbls.sql
fvankrieken: Eh that list gets chopped to a million pieces. I think this is fine, just should have somethin to filter out nulls in creating the array so that users don't see "C1, Null, Null, Null"

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
```diff
@@ -2,50 +2,93 @@ WITH pluto AS (
...
+            ELSE 'high_res'
+        END AS category_type
+    FROM lots_with_flags
+),
+
+zoning_district_flags AS (
+    SELECT
+        districts_resolved.bbl,
+        'zoning_districts' AS variable_type,
+        district_categories.label AS variable_id
```
fvankrieken, line 89: And here again - if this is the only spot where we use `district_categories`, should we just have "R1-R4 with Commercial or Manufacturing" earlier in this sql query instead of "low_res_c_or_m"? Either way we have something hardcoded, so maybe we just go with the nicer label rather than having these intermediate labels that are immediately disposed

fvankrieken: A very similar case could be made for our `variable_type`, maybe might as well use them as a label.

I do think this is more clear cut though, given that the string "low_res_c_or_m" never leaves this file really - it's created here, joined to a seed, and then left out of the returned table

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
```diff
@@ -2,50 +2,93 @@ WITH pluto AS (
...
+            ELSE zd
+        END AS zd_type
+    FROM districts_distinct
 ),
 
-zoning_districts_regrouped AS (
+districts_mapped AS (
+    SELECT
+        district_types.*,
+        district_mappings.category_type
```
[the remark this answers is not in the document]
damonmcc: definitely would be nice to have a new model to contain all this logic, and may help with the thoughts above re `green_fast_track_bbls`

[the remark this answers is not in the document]
fvankrieken: having finished looking through that's still very much my preference
- one model that does not rely on district_mappings that gets to some intermediate of district types/categories by bbl in long form
- one model (int_flags__zoning) that takes the above model and has logic to determine final categorization

damonmcc:
for readability and our future selves, gonna pull out the zoning district logic from `int_flags__zoning ` into a new model

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
fvankrieken: Going back on this - it's nice to have some internal label that never needs to change, and one simple spot where we have an external label in a single seed column

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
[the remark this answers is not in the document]
damonmcc: @fvankrieken see https://github.com/NYCPlanning/data-engineering/pull/789/commits/5b0a55656b32081d3a868fbbea304b4f5d4e3acf

products/green_fast_track/models/product/green_fast_track_bbls.sql
damonmcc: noting that we found that `array_to_string` ignores nulls and if their all null the default is an empty string

products/green_fast_track/models/intermediate/flags/int__zoning_districts.sql
```diff
@@ -0,0 +1,37 @@
...
+    FROM pluto
+),
+
+districts_distinct AS (
+    SELECT
+        bbl,
+        zd
+    FROM districts_exploded
+    WHERE (zd IS NOT null AND NOT zd_all_null) OR zd_all_null
+    GROUP BY 1, 2
```
fvankrieken, line 19: Nit - group/order by column names rather than numbers?
Also since we're not actually aggregating here, this one could just be a `SELECT DISTINCT` instead of a group by

fvankrieken: And actually, pluto build ensures that these fields are distinct from each other, so the distinct isn't needed in this CTE I don't think

fvankrieken: Well other than the all nulls I suppose. Hmm

fvankrieken: For districts exploded, could do

```sql
SELECT
    bbl,
    unnest(CASE 
        WHEN zonedist1 IS NULL THEN ARRAY['NONE']
        ELSE ARRAY[zonedist1, zonedist2, zonedist3, zonedist4]
    end) AS zd
FROM pluto
```

And then after that, only have to filter `WHERE zd IS NOT NULL`

damonmcc: since Jack is using this branch's build export and this being a nit, let's save this for later

fvankrieken: If by later, you mean still before merging!

fvankrieken: Edited above query slightly to make even simpler

Really, it might make most sense to put that `COALESCE` in `stg__pluto`

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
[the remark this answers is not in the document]
fvankrieken: Resolved!

products/green_fast_track/models/intermediate/flags/int_flags__zoning.sql
```diff
@@ -2,50 +2,69 @@ WITH pluto AS (
...
-    FROM pluto
+        zoning_districts.bbl,
+        zoning_districts.zoning_district_type,
+        district_mappings.category_type
+    FROM zoning_districts
+    LEFT JOIN district_mappings ON zoning_districts.zoning_district_type = district_mappings.zoning_district_type
 ),
 
-zoning_districts_split_exploded AS (
+lot_with_zoning_categories AS (
```
fvankrieken, line 26: This is maybe just personal preference, but I would do the group by in the last query - both are very short and readable, and one more line will still be quite legible

fvankrieken: But at a minimum, I will again protest the `GROUP BY 1`, just seems to be inviting error down the road

damonmcc:
@fvankrieken I changed the group/order bys to use column names rather than numbers and moved the last group by in `int_flags__zoning` to the last CTE

appreciate your suggestions for changing part of `int__zoning_districts` to use your preferred approaches, but it isn't worth holding up this PR which also fixes GFT being broken on main. let's save it for a follow-up

fvankrieken (approved):

PR #790, opened 2024-04-24
Workflow stub for running ingest

fvankrieken:
For use in #774 (for datasets that are large enough that I'm having issues running in my dev container. A sign that we should move away from pandas? Perhaps)


fvankrieken:
Might close this in favor of #791

PR #791, opened 2024-04-24
github action for single freeform command

fvankrieken:
Any concerns with something like this? Seems useful for dev tasks that have performance constraints

### DIFF
--- a/.github/workflows/single_command.yml
+++ b/.github/workflows/single_command.yml
@@ -1,0 +1,48 @@
+name: Run a single command
+
+on:
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


damonmcc (approved):
seems useful! I imagine needing more secret envars for some commands, but those can always be added later

alexrichey:
@fvankrieken I think there are some potential concerns around security here. Since we're loading secrets into the container, anyone who can run this has full access to all creds

fvankrieken:
Yeah, makes sense.

alexrichey:
however, would this be useful without creds?

PR #804, opened 2024-04-25
FacDB - fix issues in 24v1

sf-dcp:
Related to #461.

## What
During the 24v1 review, GIS identified this issue:
* `dfta_contracts`: SENIOR SERVICES facsubgrp dropped from 75 to zero record count

## Why
When creating the `qc_diff` table used in the QA review, we filter out records with empty values in the `geom` column. It appears that `dfta_contracts` records weren't getting geocoded by `geosupport` in the beginning of the build. Specifically, the source zip code column name changed in this data version, and our code silently failed, returning empty values from `geosupport`

**Solution**: fix zip code column name  

#### Build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8852190924/job/24310349837).

As can be seen, now `dfta_contracts` records are getting geocoded: `count_new` is not zero.
image

Old records for comparison: 
image

### DIFF
--- a/products/facilities/recipe.yml
+++ b/products/facilities/recipe.yml
@@ -82,7 +82,6 @@ inputs:
     - name: dycd_service_sites
     - name: fbop_corrections
     - name: fdny_firehouses
-    - name: nysomh_mentalhealth
     - name: dep_wwtc
       version: "20210413"
     - name: dfta_contracts
--- a/products/facilities/facdb/pipelines.py
+++ b/products/facilities/facdb/pipelines.py
@@ -130,7 +130,7 @@ def dfta_contracts(df: pd.DataFrame):
     df = Function1B(
         street_name_field="parsed_sname",
         house_number_field="parsed_hnum",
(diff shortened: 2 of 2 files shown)


products/facilities/recipe.yml
```diff
@@ -82,7 +82,6 @@ inputs:
     - name: dycd_service_sites
     - name: fbop_corrections
     - name: fdny_firehouses
-    - name: nysomh_mentalhealth
```
sf-dcp, line 85: this is a duplicate recipe. Removing it doesn't impact output.

damonmcc (approved):
nice, did u wanna build on main with the pinned versions? or gonna drop the commit now that the build on this branch succeeded?

sf-dcp:
> nice, did u wanna build on main with the pinned versions? or gonna drop the commit now that the build on this branch succeeded?

I was going to drop the commit before merging, unless you suggest otherwise?

PR #811, opened 2024-04-29
fix typo in overly-eager dev tag deletion

fvankrieken:
"dev" image tags created in PRs with changes to python, docker, etc are deleted each week in a recurring job, IF their branch no longer exists.

At least that was intended. To avoid potential collisions with real tags, all dev image tags begin with "dev-". So for this branch, there would be images "build-base:dev-fvk-docker-cleanup". However, cleanup logic was comparing this tag name to JUST the branch name, and since "dev-fvk-docker-cleanup" is not an existing branch, the tag was being deleted everytime.

### DIFF
--- a/dcpy/lifecycle/builds/clean_artifacts.py
+++ b/dcpy/lifecycle/builds/clean_artifacts.py
@@ -46,16 +46,16 @@ def delete_stale_schemas(active_build_names: list[str]):
                 logger.info(f"Keeping tests schema {database}.{build_test_schema}")
 
 
-def delete_stale_image_tags(active_build_names: list[str]):
+def delete_stale_image_tags(active_build_names: list[str]) -> None:
     logger.info(f"Potential active branch build names: {active_build_names}")
     DOCKER_IMAGES = ["dev", "build-base", "build-geosupport"]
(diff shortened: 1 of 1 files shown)


dcpy/lifecycle/builds/clean_artifacts.py
```diff
@@ -55,7 +55,7 @@ def delete_stale_image_tags(active_build_names: list[str]):
         ).json()["results"]
         dev_tags = [tag for tag in tags if tag["name"].startswith("dev-")]
         for tag in dev_tags:
-            if tag not in active_build_names:
+            if tag[4:] not in active_build_names:
```
damonmcc, line 58: seems safer to use the prefix rather than an index, mayb

```python
if tag.removeprefix("dev-") not in active_build_names:
```

I was gonna say `tag.split("dev-")[1]`, but a case like `dev-dev-my-branch` ruined that

damonmcc (approved):

fvankrieken:
mypy helped me not introduce another bug. Flaky tests now failing due to socrata - I thought I had removed them, I'll check those out in another PR

PR #813, opened 2024-04-29
mock requests.get response for socrata tests

fvankrieken:
Move the mock requests logic to top-level conftest so that it can be used in socrata tests as well as ingest


dcpy/test/connectors/socrata/test_extract.py
```diff
@@ -1,29 +1,18 @@
...
 
 from dcpy.models.connectors import socrata
 from dcpy.connectors.socrata import extract
 
+from dcpy.test.conftest import mock_request_get
 
-RESOURCES = Path(__file__).parent / "resources"
 
-
-def test_parse_version():
```
fvankrieken, line 11: Seemed a little redundant with the mocked response

alexrichey (approved):

PR #817, opened 2024-04-29
Add Historic Resources Shadows Combined

alexrichey:


products/green_fast_track/models/product/green_fast_track_bbls.sql
```diff
@@ -49,6 +49,14 @@ flags_wide AS (
             )
             AND variable_type != 'wetlands_checkzones'
         ) AS natural_resource,
+        FILTER (
```
fvankrieken, line 52: creeping on you because I had to see what commit "Maybe?" was doing. Looks like you're missing whatever array_agg comes before this `FILTER`, so sqlfluff is interpreting `FILTER` as a function name rather than a keyword

fvankrieken: On line 40 I have `bool_or` which is an aggregate function, which `FILTER` on line 41 then modifies

fvankrieken: And `any` should in fact still be lowercase

products/green_fast_track/bash/export.sh
```diff
@@ -90,15 +90,18 @@ mkdir -p output && (
     export_source sources__nyc_historic_buildings_points MULTIPOINT
     export_source sources__nyc_historic_buildings_lots
     export_source sources__nyc_historic_buildings_buffers
+    export_source sources__nyc_historic_buildings_buffers_200
```
fvankrieken, line 93: @damonmcc do you know if these exports are needed specifically, or if just the shadow buffer on line 104 is needed? I imagine that was part of your conversation leading to #826

damonmcc: only the shadow buffer is needed. but no worries on getting these source geometry export exactly right in this PR!

Jack and I are gonna touch base next week to decide what would be ideal for these exports

fvankrieken: Cool, I'll leave this then

fvankrieken:
This will be squashed. Straightforward enough to review all changes at once

products/green_fast_track/models/staging/stg__nysparks_historicplaces.sql
```diff
@@ -5,6 +5,7 @@ WITH historic_places AS (
 SELECT
     ST_TRANSFORM(wkb_geometry, 2263) AS geom,
     historicname,
+    citytown,
```
damonmcc, line 8: this should be a real word

damonmcc (approved):

Issue #769, opened 2024-04-15
Choose Data Contract Framework


sf-dcp:
Related to #650. Creating a separate issue because 1) usage of data contracts can expand beyond data library (i.e. for non-`dbt` data products and 2) to provide space for discussion.

## Motivation
As a part of the data library revamp (AKA ingest,AKA extract), we would like to introduce data contracts into data ingestion process to keep our data clean and consistent from the start. With data contracts, we can set clear rules on what data should look like and catch issues such as column name changes before data product builds, making actual builds faster. It will also provide transparency and enable cross-team collaboration, such as embedding business logic to enhance data quality. All in all, it should make everyone's life easier 🤞 

## Approach
There exist multiple frameworks for data validation. From my research, the main open source tools are [Great Expectations](https://greatexpectations.io) and [Soda](https://www.soda.io). A smaller one is [Pandera](https://pandera.readthedocs.io/en/stable/). 

Thinking about our needs, I came up with the following guidelines to evaluate the tools:

-  **Documentation quality**
-  **Easy-to-read data contracts/data checks**
    * This will help us and other teams to contribute in contract development
-  **Integrates with a variety of storages/data formats**
   * In particular, integration with local files, from S3, and database
- **Flexible data checks** 
    * Huge pro if it can do geospatial checks
    * Allow for custom data checks
- **Simple to write new data checks**
- **Integration with existing data products [new]**
    * Initially, the idea was to add data validation in the recipe ingestion process. Thinking about improving existing data products quality, it would be nice if a framework of our choice could be easily imbedded in existing pipelines; one example could be running a set of checks on final tables before exporting them. 

### Pandera
* [link](https://pandera.readthedocs.io/)
* Python-based framework
* Integrates with dataframes (pandas, geopandas, dask, etc)
* Contracts are written in python
* Good documentation
* **Pros**: 
    * Works with python data types and pandas dtypes
    * Integrates with `mypy` for typechecking
    * Can use `pydentic` classes to define table schemas and data checks 
* **Cons**: 
    * Need to define schemas for each recipe in python versus having a yaml template
    * Some data contract code is not very readable ([example](https://pandera.readthedocs.io/en/stable/checks.html))
    * having to read data into a dataframe for validation (perhaps not a con?)

### Great Expectations (GX)
* [link](https://greatexpectations.io)
* Python-based framework
* Integrates with postgres, s3, and local files via dataframes (pandas), etc ([link](https://greatexpectations.io/integrations/))
[...]

PR #818, opened 2024-04-30
[Experiment] Data validation with Soda

sf-dcp:
Related to #769.

**TLDR**: Easy-to-use framework. May not work with geospatial data checks. 

Note, this PR has 2 commits: 1st commit is the one implementing Soda, and the write-up below focuses on the 1st commit; 2nd commit shows a new `Soda` feature using data contracts instead. 

## What
Implement data quality checks for `bpl_libraries` recipe in `template-db` dataset with `Soda`. 
Checks: 
* `title` column: can't have duplicate and null values
* `region` column: values should be within a defined set of values
* `wkb_geometry` column: can't have null values and should have state projection (custom data check). 
   * geospatial check didn't work

## Where
Checks are performed in our Postgres db. Among many integrations with various sources, Soda can also be used on local files via `dask` or `spark` dataframe. 

## How
*This section describes implementation from first commit.*

Soda is very lightweight: it requires only 2 files for deploying data checks: 

* `soda_sip/configuration.yml` - contains db conn credentials: I used env variables to define them.  
 
* `soda_sip/bpl_libraries_checks.yml` - the actual data checks, aka expectations for `bpl_libraries` recipe. I also added more schema check examples that I found useful.

#### Running checks from CLI:
```bash
soda scan --data-source template_db --configuration ./configuration.yml bpl_libraries_checks.yml 
```
image

An optional argument in the CLI command is saving output in a json file which can be viewed at `soda_sip/output.json`.

## Other Soda notes

* I like how easy and quick is defining data checks and the ability of creating custom data checks with a free-form SQL statement
* I said above that Soda may not work with geospatial checks... In theory, it should because you can write free from SQL queries that are run in your engine of choice. However, I was getting a weird error: 
    
    image
    I connected with Soda in their slack channel and they said it should work. They said I may have [this](https://gis.stackexchange.com/questions/324386/psycopg2-programmingerror-function-does-not-exist) issue.
* Implementing Soda with local files via required `dask` dataframe is a bit annoying because it needs to be done in Python and `dask` uses presto SQL dialect under the hood 
* Regarding the data contract feature and how it's different from the main implementation:
    * allows to define a true data contract for a dataset with custom key-value pairs such as product owner, description, contact info, etc
    * all columns must be defined with an optional key whether a column is required.
    * as of now, must be invoked from Python


sf-dcp:
Update: geospatial checks work when you specify schema where geospatial functions live (typically in `public`):

image

All credit to @damonmcc 🎉

PR #823, opened 2024-04-30
fix missing GFT lots in exported source data

### LINKED ISSUE #815: lots missing from GFT source data exports
## problem

For many datasets we:
1. get points (raw)
2. replace with a lot polygons if one intersects (intermediate)
3. buffer the lots and remaining points (buffered)

For these datasets we export the point, lot, and buffered geometries (e.g. CATs permits, NYC Historic Buildings) so that they can been seen in the GFT app.

GIS noticed that, although the buffers are correct, some `source_xxx_lots` tables are missing lots.

damonmcc:
resolves #815 

successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/8913860231). all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-gft-source-geometry)

I found which models to change by searching for all models in `/product/sources` that select records `WHERE ST_GEOMETRYTYPE(raw_geom) = 'ST_MultiPolygon'` and changing the model they select from.

See [last commit message](https://github.com/NYCPlanning/data-engineering/pull/823/commits) for explanation of the substantial changes to `int_buffers__lpc_landmarks`. GIS is ok with (and actually likes) not seeing source points when lots where they were joined to lots.

### before
Screenshot 2024-05-01 at 12 39 44 PM

### after
Screenshot 2024-05-01 at 12 40 10 PM

### DIFF
--- a/products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
+++ b/products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
@@ -46,6 +46,7 @@ WITH all_buffers AS (
         ref('int_buffers__lpc_landmarks'),
         ref('stg__nysshpo_archaeological_buffer_areas')
     ],
+    include=["variable_type", "variable_id", "raw_geom", "buffer"],
     column_override={"raw_geom": "geometry", "buffer": "geometry"}
 ) }}
 )
--- a/products/green_fast_track/bash/export.sh
+++ b/products/green_fast_track/bash/export.sh
@@ -87,7 +87,7 @@ mkdir -p output && (
     export_source sources__natural_resources_buffer
 
     # Historic
-    export_source sources__nyc_historic_buildings_points POINT
+    export_source sources__nyc_historic_buildings_points MULTIPOINT
     export_source sources__nyc_historic_buildings_lots
     export_source sources__nyc_historic_buildings_buffers
     export_source sources__nyc_historic_districts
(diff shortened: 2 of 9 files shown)


products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
```diff
@@ -46,6 +46,7 @@ WITH all_buffers AS (
         ref('int_buffers__lpc_landmarks'),
         ref('stg__nysshpo_archaeological_buffer_areas')
     ],
+    include=["variable_type", "variable_id", "raw_geom", "buffer"],
```
fvankrieken, line 49: What does this do?

fvankrieken (approved):
Is it potentially worth using ST_MULTI on just about every data source, just to make sure we don't have other

products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
damonmcc: just noting for later:

didn't know this util creates a column called `_dbt_source_relation` "to indicate the source for each record"!! seems like a cool way to avoid declaring `variable_type` strings in these models and just use the model names to drive queries??

damonmcc:
@fvankrieken 
> Is it potentially worth using ST_MULTI on just about every data source, just to make sure we don't have other

good call, but I didn't wanna go too far since (especially historic) source data seems to be in flux at the moment and we wanna make a pass at standardizing stuff more generally

all good if we roll with this while Jack and I keep an eye out for other issues?

sf-dcp (approved):
LGTM!

products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
fvankrieken: Ooh that's interesting. Love that

Issue #816, opened 2024-04-29
change GFT source data exports


damonmcc:
to follow-up on #814, two GIS change requests are related to:
1. the columns that are needed for users and EARD to know which dataset value contributed to a question flag
2. the `product/sources` models which are needed for the app to show dataset geometries

met with GIS to confirm the details below

### columns
- level of detail: question flag X source dataset
  - e.g. `natural_resources_forever_wild`, `shadow_nat_resources_forever_wild`, `cats_permit`
- add new columns to the current final table because GFT app needs a single table (~60 columns)

### models
- level of detail: question flag X GFT geometry type
  - e.g. `natural_resources_poly`, `cats_points`, `cats_lots`
  - GFT geometry types are: point, line, lot, poly, buffer
- if geometry type is a buffer, use `ST_UNION` on all the rows
- if there are multiple rows with the same value, use `ST_UNION` to be a single row (e.g. MultiPoint)

damonmcc:
due to changes in #826:
- when a single dataset is used in two flags, we're applying the results of one flag to both

<img width="325" alt="Screenshot 2024-05-24 at 11 27 57 AM" src="https://github.com/NYCPlanning/data-engineering/assets/7444289/85f18d98-7e77-4b8b-a381-459f7b8b4ce9">

damonmcc:
after checkin with GIS and Amanda, source data columns aren't needed for public launch of the app

damonmcc:
data issues found by @jackrosacker after a build using changes from #871 

- [ ] Most (all?) point datasets have pts even after joining to lots
- [ ] Arterial hwy buffers are 90', should be 200'
- [ ] Some buffers are only buffering points, not polys as well
  - Title V buffer
  - State Facility buffer
  - Historic Res Adj buffer
- [ ] Industrial lots buffer is buffering vent towers
- [ ] Vent towers buffer is buffer industrial lots pts
- [ ] Missing sources__historic_resources_polys -> is this the same dataset as the shadows polys fc? The presence of the hist res pt fc suggests that the there should be a dedicated hist res poly fc
- [ ] Source__historic_resources name should have "points" suffix to be consistent -> holding off on re-sourcing this dataset until the name is defined
- [ ] All merged datasets should have variable_type column listing data source for each, in addition to the variable_id column (i.e. datasets like Nat Res, Hist Res, Open Space, etc.)
- [ ] Missing data
  - Shadows hist res
  - Shadows hist res buffer
[...]

PR #826, opened 2024-05-02
use GFT survey questions to define flags

### LINKED ISSUE #814: each spatial GFT survey question should have a single flag
## problem

Currently, we create a flag column for each dataset and some survey questions.

But GIS needs a single flag column and a single set of source geometry tables (e.g points, lots, buffers) for each spatial survey question.

The GFT tool will only show questions and their flag. Values which contributed to a flag will only been shown in the export table.

The exceptions to this 1-to-1 relationship are
- Project BBL: user enters this information, this isn't a flag we compute
- Project zoning: 3 questions about lots within a project, 1 column for a lot's zoning category

## approach

We're using the [GFT Data Review excel file](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/GIS/202401%20CEQR%20Type%20II%20Survey%20Build/docs/GFT%20Data%20review.xlsx?d=w20d343ed22ff4c14bda7257ebdb16fe3&csf=1&web=1&e=3SXp02) to document two things:
- all survey questions
- the mappings of survey question to dataset(s)

These tables (especially the mappings table) can inform any changes DE has to make to the GFT build.

We've had the [Data Source review excel file](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/GIS/202401%20CEQR%20Type%20II%20Survey%20Build/docs/CEQR_Type_II_Data_Source_Review.xlsx?d=wbae1537787a64188963736a89ac1df98&csf=1&web=1&e=3PZijT), but that table was used to track source data and doesn't represent the one-to-many relationship of survey question to source datasets

damonmcc:
resolves #814 

better to review each commit than all changes at once

## changes
- add seed `question_flags.csv` to declare the 1-to-1 relationship between the GFT survey questions and the GFT survey flags
  - lot zoning is the only exception because it isn't a binary flag, it's a category
- revise `variables.csv` to declare the many-to-many relationship between the survey flags and the source data
- change product models to produce a flag column and an ID column for each survey question
- change intermediate models to align with new product model columns
  - so that outputs are accurate (e.g. different buffers on the same dataset depending on the flag)
  - replaced `intermediate/buffers` with `intermediate/spatial` because they aren't all buffers and a blank slate was easier to work with

## notes
in the new `question_flags` seed:
- term "field name" is used because our output columns are FGDB fields that Jack will be choosing and aliasing
- each flag has a value (Yes/No) and there are IDs of geometries that determined the flag's value. so I ended up with `_flag` and `_flag_id`. maybe `_flag_ids` would be better though

## tests
- all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-gft-questions)
- successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9045778317) after changing product models
- successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9192926885) after changing intermediate models

## will do after this
- update `sources__` to align with new survey flags and reflect the latest requests from the GIS team (https://github.com/NYCPlanning/data-engineering/issues/816)
- review buffer values to make suer they're aligned with latest state of GIS excel sheet

### DAG focused on `int_flags__spatial` and `green_fast_track_bbls`
Screenshot 2024-05-22 at 12 07 42 PM

### a row from `test_actual__pilot_projects` showing new final columns
Screenshot 2024-05-11 at 5 09 09 PM


damonmcc:
while there are rows in `variables.csv` that aren't 1-to-1 with a survey question, [this test](https://github.com/NYCPlanning/data-engineering/actions/runs/8925270827/job/24513603847#step:6:105) will fail

products/green_fast_track/models/product/_product_models.yml
damonmcc, line 1: the changes to the `green_fast_track_bbls` columns are the core of this

products/green_fast_track/models/intermediate/flags/int_flags__all.sql
```diff
@@ -5,23 +5,43 @@
...
+        variables.flag_id_field_name,
+        all_variable_ids.variable_type,
+        variable_id,
+        distance
+    FROM all_variable_ids LEFT JOIN variables
+        ON all_variable_ids.variable_type = variables.variable_type
+)
+
+SELECT * FROM all_flags
+ORDER BY bbl ASC, flag_id_field_name ASC, variable_type ASC
```
damonmcc, line 47: `int_flags__all` now has a new column called `flag_id_field_name`

this column replaces the downstream use of `variable_type` because, when pivoting from a long to wide table, we now care about which flag (aka survery question) each `variable_id` like "Historic Lamppost" is related to rather than the variable (aka source data) it came from

products/green_fast_track/models/product/green_fast_track_bbls.sql
damonmcc, line 1: the diffs are probably hard to read, may be good to also look at the new file [here](https://github.com/NYCPlanning/data-engineering/blob/dm-gft-questions/products/green_fast_track/models/product/green_fast_track_bbls.sql)

.github/workflows/test_helper.yml
```diff
@@ -60,9 +60,8 @@ jobs:
 
           # this should probably be its own util/script
           export BUILD_ENGINE_DB=db-green-fast-track
-          cd products/green_fast_track
-          dbt seed
-          cd ../..
+          dbt deps --profiles-dir products/green_fast_track --project-dir products/green_fast_track
```
fvankrieken, line 63: Was it failing without this on your branch?

Also, I switched this from build to seed to avoid running seed tests while sqlfluffing. What functionality do you need from build?

products/green_fast_track/models/intermediate/flags/_flags_models.yml
```diff
@@ -8,13 +8,19 @@ models:
         data_type: string
         tests:
           - not_null
+      - name: flag_id_field_name
```
fvankrieken, line 11: Clarification: it seems like we're redefining what a flag is in this PR - instead of a flag being a unique combo of variable and lot, it maps 1-to-1 with survey questions.

Is Jack using the term "flag"? Or is he really thinking more just about survey questions, and "flag" is more our terminology?

If it's the former, this change makes sense. But if it's the latter, I'd sort of like to keep our definition of flag, since we have an identifier for the columns - "survey questions" (or answers to them, really). If we make flag have the same meaning, we don't have a term for unique combo of variable and bbl anymore.

fvankrieken: Or maybe this column should more just be "flag_type" or something like that since we do have 1 row per our original "flag" in this table still

fvankrieken (approved):
Couple small comments but good stuff

products/green_fast_track/models/intermediate/buffers/int_buffers__all.sql
```diff
@@ -24,7 +24,6 @@ WITH all_buffers AS (
         ref('int_buffers__us_parks_properties'),
         ref('int_buffers__waterfront_access_pow'),
         ref('int_buffers__waterfront_access_wpaa'),
-        ref('int_buffers__natural_resource_shadows'),
```
fvankrieken, line 27: Why is this being dropped?

products/green_fast_track/seeds/variables.csv
```diff
@@ -1,39 +1,55 @@
...
+historic_resources_adj,City Historic Buildings,nyc_historic_buildings,lpc_landmarks,TRUE,90,
+historic_resources_adj,State Historic Buildings,nys_historic_buildings,nysshpo_historic_buildings_points,TRUE,90,
+shadow_open_spaces,DPR Parks properties,nyc_parks_properties,dpr_parksproperties,FALSE,200,
+shadow_open_spaces,Schoolyard-to-playground,schoolyard_to_playgrounds,dpr_schoolyard_to_playgrounds,FALSE,200,
+shadow_open_spaces,POPS,pops,dcp_pops,TRUE,200,
+shadow_open_spaces,Waterfront Public Access Areas,waterfront_access_wpaa,dcp_waterfront_access_map_wpaa,FALSE,200,
+shadow_open_spaces,Publicly Owned Waterfront,waterfront_access_pow,dcp_waterfront_access_map_pow,FALSE,200,
+shadow_open_spaces,NYS Parks Property,nys_parks_properties,nysparks_parks_polygons,FALSE,200,
+shadow_open_spaces,Federal Parks property,us_parks_properties,usnps_parks,FALSE,200,
+shadow_nat_resources,NYC Parks Forever Wild,forever_wild,dcp_beaches,FALSE,200,
```
fvankrieken, line 44: I see what you mean here - but these buffer values (200) are never used, are they? So given that shadow_nat_resources has a different buffer than natural_resources, the current shadow_nat_resources column is wrong

fvankrieken: Since it isn't actually buffered

.github/workflows/test_helper.yml
fvankrieken: I think maybe I see why with the compile python reqs pr failing... is there only one test schema? I really didn't think about where this was actually running...

products/green_fast_track/seeds/variables.csv
fvankrieken: So these rows I think should just be dropped - as written, the join from the flags table to this creates duplicate rows independent of desired buffer distance. And for now, while we don't have variables that map explicitly to multiple questions, maybe keep the unique constraint on this table

.github/workflows/test_helper.yml
damonmcc: > Also, I switched this from build to seed to avoid running seed tests while sqlfluffing. What functionality do you need from build?

testing the seed tables seems worth it. since we have to build them so that `sqlfluff ... --templater=dbt` works, seems worth validating those tables

I'm sure exactly why the tables have to exist while sqlfluffing (maybe `dbt compile` is run?) so maybe all that maters is that any seed table columns used in queries exist and the contents don't matter. but testing them seems to do more good than harm

damonmcc: > is there only one test schema?

looks like there's only one for the job `check_formatting`, but the `dbt_checkpoint` uses a branch-unique schema. probably better to use the branch-unique schema for both

working on a fix in a different branch

damonmcc: @fvankrieken if https://github.com/NYCPlanning/data-engineering/pull/842 is all good, I should drop this change from this branch

products/green_fast_track/seeds/variables.csv
damonmcc: very true, this now generates incorrect `shadow_nat_resources` results

> we don't have variables that map explicitly to multiple questions

`nys_historic_buildings` and `nyc_historic_districts ` do. and these natural resources datasets do.

I was gonna say I'll restore `int_buffers__natural_resource_shadows` to it's former glory, but I think it'd be better to:
- keep these seed changes so that `variables` is unique on [`flag_id_field_name`, `variable_type`]
- let some outputs be incorrect for now
- implement the use of this new buffer column to fix those incorrect outputs

fvankrieken: I definitely think long-term this file and the way it's written make more sense, and it's a more elegant approach than the custom shadows buffer. But it's going to take some decent downstream changes to make it work: either generating multiple buffers from the same inputs programatically, which is something that we've talked about potentially being difficult in dbt, or have two staging tables for each of these datasets, or not actually have those buffers as their own models (which seems less than ideal).

Regardless, given that complexity, I don't think this should be merged to main without knowing how we're going to fix the issues it creates.

products/green_fast_track/models/intermediate/flags/_flags_models.yml
damonmcc: Jack is using the term "flag" to mean the binary answer to a single survey question

it was hard for me to name `flag_field_name` and `flag_id_field_name` in the new `question_flags.csv`:
- "field name" really seemed worth it because these are all for Jack to choose and alias columns
- each flag has a _value_ (Yes/No) and _IDs_ of geometries that resulted in the value, so I ended up with `flag` and `flag_id`. maybe `flag_ids` would be a little better though
- was tempted to lean more into `survey_question` or the `survey_question_id` values that Jack made for the app, but he cautioned that those IDs could change even if the essence of the question doesn't. so seemed good to just declare `flag_field_name` values that are 1-to-1 with survey questions

products/green_fast_track/seeds/variables.csv
damonmcc: happy to take a shot in this PR at generating flag-level intermediate models (programmatically or not) to align with this file

the end result would be:
- correct buffers and outputs
- use of this new approach
- would help with updating the `source__` export to meet Jack's needs for the map in the app: one layer per flag

products/green_fast_track/bash/export.sh
damonmcc, line 1: this is temporary. the follow-up changes to these models are so substantial that I chose to delete them all

products/green_fast_track/seeds/variables.csv
damonmcc: @fvankrieken the new commit most related to this is https://github.com/NYCPlanning/data-engineering/pull/826/commits/a3b9cae04d66778bc9e351733f91bbcb2e44a80c

products/green_fast_track/models/intermediate/flags/int_flags__all.sql
damonmcc, line 1: `int_flags__all` now has a new column `flag_id_field_name`

this column replaces the downstream use of `variable_type` because, when we pivot from a long table to a wide table, we now care about which flag (aka survey question) each variable_id like "Historic Lamppost" is related to rather than the variable (aka source data) it came from

products/green_fast_track/models/intermediate/flags/int_flags__spatial.sql
```diff
@@ -75,3 +75,4 @@ SELECT
         ELSE ST_DISTANCE(bbl_geom, raw_geom)
     END AS distance
 FROM joined_hexes
+ORDER BY bbl ASC, variable_type ASC, variable_id ASC
```
fvankrieken, line 78: I might keep this ordering in the aggregate in `green_fast_track_bbls` - postgres makes no guarantee of ordering of results returned by a query, and depending on how the compiler is feeling in terms of optimizing that aggregation, I think it makes sense to explicitly order where we want things ordered

fvankrieken:
Went commit-by-commit and looks good, would like a chance this afternoon to check out the codebase and poke around with how things are laid out now

products/green_fast_track/models/intermediate/flags/int_flags__spatial.sql
damonmcc: see [be14145](https://github.com/NYCPlanning/data-engineering/pull/826/commits/be14145e946325b8f42ac056b9a6553687cb3c46). outputs didn't change (a good thing) but I definitely like the extra confidence in the ordering

damonmcc:
@fvankrieken sounds good! hope the DAG screenshot is helpful and happy to walk through anything

I may start a dev branch off of this soon to get started on the changes to `source__` models

fvankrieken (approved):

PR #828, opened 2024-05-04
Compile python requirements

fvankrieken:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #829, opened 2024-05-06
bump forked sqlfluff to 3.0

fvankrieken:
Resolves issue in #828

Basically, it seems that sqlfluff-templater-dbt [used a deprecated function](https://github.com/sqlfluff/sqlfluff/blob/3.0.5/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py#L166). They only used it in the case that the dbt version was < 1.5, but [it was imported regardless](https://github.com/sqlfluff/sqlfluff/blob/3.0.5/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py#L148). In dbt-core 1.7.14, this [function was dropped](https://github.com/dbt-labs/dbt-core/compare/v1.7.13...v1.7.14#diff-a2ba8d46e3d69f8874c4c46f0d6a1d413f70879c405f82230e63d7c2bdb255beL2) from a submodules exports (why they specifically did this in a patch, I do not know). Luckily, on main sqlfluff templater dbt [only imports this function if using an old version of dbt](https://github.com/sqlfluff/sqlfluff/blob/main/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py#L165). So syncing my fork with sqlfluff@main and using the fork for the templater as well fixed our issue


fvankrieken:
Running nightly_qa job since this has recompiled python reqs

https://github.com/NYCPlanning/data-engineering/actions/runs/8974832482

.github/workflows/test_helper.yml
```diff
@@ -56,7 +56,12 @@ jobs:
           sqlfluff lint products/knownprojects/sql --templater=jinja
           sqlfluff lint products/pluto/pluto_build/sql/
           sqlfluff lint products/zoningtaxlots/sql/
+
+          # this should probably be its own util/script
           export BUILD_ENGINE_DB=db-green-fast-track
+          cd products/green_fast_track
+          dbt seed
```
alexrichey, line 63: Does sqlfluff need the seed data in tables to validate/fix?

alexrichey (approved):
💪

PR #830, opened 2024-05-07
Fix nightly qa workflow

fvankrieken:
cleanup from #829. Fixing [failed workflow](https://github.com/NYCPlanning/data-engineering/actions/runs/8979959749) due to invalid template on non-workflow dispatch trigger

Also, turns out I needed to clean up how boolean inputs to `workflow_call` were being handles - turns out they aren't handled very well. Kept getting "misformatted yml" when I tried to specify `input: false` becuase "false" there is a string not a bool... even if I use `${{ input.value == 'true'}}`. So for workflow calls made them strings, and logic will work around that.

Despite all these canceled/failed actions, seems that all is working.

[Nightly QA via push trigger (build via workflow call](https://github.com/NYCPlanning/data-engineering/actions/runs/8986624237) - canceled, but it worked as expected

[Running with workflow dispatch to make sure it attempts to use dev images when specified](https://github.com/NYCPlanning/data-engineering/actions/runs/8986627814) are properly used (fails because dev image doesn't exist for this branch)

[Build with workflow dispatch - don't use dev image](https://github.com/NYCPlanning/data-engineering/actions/runs/8986711315/job/24683508803). Canceled because images are successfully grabbed

[Build with workflow dispatch - use dev image](https://github.com/NYCPlanning/data-engineering/actions/runs/8986761872) fails because they don't exist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,10 +72,11 @@ on:
         type: string
         required: true
       dev_image:
-        type: boolean
+        type: string
         default: false
+        required: false
       repeat:
-        type: boolean
+        type: string
         default: false
       build_name:
         type: string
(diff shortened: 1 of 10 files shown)


fvankrieken:
So to elaborate, part of the issue is that when I pass things to workflow calls, they're getting handled as strings. Workflow dispatch inputs DO get handled as bools... but NOT default values. So our build job having both workflow dispatch and workflow call triggers leads to some ugly long bool statements (input == 'true' || input == true) - the second is needed as written because the string 'false' is truthy. But a boolean value of `true` would be false for `true == 'true'`

alexrichey (approved):

PR #832, opened 2024-05-08
Ar facilities automation

alexrichey:
This PR does a lot, basically everything required for me to have confidence pushing facilities to Socrata. it's probably easiest to review commit by commit.

- Moves code around to accommodate our new folder structure for lifecycle and models 
- Renames Product Metadata to align with our new terminology around `products` and `dataset files`
- Adds Shapefile validation
- Adds rowcount validation between different dataset files (e.g. csvs and shapefiles)
- Adds attachment validation (just checking that they exist)
- Adds Typer CLI to the very top level of `dcpy` so that we get nice hierarchical layout (and we can remove the spurious "don't invoke me" commands)
image
- Adds validation unit test with an actual shapefile. 
- Refactors connectors.edm.package to align to new conventions, including our new consensus around where package files should live, as well as the function of edm connector modules.

### Validation
Here's what you'll see, using template_db as an example:
```
['templatedb_points.shp.zip',
 'COLUMN_MISMATCH',
 "Invalid column(s) found in source data: {'geometry', 'wkb_geomet', "
 "'geometry_e'}."]
['templatedb_points.shp.zip',
 'COLUMN_MISMATCH',
 "Column(s) missing from source data: {'the_geom'}."]
['templatedb_points.shp.zip',
 'INVALID_DATA',
 'Column bbl contains 218 invalid record(s), for example: None']
['templatedb_polygons.shp.zip',
 'COLUMN_MISMATCH',
 "Invalid column(s) found in source data: {'geometry', 'wkb_geomet', "
 "'geometry_e'}."]
['templatedb_polygons.shp.zip',
 'COLUMN_MISMATCH',
 "Column(s) missing from source data: {'the_geom'}."]
['templatedb_polygons.shp.zip',
 'INVALID_DATA',
 'Column bbl contains 2372 invalid record(s), for example: None']
```


dcpy/lifecycle/distribute/socrata.py
```diff
@@ -78,25 +78,16 @@ def _dist_from_local(
             raise Exception("Only shapefiles have been implemented so far")
 
 
-@app.command("from_s3")
+@socrata_app.command("from_s3")
 def _dist_from_s3(
-    product_name: str = typer.Option(
```
alexrichey, line 83: Something that seems obvious in retrospect: required args shouldn't be typer.Options. It looks like we `should` be using [annotations](https://typer.tiangolo.com/tutorial/arguments/help/#add-a-help-text-for-a-cli-argument) for these. TODO for later.

fvankrieken: So their docs have

`name: Annotated[str, typer.Argument(help="The name of the user to greet")]`

but I think you can also just do

`name: str = typer.Argument(help="The name of the user to greet")`

dcpy/lifecycle/distribute/socrata.py
```diff
@@ -48,55 +48,52 @@ def _dist_from_local(
...
-    dataset_destination_id: str = typer.Option(
+    product_name: str,
+    version: str,
+    dataset_destination_id: str,
+    dataset: str = typer.Option(
         None,
-        "-d",
-        "--dest",
-        help="Dataset Destination Id",
+        "-d",  # d is already taken, unfortunately
```
damonmcc, line 94: but this still uses `-d`?

dcpy/lifecycle/package/validate.py
```diff
@@ -0,0 +1,302 @@
...
+                    e.dataset_file.filename if e.dataset_file else None,
+                    e.error_type.value,
+                    e.message,
+                ]
+            )
+
+
+def _is_valid_wkb(g):
+    if not g:
+        return True
```
damonmcc, line 67: seems like this should `return False` right?

dcpy/lifecycle/package/validate.py
```diff
@@ -0,0 +1,302 @@
...
+
+def _is_int(s):
+    if s[0] in ("-", "+"):
+        return s[1:].isdigit()
+    return s.isdigit()
+
+
+def _is_geom_point(s):
+    try:
+        return wkt.loads(s).geom_type == "Point"
```
damonmcc, line 91: (thought for future) we'll likely see datasets with Point/MultiPoint values in a single column, but happy to wait-and-see and address that by extending tests later

damonmcc:
awesome! left comments on a couple small things

dcpy/lifecycle/package/validate.py
```diff
@@ -8,13 +9,20 @@
...
+class ErrorType:
     INVALID_DATA = "INVALID_DATA"
     NULLS_FOUND = "NULLS_FOUND"
     COLUMM_MISMATCH = "COLUMN_MISMATCH"
     INVALID_METADATA = "INVALID_METADATA"
 
 
+@dataclass
+class ValidationError:
+    error_type: ErrorType
```
fvankrieken, line 25: Blanking on how simple classes like this work in python - should `ErrorType` be an enum? Or will this work as written

fvankrieken: Is logic for `Enum` classes more around conversion to their types? Like having a built-in `.value` function

dcpy/connectors/edm/packaging.py
```diff
@@ -10,56 +7,29 @@
...
-DOWNLOAD_ROOT_PATH = Path(".publishing")
+DOWNLOAD_ROOT_PATH = Path(".package")
 OUTPUT_ROOT_PATH = Path(".output")
 
 
 @dataclass
 class PackageKey(publishing.ProductKey):
     product: str
     version: str
+    dataset: str
```
fvankrieken, line 22: Isn't this more a PackageDatasetKey then? Just that PackageKey sounds like it is a unique key for a package, not a dataset within it

dcpy/lifecycle/distribute/socrata.py
alexrichey: ha, that was a tongue-in-cheek comment that become unnecessary. Will fix ✂️

dcpy/lifecycle/package/validate.py
alexrichey: So in validation, I'd separated out the null checks from the type-checks. This is really just meant to remove non-null invalids, so they're not fed into wkb.loads

dcpy/lifecycle/package/validate.py
alexrichey: Apologies, this was sloppy commiting. ErrorType is enum'd in a later commit

dcpy/connectors/edm/packaging.py
alexrichey: That's a good point. We really have two types of Package: The `ProductPackage` and the `DatasetPackage`. Will change that to make it clear here.

damonmcc (approved):

Issue #831, opened 2024-05-07
PLUTO 24v2


fvankrieken:
### Product Name

pluto

### Build Version

24v2

### Status of Update

- [x] source data extracted
- [x] draft build succeeded
- [x] draft build passed QA
- [x] data packaged and distributed

fvankrieken:
## Data loading

### Manual Updates

#### Updated 2x a year typically in June and December
- [ ] **dcp_colp** (check [here](https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-colp.page)) 
#### Currently updated with each new release of a major version of PLUTO - important to make sure this is up to date
- [ ] **pluto_corrections** (pulling from bytes, must update when there's updates to **pluto_input_research**)

### Automated Updates

#### Open data automated pull

> Check [here](https://github.com/NYCPlanning/db-data-library/actions/workflows/open-data.yml) to see the latest run

- [ ] **dsny_frequencies**
- [ ] **dpr_greenthumb**
- [ ] **lpc_historic_districts**
- [ ] **lpc_landmarks**

#### DOF Automated Pull and Number of Buildings

- [ ] **pluto_pts** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_pts.yml))
- [ ] **pluto_input_geocodes** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_pts.yml))
- [ ] **pluto_input_cama_dof** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_cama.yml))
- [ ] **pluto_input_numbldgs** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_numbldgs.yml))

### Updated with Quarterly updates (check [here](https://github.com/NYCPlanning/db-data-library/actions/workflows/quaterly-updates.yml))

- [ ] **dcp_cdboundaries_wi**
- [ ] **dcp_ct2010**
- [ ] **dcp_cb2010**
- [ ] **dcp_ct2020**
- [ ] **dcp_cb2020**
- [ ] **dcp_school_districts**  
- [ ] **dcp_councildistricts_wi**  
- [ ] **dcp_firecompanies**  
- [ ] **dcp_policeprecincts**
- [ ] **dcp_healthareas**  
- [ ] **dcp_healthcenters**

### Updated with Zoning Taxlots 

[...]

PR #835, opened 2024-05-08
PLUTO QA enhancements, expand `dcpy` repeat build functionality

### LINKED ISSUE #538: PLUTO QA enhancements
### GIS requests after QA of PLUTO 24v1
- Label value units where possible in graph tooltips and axes (% vs sq ft vs record count, etc.)
- Table of BBLs where bldgarea/lotarea > numfloors*2 aka “buildings containing unreasonably small apartments”
  - Hard to manually QA when there are ~400 records
  - One improvement would be if the report could identify new records added to this list based on the current build, hopefully it would be less daunting of a task if it was only 5-10 records each cycle.
  - Either filter to exclude or flag NYCHA BBLs

### LINKED ISSUE #657: PLUTO: Incorporate manual research
During the 24v1 creation records were identified where should be modified based on research:
1) 5008640515 building area value changed from 250 to 28500
Research top offenders from Unreasonably small apartments table and make modifications

fvankrieken:
Does NOT actually close #657, 

Closes #538

Part of #831 

Ended being a bunch of little stuff in here, so will go commit by commit

### QAQC Enhancements (#538) See issue for screenshots
1. Move logic to a component. Mainly wanted as part of commit 2 but this makes diffs in commit 2 easier to see
2. Add selectbox that lets you choose whether the other "diff" being shown is the previous, or previous of the same build type (major vs minor). The current build is a good example. We built 24v2. The prev is 24v1.1, so our "current" diff is "24v2 - 24v1.1". The former default, "previous previous" diff would be "24v1.1 - 24v1". But if we want to compare this to the last "major" diff instead, we can now do that, where the comparison diff will be "24v1 - 23v3.1" instead.
    There are cases where these would be the same - if we didn't have a minor last build and current was "24v2 - 24v1", either option would be "24v1 - 23v3.1". Same if we have 24v2.2 - both will be "24v2.1 - 24v2"
    This works for minors as well. 24v1.1 (with diff "24v1.1 - 24v1") can either look at diff "24v1 - 23v3.1" or "23v3.1 - 23v3"
    This commit includes a lot of renaming - instead of just "v1" "v2" and "v3" thought it's best to describe. So "version"/"v", "previous", "comparison", and "comparison_previous" seemed intuitive to me. "previous" always describing the directly preceding build, and "comparison" describing the build whose diff we want to compare to.
3. Adding new fields to outlier reports so that GIS can filter them out of the tables. So adding an additional column to the QA outputs as well as a checkbox in the app to filter rows. This checkbox is grayed out (with a useful tooltip) if the build doesn't have the new columns
4. add a few comments and axes labels as requested

### Non-correction or QA changes
5. don't vacuum in pluto build. This took a little while, and was added to attempt to reclaim space in constrained github runner temp postgres instances. Our db seems to have aggressive autovacuuming set up, so this seems unneeded
6. we've moved away from using .envs in this manner - I personally would rather run `set_env` manually before a build. Maybe worth broader discussion
7. Fix a bug introduced in #801 (or rather not entirely fixed - the fix was needed in 3 lines, I only added it on 2)
8. Rework "repeat" build functionality. I wanted to be able to repeat a draft, rather than only published versions, since once GIS begins QAing a dataset, it seems like we should be locking versions of all input datasets as to not introduce potentially new errors. This was tough with the way I initially implemented this. I decoupled normal recipe-planning and this repeat logic since they require fairly different inputs, and we can't have dynamic inputs to github actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,10 +34,6 @@ on:
         description: "Version of dataset being built (overrides version defined in recipe)"
         type: string
         required: false
-      repeat:
-        description: "Repeat existing build?"
-        type: boolean
-        default: false
       dev_image: 
         description: "Use dev image specific to this branch (if exists)"
         type: boolean
@@ -75,20 +71,24 @@ on:
         type: string
         default: false
         required: false
-      repeat:
-        type: string
-        default: false
       build_name:
         type: string
         required: false
+
+
+env:
+  tag: ${{ (inputs.dev_image == 'true' || inputs.dev_image == true) && format('dev-{0}', github.head_ref || github.ref_name) || 'latest' }}
+  build_name: ${{ inputs.build_name || github.head_ref || github.ref_name }}
+  version: ${{ github.event.inputs.version && format('--version {0}', github.event.inputs.version) || '' }}
 
 jobs:
   health_check:
     name: Health Check
     runs-on: ubuntu-latest
     outputs:
-      tag: ${{ (inputs.dev_image == 'true' || inputs.dev_image == true) && format('dev-{0}', github.head_ref || github.ref_name) || 'latest' }}
-      build_name: ${{ inputs.build_name || github.head_ref || github.ref_name }}
+      plan_command: recipe --recipe-path ./${{ inputs.recipe_file }}.yml ${{ env.version }}
+      tag: ${{ env.tag }}
+      build_name: ${{ env.build_name }}
     steps:
       - name: Check inputs
         run: |
@@ -104,8 +104,7 @@ jobs:
       image_tag: ${{ needs.health_check.outputs.tag }}
       build_name: ${{ needs.health_check.outputs.build_name }}
       recipe_file: ${{ inputs.recipe_file }}
-      version: ${{ inputs.version }}
-      repeat: ${{ inputs.repeat }}
+      plan_command: ${{ needs.health_check.outputs.plan_command }}
       logging_level: ${{ inputs.logging_level }}
       build_note: ${{ inputs.build_note }}
   cbbr:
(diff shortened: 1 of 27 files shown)


apps/qa/src/pages/pluto/components/outlier_report.py
```diff
@@ -1,46 +1,85 @@
...
             st.write(f"There are {df.shape[0]} outliers in total.")
-            st.info(self.info_dict[field])
+            if filter:
+                df = filter(df)
 
-            with st.expander("See table"):
+            with st.expander("Show table"):
                 AgGrid(df)
 
+    def building_area_increase(self):
```
fvankrieken, line 44: Made these their own functions because of need for custom logic for filtering datasets

.github/workflows/build.yml
```diff
@@ -75,20 +71,24 @@ on:
...
+
 jobs:
   health_check:
     name: Health Check
     runs-on: ubuntu-latest
     outputs:
-      tag: ${{ (inputs.dev_image == 'true' || inputs.dev_image == true) && format('dev-{0}', github.head_ref || github.ref_name) || 'latest' }}
-      build_name: ${{ inputs.build_name || github.head_ref || github.ref_name }}
+      plan_command: recipe --recipe-path ./${{ inputs.recipe_file }}.yml ${{ env.version }}
+      tag: ${{ env.tag }}
```
fvankrieken, line 90: env vars cannot be passed to workflow_calls (they're just not accessible at that part of the workflow), so they need to be here. But I thought it's a bit cleaner to have them defined up higher just to make them a bit clearer.

However, the "plan_command" had to be defined here because it depends on those vars

fvankrieken:
Passing regular build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9098664578)

Passing repeat build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9098664578)

dcpy/lifecycle/builds/plan.py
```diff
@@ -248,7 +270,47 @@ def _cli_wrapper_plan_recipe(
...
-    plan(recipe_path, version, repeat)
+    plan(recipe_path, version)
+
+
+@app.command("repeat")
+def _cli_wrapper_repeat_recipe(
+    product: str = typer.Argument(help="Name of the product to build"),
+    product_type: str = typer.Option(
+        None, "--product-type", "-t", help="Product/build type ('draft' or 'publish')"
+    ),
```
damonmcc, line 281: (just a thought, not worth holding this up for)

this seems more like `version_type` since it's used to pick a type of `ProductKey`

damonmcc (approved):
LGTM! I like the `plan_command` concept

🤞🏾 for nightly QA after this is merged

dcpy/lifecycle/builds/plan.py
fvankrieken: Yeah this was a tough one to name. Literally, more "product instance type" but that doesn't roll off the tongue or keyboard

fvankrieken:
> LGTM! I like the `plan_command` concept
> 
> 🤞🏾 for nightly QA after this is merged

Nightly QA is passing already!

fvankrieken:
Just rebased main, will make sure tests pass before merging

PR #841, opened 2024-05-13
Add Socrata distribution README

alexrichey:
A quick README for distributing to Socrata. @fvankrieken @sf-dcp @damonmcc let me know what information would be useful in order to get started automating socrata distributions.

Also, if you have any suggestions for formatting, etc. I like the idea of having README's at relevant levels in dcpy, then perhaps we can link them together from a top-level README

### DIFF
--- a/dcpy/lifecycle/distribute/README.md
+++ b/dcpy/lifecycle/distribute/README.md
@@ -70,6 +70,6 @@ INFO:dcpy:Finished syncing product to Socrata, but did not publish. Find revisio
 INFO:dcpy:            here https://data.cityofnewyork.us/d/b7pm-uzu7/revisions/32
 ```
 
-[ TODO: Image Here ]
 Follow the provided link. Here you can review the modified data and metadata. Hit `Update` in the top right to apply the revision. 
+![template_db_socrata](https://github.com/NYCPlanning/data-engineering/assets/11164730/b0c24251-00e3-4be1-99a6-6cf015240cc6)
 
--- a/dcpy/lifecycle/distribute/README.md
+++ b/dcpy/lifecycle/distribute/README.md
@@ -1,10 +1,12 @@
 # Distribute
 
 ## Terms
-- `Product`: a suite of datasets. E.g. LION, which contains multiple `datasets`, such as "2010 Census Blocks" or "Community Districts (water areas Included)"
-- `Dataset` : roughly analogous to what you'd find on a Socrata page. Effectively it's metadata, attachments, and dataset files. 
-- `Dataset File`: actual file for a dataset, e.g. a csv, or a shapefile. Note: there may be slight variations between `dataset files` for the same `dataset`. e.g. columns in the shapefile for PLUTO will have slightly different columns than the csv. 
+- `Dataset`: Abstractly, a single table of data and metadata about it.
+- `Product`: A suite of datasets. E.g. LION, which contains multiple `datasets`, such as "2010 Census Blocks" or "Community Districts (water areas Included)"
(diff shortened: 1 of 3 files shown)


fvankrieken:
Do you have an example revision that you've pushed, but not published? Just to see more practically how to interact with it in socrata

alexrichey:
> Do you have an example revision that you've pushed, but not published? Just to see more practically how to interact with it in socrata

@fvankrieken Let me add an example of that whole flow

dcpy/lifecycle/distribute/README.md
```diff
@@ -0,0 +1,57 @@
+# Distribute
+
+## Terms
```
damonmcc, line 3: it'd be helpful to add `Dataset Package` to the list of terms, maybe last

damonmcc (approved):
lgtm! just had a suggestion about the Terms section

dcpy/lifecycle/distribute/README.md
alexrichey: @damonmcc for sure. I'm curious your thoughts regarding `Dataset` vs `Dataset Package`. It sort of feels like Dataset is an abstract concept or standard, whereas the Package is the concrete file instantiation. Is that too heady? 😂

damonmcc: agreed that they feel distinct. to me, Dataset is a concept and Package is a set of files

a Package could have multiple Datasets in it right? so a Product has one or more Packages. each Package has one or more Datasets

alexrichey: Good question... I think there are probably two package concepts, the `product package` which would contain potentially multiple `dataset package`s. So for something like PLUTO you'd have a folder for e.g. 24v2 (the `product package`) which would contain a `dataset package` for the main files, and another `dataset package` for the changefile dataset

damonmcc: that makes sense. now I feel like there are two qualities of a Package (either product or dataset):
- a collection of files
- everything a distribution endpoint needs for an update

sf-dcp: If we leaning towards using "dataset package", I agree on adding Dataset Package to Terms... For example, reading the red portion below makes me check the Terms section:

image

Otherwise, I'm personally indifferent about dataset vs dataset package as long as it's explicitly defined.

dcpy/lifecycle/distribute/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Note that unless you explicitly specify `--publish`, only a draft will be created on Socrata, and you'll need to manually apply it via the socrata GUI. The output of the command will tell you the revision number, which you can find via search on Socrata. There you can review data changes, metadata changes, etc, before actually publishing. 
+
+##### Quick Tip
+If package validation fails, you can validate locally by running the validation commands locally:
+```
+python -m dcpy.cli lifecycle package validate [your package path here]
+```
+
+#### Pushing from S3 via Github Action:
```
sf-dcp, line 46: tiiiiiny rec to enumerate steps :)

dcpy/lifecycle/distribute/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+So the first step is to assemble the `Dataset Package`.
+
+#### Dataset Package Structure
+The package must be structured like so:
+- `/attachments/`: folder containing all attachments (except metadata.yml)
+- `/dataset_files/`: folder containing data files for the dataset, e.g. the shapefile, the csv, etc.
+- `metadata.yml`: describes all columns, destinations, expected attachments, and expected dataset files. It also drives all the package-level validation.
+
+#### Pushing from S3 via CLI:
```
sf-dcp, line 22: tiiiiiny rec to enumerate steps :)

sf-dcp (approved):
Love the Readme writeup! Have minor styling recs :)

PR #846, opened 2024-05-14
Add new historical footprints GR test to app

### LINKED ISSUE #812: new GRU QAQC report
Geographic research would like to add a new QAQC report to the portfolio of existing [GR QAQC reports](https://de-qaqc.nycplanningdigital.com/?page=GRU+QAQC).

This new report would compare Historic Footprint BIN vs PAD BIN. This is essentially a version of the existing BIN vs PAD BIN, but using historical data.

Geographic Research would like to be able to run this new report after 24B is released on May 5 to clean up the data ahead of the 24C release.

[This](https://data.cityofnewyork.us/Housing-Development/Building-Footprints-Historical-Shape/s5zg-yzea) is the best place for DE to get historical footprint data (already in edm-recipes)

fvankrieken:
Closes #812 
Follow up to https://github.com/NYCPlanning/db-gru-qaqc/pull/267 - Output (from new check in linked PR) has already been confirmed as what's needed (for now at least) by VB. This gets the new check in the dashboard

Check added to dashboard (both screenshots running app locally)

image

And successfully triggered

image

### DIFF
--- a/.github/workflows/data_library_open_data.yml
+++ b/.github/workflows/data_library_open_data.yml
@@ -22,56 +22,15 @@ jobs:
     strategy:
       matrix:
         dataset:
-          - council_members
-          - dca_operatingbusinesses
-          - dcla_culturalinstitutions
-          - dcp_edeseignation_csv
-          - dep_cats_permits
-          - dfta_contracts
-          - doe_busroutesgarages
-          - doe_eszones
-          - doe_hszones
-          - doe_mszones
-          - dohmh_daycare
-          - doit_evictions
-          - doitt_buildingcentroids
           - doitt_buildingfootprints
           - doitt_buildingfootprints_historical
-          - dot_pedplazas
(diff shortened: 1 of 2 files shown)


.github/workflows/data_library_open_data.yml
```diff
@@ -22,13 +22,57 @@ jobs:
     strategy:
```
fvankrieken, line 22: This is all opendata datasets. Does this seem worthwhile? Could wait. But I figure we might as well be pulling socrata data more regularly.

Main counterpoint might be to just wait until we roll out ingest - then we'll have a better record of when data is being updated

fvankrieken: Other argument to wait - we maybe add functionality to check for updates, and only update those that have actually changed.

damonmcc (approved):

PR #851, opened 2024-05-16
update Housing DB version in EDDE

damonmcc:


fvankrieken (changes requested):
Could be a bit cleaner

fvankrieken (approved):

PR #853, opened 2024-05-17
Ingest: expand data formats in `to_parquet` fn

sf-dcp:
Related to #631.

### Major changes in this PR:
1. Add `json`, `geojson`, and `excel` data formats to ingest module. 
2. Abstract code converting pandas dataframe to a geodataframe into a separate function. We convert df to gdf when reading in non-geospatial data formats (csv, excel, json) that do have geospatial data. For example, a csv that has a point geometry column can't be read directly into geopandas (geopandas only likes geospatial data formats); instead, it's read into pandas and then into geopandas.
3. Create moooooore tests! Most of the changed files in this PR are regenerated test data.

### A note about `json` implementation:
* When compared to other data formats like `csv` and `shapefile`, it's a a bit trickier reading in a json file into a pandas dataframe because: 
    * Same data can be represented in many different schemas. Here is an example of equivalent datasets: 
    ```python
    # each object is a column
    json_str = {
        "col 1": {
            "row 1": "a",
            "row 2": "c"
        },
        "col 2": {
            "row 1": "b",
            "row 2": "d"
        }
    }

    # each object is a row (more common representation of data)
    json_str = [
        {
            "col 1": "a",
            "col 2": "b"
        },
        {
            "col 1": "c",
            "col 2": "d"
        }
    ]
    ```
    *  Nested objects like this:
    ```python
    json_str = [
            {
                "boro_code": 4,
                "location": {
                    "bbl": 4469310598
                },
                "details": {
                    "text": "GsifrlkxmckyxrKHjGsr",
                    "wkt": null
                }
            },
            {
                "boro_code": null,
                "location": {
                    "bbl": 5192630318
                },
                "details": {
                    "text": null,
                    "wkt": "POINT (10.3894635 -175.008089)"
                }
            }
        ]
    ```

* To address these challenges we can read in a json doc using 2 `pandas` methods: `read_json()` and `json_normalize()`:
    * `read_json()` is a more generic fn that reads in json data. You can specify in the signature whether the underlying data is represented through rows or columns. This fn doesn't explode nested records. 
    *  `json_normalize()` helps unpack nested objects when reading data into a dataframe and it expects data to be in records format, i.e. each record is a row (see the more common representation above). 
* After we read json data into dataframe, we save in a `parquet` format. Fun fact is that `parquet` doesn't accept nested values and `pyarrow` engine will complain. And given that our generalized approach may not catch some nested values (say we didn't specify nested columns in a recipe), I convert nested values into validated json strings prior to saving to parquet. Then further cleaning can take place either in preprocessing steps of ingest or in a data pipeline itself.


dcpy/models/file.py
```diff
@@ -56,12 +56,32 @@ class Geodatabase(BaseModel, extra="forbid"):
     crs: str
 
 
-# TODO: implement JSON and GEOJSON
 class Json(BaseModel):
     type: Literal["json"]
-    normalize: dict = {}
+    json_read_meta: JsonMeta
```
sf-dcp, line 61: @fvankrieken (and others), would love to hear your thoughts about `json_read_meta` implementation, specifically about `JsonNormalize` and `ReadJson` classes

dcpy/lifecycle/ingest/templates/dpr_capitalprojects.yml
```diff
@@ -5,20 +5,18 @@ source:
...
-    - Location
-    meta:
-    - TrackerID
-    - FMSID
-    - Title
-    - TotalFunding
+  json_read_meta:
+    json_read_fn: normalize
+    json_read_kwargs:
+      {
```
fvankrieken, line 11: I was going to say this could just be yml, but there's something nice about making it look a little more like actual function args. Definitely at least like one-lining the arrays, undecided on the brackets

dcpy/models/file.py
fvankrieken: Since the two types have the same args and are required, I would lean with having the subfields (read_fn and read_kwargs) just be attributes of the Json class here. (with `json_read_fn`: Literal["read_json", "json_normalize"]`)

dcpy/utils/data.py
```diff
@@ -81,56 +83,39 @@ def read_data_to_df(
...
+                    json_str = json.load(f)
+                df = pd.json_normalize(
+                    json_str, **data_format.json_read_meta.json_read_kwargs
+                )
+
             # Feels a tad hacky to have this here instead of in preprocessing
             # But json columns were causing parquet issues
             if data_format.columns:
-                gdf = gdf[data_format.columns]
+                df = df[data_format.columns]
```
fvankrieken, line 113: If we're coercing any remaining json fields to strings, I'd say we can drop this and let preprocessing take care of it.

Though I'm not seeing where you're doing that?

fvankrieken:
A couple comments, but this looks great

fvankrieken:
Also just one important note on use case for json_normalize - when `record_path` is provided, it means that an array must be found at that path, from which to normalize(/explode). So something like

```json
[
    {
        "agency": "nypl",
        "locations": [
            { "location": "POINT (10.3894635 -175.008089)" },
            { "location": "POINT (10.3894635 -175.008089)" }
        ]
    },
    {
        "agency": "bpl",
        "locations": [
            { "location": "POINT (10.3894635 -175.008089)" },
            { "location": "POINT (10.3894635 -175.008089)" }
        ]
    }
]
```
Would be expanded to four rows with `record_path=["locations", "location"]`

dcpy/models/file.py
```diff
@@ -37,7 +37,6 @@ class Xlsx(BaseModel, extra="forbid"):
     type: Literal["xlsx"]
     unzipped_filename: str | None = None
     tab_name: str
```
fvankrieken, line 39: Can we call this sheet_name?

dcpy/models/file.py
```diff
@@ -56,12 +55,32 @@ class Geodatabase(BaseModel, extra="forbid"):
     crs: str
 
 
-# TODO: implement JSON and GEOJSON
 class Json(BaseModel):
     type: Literal["json"]
-    normalize: dict = {}
+    json_read_meta: JsonMeta
     columns: list[str] = []
```
fvankrieken, line 61: And per my comment in `data.py`, maybe we just scrap this attribute entirely

dcpy/lifecycle/ingest/templates/bpl_libraries.yml
```diff
@@ -6,14 +6,13 @@ source:
...
+  json_read_kwargs: { "record_path": ["locations"] }
 processing_steps:
-- name: clean_column_names
-  args:
-    replace: 
-      "data.": ""
+  - name: clean_column_names
+    args:
+      replace:
+        "data.": ""
```
fvankrieken, line 15: Are you using a yml linter/formatter? And if so which one 😬

dcpy/models/file.py
```diff
@@ -56,12 +55,19 @@ class Geodatabase(BaseModel, extra="forbid"):
     crs: str
 
 
-# TODO: implement JSON and GEOJSON
 class Json(BaseModel):
```
fvankrieken, line 58: Also, we should make sure these all have `extra="forbid"` (my bad - I added this one)

dcpy/utils/data.py
fvankrieken: nypl_libraries is the dataset that needs this, so you can use that to test.

Also, an fyi - I changed the cli slightly. So it'll be
- `python3 -m dcpy.lifecycle.ingest.run nypl_libraries -s` on your branch right now
- `python3 -m dcpy.cli lifecycle ingest nypl_libraries -s` if you rebase main into your branch

sf-dcp: Successfully extracted and converted to parquet:

image

dcpy/lifecycle/ingest/templates/bpl_libraries.yml
sf-dcp: I use this vscode yaml extension: 

image


cc @damonmcc

fvankrieken (approved):

PR #857, opened 2024-05-20
Reproject ingested datasets

fvankrieken:
Ended up doing a handful of things, don't be put off by the thousand lines, there's a bunch of generated stuff in one commit. Definitely one to go by commit. This will be squashed

### Minor stuff
1. Actually archive datasets at the end of `ingest.run`! I added a helper since this is very similar to archiving a raw dataset Both are tested.
2. add pyarrow-stubs for mypy, recompile requirements
3. remove an outdated cli target
4. put extracted zip files during ingest in the same folder as the file being unzipped
5. use zip file logic for dcp_pad

### geo stuff
6. BIG models. I used [datamodel-codegen](https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/) to turn the json schemas for geoparquet and projjson (used within geoparquet) to create pydantic models, then edited them a bunch. Seemed nice to be able to read CRS from a geoparquet file without opening it, but is also overkill. I'm happy to not fully check this in, but this seems like a nice thing to base our own geospatial stuff around. Open to thoughts
    - I think a good question here is - do we see value in having this? Should I just go ahead and pare this down? It is nice to have the geoparquet metadata read in structured, but it might be way overkill
    - One good sized tweak I made from the codegen output was going from [RootModel](https://docs.pydantic.dev/latest/concepts/models/#rootmodel-and-custom-root-types) -> TypeAlias. RootModel seems to be Pydantic's preferred route of handling these sorts of union types, but it gets clunky when nested (so in some cases you need `obj.id`, others you need `obj.root.id`. Could maybe get around it with methods for the class but it was annoying)
7. Check whether a parquet is geo or not, and load it as a df or geodf!
8. actually reproject dataframe
9. alter geom definition for bpl_libraries to work - theirs isn't actually wkt (which would be "Point(-74.0 40.0)" but instead just x, y ("-74.0, 40.0"). I've added this as a possible geom format.
    This also moves around a bunch of our geospatial code - moves a few things into models, consolidates to one "df to gdf" function, and makes sure new functionality is tested.
10. Code affected by 9 outside of dcpy
11. Clean up duplicate postgres utils for reading in table as df as well as add a read_gdf method - templatedb was reading in table as df and then converting, and old method was covering up errors.
12. remove a bunch of print statements


dcpy/models/geospatial/projection.py
```diff
@@ -450,16 +448,15 @@ class TemporalCrs(BaseModel, extra="forbid"):
...
-    root: (
-        GeodeticReferenceFrame
-        | VerticalReferenceFrame
-        | DynamicGeodeticReferenceFrame
-        | DynamicVerticalReferenceFrame
-        | TemporalDatum
-        | ParametricDatum
-        | EngineeringDatum
-    )
+Datum: TypeAlias = (
```
fvankrieken, line 451: RootModel seems to be Pydantic's preferred way of dealing with this sort of thing, but it was a bit annoying to work with

dcpy/models/geospatial/geometry.py
```diff
@@ -0,0 +1,39 @@
...
+from pydantic import BaseModel, Field
+
+
+class StandardGeometryFormat(str, Enum):
+    wkt = "wkt"
+    wkb = "wkb"
+
+
+class PointXYStr(BaseModel, extra="forbid"):
+    point_xy_str: str  # = Field(pattern=r"^(?=[^x]*x[^x]*)(?=[^y]*y[^y])$")
```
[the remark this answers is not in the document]
fvankrieken: This is resolved - went with a validator function

dcpy/models/geospatial/geometry.py
```diff
@@ -0,0 +1,39 @@
...
+        if re_match:
+            return f"Point({re_match['y']} {re_match['x']})"
+        else:
+            raise ValueError(
+                f"Geom field '{s}' does not match format '{self.point_xy_str}"
+            )
+
+
+class PointXYRegex(BaseModel):
+    point_xy_regex: str  # = Field(pattern=r"^(?=.*\(\?P\<x\>\\d\+\\\.\\d\+\))(?=.*\(\?P\<y\>\\d\+\\\.\\d\+\)).*$")
```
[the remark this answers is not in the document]
fvankrieken: I will say that writing this regex felt like my greatest technical achievement while at DCP - a regex that matches a string that is a valid regex pattern with two named capture groups x and y, where the capture groups are floats

dcpy/models/lifecycle/ingest.py
```diff
@@ -53,11 +53,14 @@ class Template(BaseModel, extra="forbid", arbitrary_types_allowed=True):
     name: str
     acl: recipes.ValidAclValues
 
+    target_crs: int | None = None
```
fvankrieken, line 56: Bigger picture issue here - we should sit down and think about if we'd like to restructure these template/config objects at all. Wanted to keep this fairly atomic for now and not shuffle things. But that'll be an important step before we wrap up an MVP of the new ingest code

dcpy/lifecycle/ingest/transform.py
```diff
@@ -264,4 +264,5 @@ def preprocess(
     compiled_steps = validate_processing_steps(config.name, config.processing_steps)
     for step in compiled_steps:
         df = step(df)
+    print(df.columns)
```
fvankrieken: This is removed in final commit

sf-dcp:
holding off my review until the tests are passing

fvankrieken:
> holding off my review until the tests are passing

Only test failing is the one missing credentials that's failing on main, this is good to go

dcpy/connectors/edm/recipes.py
```diff
@@ -28,29 +28,27 @@
...
-    ### this could either take a config object and path to a file, and archive them both in a folder together,
-    ### or take path to folder that takes dumped config as well as file output and loads whole folder
-    ### likely the latter in case of multiple output formats. Not necessarily something we want to be doing long term
-    ### but for flexibility for now it might make sense
-    raise NotImplemented
+def archive_raw_dataset(config: ingest.Config, file_path: Path):
+    _archive_dataset(config, file_path, config.raw_dataset_key.s3_path(RAW_FOLDER))
+
+
+def archive_dataset(config: ingest.Config, file_path: Path):
```
sf-dcp, line 50: Could you please explain the need for `_archive_dataset()`, `archive_raw_dataset()`, and `archive_dataset()`? Seems somewhat redundant...

dcpy/lifecycle/ingest/run.py
```diff
@@ -35,32 +35,21 @@ def run(
...
+    if not skip_archival:
+        recipes.archive_dataset(config, staging_dir / config.filename)
 
 
 app = typer.Typer(add_completion=False)
 
 
-@app.command("run")
+@app.command()
 def _cli_wrapper_run(
```
sf-dcp, line 52: Since we changed the structure from where the CLI command is triggered, perhaps just move the code below into `dcpy/lifecycle/_cli.py` for simplicity?

python/constraints.txt
```diff
@@ -456,6 +456,8 @@ pyarrow==16.1.0
     # via
     #   -r python/requirements.in
     #   streamlit
+pyarrow-stubs==10.0.1.7
```
sf-dcp, line 459: perhaps I will answer my question later in the review... if not, curious where do we use this stub?

dcpy/connectors/edm/recipes.py
fvankrieken: Yeah that's fair. When I started, I wrote them as separate functions before realizin how much code was shared.

I think to some extent, it's nice that when archiving from `ingest` code, you don't have to worry about things like s3 prefixes, and just have atomic operations that make sense in the context of the "run". At the beginning, simply a command to archive the raw dataset, at the end, a command to archive the processed dataset, just based on the config and file.

Main difference in functionality is that raw datasets go `raw_datasets/{timestamp}/{raw_filename}` whereas processed are `datasets/{version}/{dataset_name}.parquet`.

Could definitely do it with a boolean value instead (`raw: bool = False`), or just pass the s3 prefix. I do prefer not the latter personally, just a tad.

python/constraints.txt
fvankrieken: When reading parquet metadata

dcpy/lifecycle/ingest/run.py
fvankrieken: Maybe. This is following Alex's structure for his (to some extent - I obviously tweaked it a bit since there's only one ingest command for now). But I think the idea is that cli file really just has the links to the sub-apps/commands and not logic specific to those commands.

Definitely something that's open for discussion, but I'd like to keep the same convention for now

fvankrieken: I think I might lean with what you're suggesting though. Warrants a broader discussion!

alexrichey: Yeah, could see either way. Open to chat, but maybe we just let things evolve, and make a decision when we know better what we need

dcpy/utils/mapping.py
alexrichey, line 1: Maybe this warrants a `utils/geo/` folder? I think I'll probably forever mistake this for functional utils around sequences

dcpy/lifecycle/ingest/transform.py
```diff
@@ -248,18 +247,21 @@ def validate_processing_steps(
...
-    if len(compiled_steps) == 0:
-        shutil.copy(input_path, output_path)
-    else:
-        df = pd.read_parquet(input_path)
-        for step in compiled_steps:
-            df = step(df)
-        df.to_parquet(output_path)
+    """Reprojects if needed, then validates and runs preprocessing steps defined in config object"""
+    df = data.read_parquet(input_path)
+    if config.target_crs:
```
alexrichey, line 257: Should this just be a processing step? (ie should `validate_processing_steps` just tack this on at the beginning of the `processing_steps`?)

fvankrieken: Yeah maybe. It definitely shouldn't be in the templates, but it probably makes sense to use the same machinery to run it

dcpy/models/geospatial/parquet.py
alexrichey, line 1: Super cool that they're autogenerated. Curious what part wasn't, though (kinda spitballing on whether it'd be possible just have pure generated artifacts that we can just monkey patch at runtime)

alexrichey (approved):
Lots of great stuff! Didn't delve too deep into the details - I'm sure it's all in order. 

On bigger-picture thought: I hadn't seen the logging stuff in recipes before. Wonder if it makes sense to generalize that into something like `lifecycle events`, and convert the insert/read code into a connector?

dcpy/models/geospatial/parquet.py
fvankrieken: There are some random things that didn't work entirely - the "mutually exclusive id or ids" rule seemed to make sense in the jsonschema, but the class originally just had "pass" as its body (and then `Id`, `Ids` were fields of classes which inherited this class). So there was no actual validation that these two fields were mutually exclusive.

I also swapped out `RootModel`s for `TypeAlias`es. `RootModel` seems to be pydantic's preferred way of having a union class, but as written it felt a little clunky to use.

fvankrieken: If you're interested, could regenerate it to show the diffs (probably should have just done that in the first place, with separate commits)

products/template/build_scripts/export.py
```diff
@@ -30,35 +32,30 @@ def export():
...
     # export builds tables
     for table_name, file_types in BUILD_TABLES.items():
-        data = PG_CLIENT.get_table(table_name)
         file_path = OUTPUT_DIR / table_name
         for file_type in file_types:
             logger.info(
                 f"Exporting table\n\t{table_name}\n\tas a {file_type} to\n\t{OUTPUT_DIR}"
             )
             if file_type == "csv":
+                data = PG_CLIENT.read_table_df(table_name)
```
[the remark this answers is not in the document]
fvankrieken: Probably!

dcpy/models/geospatial/parquet.py
alexrichey: no worries about regenerating! can cross that bridge if regenerating after modifications becomes a PITA

dcpy/connectors/edm/recipes.py
sf-dcp: I see, thanks for the description. I don't have a strong a preference for either having the 2 separate functions for `ingest` archiving or 1 function with an input param, so I defer to you. It would be helpful to add function docstring(s) regardless of the choice.

dcpy/lifecycle/ingest/run.py
sf-dcp: Makes sense why you changed it, @fvankrieken . Agree, would be nice to have the same convention across the board.

dcpy/test/utils/test_data.py
```diff
@@ -72,3 +76,11 @@ def test_serialize_nested_objects():
         for value in serialized_df[col]:
             assert isinstance(value, str)
             assert value.startswith("{")
+
+
+def test_read_parquet():
```
sf-dcp, line 81: 👀👍

sf-dcp:
I have some thoughts on the #6 commit. Geoparquet is a comparatively new data format, and I think it’s fair to say it may undergo future iterations, including its metadata schema. If true, it adds overhead for us to maintain the pydentic classes to align with the schema. Like Alex mentioned, perhaps we could do it on the fly using something like [this](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.FileMetaData.html#pyarrow.parquet.FileMetaData)?

fvankrieken:
> I have some thoughts on the #6 commit. Geoparquet is a comparatively new data format, and I think it’s fair to say it may undergo future iterations, including its metadata schema. If true, it adds overhead for us to maintain the pydentic classes to align with the schema. Like Alex mentioned, perhaps we could do it on the fly using something like [this](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.FileMetaData.html#pyarrow.parquet.FileMetaData)?

I think that makes some sense. Most of this generated code is not from geoparquet specification though, it's from [projjson](https://proj.org/en/9.4/specifications/projjson.html) (which is a subfield of the geoparquet metadata), which is a bit more mature. Geoparquet is pretty minimal - just the `Columns` class and `GeoParquet` class in commit 6.

And bizarre seeing you link that `FileMetaData` class, because I was about to say "well that's what I used"... but I don't see it in my code anywhere? For some reason I have that field as "Any" in `models/.../parquet.py`. That's supposed to be this exact object. See [this line](https://github.com/NYCPlanning/data-engineering/pull/857/commits/5ddfa2341c4b0d0bff8f37265824d72662eaa8a6#diff-d0757076c3a5af3b1f2db1e2302da41d93957bb7cf30f07ef75468671bd99fb1R163), that returns a `FileMetaData` that gets fed into the `parquet.MetaData` object that I've defined. It doesn't contain geospatial info though - that just is part of an unstructured dict (from python's perspective). We could do a lighter-weight pydantic model that ignores extra args, and just gets the "crs" of `Columns` rather than trying to get the whole column

fvankrieken:
Though to my point that projjson is more "mature" - it is, though its json schema itself is only v0.7, meaning they might be refining that still. Might still be best to do something a little more lightweight/dynamic

dcpy/models/lifecycle/ingest.py
sf-dcp: Agree. Let's do it next week. 

I do have a question about this `target_crs` attribute. In the Template, it's of int data type; in Config, it's a string. Why is that? 

Note, in the `file.Format`, crs is of a string type like `EPSG:4326`.

dcpy/utils/geospatial.py
```diff
@@ -166,3 +166,18 @@ def read_parquet_metadata(filepath: Path) -> geoparquet.MetaData:
     geo = parquet_metadata.metadata[geoparquet.GEOPARQUET_METADATA_KEY]
     geo_parquet = geoparquet.GeoParquet(**json.loads(geo))
     return geoparquet.MetaData(file_metadata=parquet_metadata, geo_parquet=geo_parquet)
+
+
+def reproject_gdf(
+    df: gpd.GeoDataFrame, target_crs: str, source_crs: str | None = None
+) -> gpd.GeoDataFrame:
+    if not df.crs:
```
sf-dcp: Not sure if this is needed here. You can't read data into a gpd data frame if it's not geospatial

fvankrieken: You can certainly have geometry data without having a crs set. Not geoparquet, so we'll be safe when reading from that, but in general [gdf.crs can return None](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.crs.html)

dcpy/models/file.py
```diff
@@ -2,8 +2,10 @@
 from pydantic import BaseModel
 from typing import Literal, TypeAlias
 
+from dcpy.models.geospatial import geometry
 
-class Geometry(BaseModel):
+
+class Geometry(BaseModel, extra="forbid"):
```
sf-dcp, line 8: not urgent at all, but curious what you think. Should we move this class to `models/geospatial/geometry.py`? Seems like it belongs there along with `GeometryFormat`

dcpy/models/geospatial/geometry.py
```diff
@@ -0,0 +1,59 @@
...
+    Or a column with value "-74.0, 40.7" would be "x y"
+
+    The wkt function of this class assumes that any string to be parsed has x and y
+    each represented by a decimal (digits, followed by period, followed by digits)
+    For now, this does not need to be more flexible, but this could be another optional input
+    """
+
+    point_xy_str: str
+
+    def wkt(self, s: str) -> str:
```
sf-dcp, line 33: Nice addition to the geospatial functionality! Using regex for parsing out a geometry feels a bit dangerous, however. I think it should require some validation from the user side. At a minimum. Perhaps printing a few records asking to match would help.

products/template/build_scripts/export.py
```diff
@@ -40,11 +42,14 @@ def export():
                 data.to_csv(file_path.with_suffix(".csv"), index=False)
             elif "shapefile" in file_type:
                 shapefile_directory = OUTPUT_DIR
-                geodata = geospatial.convert_to_geodata(
+                geom = Geometry(
```
sf-dcp: niceee

dcpy/utils/postgres.py
```diff
@@ -118,6 +119,22 @@ def read_sql_table(
         else:
             return pd.read_sql_table(table_name=table_name, con=conn, **kwargs)
 
+    def read_table_gdf(
```
sf-dcp, line 122: niceee

dcpy/utils/geospatial.py
sf-dcp: oh so it's probably similar to crs not being set on a geometry column in Postgres in which case it would return zero. Thank you for clarifying

sf-dcp:
> Though to my point that projjson is more "mature" - it is, though its json schema itself is only v0.7, meaning they might be refining that still. Might still be best to do something a little more lightweight/dynamic

Could you please explain why you leaning towards using a class when reading in parquet metadata? Maybe I'm missing the point. 

Let me explain my thought process. We only want a relatively small subset of information from parquet metadata, so we don't need to know about other objects. And, when parquet metadata schema changes in the future, we will get an error querying dict keys we need regardless whether we use a class or the raw data representation. I think using a pydantic class (or any other validation class) is more applicable for data that we control, like our recipe templates. And in a situation with parquet, it seems to only add overhead for us.

fvankrieken:
> Could you please explain why you leaning towards using a class when reading in parquet metadata? Maybe I'm missing the point.
> 
> Let me explain my thought process. We only want a relatively small subset of information from parquet metadata, so we don't need to know about other objects. And, when parquet metadata schema changes in the future, we will get an error querying dict keys we need regardless whether we use a class or the raw data representation. I think using a pydantic class (or any other validation class) is more applicable for data that we control, like our recipe templates. And in a situation with parquet, it seems to only add overhead for us.

Doesn't need to be a class I guess, mainly I just like type safety. If we read in GeoParquet metadata, we should be able to read `crs` or `epsg_id` or something like that and know it's present and a string (or a more complex object), and throw an error if we can't find it, since having this information is part of the GeoParquet spec, and we rely on it for reading in geoparquet files to geopandas. Pydantic offers nice utilities around reading in dictionaries and having some sort of error handling when it doesn't match the structure you expect, and we use it all over the code base, hence why I used it here. The codegen tool I used also specifically creates pydantic objects from jsonschema.

If we start working more with different types of CRSs, it's nice to have these as type-safe proper python objects, and even better if they align with accepted standards for these things like projjson (which is already a standard being used by geoparquet). I agree that this might not be useful or practical right now in terms of the overhead it adds (in terms of the huge projjson file, the complete `Columns` object within the geoparquet metadata)

dcpy/models/geospatial/geometry.py
fvankrieken: Can you elaborate on "feels a bit dangerous"?

dcpy/models/file.py
fvankrieken: Yeah I was wondering a bit about this. It is fairly specific to tabular geometry - a set of columns denoting a definition of geometry. So still maybe, but I might want to call if something slightly more specific that just `Geometry`. But I don't quite know what that name is (a classic tale)

fvankrieken: But it certainly belongs more over there than here

dcpy/models/geospatial/geometry.py
sf-dcp: Sure, I referred to the possibility of reading in incorrect geospatial data successfully. Say the column of interest is not of a real geospatial data format and may have errors (perhaps it was entered manually). And the worst case where it won't throw a failure, but we end up with wrong data. idk, maybe my fear is invalid.

fvankrieken: Yeah I get that. But we still have to read in data somehow. I hope that this makes things a little simple twofold
- specifying that the creator of the template file has to choose a specific string format "point_xy_str" rather than provide their own regex, so that there's less room for weirdness and validating the input. In this case, the string provided can only have one "x" and one "y" and needs both. If source data comes in like "70.0, 40.0", I don't think we're going to get much safer in terms of how we want to tell the template how to convert it into a point. It also (currently) won't support multiple formats, so errors will get thrown if things are inconsistent. Which brings me to my second hope of keeping things simple/not dangerous
- This at least will throw errors when the column doesn't match the provided format, rather than try to handle them in some way. We won't catch anything like flipflopped lat and long, but we also don't catch that currently in cases where we have separate lat long columns (like that one facdb dataset...). For this stuff, I hope we can start catching more weirdness with data expectations - whether all rows with geom are in NYC, stuff like that.

fvankrieken:
> > Could you please explain why you leaning towards using a class when reading in parquet metadata? Maybe I'm missing the point.
> > Let me explain my thought process. We only want a relatively small subset of information from parquet metadata, so we don't need to know about other objects. And, when parquet metadata schema changes in the future, we will get an error querying dict keys we need regardless whether we use a class or the raw data representation. I think using a pydantic class (or any other validation class) is more applicable for data that we control, like our recipe templates. And in a situation with parquet, it seems to only add overhead for us.
> 
> Doesn't need to be a class I guess, mainly I just like type safety. If we read in GeoParquet metadata, we should be able to read `crs` or `epsg_id` or something like that and know it's present and a string (or a more complex object), and throw an error if we can't find it, since having this information is part of the GeoParquet spec, and we rely on it for reading in geoparquet files to geopandas. Pydantic offers nice utilities around reading in dictionaries and having some sort of error handling when it doesn't match the structure you expect, and we use it all over the code base, hence why I used it here. The codegen tool I used also specifically creates pydantic objects from jsonschema.
> 
> If we start working more with different types of CRSs, it's nice to have these as type-safe proper python objects, and even better if they align with accepted standards for these things like projjson (which is already a standard being used by geoparquet). I agree that this might not be useful or practical right now in terms of the overhead it adds (in terms of the huge projjson file, the complete `Columns` object within the geoparquet metadata)

One cool use case - I think crs's, datums etc typically have associated bounding boxes in these objects. That would make it really easy to add a check to see if all geoms in the dataset are in its (supposed) crs's limits, maybe giving us a really simple (if not exhaustive) way to catch incorrect projections, or at least entirely invalid ones. Won't catch things like wrongly assuming a wgs84 dataset is state plain (since state plain's bounding box is much bigger), but at least it would catch things in the other direction. Maybe a little less useful than I thought when I started writing this comment

sf-dcp:
Regarding the geospatial check you described: we will be doing validation step on an entire dataset as a part of `ingest` and it's probably best to keep all checks in one "place" using one utility/framework rather than referring to metadata info for validation. 

Thanks for sharing your perspective on using pydentic for metadata. I still think it's more than what we need, at least as of today. I'm also okay with it if you strongly believe it's a better approach. 

A side rec to break PRs into smaller ones to make review faster. Otherwise, good stuff

sf-dcp (approved):

dcpy/models/geospatial/parquet.py
```diff
@@ -0,0 +1,65 @@
+from pyarrow.parquet import FileMetaData
```
fvankrieken, line 1: @sf-dcp - I cleaned this up and dropped the complicated projjson, and I think this shows a good example of how I'd like to use this. See [this line](https://github.com/NYCPlanning/data-engineering/pull/857/files#diff-6de4d995b5b185d7999972afa1df26cb154ba629c4c6c5defaa7216a25eda219R137) in tests. It's nice to load a dict into an object where then I can quickly use computed properties that would be tougher to parse using dicts, and just quickly get the string "EPSG:4326" from the geoparquet metadata.

dcpy/lifecycle/ingest/transform.py
fvankrieken: See [here](https://github.com/NYCPlanning/data-engineering/pull/857/commits/9cb04d05b1c5c69b1b7d782a3228f540de405e5b). Basically, when converting from template to config, if target_crs is specified this is added as the first processing step. Could be argument to put it last. Anyways, this makes things a little cleaner imo - `transform` doesn't need to treat this as a special case now

dcpy/utils/mapping.py
fvankrieken: Addressed

sf-dcp:
> Added a couple commits to address comments!

LGTM!

PR #867, opened 2024-05-26
[old] fix GFT use of single dataset in multiple flags

damonmcc:
## problem

due to changes in https://github.com/NYCPlanning/data-engineering/pull/826, when a single dataset is used in multiple flags, we're applying the results of one flag to all of them

### screenshot of `int_flags_all` grouped by `flag_id_field_name`

shows identical counts for different flags despite their use of different geometries

Screenshot 2024-05-24 at 11 27 57 AM

## changes

- add tests to primary output `green_fast_track_bbls` to ensure the results vary for different flags' use of `nyc_historic_buildings`
- add a lot to the pilot projects as `Buffer Test` to ensure results vary for different flags' use of `nyc_historic_buildings`
- add and use seed column `variables.flag_variable_type` to be a unique id for a dataset's use in a flag
  - so that flags can be correctly computed, rather than only using `variables.variable_type`

### example of new seed column `variables.flag_variable_type`
| flag_variable_type | flag_id_field_name | variable_type |
|--------|--------|--------|
| hr_nyc_historic_buildings | historic_resources | nyc_historic_buildings |
| hra_nyc_historic_buildings | historic_resources_adj | nyc_historic_buildings |
| shr_nyc_historic_buildings | shadow_hist_resources | nyc_historic_buildings | 

## tests

A test failing (via warning) [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9246003972/job/25433065807#step:8:728) showing that the number of lots flagged for `historic_resources` is the same as those flagged for `historic_resources_adj`. this should not be true because `historic_resources_adj` covers a larger area by definition

Pilot project test [failing before](https://github.com/NYCPlanning/data-engineering/actions/runs/9246273299/job/25433626109#step:8:728) fix and [passing after] fix.

### screenshot of `Buffer Test` pilot project lot and nearby geometries related to `nyc_historic_buildings`

shows a lot that only touches buffers related to `Shadows Historic Resources`. but due to the bug, this lot has a `Yes` value for other flags which use a common source dataset (`nyc_historic_buildings`)

Screenshot 2024-05-26 at 5 07 38 PM


damonmcc:
closing in favor of approach in #868

PR #868, opened 2024-05-27
fix GFT use of single dataset in multiple flags (alt)

damonmcc:


products/green_fast_track/models/intermediate/flags/int_flags__all.sql
damonmcc, line 1: this is where setting `flag_id_field_name` makes a difference. before, the `LEFT JOIN` here was causing the bug because `variables.variable_type` isn't unique


now, instead of joining on `variables` to get `flag_id_field_name`, this model get it's from the unioned flag models and passes it to `green_fast_track_bbls` to be used [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/green_fast_track/models/product/green_fast_track_bbls.sql#L46-L64)

fvankrieken (approved):

PR #873, opened 2024-05-28
Get latest gis dataset based on folder name

fvankrieken:
baby pr - we shouldn't rely on what's in staging folder ideally

This logic is currently only used in ingest

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -385,12 +385,19 @@ def get_latest_gis_dataset_version(dataset_name: str) -> str:
     Get latest version of GIS-published dataset in edm-publishing/datasets
     assuming versions are sortable
     """
-    gis_version_formats = [r"\d{2}[A-Z]", r"\d{8}"]
-    subfolders = [
-        f
-        for f in s3.get_subfolders(BUCKET, f"datasets/{dataset_name}/")
-        if any([re.match(p, f) for p in gis_version_formats])
-    ]
+    gis_version_formats = [r"^\d{2}[A-Z]$", r"^\d{8}$"]
+    subfolders = []
+    matched_formats = set()
+    for f in s3.get_subfolders(BUCKET, f"datasets/{dataset_name}"):
+        for p in gis_version_formats:
+            if re.match(p, f):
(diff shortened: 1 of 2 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -380,21 +381,31 @@ def _assert_gis_dataset_exists(name: str, version: str):
...
-    ## by that logic, should likely try to parse str version to dcpy version object
-    ## so we don't have strange cases of an unexpected subfolder (that starts with an _)
-    ## this would let us not have to supply versions for geosupport datasets ("24a", etc)
-    metadata = s3.get_metadata(BUCKET, _gis_dataset_path(dataset_name, "staging"))
-    version = metadata.last_modified.strftime("%Y%m%d")
+    """
+    Get latest version of GIS-published dataset in edm-publishing/datasets
+    assuming versions are sortable
+    """
+    gis_version_formats = [r"^\d{2}[A-Z]$", r"^\d{8}$"]
```
alexrichey, line 388: Should we move this over to dcpy.utils.versions?

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -24,7 +24,7 @@ def add_gis_datasets(create_buckets):
     test_objects = [
         f"datasets/{TEST_GIS_DATASET}/staging/{TEST_GIS_DATASET}.zip",
         f"datasets/{TEST_GIS_DATASET}/{today}/{TEST_GIS_DATASET}.zip",
-        f"datasets/{TEST_GIS_DATASET}/{TEST_VERSION.upper()}/{TEST_GIS_DATASET}.zip",
+        f"datasets/{TEST_GIS_DATASET}/20240101/{TEST_GIS_DATASET}.zip",
```
alexrichey, line 27: nit: let's pull this out into a constant

alexrichey (approved):
Two optional nits, but looks good!

dcpy/connectors/edm/publishing.py
fvankrieken: Maybe. We need to make that more consistent and really implement it across things, so I'd rather just leave for now

PR #876, opened 2024-05-30
fix gft exports

damonmcc:
related to #816, resolves bugs listed in [this comment](https://github.com/NYCPlanning/data-engineering/issues/816#issuecomment-2138415074) on that issue

## tests

latest build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9308941083), all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-fix-gft-exports)

reviewed outputs with Jack and confirmed all issues are fixed

### QGIS screenshots with latest FGDB

corrected Shadow Historic Resources exports

Screenshot 2024-05-30 at 11 15 13 AM

corrected Vent Towers and Industrial Lots

Screenshot 2024-05-30 at 2 18 17 PM


products/green_fast_track/models/intermediate/spatial/shadows/int_spatial__shadow_hist_resources.sql
```diff
@@ -1,13 +1,12 @@
-WITH all_natural_resources AS (
+WITH all_shadow_resources AS (
 {{ dbt_utils.union_relations(
     relations=[
-        ref('stg__lpc_landmarks'),
-        ref('stg__nysshpo_historic_buildings'),
+        ref('int_spatial__historic_resources'),
```
fvankrieken, line 4: This is I think why I'm a little less crazy about the name of `int_spatial__historic_resources` - lpc_scenic_landmarks seems like a historic resource to me. 

Maybe `historic_lots`? `historic_landmarks_and_buildings`? Just something a little more specific would be helpful, until we're at this dataset, which really is all historic resources.

Lemme know if you disagree obvs, I'm coming in with less context

products/green_fast_track/models/product/source/air_quality/source__cats_permit_points.sql
```diff
@@ -3,3 +3,4 @@ SELECT
     variable_id,
     raw_geom
 FROM {{ ref("int_spatial__cats_permits") }}
+WHERE lot_geom IS NULL
```
fvankrieken, line 6: This seems like a consequence of your separating raw_geom and lot_geom - before, when raw_geom has mixed geometry, only points are exported when specified as points.

This might be a bit more legible though.

damonmcc: correct, when raw_geom had multiple types, this selection would be via `WHERE ST_GEOMETRYTYPE(raw_geom) = 'ST_MultiPoint'`

but the separate columns change was before these exports and made the actual build logic more legible/traceable too ([`int_spatial__all`](https://github.com/NYCPlanning/data-engineering/blob/dm-fix-gft-exports/products/green_fast_track/models/intermediate/spatial/int_spatial__all.sql#L42))

damonmcc: using geometry types (in my experience) is more prone to errors and headaches than using separate columns for these distinct entities (points, lots, polys, buffers)

products/green_fast_track/models/intermediate/spatial/shadows/int_spatial__shadow_hist_resources.sql
damonmcc: (per chat in-person) EARD/Planning Support named the questions and decided which sources go into each question, unfortunately Shadow Historic Resources _isn't_ just "buffer Historic Resources"

separate from these changes, we have a discrete task to review alignment of the build with the intentions of EARD/Planning Support. if it turns out that Shadow Historic Resources is just "buffer Historic Resources", this idea makes more sense

that review should center on our [`variables.csv`](https://github.com/NYCPlanning/data-engineering/blob/main/products/green_fast_track/seeds/variables.csv) because the [Source Data excel file](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BBAE15377-87A6-4188-9637-36A89AC1DF98%7D&file=CEQR_Type_II_Data_Source_Review.xlsx&action=default&mobileredirect=true) isn't really as clear about "source data in each question"

fvankrieken (approved):

PR #877, opened 2024-05-30
GFT: Filter CATS permits that start with `PA` or `PB` permit id

### LINKED ISSUE #872: change GFT CATS permits
Filter CATS permits to only include PA/PB records
- involves querying the "applicationid" column to read the first two characters of the alphanumeric id, and exclude anything that doesn't start with PA or PB

sf-dcp:
Fixes #872.

I also moved filtering by status from staging to intermediate model as it makes more sense to have it there. 

Build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9308602666/job/25622348258)


damonmcc (approved):

sf-dcp:
Before filtering, the # of records with undesired permits is non-zero:
image

After filtering, the # of records in the intermediate table is zero as expected:

image

Issue #664, opened 2024-03-08
add remaining Green Fast Track variables


damonmcc:
see meeting notes [here](https://github.com/NYCPlanning/edm-overview/issues/1012#issuecomment-1967427682) and source data spreadsheet [here](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BBAE15377-87A6-4188-9637-36A89AC1DF98%7D&file=CEQR_Type_II_Data_Source_Review.xlsx&action=default&mobileredirect=true)

### extract new source data

- [x] #674
- [x] #672
- [x] #673
- [x] Exposed Rail Yards
- [x] NYCDOB Tidal Wetland
- [x] NYCDOB Freshwater Wetland
- [x] NYCDOB Coastal Erosion Hazard Area

### adding logic for new variables
- [x] #732 (added by @sf-dcp on 4/3/24)
- [x] Natural Resources
- [x] Historic Resources
- [x] Exposed Rail Yards
- [x] NYCDOB Tidal Wetland
- [x] NYCDOB Freshwater Wetland
- [x] NYCDOB Coastal Erosion Hazard Area

### exporting new source data
- [x] export all new source data in FGDB

---

### GFT dataset details
- The primary GFT table has a row for each PLUTO lot and columns to represent each "GFT variable". These variables are grouped into "CEQR categories".
- For each variable, there's a single binary flag column and at least one value column.
- When a variable flag is "Yes", the variable value columns show the source data values that contribute to the flag.
- Each GFT variable has one source dataset.
- `Category -> Variables -> Source`

### more GFT dataset details
- We have a CSV to describe all GFT variables [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/green_fast_track/seeds/variables.csv).
  - It maps each variable to a category.
  - Each variable has at least source.
  - `Category -> Variables`
- Other than combing through sql code, we don't yet have the `Variable -> Sources` mappings documented (similar to the [source data excel file](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BBAE15377-87A6-4188-9637-36A89AC1DF98%7D&file=CEQR_Type_II_Data_Source_Review.xlsx&action=default&mobileredirect=true)).

### Potential approach to confirming existing variables
[...]

PR #878, opened 2024-05-30
Add (hopefully) final gft sources

fvankrieken:
#664


products/green_fast_track/seeds/variables.csv
damonmcc, line 1: I think this is missing
- 3 new rows for `shadow_natural_resources` for the new DOB data
- 2 new rows for `shadow_natural_resources` for `rec` and `snwa`

products/green_fast_track/models/intermediate/spatial/shadows/int_spatial__shadow_nat_resources.sql
```diff
@@ -1,15 +1,8 @@
...
-        ref('stg__nysdec_tidal_wetlands'),
-        ref('stg__nysdec_priority_lakes'),
-        ref('stg__nysdec_priority_estuaries'),
-        ref('stg__nysdec_priority_streams'),
-        ref('stg__dcp_beaches'),
-        ref('stg__nysdec_natural_heritage_communities'),
-        ref('stg__usfws_nyc_wetlands'),
-        ref('stg__dpr_forever_wild'),
+        ref('int_spatial__natural_resources'),
+        ref('int_flags__dob_natural_resources'),
```
damonmcc, line 5: I think this should only be `int_spatial__natural_resources` since `int_flags__dob_natural_resources` should be going into that right?

fvankrieken: It is not - see comment here https://github.com/NYCPlanning/data-engineering/pull/878/files/0c8c7a0a949ac627251f4e3ee9b1458df5425781#diff-c227eefe4b635026c8543a11f3cde0b647b9957f322567a2911c3850997268c9R7

fvankrieken: Basically, I didn't add it to int_spatial__natural_resources because it doesn't need a spatial join to determine flags since they're provided tabularly

damonmcc: that makes sense. then `int_flags__dob_natural_resources` shouldn't be in this model too then right?

it's used in `int_flags__all`

fvankrieken: It should be in this model - Jack and Matt were both explicit that they want the lots buffered as part of the shadow dataset. 

So 
- plain natural resources - just join bbl
- shadows - join buffered bbl lot geom

products/green_fast_track/models/intermediate/flags/int_flags__dob_natural_resources.sql
```diff
@@ -0,0 +1,57 @@
+-- this file should maybe be broken up, as it's doing two things
+-- 1. supplying flags that go straight into int_flags__all
+-- 2. supplying geoms that go into natural_resource shadows
+-- 3. supplying geoms that go into natural_resource source outputs
```
damonmcc, line 4: since in a comment below you said this model doesn't feed into `int_spatial__natural_resources` because it doesn't need a spatial join to determine flags since they're provided tabularly, who do the geoms need to go into Shadows Natural Resources?

products/green_fast_track/models/intermediate/spatial/shadows/int_spatial__shadow_nat_resources.sql
damonmcc: ahhh ok thank you. so either a project includes on of these lots or it's "shadow adjacent" to one

products/green_fast_track/seeds/variables.csv
damonmcc, line 1: (thought for later)

to (my) confusion around tabular vs spatial use of the new DOB source data, this seed table could maybe better represent that distinction

before this new data, everything related to `natural_resources` was spatial

products/green_fast_track/models/intermediate/spatial/shadows/int_spatial__shadow_nat_resources.sql
fvankrieken: Precisely

products/green_fast_track/seeds/variables.csv
fvankrieken: Yeah we definitely need some sort of slight adjustment to make it a bit cleaner. It is just kind of a complex web though

damonmcc (approved):
not sure if u wanted to add more but that build passed so LGTM! thanks for tackling this

PR #879, opened 2024-05-31
fix GFT Shadows Historic Resources

damonmcc:
per GIS comment [here](https://github.com/NYCPlanning/data-engineering/issues/816#issuecomment-2140920024)

> I reviewed the `variables.csv` and here are my additional comments besides **removing Scenic Landmarks from Historic Shadows**

Screenshot showing new source raw geometry types contributing to Shadow Natural Resources below (all points due to removal of Scenic Landmarks)

Screenshot 2024-05-31 at 9 20 57 AM

### DIFF
--- a/products/green_fast_track/analyses/pluto_zoning_district_groups.sql
+++ b/products/green_fast_track/analyses/pluto_zoning_district_groups.sql
@@ -15,7 +15,7 @@ group_by_rollup AS (
         zonedist4,
         count(*) AS zoning_combo_count
     FROM pluto
-    GROUP BY rollup(zonedist1, zonedist2, zonedist3, zonedist4) -- noqa: PRS
+    GROUP BY ROLLUP(zonedist1, zonedist2, zonedist3, zonedist4) -- noqa: PRS, CP03
 )
 
 SELECT
--- a/products/green_fast_track/analyses/pluto_zoning_district_groups.sql
+++ b/products/green_fast_track/analyses/pluto_zoning_district_groups.sql
@@ -15,7 +15,7 @@ group_by_rollup AS (
         zonedist4,
         count(*) AS zoning_combo_count
     FROM pluto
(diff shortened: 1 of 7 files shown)


products/green_fast_track/models/product/source/_source_models.yml
```diff
@@ -57,5 +57,4 @@ models:
 
   - name: source__shadow_hist_resources_points
   - name: source__shadow_hist_resources_lots
-  - name: source__shadow_hist_resources_polys
```
fvankrieken, line 60: I take it there are no polys after this? The two datasets going into historical_resources are just points and joined lots

damonmcc: indeed, just like `historic_resources` and `historic_resources_adj`. double checking right now

damonmcc: added a screenshot above shoring that is all points now

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
```diff
@@ -15,7 +15,7 @@ group_by_rollup AS (
         zonedist4,
         count(*) AS zoning_combo_count
     FROM pluto
-    GROUP BY ROLLUP (zonedist1, zonedist2, zonedist3, zonedist4) -- noqa: PRS
+    GROUP BY rollup(zonedist1, zonedist2, zonedist3, zonedist4) -- noqa: PRS
```
fvankrieken, line 18: nit but I think rollup is a keyword not a function so it should be uppercase

fvankrieken (approved):

products/green_fast_track/analyses/pluto_zoning_district_groups.sql
damonmcc: yes! thank u, i forgot that sqlfluf is overzealous here

PR #888, opened 2024-06-05
Move Metadata to new Repo

alexrichey:
[It lives here now!](https://github.com/NYCPlanning/product-metadata/)


alexrichey:
> Is there any code right now that expects these metadata files to live where they currently are?

@fvankrieken Nope! The Metadata files are going to be packaged up with the product+datasets on s3, so once we implement packaging, the location will start to matter.

damonmcc:
@alexrichey 

> The Metadata files are going to be packaged up with the product+datasets on s3, so once we implement packaging, the location will start to matter.

so right now for a CI or local use of `python3 -m dcpy.cli lifecycle distribute from_s3 ...` to succeed, a metadata file has to be manually put in the relevant folder?

alexrichey:
@damonmcc Yes, the metadata file will be part of the package on s3, e.g. 
image

That's the metadata file that the distribution CLI will use (though you override for testing purposes).

Issue #800, opened 2024-04-25
 Opendata: Automate Push -  First Batch


alexrichey:
Automate and Push the following datasets, in roughly this order:

- [x] Capital Projects Database (CPDB)
  - [x] [commitments](https://data.cityofnewyork.us/City-Government/Capital-Projects-Database-CPDB-Commitments/djxg-kcfi/about_data)
  - [x] [projects](https://data.cityofnewyork.us/City-Government/Capital-Projects-Database-CPDB-Projects/fi59-268w/about_data)
  - [x] [project points](https://data.cityofnewyork.us/City-Government/Capital-Projects-Database-CPDB-Projects-Points-/h2ic-zdws/about_data)
  - [x] [project polygons](https://data.cityofnewyork.us/City-Government/Capital-Projects-Database-CPDB-Projects-Polygons-/9jkp-n57r/about_data)
- [x] LION/Districts Quarterly Release
- [x] Zoning Application Portal (ZAP)
  - [x] [projects](https://data.cityofnewyork.us/City-Government/Zoning-Application-Portal-ZAP-Project-Data/hgx4-8ukb/about_data)
  - [x] [bbls](https://data.cityofnewyork.us/City-Government/Zoning-Application-Portal-ZAP-BBL/2iga-a6mk/about_data)
- [x] Mandatory Inclusionary Housing (MIH)
- [x] E Designations (EDes)

alexrichey:
bumping, so this is at the top of my notifications

damonmcc:
pushed CPDB Commitments via automation. moving on to the 3 Projects datasets

I'm pushing 23adopt first and then will push 24prelim. the columns are the same and it will be informative to see the data change

alexrichey:
thanks @damonmcc - if you can also update the spreadsheet, that'd be cool. Just confirming, for these we need to do two things 
1. Fulfil l our role as OTI and update the old-datasets
2. Time-permitting, create a new dataset to which we can automate the push. 

Sounds like you did #2 for Commitments? Did you also update the old version?

damonmcc:
@alexrichey
the push for [CPDB - Commitments](https://data.cityofnewyork.us/City-Government/Capital-Projects-Database-CPDB-Commitments/djxg-kcfi/about_data) was via automation (CLI) so a new dataset is needed for it right?

damonmcc:
[...]

PR #890, opened 2024-06-05
socrata distribution for CPDB

damonmcc:
related to data update #460 and issue #800 

these changes ended up just being QOL stuff while creating and using CPDB's new metadata files

see PR https://github.com/NYCPlanning/product-metadata/pull/2 for use of these changes with CPDB metadata files

### DIFF
--- a/dcpy/connectors/socrata/publish.py
+++ b/dcpy/connectors/socrata/publish.py
@@ -134,7 +134,7 @@ class Socrata:
 
                 logger.info(
                     f"""Updating Columns at {self._column_update_endpoint}
-                    Columns in uploaded data: {self.column_names}
+                    Columns from dataset page: {self.column_names}
                     Columns from our metadata: {expected_api_names}
                 """
                 )
@@ -142,8 +142,8 @@ class Socrata:
                 assert self.column_names == set(
                     expected_api_names
                 ), f"""The field names in the uploaded data do not match our metadata.
-                - Present in our metadata, but not uploaded data: {set(expected_api_names) - set(self.column_names)}
-                - Present in uploaded data, but not our metadata: {set(self.column_names) - set(expected_api_names)}
+                - Present in our metadata, but not dataset page: {set(expected_api_names) - set(self.column_names)}
+                - Present in dataset page, but not our metadata: {set(self.column_names) - set(expected_api_names)}
                 """
 
                 for uploaded_col in self.soc_output_columns:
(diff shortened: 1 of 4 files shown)


dcpy/connectors/socrata/publish.py
```diff
@@ -120,28 +120,30 @@ def _column_update_endpoint(self):
                 )
 
             @property
-            def column_names(self) -> set[str]:
-                return {c["field_name"] for c in self.soc_output_columns}
+            def column_names(self) -> list[str]:
```
fvankrieken, line 123: Is there reason we'd want this here and not have the underlying `self.soc_output_columns` have the same ordering?

dcpy/lifecycle/package/validate.py
```diff
@@ -282,15 +286,6 @@ def validate_package(
                 )
             )
 
-    unique_row_counts = {v.stats.row_count for v in validations}
```
fvankrieken, line 285: What's the story here?

damonmcc: see [commit message](https://github.com/NYCPlanning/data-engineering/pull/890/commits/c820a4992f674fec6eebdcb7fb9974b30dbd66e1)

damonmcc: it feels cruel to fully delete it, but maybe this is a check we'd rather save for our upcoming `pandera`ish approach to checks like this?

dcpy/connectors/socrata/publish.py
damonmcc: [looks like](https://github.com/NYCPlanning/data-engineering/blob/dm-cpdb-socrata/dcpy/connectors/socrata/publish.py#L166-L170) when we actually push column metadata we'd probably rather have them in the manual order they're set to in socrata (ideally the same order as in `metadata.yml`) rather than sorted alphabetically

this sort was really just for logging QOL

dcpy/lifecycle/package/validate.py
alexrichey: I do think that this probably a check worth saving, but with an option to ignore differing row counts. Perhaps we could set a flag in the metadata at the package level about whether row counts should be the same. However, the pandera approach isn't going to help at all with this, as it's a check that goes across dataset_files. 

So I'd leave this line in place, but not throw an exception during validation if there are differing row counts.

alexrichey (approved):

PR #894, opened 2024-06-07
ignore .package folder

alexrichey:


damonmcc:
thanks!

alexrichey:
FYI, yolo'ing this type of thing into main because it just doesn't warrant bothering other folks about it

PR #901, opened 2024-06-10
Ar build 2024 colp

alexrichey:


products/colp/recipe.yml
```diff
@@ -1,13 +1,11 @@
 name: COLP
-version: "202311"
+version: "2024-06-01"
```
fvankrieken, line 2: We maybe should check in with Amanda and settle on a versioning schema. But for now seems like we've done this simple date string, so seems fine

fvankrieken (approved):

PR #920, opened 2024-06-17
ESRI PDF Parser

alexrichey:
Add a very janky parser to make column metadata from the contents of an ESRI pdf. Also updates the documentation to list the ways one could generate metadata. 

I wouldn't waste too much time reviewing the logic of the pdf-text parser. If this is something we continue on with, we might want to actually use a pdf-parser, or just grab the data from an FGDB or somehow from ESRI. (There's got be a better way!)

@damonmcc 👆in reference to your question about how to export md.

### DIFF
--- a/dcpy/lifecycle/package/esri.py
+++ b/dcpy/lifecycle/package/esri.py
@@ -1,0 +1,84 @@
+from pathlib import Path
+import re
+import typer
+import yaml
+
+import dcpy.models.product.dataset.metadata as models
+from dcpy.utils.logging import logger
+
+
+def _parse_raw_column(field_raw) -> models.Column:
+    """
+    Parse a column value from ESRI metadata pdfs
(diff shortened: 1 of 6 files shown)


dcpy/lifecycle/package/esri.py
```diff
@@ -0,0 +1,84 @@
...
+    field_token = "Field "
+    alias_token = "Alias "
+    type_token = "Data type"
+    desc_token = "Field description\n"
+    iters = 0
+    while True:
+        next_comma = field_raw.find(",")
+        next_chunk = field_raw[0:next_comma]
+
+        PROBABLY_AN_INFINITE_LOOP_COUNTER = 100
```
fvankrieken, line 31: A+ variable name

fvankrieken (approved):

PR #922, opened 2024-06-17
ztl dbt initial setup

### LINKED ISSUE #910: DBT ZTL: set up dbt project with sources
## todos
- [x] create `project.yml`
- [x] create `models` folder (used the existing `sql` folder instead)
- [x] create `_sources.yml`
- [x] create `profile.yml`
- [x] create `packages.yml`
- [x] ztl build and load 
- [x] confirm the existence of ztl datasets in DBeaver
- [x] run `dbt build`
- [x] add expectations of source data 

## background/description

As a primary step of getting ztl ready for dbt, we'll need to get the project set up and configured in `products/zoningtaxlots`. Luckily, `products/green_fast_track` offers a great template for what's needed. 

Project, profiles, and packages yml files should be created. For now this is a bit boilerplate - ideally we want to share some of these things across products, but for now this keeps things simple for getting ztl running with dbt

Additionally, a `models` folder should be added, for now just with `_sources.yml`. This file is meant to tell dbt what resources/tables/etc will be available at the time of running dbt - the environment before dbt is invoked. For us, this means whatever tables we've loaded using `recipe.yml` and `dcpy.lifecycle.builds.plan/load`. This will take a little more work. To start, it can just be set up with names of sources. If `load` command is run, you should then be able to run `dbt build` and have it at least verify that sources exist. At that point, if all else is working, expectations of source data can be added. The easiest way to do this will be, after running `load` command on a clean database, inspecting the tables in DBeaver. Note that we don't necessarily need to specify all columns in each source dataset as required/expected, just the ones used in the build. This is a little harder to find, and will take some digging through sql files.

HengJiang0206:
resolves #910 

Set up and configured dbt for ZTL, created `project.yaml`, `profile.yaml`, `packages.yaml`,`_sources.yaml`, and the `models` folder. Table expectations added to _sources and tested with `dbt build`. 

Screenshot 2024-06-20 at 12 28 10
Screenshot 2024-06-20 at 12 28 35


products/zoningtaxlots/models/_sources.yml
HengJiang0206, line 1: @fvankrieken I added some expectations based on the SQL queries and table structures. I'm trying to figure out where to use `dbt_utils.at_least_one`. I also didn't find where `dcp_specialpurposesubdistricts` was referenced in the queries. There are 2 failed tests: `Failure in test source_not_null_recipe_sources_dof_dtm_wkb_geometry` and `Failure in test source_unique_recipe_sources_dcp_zoningmapindex_zoning_map`.

fvankrieken: Hmm. Well the specialpurposesubdistricts is the first interesting find here. It must have been dropped from ztl at some point (maybe in favor of dcp_specialpurpose) but wasn't removed.

Re the tests - I would run some queries, and see what fails. The `not_null` geometry seems like that may just be a requirement that we can't actually expect. This seems like a case to use `at_least_one` - see [here](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#at_least_one-source) on usage.

The other one, the `unique` one is worth maybe querying the db just to see what the data looks like. How many rows have duplicate values, does it seem like an isolate case, or is this truly not a unique field? Probably we'll just drop the constraint, but it'd be good to know what the data look like first

fvankrieken: Ooh good point, I always forget that

HengJiang0206: Screenshot 2024-06-18 at 13 12 22


@fvankrieken Thank you, Finn. I just checked `dcp_zoningmapindex`, where `zoning_map` is not unique. It seems like some zoning_map values were repeated in the last few records. I will go ahead and drop this constraint for now.

fvankrieken: What do these look like when visualized spatially? As in, just select where zoning_map = 30D, and then view in the geospatial tab of dbeaver?

HengJiang0206: Screenshot 2024-06-18 at 13 24 40
@fvankrieken

damonmcc: i guess `zoning_map + section` might be unique? I preferred way to test a compound key is

```yml
tests:
  - dbt_utils.unique_combination_of_columns:
      name: table_name__compound_key
      combination_of_columns:
        - column_a
        - column_b
```

fvankrieken: Fascinating. Okay, go ahead and drop the constraint

fvankrieken: https://www.nyc.gov/assets/planning/pages/zoning-index-map/index.html

fvankrieken: @damonmcc there are 2 section 38 index 30Ds

HengJiang0206: @damonmcc @fvankrieken It seems only the entries with the smaller ogc_fid match the indexes on the map.

fvankrieken: For 30d, it looks like they all match - there's the second section labelled 30d in the link I shared

HengJiang0206: @damonmcc Thanks for the compound key! That's very helpful.

HengJiang0206: @fvankrieken Oh now I see, they all match the map. Sorry, I didn't zoom in close enough.

products/zoningtaxlots/packages.yml
```diff
@@ -0,0 +1,5 @@
+packages:
+  - package: dbt-labs/dbt_utils
+    version: 1.1.1
+  - package: calogica/dbt_expectations
+    version: 0.10.3
```
damonmcc, line 5: nit: we like having new lines at the end of files like this

damonmcc (approved):
LGTM!

Issue #928, opened 2024-06-18
DBT ZTL: Refine initial DAG, models


fvankrieken:
After we have a working dag/dbt project from #925, it'd be good to evaluate our DAG/sql and iterate on our models. Did we compress too much and now have queries that are too busy and hard to read? Or are some models maybe too discrete, and could be grouped further? This can be a bit of a gray area, but to me it comes down to largely
  - how readable is the DAG that dbt generates?
  - using that DAG as a reference, how intuitive is it to read through the repo and figure out what the transformations are doing?

There's also likely a macro to be written. The code for all the "per area" files/models is very repetitive and likely could have some of it pulled out

fvankrieken:
So there are a few parts to this.
- [ ] simplify staging models. I think this will be pretty minimal, but we should glance over and see if there are simple changes to make - do two consecutive CTEs do very little differently, and could we combine them? Stuff along those lines
- [ ] simplify intermediate models. This has a few parts
  - move logic from "dcp_zoningtaxlots" table upstream. There's a bunch of logic relating to some final filtering, altering etc that ideally could happen upstream from this final join table. 
  - Simplify/clean existing code in intermediate models. Evaluate final columns selected - are these used downstream? Are these helpful as persisted intermediate columns for debugging?
  - Edit constraints/tests. Ideally we want to be as constraining as possible here (unique set of columns ideally is just 2 columns)

HengJiang0206:
Todos:

- [ ] Move `commercialoverlay` pivoting logic upstream
- [ ] Move `limitedheightdistrict` column setup upstream
- [ ] Move `zoningmapnumber` column setup upstream 
- [ ] Move `zoningmapcode` column setup upstream
- [ ] Move `zoningdistricts` pivoting logic upstream
- [ ] Create a macro for setting nulls  for parks
- [ ] Move `commercialoverlay` drop duplicate logic upstream
- [ ] Create per area macro

PR #936, opened 2024-06-21
create dbt tables for ZTL 

### LINKED ISSUE #925: DBT ZTL: Implement DAG as dbt models and validate outputs
Follow up to #924 . Once we have a second draft of a dag, it's time to implement! This has two main components

- port over sql. This should happen in parallel - we can keep the `sql` folder as is, and move over and adjust relevant code and queries to create the new models as specified in our dag. This really has two parts, because one is much more complex than the other
  - [x] create staging tables. Again, reference our [dbt practices section](https://github.com/NYCPlanning/data-engineering/wiki/Code-Conventions#model-foldersfile-structure), but this is fairly straightforward - any table that's really a slight tweak of a source table should be a staging table. So if a column is simply renamed, there should probably be a `stg__dcp_zoningdistricts` that selects from `dcp_zoningdistricts`. Use `products/green_fast_track` as an example here.
  - [x] implement "int__zoningdistricts". This model should be equivalent to `lotzoneperorder` in `area_zoningdistrict.sql`. It may take an upstream intermediate model or two as well.
- [x] implement remaining"target" int tables (the highest number of each one in your flow chart) one at a time. These tables can maybe stay long format, not worrying about ordering and updating dcp_zoningtaxlot
- [x] implement dcp_zoning_taxlot/dcp_zoning_taxlot_export - our product table(s)
- [ ] validate. Does the output match up with the output of the old sql? This is also a bit of a complex question to answer. We have some queries built in to ztl qa to address these sorts of changes luckily

HengJiang0206:
resolves #925 

Creates stg tables for ZTL, `dbt build` completed successfully: 

Screenshot 2024-06-21 at 13 58 06

Creates preliminary int tables for ZTL, `dbt build` completed successfully: 

Screenshot 2024-06-21 at 17 36 25

Create preliminary SQL queries for the product table. `dbt build` completed successfully.
Screenshot 2024-06-24 at 17 17 58

** Modifications made to product table and int tables 
The primary key for the product table is `dtm_id`.  `_intermediate_models.yml` was modified to add table expectations that `dtm_id` or the combination of `dtm_id` and the joined column ids are unique, avoiding duplicate records. When pivoting the int tables using `row_numbers`, corrections were made to the queries to ensure that the joining process doesn't create duplicate `dtm_id.` The product table passed all tests and has `dtm_id` as its unique identifier.


products/zoningtaxlots/models/intermediate/zoning-tax-lots.sql
fvankrieken, line 1: Can you rename this "int__zoningtaxlots"?

products/zoningtaxlots/models/product/dcp-zoning-tax-lots-export.sql
fvankrieken, line 1: Can we rename this just "zoningtaxlots"?

products/zoningtaxlots/models/intermediate/_int_models.yml
```diff
@@ -0,0 +1,79 @@
...
+  - name: int__validdtm
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - bbl
+            - dtm_id
+  - name: int__zoningdistricts
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
```
fvankrieken, line 52: I had originally put this in #928, but I think actually at this point already restrict these tests a bit. The more columns are added to this, the less stringent it is. So for each of these, think about conceptually and maybe query - what actually __should__ identify a unique row here?

fvankrieken: Just to rephrase - this test is essentially declaring a compound unique key/identifier. So thinking about what columns rows are defined/identified by vs what columns are more properties of the entity defined by the row

products/zoningtaxlots/models/intermediate/int__zoningmapindex.sql
```diff
@@ -0,0 +1,83 @@
+{{ config(
+    materialized = 'table'
+) }}
+
+WITH dof_dtm AS (
+    SELECT * FROM {{ ref('stg__dof_dtm') }}
+),
+
+validdtm AS (
```
fvankrieken, line 9: I know this probably came from the source query, but lets refer to `int__valid_dtm` here

HengJiang0206: @fvankrieken `int__valid_dtm` has an additional filter `WHERE ST_GEOMETRYTYPE(ST_MAKEVALID(geom)) = 'ST_MultiPolygon'` , which is slightly different from the `validdtm` here.  Should I check if that makes a difference?

products/zoningtaxlots/models/intermediate/_int_models.yml
HengJiang0206: @fvankrieken I agree! For those where we use `row_number` to pivot, I think it probably makes sense to just test the unique combinations of `dtm_id`, `bbl`, `row_number`, and the pivoting condition column if there are any (for example some specified `perbblgeom >= 10`)?

products/zoningtaxlots/models/intermediate/int__zoningmapindex.sql
fvankrieken: Yup - see if that filters the row count anymore. I imagine there are rows in dtm without geom, but those wouldn't be joined anyways, so it depends if there are any dtm rows with geom that are not multipolygon

products/zoningtaxlots/models/intermediate/_int_models.yml
fvankrieken: So because of dtm weirdness, bbl isn't actually unique. So let's use `dtm_id` and __not__ `bbl`. 

`row_number` feels more of an attribute - really, these rows are mappings of dtm_id to whatever entity is represented in the other table. So in this case, I would say just `dtm_id` and `zonedist`. I don't know that this will be valid for all other tables (with their analagous columns like "zoning_map" for int__zoningmapindex, but we should do this pair of columns when possible, and then when needed (as in, if this combo is NOT unique), then include row_number as well

products/zoningtaxlots/models/intermediate/int__zoningmapindex.sql
HengJiang0206: @fvankrieken `int__valid_dtm` does have one less row than `stg__dof_dtm`, seems that one row got filtered out by the multi-polygon restriction.

fvankrieken: Ooh. What is it?

HengJiang0206: Oh actually it got filtered out for geom is null. I guess I can use `int__valid_dtm` then since the null won't get joined anyways? 

Screenshot 2024-06-27 at 16 30 21

products/zoningtaxlots/models/intermediate/int__zoningtaxlots.sql
```diff
@@ -0,0 +1,556 @@
...
+        boroughcode,
+        taxblock,
+        taxlot,
+        area
+    FROM insertion
+),
+
+-- add commercialoverlay
+
+add_comm1 AS (
```
fvankrieken, line 69: One final request - can we group these by category? So for the "add_comm" queries, instead of one CTE per row number, do something with
```sql
FROM pop_bbl AS ztl
LEFT JOIN commercialoverlay AS co1
    ON ztl.dtm_id = co.dtm_id AND co.row_number = 1
LEFT JOIN commercialoverlay AS co2
    ON ztl.dtm_id = co.dtm_id AND co.row_number = 2
LEFT JOIN commercialoverlay AS co3
    ON ztl.dtm_id = co.dtm_id AND co.row_number = 3
```

I'd like to clean up this file a bit more, but think that'll take a bit of upstream logic as specified in #928

fvankrieken: And same for specialdist and zoningdist

pyproject.toml
```diff
@@ -48,6 +48,14 @@ CAPTURE_DATE_PREV = "2023-07-01"
 internal_columns = '"Job_Number","Job_Type"'
 external_columns = '"Job_Number","Job_Type"'
 
+<<<<<<< HEAD
```
fvankrieken, line 51: You've got a little rebase detritus here - just delete this whole segment (lines 51-58, inclusive)

fvankrieken (approved):
Awesome. Let's wait for tests to pass, then merge

Let's squash and merge, making the message a bit more accurate since the scope of this has changed a bit, and paring down the description to the main things added

PR #945, opened 2024-06-24
Add XLSX content type for Socrata attachments

alexrichey:
Successful push [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9650060443/job/26614891450)


damonmcc (approved):
nice! just curious, where'd you find the incantation of `application/vnd.ms-excel`?

alexrichey:
@damonmcc from skimming [this](https://stackoverflow.com/questions/2937465/what-is-correct-content-type-for-excel-files) too quickly. This should probably just be `application/octet-stream`. Going to make that change real quick

PR #946, opened 2024-06-24
Use octet-stream for xls(x) socrata attachments

alexrichey:


dcpy/connectors/socrata/publish.py
```diff
@@ -329,7 +329,8 @@ def upload_attachment(
         content_types = {
             ".csv": "text/csv",
             ".pdf": "application/pdf",
-            ".xlsx": "application/vnd.ms-excel",
+            ".xlsx": "application/octet-stream",
+            ".xls": "application/octet-stream",
```
fvankrieken, line 333: Are we pushing any `xls`s??

alexrichey: not yet - but that was the only other format I could envision, other than various textfile formats

damonmcc (approved):
our `version.txt` isn't the most elegant thing in the world, but since Open Data doesn't have a metadata field to represent a dataset's version I'd be in favor of attaching it

alexrichey:
> our `version.txt` isn't the most elegant thing in the world. but since Open Data doesn't have a metadata field to represent a dataset's version, I'd be in favor of attaching it

This is a really intriguing thought... I think our versions.txt contains data that we absolutely should surface, but do you think it would be comprehensible to external folks? I wonder if we could fold this into packaging, perhaps a little `source data` section in the README detailing the various sources 🤔

damonmcc:
I think `version.txt` just has a single string in it (like `23v2`) and so seems pretty comprehensible, especially since that value should align with what external people would see on Bytes

`source_data_versions.txt` on the other hand lol those dataset names wouldn't make much sense to people. but (as a member of the public) I'd always prefer confusing accuracy over zero details about source data

PR #951, opened 2024-06-25
Add ability to template destination filenames

alexrichey:
Add ability to set destination filename override. Successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9670094962)

There's an argument that we should perhaps just jinja template the entire metadata.yml file, rather than targeting individual fields. But we can cross that bridge when we start implementing re-usable metadata components (e.g. the column definition for BBLs)


damonmcc (approved):
nice!

your note about templating the entire metadata.yml file (especially for column descriptions) reminded me of [dbt docs](https://docs.getdbt.com/reference/resource-properties/description#use-a-docs-block-in-a-description)

PR #950, opened 2024-06-25
bump python from 3.11 to 3.12

damonmcc:
active support for 3.11 [ended](https://endoflife.date/python) ~3 months ago

## steps taken
- change the `PYTHON_VERSION` variable which is used by our docker files
- run action to build our `base` image using these changes [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9668571351)
  - !! after doing this, every branch will be using this base image even before this is merged to `main` !!
- re-run docker image builds in this PR's tests [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9668446567/job/26673944389#step:4:286) to use new base image
  - old base image [tag](https://github.com/NYCPlanning/data-engineering/actions/runs/9668446567/job/26672648518#step:4:286) and new base image [tag](https://github.com/NYCPlanning/data-engineering/actions/runs/9668446567/job/26673944389#step:4:286)
- compile python packages using this branch
- run all builds [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9682659915) using new python and package versions (in-progress)

PR #952, opened 2024-06-25
Build dev "base" docker image in pr tests when needed

fvankrieken:
Squasher - just look at the overall changes

### Recap: What images we maintain

We have 5 docker images

**base** - image with python installed, our standard command line utilities installed, and (currently) my forked gdal installed. This was created because we needed my gdal fork due to a bug in 3.8 (which seems maybe fixed now), and it takes a long time to build gdal from source.

base feeds into 3 of our other four images. Let's call these our "production" images

- **build-base**: base image, with python packages needed for build and dcpy installed
- **build-geosupport**: ", geosupport desktop and python installed
- **dev**: ", with tools like sqlfluff, black, mypy, etc. Used for dev container and running tests

Then we also have **docker-geosupport**. This is a bare container with python and geosupport. This is potentially slightly more public-facing. It doesn't pull from build base, nor do we build "dev" versions of it.

### Motivation
In #390 I added some logic to build our three "production" images as part of PR tests when relevant code changes have occured, to allow testing of logic with new python requirements (or other changes to production environment). However, at some point I pulled out our "base" image so that we didn't have to build gdal from source each time we built one of these images. Our "base" image seemed static enough, so for the time I simply did not add any functionality for using a dev "base" image. 

The downside of course is that this makes isolated testing of things like #950 (or bumping gdal versions, etc) tough.

### Changes

Basically, there's a bit of a decision tree now when we have a pr test
```mermaid
flowchart
1[Changes to base image in pr?] --Yes--> 2[Changes to base image since last push?]
1 --No--> 3[Use base:latest]
2 --Yes--> 11["Build and use base:dev-{branch}"]
3 --> 6[Changes to docker/python folders in pr?]
6 --No--> 7[Use build-geosupport/dev:latest]
6 --Yes--> 8[Changes to docker/python folders since last push?]
8 --No--> 9["Use existing build-geosupport/dev:dev-{branch}"]
8 --Yes--> 10["Build and use build-geosupport/dev:dev-{branch}"]
2 --No--> 4["Use existing base:dev-{branch}"]
4 --> 8
11 --> 8
```

In terms of practical changes, this meant
- modifying logic in `docker` folder to put all "base" image-related logic in its own folder, so that we can tell from path-filter changes if changes to the base image have occured
- tweaking the test action to follow our decision tree

[Passing action building all dev images](https://github.com/NYCPlanning/data-engineering/actions/runs/9681237466)
[Passing action using existing dev images](https://github.com/NYCPlanning/data-engineering/actions/runs/9680828784)
[Passing action using existing base, rebuilding other images](https://github.com/NYCPlanning/data-engineering/actions/runs/9680997604)

[Line in build](https://github.com/NYCPlanning/data-engineering/actions/runs/9680997604/job/26710565996#step:4:272) showing that dev non-base images pull from correctly supplied base image tag


fvankrieken:
Closing and recreating to rerun original tests

.github/workflows/test.yml
```diff
@@ -69,26 +71,76 @@ jobs:
...
               - python/**
               - docker/**
+            docker_base:
+              - docker/base/**
           treat-pr-as-push: true
 
-  set_docker_image_tag:
-    name: Set docker image tag
+  base_image:
+    name: >
```
fvankrieken, line 79: This is hideous, I know. But it ends up being a nice label of what base image is being used/built.

It'd be cleaner if I made "set image tag" step its own job, but I'd rather have a cleaner dag on the [test worfklow](https://github.com/NYCPlanning/data-engineering/actions/runs/9681237466) and a nice name than simplify this horrible-looking expression

.github/workflows/test.yml
```diff
@@ -69,26 +71,76 @@ jobs:
             docker:
               - python/**
               - docker/**
+            docker_base:
+              - docker/base/**
```
fvankrieken, line 75: As a reminder, this second step here is checking to see if any changes have happened since the last push rather than in the pr

fvankrieken:
I should wait until #950 is merged to actually merge this to cleanup movement of a variable

fvankrieken:
Actually ready for review now

docker/base/setup.sh
fvankrieken, line 1: changes here are just cut/pasted from `docker/config.sh`

damonmcc (approved):
nice! two thoughts:
- that diagram should probably be added to the wiki
- are these images built and published when relevant PRs are merged to main?

fvankrieken:
Ooh good catch - the "production" images are, base is not. I'll add that

fvankrieken:
> nice! two thoughts:
> 
> * that diagram should probably be added to the wiki
> * are these images built and published when relevant PRs are merged to main?

New commit addresses the latter. I opted to build base on any push to main with changes to docker/python folders, not worrying about whether it changes base specifically. Figured it's not needed to run fast, so just keeping the workflow file simple and not bringing in `path-filters` action made sense.

[Run just building a production image](https://github.com/NYCPlanning/data-engineering/actions/runs/9685450072)
[Run building base as well](https://github.com/NYCPlanning/data-engineering/actions/runs/9685475204)

alexrichey (approved):
Looks good!

PR #956, opened 2024-06-26
Experiment: Create tags when publishing

fvankrieken:
Thinking about ingest rollout has gotten me thinking about stability of code and ability to recreate older builds. Personally, I'd like us to be able to change the projection of our source datasets to be consistently in state plane, or not worry too much if we want to clean up column names in a source dataset at some point. However, those would likely both require code changes in, say ztl build code. These changes would produce the desired results with new data, but if we tried to rerun an old build with our current repo, we'd obviously not be able to due to breaking changes.

We do track the commit which produced a dataset, but often these commits don't exist in the repo anymore. Github seems to [retain them](https://github.com/NYCPlanning/data-engineering/commit/7330f0700a8190ae904132da2a5872bcdccb88a2) for some amount of time, see that example of the commit that built FacDB 24v1. But it'd be a pain to try to run code from that commit - we can't just check it out, since it's not a part of any tree.

So, it'd make a lot of sense to create tags when we publish, cementing the code that built a specific published version. This is a first draft, really mean this to be a conversation starter more than anything else. For now, went with convention of tag names being "{product}__{version}" - most special characters aren't allowed.

[Here's the action that published a tag](https://github.com/NYCPlanning/data-engineering/actions/runs/9685375521/job/26725313608)
[Here's the tag](https://github.com/NYCPlanning/data-engineering/releases/tag/db-template__0.0.2)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -22,10 +22,6 @@ on:
         description: "Build folder in drafts to publish"
         type: string
         required: true
-      publishing_version:
-        description: "Data product release version"
-        type: string
-        required: true
       latest:
         description: "Publish to 'latest' folder?"
         type: boolean
@@ -70,13 +66,34 @@ jobs:
         working-directory: ./
         run: ./bash/docker_container_setup.sh
 
-      - name: publish
-        env:
-          latest: ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
+      - name: Download Metadata
+        run: |
+          python3 -m dcpy.connectors.edm.publishing download_file \
+            --filepath build_metadata.json \
+            --product ${{ inputs.product }} \
+            --draft \
+            --version ${{inputs.draft_build}}
+
+      - name: Set version and build sha
+        run: |
+          echo "version=$(cat build_metadata.json | jq -r .version)" >> $GITHUB_ENV
+          echo "commit=$(cat build_metadata.json | jq -r .commit)" >> $GITHUB_ENV
+
+      - name: Publish
         run: |
           python3 -m dcpy.connectors.edm.publishing publish \
             --product ${{ inputs.product }} \
             --build ${{ inputs.draft_build }} \
-            --version ${{ inputs.publishing_version }} \
             --acl ${{ inputs.product_acl }} \
-            $latest
+            ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
+
+      - name: create tag
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/${{ inputs.product }}__${{ env.version }}',
+              sha: '${{ env.commit }}'
+            })
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -495,7 +495,7 @@ def _cli_wrapper_download_file(
     ),
 ):
     if draft:
-        key = DraftKey(product=product, build=version)
+        key: ProductKey = DraftKey(product=product, build=version)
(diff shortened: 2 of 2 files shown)


fvankrieken:
@damonmcc @alexrichey @sf-dcp would love your thoughts

alexrichey:
@fvankrieken Ha, I was just thinking about this, in regards to product change-logs in metadata. I think it's a great idea to experiment with.

sf-dcp:
Nice! Curious about these things: 
1. How one can interpret the version? In your example, it's 0.02... Is this a second version of the code? 
2. Are these tags specific to changes in the entire repo or to the specific code used to build a certain product?
3. Are there storage concerns if we store code artifacts?

fvankrieken:
> Nice! Curious about these things:
> 
> 1. How one can interpret the version? In your example, it's 0.02... Is this a second version of the code?
> 2. Are these tags specific to changes in the entire repo or to the specific code used to build a certain product?
> 3. Are there storage concerns if we store code artifacts?

1. Version is version as specified in `build_metadata.json`. That's what currently is listed in templatedb's `recipe.yml`. So for pluto 24v2, the tag would be `db-pluto__24v2`
2. The latter - it just takes the commit that's in `build_metadata.json` and immortalizes it
3. A little hard to tell, but I don't think it's an issue. [This SO answer](https://stackoverflow.com/questions/10612742/will-too-many-tags-cause-any-issues-in-git) doesn't think so 🤷

alexrichey:
I'm wondering if we'd want to tag `main` only. Mostly so that the tagged commit is still connected to our linear commit history, and in theory you could easily blame between tags/releases. e.g. if we had a tag for @Pluto24v1 and @Pluto24v2, and they were commits from deleted branch, you could aways diff between them, but not sure if you could see what commits contributed to those changes. (or could you? Not sure how smart git would be in that case)

alexrichey (approved):

alexrichey:
@fvankrieken seems pretty low-risk, and tags are easily queryable/deletable if want to go a different direction or clean up later, so 👍. I do have a hunch that building/tagging from `main` is our eventual goal, but good with this for now.

.github/workflows/publish.yml
```diff
@@ -22,10 +22,6 @@ on:
         description: "Build folder in drafts to publish"
         type: string
         required: true
-      publishing_version:
-        description: "Data product release version"
-        type: string
-        required: true
```
damonmcc, line 28: dropping this because it's better to infer it from metadata than force someone to always but the correct value here?

fvankrieken:
An important note is that operationally, we should not delete branches if outputs from that branch are in review as a `draft`. I think in general we do a good job of that anyways

.github/workflows/publish.yml
fvankrieken: Precisely

fvankrieken:
> > Nice! Curious about these things:
> > 
> > 1. How one can interpret the version? In your example, it's 0.02... Is this a second version of the code?
> > 2. Are these tags specific to changes in the entire repo or to the specific code used to build a certain product?
> > 3. Are there storage concerns if we store code artifacts?
> 
> 1. Version is version as specified in `build_metadata.json`. That's what currently is listed in templatedb's `recipe.yml`. So for pluto 24v2, the tag would be `db-pluto__24v2`
> 2. The latter - it just takes the commit that's in `build_metadata.json` and immortalizes it
> 3. A little hard to tell, but I don't think it's an issue. [This SO answer](https://stackoverflow.com/questions/10612742/will-too-many-tags-cause-any-issues-in-git) doesn't think so 🤷

@sf-dcp I answered 2 a little oddly - it's specific to the entire repo, but not really to "changes" - just checks out/saves a specific commit (whichever commit was checked out to build the product)

sf-dcp (approved):
👌

PR #959, opened 2024-06-27
add bash script to run sqlfluff dbt for multiple projects

fvankrieken:
@HengJiang0206 has run up in issues in #936 that it's awkward to run sqlfluff with dbt templater for a different project. This PR creates a script to make it a relatively simple call to run dbt in one step.

This removes the sqlfluff dbt templater section from pyproject.toml and instead only adds it in when we want to run it. This is an annoying hack since these configs can't just be passed to sqlfluff (maybe I need add this to my fork...)

I originally was going to do something like yq to edit the toml, but yq doesn't support things like arrays in toml yet and installing another tool seemed like more hassle than it was worth, so I just append to and then drop the end of the pyproject.toml with basic bash function calls

### DIFF
--- a/bash/sqlfluff_dbt.sh
+++ b/bash/sqlfluff_dbt.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+command=$1
+project=$2
+dir="products/${project}"
+folder=${3:-models}
+
+if [ -z $project ]
+then
+    echo "supply both sqlfluff command (lint/fix) and project folder"
+    exit 1
+fi
+
+if [ -eq "$project" "zoningtaxlots" ]
+then
+    export BUILD_ENGINE_DB="db-ztl"
+else
+    export BUILD_ENGINE_DB="db-$(echo "$project" | tr '_' '-')"
+fi
+
+echo "
+[tool.sqlfluff.templater.dbt]
+dialect = \"postgres\"
+project_dir = \"${dir}\"
+profiles_dir = \"${dir}\"" >> pyproject.toml
+
+dbt deps --profiles-dir $dir --project-dir $dir
+dbt build --select config.materialized:seed --indirect-selection=cautious --full-refresh --profiles-dir $dir --project-dir $dir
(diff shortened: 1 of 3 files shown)


fvankrieken:
@damonmcc do you think it'd be worth trying to do something similar around "check dbt files" job in tests? What exactly is going on there again?

bash/sqlfluff_dbt.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo "supply both sqlfluff command (lint/fix) and project folder"
+    exit 1
+fi
+
+if [ -eq "$project" "zoningtaxlots" ]
+then
+    export BUILD_ENGINE_DB="db-ztl"
+else
+    export BUILD_ENGINE_DB="db-$(echo "$project" | tr '_' '-')"
+fi
```
damonmcc, line 19: this if/else is because our db names aren't 1-to-1 with our data product directory names?

fvankrieken: Precisely

damonmcc:
> @damonmcc do you think it'd be worth trying to do something similar around "check dbt files" job in tests? What exactly is going on there again?

seems worth doing something similar! maybe just moving `pre-commit run --all-files` to the end of the new bash script and deleting the entire job in `test_helper.yml` (I'll approve this PR tho, no need to hold up this useful change). so:

```bash
dbt deps --profiles-dir $dir --project-dir $dir
dbt build --select config.materialized:seed --indirect-selection=cautious --full-refresh --profiles-dir $dir --project-dir $dir
sqlfluff $command $dir/$folder --templater=dbt
pre-commit run --all-files
```

that job is doing very "dbt project" specific checks, rather than linting sql files to check for bad formatting. it's using [`dbt-checkpoint`](https://github.com/dbt-checkpoint/dbt-checkpoint) to enforce some conventions we've chosen like "All models have a properties file" or "All staging models have at least 1 child model"

damonmcc (approved):

PR #968, opened 2024-06-28
Modify ZTL QA logic

### LINKED ISSUE #926: Modify ZTL QA logic to not depend on state of database
This is a bit orthogonal, but still including as part of the dbt ztl work. The logic here is what currently is run by `bash/03_qaqc_upload.sh`. 

## QA Logic
What happens big picture is that the current build of ztl is compared to the last build. Given these two tables in the database, queries are run to count things like total number of changes by column, a list of specific bbls with changes in specific fields, what bbls are new/vanished, stuff along those lines. All the sql to generate these outputs lives in `products/zoningtaxlots/sql/qaqc`.

What needs to change is __how__ this happens

ZTL QA logic currently 
- dumps dcp_zoningtaxlot table from the current build (from `edm-data/db-ztl.{schema}`) to local csv
- loads this csv it to default-db.db-ztl.{version} in edm-data. We have moved away from doing logic in `default-db`. This also doesn't isolate anything by build - if you run on main, it creates this table. If you then run on your dev branch, it replaces this table.
image
- compares it to default-db.db-ztl.{prev version} in edm-data using scripts in `qaqc` folder
- dumps out qa output from this comparison to include in build exports

Really, what we should do instead is more akin to pluto, and do this in the scope of a build (and build db/schema) rather than do it in a more permanent place
- import previous build of ztl as one of inputs (see Pluto as example)
- create qc comparison tables as build/analysis dbt models (or product? @damonmcc curious on your thoughts. These files should either go in `models/product/qa` or in `analysis` folder at same level as `models` folder)
- export from build engine rather than this persisted db

HengJiang0206:
resolves #926 

*** To be resolved:

- [x] `in_bbldiffs.sql`
- [x] `recipe.yaml` : add `version_env_var`
- [x] `:VERSION ` and `:PREV_VERSION` references


products/zoningtaxlots/recipe.yml
```diff
@@ -12,3 +12,4 @@ inputs:
     - name: dof_dtm
     - name: dcp_zoningdistricts
     - name: dcp_zoningmapindex
+    - name: dcp_zoningtaxlots
```
fvankrieken, line 15: let's add an `import_as` clause like in Pluto - something to denote that this table is the previous version.

Let's also add back in the `version_env_var` - there's a bash script that sets this, which is why you were having an error before. Happy to talk about this on a call tomorrow - for now you could just run `export VERSION_PREV=2024-05-01` from your terminal and things should go smoothly

fvankrieken: Actually let's use the real script.

After running the `plan` command (assuming you're in products/zoningtaxlots), run
```
source ../../bash/export_recipe_env.sh recipe.lock.yml 
```

This will both set the `VERSION` and `VERSION_PREV` env vars

products/zoningtaxlots/models/product/qa/qa_bbl.sql
```diff
@@ -0,0 +1,29 @@
...
+
+prev_version AS (
+    SELECT * FROM {{ source('recipe_sources', 'dcp_zoningtaxlots') }}
+),
+
+bbl AS (
+    SELECT
+        sum(CASE WHEN bblnew IS null THEN 1 ELSE 0 END) AS removed,
+        sum(CASE WHEN bblold IS null THEN 1 ELSE 0 END) AS added,
+        'new_version' AS version,
```
fvankrieken, line 17: For these, we should use the env vars that we set in my comment below. See this guide for using [env vars](https://docs.getdbt.com/docs/build/environment-variables) within dbt

fvankrieken: For this query, let's also simplify it - get rid of the nested query and do this in a single select/join instead

products/zoningtaxlots/models/product/qa/qa_bbl_area_change
fvankrieken, line 1: this file has no extension!

products/zoningtaxlots/models/product/qa/qa_bbl_area_change
```diff
@@ -0,0 +1,33 @@
...
+WITH new_version AS (
+    SELECT * FROM {{ ref('int__zoningtaxlots') }}
+),
+
+prev_version AS (
+    SELECT * FROM {{ source('recipe_sources', 'dcp_zoningtaxlots') }}
+),
+
+qaqc_bblareachange AS (
+    SELECT
```
fvankrieken, line 14: Let's also collapse the nested query in this one

products/zoningtaxlots/models/product/qa/qa_mismatch.sql
```diff
@@ -0,0 +1,100 @@
...
+
+prev_version AS (
+    SELECT * FROM {{ source('recipe_sources', 'dcp_zoningtaxlots') }}
+),
+
+mismatch AS (
+    SELECT
+        count(*) AS total,
+        sum(
+            (a.borough_code != b.borough_code OR (a.borough_code IS NULL AND b.borough_code IS NOT NULL))::integer
```
fvankrieken, line 17: sqlfluff is doing it's best, but what might make a lot of this cleaner is some macros.

There's some similar stuff in queries below in qa_bbldiffs. It seems that we could use a dbt macro that takes in
- left table
- right table
- left table column
- right table column (default to left table column)

which when called `{{ not_equal_or_null(a, b, borough_code) }}` would give something like the clause above. Or maybe additionally have a boolean flag argument - down in bbldiffs, we want the clause above with an additional `OR`

products/zoningtaxlots/models/product/qa/qa_null.sql
```diff
@@ -0,0 +1,223 @@
...
+),
+
+prev_version AS (
+    SELECT * FROM {{ source('recipe_sources', 'dcp_zoningtaxlots') }}
+),
+
+
+
+newnull AS (
+    SELECT
```
fvankrieken, line 16: This section seems prime for maybe not a macro but just some jinja templating - iterate through a list of columns and union them. There might be some existing dbt function which does something like this, or you can write a for loop.

`dbt compile --select ...` is a very helpful command for this -> it spits out the "final" query that it's going to run

products/zoningtaxlots/models/product/qa/qa_out_bbldiffs.sql
fvankrieken, line 1: let's rename this just "qa_bbldiffs.sql"

products/zoningtaxlots/recipe.yml
HengJiang0206: @fvankrieken I got the error that `Dataset dcp_zoningtaxlots requires version env var: VERSION_PREV` when running `plan`

HengJiang0206: @damonmcc

damonmcc: (have been chatting in Teams) you've used a temporary workaround and I'm working on a fix related to getting previous versions

fvankrieken:
Nice. This is looking a lot cleaner. Sorry for the hassle around versioning stuff, I hadn't realized there might be breaking issues

products/zoningtaxlots/recipe.yml
fvankrieken: With #973 merged, hopefully this can be unblocked when you rebase main into this branch!

fvankrieken (approved):

PR #972, opened 2024-07-01
Refine initial DAG, models for ZTL

### LINKED ISSUE #928: DBT ZTL: Refine initial DAG, models

### LINKED ISSUE #965: DBT ZTL: create specialdistrict_priority seed file
Part of #928 but seemed distinct enough to warrant its own issue

Similar to zoning district priority seed, we do some manual reordering of special districts in ztl.

Right now, this reordering happens when [joining/updating the final table](https://github.com/NYCPlanning/data-engineering/pull/936/files#diff-bf536787e5e8a542639487eb7408dd58f06bb52c8ec845ae55319760a7bb47f0R162). Ideally, this happens upstream in `int__specialdistricts`, where the final table simply has the correct row numbers.

To do this, we should take the ordering specified in the linked line and put it in a seed file. Since these are only "relative" orderings, we can do this by having two columns, sdlabel1 and sdlabel2 (or something similar) that has these pairings defined. Then, this seed can be left joined in `int__specialdistricts` and used while ordering the rows. This is going to be a little tricky to do fully generalized (only reordering these pairs and not affecting edge cases of 3 or more special districts), but it's certainly possible. I think we'll maybe need a step of grouping by these pairings when possible, but regardless of approach it'll take some creativity and tinkering

HengJiang0206:
resolves #965 
resolves #928 

Logic leading up to the pivots was moved to the upstream tables. Tables were left normalized before the final `int__zoningtaxlots` final integration. 

The revision resulted in 60 updated rows compared with the reference table in the `nightly_qa` schema:
- 58 rows had opposite `specialdistrict1` and `specialdistrict2` orders due to change in `specialdistrict_priority`
- 2 rows had `commercialoverlay2` dropped, potentially due to syntax variations in calculations.


products/zoningtaxlots/models/intermediate/int__limitedheight.sql
```diff
@@ -33,18 +33,19 @@ limitedheightper AS (
 limitedheightperorder AS (
```
fvankrieken, line 33: Wow this one had a lot of unused stuff happening.

Looks like this can be simplified a bit further - the two separate CTEs here aren't really doing anything distinct, so we can merge them.

Also, we can simplify the result. The CASE statement is making the value to be joined to our final table either NULL or the appropriate value. But the output of this model is meant to be a table of pairs that gets joined, so NULLs aren't really needed. Meaning the boolean value in the CASE statement can really just be in a WHERE clause instead to filter the rows in the model

products/zoningtaxlots/models/intermediate/_int_models.yml
```diff
@@ -2,12 +2,12 @@ version: 2
 
 models:
   - name: int__commercialoverlay
-    tests:
-      - dbt_utils.unique_combination_of_columns:
-          combination_of_columns:
-            - dtm_id
-            - overlay
-            - row_number
+    columns:
```
fvankrieken, line 5: And with the tweak I'm suggesting in commercial overlay, this should go back to unique combo of dtm_id and overlay

products/zoningtaxlots/models/intermediate/int__zoningdistricts.sql
```diff
@@ -147,17 +147,28 @@ new_order AS (
 lotzoneperorder AS (
```
fvankrieken, line 147: And sorry - I do think here it's also better not to pivot. These intermediate tables are useful for debugging

products/zoningtaxlots/models/intermediate/int__specialpurpose.sql
fvankrieken, line 1: For this, I'd like to incorporate the priority seed slightly higher upstream, thinking about it more as an input to how we're ordering rows (as in, used somehow in the ROW_NUMBER clause that sets the final row number column) rather than something used at the end to correct what's produced in the upstream logic

products/zoningtaxlots/models/intermediate/int__commercialoverlay.sql
```diff
@@ -47,10 +47,43 @@ commoverlayperorder AS (
     FROM commoverlayper
 ),
 
-filter AS (
-    SELECT *
-    FROM commoverlayperorder
+filtered AS (
+    SELECT * FROM commoverlayperorder
     WHERE perbblgeom >= 10 OR perzonegeom >= 50
```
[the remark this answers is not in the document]
HengJiang0206: I tried doing those in one CTE. Since I can't refer to the alias specified in the SELECT clause in WHERE clause, I did `WHERE (segbblgeom / allbblgeom) * 100 >= 10 OR (segzonegeom / allzonegeom) * 100 >=  50`. I am unsure if it was because of the rounding or differences in SQL calculation, but this produced different results from referring to aliases in a separate query (one row was different).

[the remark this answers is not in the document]
fvankrieken: Let's include it, and look at the difference. I wouldn't be shocked if this is an edge case where we'd be correcting something.

products/zoningtaxlots/models/intermediate/int__specialpurpose.sql
HengJiang0206: @fvankrieken I can't think of an optimal way to do this, do I find a way to incorporate this order into the row number window function?

fvankrieken: Yeah sorry this one is a little funky. Let's make it a bit easier - 

For the priority csv, let's structure it more like zonedist priority, like so
```csv
spdist,priority
CL,1
MiD,2
"125th",3
EHC,4
EC-5,5
EC-6,6
TA,7
MX-1,8
MX-16,9
G,10
```

Let's left join this in `specialpurposeperorder`, and the row number function can be
```sql
(PARTITION BY dtm_id ORDER BY segbblgeom DESC, priority, sdlabel)
```

fvankrieken: This is potentially going to add 32 rows to bbldiffs - this is fine. It seems like there's actually a bug in the current code, that luckily hasn't been a huge issue due to very few tax lots being in multiple special districts, outside of the ones in this list

HengJiang0206: @fvankrieken Yes, I noticed the previous list didn't include all sdlbl. It left the orders involving the undefined ones untouched.

products/zoningtaxlots/models/intermediate/_int_models.yml
HengJiang0206: @fvankrieken It didn't pass this test, maybe I have to either deduplicate before adding row numbers or keep the deduplicate CTEs.

fvankrieken: Gotcha. Let's do a group by before the ordering in that case (grouping by `dtm_id, overlay`)

fvankrieken: And unioning/summing areas in the group by (similar to how we do for both special districts and zoning districts)

products/zoningtaxlots/models/intermediate/int__commercialoverlay.sql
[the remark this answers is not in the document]
HengJiang0206: @fvankrieken This results in 2 different rows according to `bbldiff`

products/zoningtaxlots/models/intermediate/int__specialpurpose.sql
HengJiang0206: @fvankrieken This results in 58 rows of SDs being in different orders

products/zoningtaxlots/models/intermediate/int__zoningtaxlots.sql
```diff
@@ -179,34 +144,16 @@ add_zoningmapnumber AS (
...
-        a.area,
-        a.commercialoverlay1,
-        a.commercialoverlay2,
-        a.specialdistrict1,
-        a.specialdistrict2,
-        a.specialdistrict3,
-        a.limitedheightdistrict,
-        a.zoningmapnumber,
+            WHEN b1.perbblgeom >= 10 THEN b1.zoning_map
+        END) AS zoningmapnumber,
```
fvankrieken, line 148: Yeah this seems like the one where it's fine to be leaving this bit of logic here, since it's a tad funky (and just different enough from a super simple pivot)

fvankrieken:
This is looking good! A couple more notes
- Can you add back in columns bbl, segbblgeom, and segzonegeom to the intermediate tables? While they're not used downstream, they are useful in debugging or looking into data oddities
- Let's add list of columns to the _int_models.yml for all int models. Don't need data type or constraints, but just columns that should be present
- inwoodrezoning should have the bbl filtering moved to its int table from the big join table, and then use dtm_id instead of bbl
- zoningtaxlots can be simplified a bit more, now that we've moved a lot of the logic upstream
  - combine "insertion" and "pop_bbl" into one CTE
  - combine the next CTEs all the way up until "park" to one CTE. These are all just joining with no real other logic, and we get a lot of duplicated select statements

I'm gonna check out those commercial overlay diffs too. But then I think we'll be good to go

HengJiang0206:
> Can you share your query for commercialoverlay2? I'm not seeing them just comparing your build to nightly_qa

@fvankrieken Did you mean the different rows? Or the table queries ?

fvankrieken:
> > Can you share your query for commercialoverlay2? I'm not seeing them just comparing your build to nightly_qa
> 
> @fvankrieken Did you mean the different rows? Or the table queries ?

Either one!

HengJiang0206:
> > > Can you share your query for commercialoverlay2? I'm not seeing them just comparing your build to nightly_qa
> > 
> > 
> > @fvankrieken Did you mean the different rows? Or the table queries ?
> 
> Either one!

Screenshot 2024-07-05 at 09 50 00

fvankrieken:
Gotcha. So they were null before and now they are not. This makes sense - we're applying the filter by area after the grouping now rather than before, so glancing at the data these are cases where a lot intersects a few discrete commercial overlay segments that collectively make it over the threshold.

This is something that should be implemented regardless, so these two changes are okay

fvankrieken (approved):
Good stuff. Passing your latest run of `dbt build`?

If so, go ahead and squash and merge!

PR #973, opened 2024-07-02
add `versions.Date`

damonmcc:
## motivations

during #968, doing

```yaml
- name: dcp_zoningdistricts:
   version_env_var: VERSION_PREV`
```

in `zoningtaxlots/recipe.yml` fails with

```bash
❯ python3 -m dcpy.lifecycle.builds.plan recipe --recipe-path recipe.yml
INFO:dcpy:Planning recipe from recipe.yml
INFO:dcpy:published_version_strings=['latest', '2024-05-01', '2024-04-01', '2023-12-01', '2023-10-01', '2023-08-01', '2023-07-01', '2023-06-01', '2023-05-01', '2023-04-01', '2023-02-01', '2023-01-01', '2022-12-01', '2022-11-01', '2022-10-01', '2022-09-01', '2022-08-01', '2022-07-01', '2022-06-01', '2022-05-01', '2022-04-01', '2022-03-01', '2022-02-01', '2022-01-01', '2021-11-01', '2021-10-01', '2021-09-01', '2021-08-01', '2021-07-01', '2021-06-01', '2021-05-01', '2021-04-01', '2021-03-01', '2021-02-01', '2021-01-01', '2020-12-01', '2020-11-01', '2020-10-01', '2020-09-01', '2020-08-01', '2020-06-01', '2020-05-01', '2020-04-01', '2020-03-13', '2020-03-10', '2020-03-09', '2020-03-02', '2020-03-01', '2020-02-22', '2020-02-18']
ERROR:dcpy:Error: Tried to parse version 2020-03-13 but it did not match the expected format
INFO:dcpy:previous_recipe=None
INFO:dcpy:Export envars: {'VERSION': '2024-07-01'}
```

this is because published versions of ZTL are not only the first day of the month, some are just normal dates like `2020-03-13`. and in [`versions.py`](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/utils/versions.py) we don't have a model for this simpler "date" format

## changes
- add `Date` to `utils.versions` and add tests for it
- ensure `get_previous_version` in `connectors.edm.publishing` can handle a mix of version types
- improve logging

## tests

PR tests passing and (using temporary changes to ZTL's `recipe.yml`) I ran `builds.plan` locally and got:

Screenshot 2024-07-02 at 3 16 18 PM

## thoughts

There are risks in having both `FirstOfMonth` and `Date`. For example, if we publish a dataset on 5/1, does that mean we _always_ publish on the first of a month?

I think we should eventually drop `FirstOfMonth` and either replace it with `Month` or use `Date` with a dataset-specific bump number.


fvankrieken:
We should scrap first of month, though it is the current convention for ZTL. The reason for having it as a specific type was to be able to programmatically determine the "previous" version (2024-06-01 has 2024-05-01 as prev)

But version types should definitely be fully distinguishable. I'd be happy to make a "month" version type and move to that moving forwards with ZTL. The "publish" versions are just for us anyways, but it's also a good chance to align ZTL versions with what GIS actually publishes, in general I think our "2024-06-01" gets published as "May 2024" (NOT June 2024) (link address in bytes for this version is https://s-media.nyc.gov/agencies/dcp/assets/files/zip/data-tools/bytes/nyczoningtaxlotdb_20240531.zip - not sure if that date has some sort of convention or if it just happens to have been published then)

dcpy/test/utils/test_versions.py
```diff
@@ -22,6 +23,40 @@ def test_parsing_invalid_version(self):
...
+                ],
+                [
+                    versions.FirstOfMonth(year=23, month=1),
+                    versions.Date(year=23, month=1, day=2),
+                ],
+            ],
+        ]:
+            self.assertEqual(sorted, versions.sort(version_list))
+
+    def test_sort_valid_invalid_versions(self):
```
sf-dcp, line 51: should rename `valid_invalid`

sf-dcp (approved):

dcpy/utils/versions.py
```diff
@@ -89,6 +89,38 @@ def bump(
...
+    month: int
+    day: int
+
+    @property
+    def label(self) -> str:
+        return f"{self.year}-{self.month:02d}-{self.day:02d}"
+
+    def bump(
+        self, version_subtype: VersionSubType | None = None, bump_by: int | None = None
+    ) -> Date:
```
fvankrieken, line 104: Should we just return a NotImplemented error? And leave this unimplemented. I don't think we currently bump anything that would take this format, and maybe we don't want to out of ambiguity (right now only bumps by days rather than months, and I can't think of a case where we'd want to bump the number of days programmatically)

dcpy/utils/versions.py
```diff
@@ -89,6 +89,38 @@ def bump(
...
+                f"Version subtype {version_subtype} not applicable for Date versions"
+            )
+        bump_by = bump_by or 1
+        current_date = datetime.strptime(self.label, "%Y-%m-%d")
+        end_date = current_date + timedelta(days=bump_by)
+        return Date(year=end_date.year, month=end_date.month, day=end_date.day)
+
+    @staticmethod
+    def generate() -> Date:
+        version = parse(date.today().strftime("%Y-%m-%d"))
```
alexrichey, line 116: should we just call this something like `from_todays_date()` and just use this as a static constructor? (as opposed to calling the version parser below)

alexrichey (approved):
Looks good. I do wonder about our abstraction for these date'ish version types in general... We've got `Quarter`, `FirstOfMonth`, and now `Date`. It seems like these are all actually just Date types, with a a different minimum duration between them. E.g. 
Date = 1 day
FirstOfMonth = 1 month
Quarter = 3 months. 

So if we just treat these types as (our implementation of) Date, with a duration, we could bypass any custom sorting. Definitely a rabbit hole to go down, here.

damonmcc:
made requested changes!

gonna hold off on scrapping `FirstOfMonth` from`verisons.py` and `first_of_month` version strategy from recipes so we can make that change as part of bigger discussion about these versions, bumping them, etc.

dcpy/utils/versions.py
alexrichey: FYI, looks like `FirstOfMonth` has that same `generate` circuitous parsing logic. Would be nice to apply the same convention there. (not a blocker)

dcpy/test/utils/test_versions.py
```diff
@@ -22,6 +23,40 @@ def test_parsing_invalid_version(self):
...
+                ],
+                [
+                    versions.MajorMinor(year=23, major=1),
+                    versions.MajorMinor(year=23, major=2),
+                ],
+            ],
+            [
+                [
+                    versions.Date(year=23, month=1, day=2),
+                    versions.FirstOfMonth(year=23, month=1),
```
fvankrieken, line 41: Can you include a few more test cases in this list? I'd like to make sure that first of months correctly sorts both before and after `Date`s in between them (and vice-versa)

dcpy/test/utils/test_versions.py
```diff
@@ -22,6 +23,40 @@ def test_parsing_invalid_version(self):
...
+                ],
+            ],
+        ]:
+            self.assertEqual(sorted, versions.sort(version_list))
+
+    def test_sort_invalid_versions(self):
+        with self.assertRaises(TypeError):
+            versions.sort(
+                [
+                    versions.FirstOfMonth(year=23, month=1),
```
fvankrieken, line 55: This is reassuring. Thank you for adding this test case!

fvankrieken:
> Looks good. I do wonder about our abstraction for these date'ish version types in general... We've got `Quarter`, `FirstOfMonth`, and now `Date`. It seems like these are all actually just Date types, with a a different minimum duration between them. E.g. Date = 1 day FirstOfMonth = 1 month Quarter = 3 months.
> 
> So if we just treat these types as (our implementation of) Date, with a duration, we could bypass any custom sorting. Definitely a rabbit hole to go down, here.

This really does seem like the way to go longer term - all dates under the hood, with different conventions for formatting.

Main exception would be CPDB - we have "{year}{stage of ccp}".

And I guess pluto? Majors are meant to be quarterly... but aren't entirely (or necessarily)

dcpy/test/utils/test_versions.py
damonmcc: added!

dcpy/utils/versions.py
```diff
@@ -177,6 +177,23 @@ def parse(v: str) -> Version:
...
 
+def sort(versions: list[Version]) -> list[Version]:
+    version_types = set([type(v) for v in versions])
+    # can't compare different types
+    if len(version_types) != 1:
+        # only handle date-like versions
+        if version_types.issubset(set([Date, FirstOfMonth])):
+            return sorted(
+                versions,
+                key=lambda version: datetime.strptime(version.label, "%y-%m-%d"),
```
fvankrieken, line 188: Not sure how tests are passing - this line throws an error when I try to run anything locally. %y should b %Y. @HengJiang0206 if you fix this in your branch, I think things should start working. Sorry for the mysterious error messages

fvankrieken: Ah - versions in the real world are %Y, while in the test cases we have %y (24 instead of 2024, etc)

fvankrieken: I guess we should have just stored a `date` object within the `Date` version

HengJiang0206: @fvankrieken It worked! Thanks !

damonmcc: @fvankrieken would love to store a `date` object in these `Version` dataclasses

Issue #960, opened 2024-06-27
new directory structure in DO like `build` -> `draft` -> `publishing`


sf-dcp:
pretty much #825

fvankrieken:
A couple thoughts on random features that would be nice

- when ztl data loading action runs and builds ztl, it'd be nice to immediately promote the build to a draft, since GIS runs this job with the explicit purpose of generating a draft
- maybe in general, it'd be nice to have a checkbox in build.yml allowing to directly promote to draft

sf-dcp:
> A couple thoughts on random features that would be nice
> 
> * when ztl data loading action runs and builds ztl, it'd be nice to immediately promote the build to a draft, since GIS runs this job with the explicit purpose of generating a draft
> * maybe in general, it'd be nice to have a checkbox in build.yml allowing to directly promote to draft

Could you clarify if both of these or one of these bullet points should be true?

fvankrieken:
I would say both - for the ztl dataloading action, we don't need to provide the choice exposed to users, just automatically do it. For `build.yml` it should be a boolean input

Issue #825, opened 2024-05-01
DE<>GIS Dataset Review Process + Internal Publishing Versions


alexrichey:
[Related Discussion](https://github.com/NYCPlanning/data-engineering/issues/785)

### Terms
`Build`: outputs of a build process, e.g. files from the build of Pluto

`Draft`: A build that has been promoted to the drafts folder, that is ready for review, and is immutable (ie can't be tinkered with or deleted, except for in rare circumstance). Has a `draft-version`

`Draft Version`: a sequential version, composed of a number and a summary. e.g. 1-initial-build, 2-fix-the-bug

`Published Draft`: a `Draft` moved to `publish` folder after review and approval.

`Published Version`: a sequential number, used in the case of republishing. e.g. 1,2,3. It might make sense to format them as v1, v2, etc.?

### Code Changes Required

- Modify edm-publishing models to account for the `draft-version`. 

- Add DraftVersion functionality to dcpy.utils.versions

- Change builds to output to `edm-publishing / {product-name} / builds / {build-name}`
    Also modify s3 cleanup scripts to clear out the `builds` folder, not the `drafts` folder.

- Add GH Action to `Promote to Draft`. The Action should take in a draft summary (e.g. "fix the bug") and dcpy should infer the next draft number, to create the draft-version. This should also create 'Review Required' GH Issue, and assign to GIS. 

- Modify the Publish GH Action to either publish latest or take a specific draft number. 

- There needs to exist a link between what's in Drafts and what's in `publish`, because those versions might not be identical. E.g. When we publish `draft: 3-fix-the-thing` -> `publish: (v)1` we need to document that pub v1 came from draft 3. [TBD on this one]

- Clean out the drafts folders. No moving data migration required, I don't think... 

### BONUS CODE
- Let's start preparing to kill the `latest` folder. Add an endpoint to qaqc to point to a product latest folder.

sf-dcp:
I have attachment issues with `latest` 🥺

fvankrieken:
One random requirement - `publish` action should definitely not take version as an arg anymore, it should read it from `build_metadata.json`

sf-dcp:
> One random requirement - `publish` action should definitely not take version as an arg anymore, it should read it from `build_metadata.json`

We actually do need a version when publishing with new reorg because the new `draft` folder will look like this: 
`draft/`
* `24v2/`
    *  `1-initial/`
    *  `2-fix-somthing/`
* `24v3`
[...]

PR #975, opened 2024-07-03
DO Directory Restructure : implement new `publishing`

sf-dcp:
Related to #825 and #960.

## What
1. Add `overwrite` option in GH publishing action. Why?
    * During a publishing action, suppose something fails **before** publishing to `latest` but **after** publishing to `<version>` folder. In this case, `latest` doesn't have up to date version. If we were to re-publish same data, then we would be creating a new, patched `<version>` folder. 
    **Solution:** add a button in GHA to overwrite folder contents if the version already exists. If overwriting is not selected, bump the version by patch type. 

2. Modify logic promoting to `latest` folder: if the draft has an older version when compared to the data in `latest`, throw an error and don't promote the draft.

3. Modify current versioning to allow patches in `publish`. Below are 3 scenarios when promoting to `publish` and the new logic implemented in this PR:
    * The version doesn't exist in `publish` folder - promote to `publish` with the input version. No changes to current behavior.         
    * The version already exists in `publish` and `overwrite` is selected - overwrite existing version folder with the new contents.
    * The version already exists in `publish` and `overwrite` is NOT selected - save data contents to a new, patched version folder.  

I recommend reviewing commit by commit. 

## New version examples
* `24v3` -> `24v3.0.1` 
* `24v3.2` -> `24v3.2.1`
* `24Q1` -> `24Q1.1`
* `2024-03-01` -> `2024-03-01.1`

## 🚨 Feedback needed
* For non-patched `MajorMinor` versions, do we prefer `24v3` or `24v3.0`? 
* Overall version patch format consisting of  `<version>.<patch num>`? 
* Thoughts on new logic for promoting to `latest`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -90,7 +90,7 @@ jobs:
             --product ${{ inputs.product }} \
             --build ${{ inputs.draft_build }} \
             --acl ${{ inputs.product_acl }} \
-            --is_patch ${{ inputs.is_patch }} \
+            ${{ github.event.inputs.is_patch == 'true' && '--is_patch' || '' }} \
             ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
 
       - name: create tag
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,11 @@ on:
         type: boolean
         default: true
         required: true
+      overwrite_if_exists:
+        description: "Overwrite contents if version exists?"
+        type: boolean
+        default: false
+        required: true
       product_acl:
         description: "Set access control list for output in DO"
         type: choice
@@ -85,6 +90,7 @@ jobs:
             --product ${{ inputs.product }} \
             --build ${{ inputs.draft_build }} \
             --acl ${{ inputs.product_acl }} \
+            --overwrite_if_exists ${{ inputs.overwrite_if_exists }} \
             ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
 
       - name: create tag
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,8 +27,8 @@ on:
         type: boolean
         default: true
         required: true
-      overwrite_if_exists:
-        description: "Overwrite contents if version exists?"
+      is_patch:
+        description: "Create a patched version?"
         type: boolean
         default: false
         required: true
@@ -90,7 +90,7 @@ jobs:
             --product ${{ inputs.product }} \
             --build ${{ inputs.draft_build }} \
             --acl ${{ inputs.product_acl }} \
-            --overwrite_if_exists ${{ inputs.overwrite_if_exists }} \
+            --is_patch ${{ inputs.is_patch }} \
             ${{ github.event.inputs.latest == 'true' && '--latest' || '' }}
 
       - name: create tag
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -214,7 +214,6 @@ def publish(
     By default, keeps draft output folder"""
     if version is None:
         version = get_version(draft_key)
-        assert version is not None
 
     version_already_published = version in get_published_versions(
         product=draft_key.product
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -216,7 +216,9 @@ def publish(
     )
     # Bump version if input version already exists and overwrite isn't allowed
     if version_already_published and not overwrite_if_exists:
-        version = versions.bump(previous_version=version, bump_type="patch", bump_by=1)
+        version = versions.bump(
+            previous_version=version, bump_type=versions.VersionSubType.patch, bump_by=1
+        )
 
     source = draft_key.path + "/"
     target = f"{draft_key.product}/publish/{version}/"
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -38,22 +38,19 @@ def get_build_metadata(product_key: ProductKey) -> BuildMetadata:
     return BuildMetadata(**yaml.safe_load(file_content))
 
 
-def get_version(product_key: ProductKey) -> str | None:
-    """
-    Given product key, gets version. None is returned when metadata file
-    for given product_key doesn't exist.
-    """
-    try:
-        return get_build_metadata(product_key).version
-    except ClientError:
-        return None
+def get_version(product_key: ProductKey) -> str:
+    """Given product key, gets version."""
+    return get_build_metadata(product_key).version
 
 
 def get_latest_version(product: str) -> str | None:
     """Given product name, gets latest version
     Assumes existence of build_metadata.json in output folder
     """
-    return get_version(PublishKey(product, "latest"))
+    try:
+        return get_version(PublishKey(product, "latest"))
+    except ClientError:
+        return None
 
 
 def get_published_versions(product: str) -> list[str]:
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -42,7 +42,7 @@ def get_version(product_key: ProductKey) -> str:
 
 def get_latest_version(product: str) -> str:
     """Given product name, gets latest version
-    Assumes existence of version.txt in output folder
+    Assumes existence of build_metadata.json in output folder
     """
     return get_version(PublishKey(product, "latest"))
 
@@ -211,6 +211,8 @@ def publish(
     target = f"{draft_key.product}/publish/{version}/"
     s3.copy_folder(BUCKET, source, target, acl, max_files=max_files)
     if latest:
+        # TODO: if the version is is older than existing latest, pass. provide logging message
+        # TODO: do we want raise an error or a logging message suffices?
         s3.copy_folder(
             BUCKET,
             source,
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -214,13 +214,8 @@ def publish(
     version_already_published = version in get_published_versions(
         product=draft_key.product
     )
-    # Exit if input version already exists and overwrite isn't allowed
+    # Bump version if input version already exists and overwrite isn't allowed
     if version_already_published and not overwrite_if_exists:
-        raise FileExistsError(
-            f"Version {version} already exists in s3. If you wish to overwrite folder contents, select 'overwrite'."
-        )
-    # Bump version if input version already exists in s3
-    elif version_already_published:
         version = versions.bump(previous_version=version, bump_type="patch", bump_by=1)
 
     source = draft_key.path + "/"
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -204,7 +204,7 @@ def publish(
     keep_draft: bool = True,
     max_files: int = s3.MAX_FILE_COUNT,
     latest: bool = False,
-    overwrite_if_exists: bool = False,
+    is_patch: bool = False,
 ) -> None:
     """Publishes a specific draft build of a data product
     By default, keeps draft output folder"""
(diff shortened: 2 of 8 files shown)


sf-dcp:
Tests started failing after rebase. Converting to draft until issues are resolved.

dcpy/connectors/edm/publishing.py
```diff
@@ -202,22 +204,53 @@ def publish(
     keep_draft: bool = True,
     max_files: int = s3.MAX_FILE_COUNT,
     latest: bool = False,
```
alexrichey, line 206: I'm curious whether you think we should keep this flag? Ie is there a case where we'd actually want to publish a later version, and _not_ update the latest folder?

fvankrieken: Rare case, but if we publish 24v2 and need to patch some horrible error in 24v1.1? Say a completely nonsense bbl with 100000 units that we want to correct for posterity

dcpy/connectors/edm/publishing.py
```diff
@@ -202,22 +204,53 @@ def publish(
...
     By default, keeps draft output folder"""
     if version is None:
         version = get_version(draft_key)
+
+    version_already_published = version in get_published_versions(
+        product=draft_key.product
+    )
+    # Bump version if input version already exists and overwrite isn't allowed
+    if version_already_published and not overwrite_if_exists:
+        version = versions.bump(
```
alexrichey, line 219: I'm hesitant about this. I'd be confused if I called this function to publish a version, and then a different version was published. I think I'd just prefer it to throw an error letting me know that the version already exists.

.github/workflows/publish.yml
```diff
@@ -27,6 +27,11 @@ on:
         type: boolean
         default: true
         required: true
+      overwrite_if_exists:
```
fvankrieken, line 30: I don't think we want this - personally, I'd rather have ability to simply/easily copy an existing `publish` version to `latest` rather than re-publish a draft to `{version}` and `latest` folders.

And if something goes horribly wrong and a folder gets half-generated correctly, we can just delete it manually

dcpy/connectors/edm/publishing.py
fvankrieken: Hmm. Should `bump`ing a patch version then maybe _only_ happen when we're "repeating" a build? We have functionality around that, so maybe at time of `plan`, we have
- repeat from `build`/`draft` folder -> keep version of that build/draft
- repeat from `publish` folder -> bump `patch`

dcpy/connectors/edm/publishing.py
alexrichey: @fvankrieken Sorry, to be clear, what I mean is if we infer that something is the latest or not programmatically, why has this optional at all? So in the case, we'd detect that 24v1.1 isn't latest (which, for us really means latest version, as opposed to latest built) and just not push it to the `latest` folder.

fvankrieken: OH great point. Sounds good to me

dcpy/connectors/edm/publishing.py
alexrichey: @fvankrieken I'm intrigued at this idea of separating code vs input data in versions. So you could look at two versions side by side and see at a glance that the code was bumped, but not the input data (or vice versa). Not concrete thoughts - just intrigued.

dcpy/connectors/edm/publishing.py
```diff
@@ -202,22 +204,53 @@ def publish(
...
-            max_files=max_files,
-        )
+        try:
+            latest_version = get_latest_version(draft_key.product)
+
+            # Both latest_version and version are expected to be of same version schema
+            after_latest_version = versions.is_newer(
+                version_1=version, version_2=latest_version
+            )
+        except ClientError:
```
alexrichey, line 237: Is this what we expect to happen when there are no existing versions? (if so, we really should handle this exception in `get_latest_version`)

dcpy/connectors/edm/publishing.py
damonmcc: >  if a `patch` means code changes but not source data version changes

I think it could mean either. it seems like what we're going for with a patch is "oops, we have to re-distribute a release. the outside world will use the same version, but we don't want to overwrite our internal record of what was distributed the first time"

so whether because we had change code or we because we used the wrong version of source data, seems like either could be the motivation for a patch

damonmcc: I think I agree that bumping for a patch should happen during the `plan` step. that's when we already bump the version via a product's `VersionStrategy` and the logic could go [here](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/lifecycle/builds/plan.py#L36)

seems like the plan logic could be something like:
```python
if this_is_a_patch:
  assert version_from_github_action is not None
  # bump the patch number of version_from_github_action and use that
else:
  if version_from_github_action is None:
    # use recipe.version_strategy to bump latest version and use that
  else:
    # use version_from_github_action
```

please forgive the nested `if/else`s, it's pseudocode. the logic above ignores `recipe.version` because we'd probably rather stop putting that in our checked-in recipe files

damonmcc: just added a comment to the versions discussions [here](https://github.com/NYCPlanning/data-engineering/discussions/984#discussioncomment-10054943)

dcpy/connectors/edm/publishing.py
sf-dcp: My thoughts around this: suppose we manually entered incorrect *future* version in the GHA, resulting in unwanted, updated `latest` version. Then, as far as I know, we don't have means to trace what version we had in latest before. Ideas?

fvankrieken: Related to "overwrite if exists" comments - I think just having some amount of functionality to either
- manually specify a published version as latest (just overwrite "latest" folder)
- reset "latest" (look at published versions, programmatically determine latest, update latest folder)

So just keeping functionality so that what we do to "latest" is somewhat flexible (since we overwrite it all the time and can't get around its mutability), but not having that flexibility around published versions

dcpy/connectors/edm/publishing.py
sf-dcp: I would like to revisit this discussion over a call/in-person to make sure I will be making correct changes to the PR

.github/workflows/publish.yml
sf-dcp: @fvankrieken, what if we chose incorrect draft build when publishing a version? Delete the generated version folder and republish since it wouldn't be a patch? Trying to think of our steps when a version must be overwritten

fvankrieken: Hmm. I agree that's a case that we should think about operationally. But it hasn't happened in my memory, and in that rare case, I think it's okay to "deviate" from our standard operations and delete a folder manually. I think my stance is stemming from feeling that having this checkbox seems to suggest that this option is "valid" in our standard workflow, which it really shouldn't be. So I'd rather err on side of caution in terms of devs not accidentally deviating from standard workflow in the case of a mis-publishing

fvankrieken: Had a typo in the last sentence which really ended up saying the opposite of what I meant - edited

dcpy/connectors/edm/publishing.py
fvankrieken: So I think I went off track a little bit in my comments, and looking back I don't think significant change is needed.

Again, this is a rare case, and per our discussions in person, I think we intended to append a patch when needed, and while my idealized case sounds nice, I think @sf-dcp 's points (in chat outside of github) around how/when we might need to do this maybe means that something more like this will be easier/necessary.

I do think version should not be an option to supply here via gha - I think it only should be read with the logic in line 212. Manually specifying version at time of publishing seems like asking for manual errors.

Also to Alex's point - maybe it's a tad risky to just automatically "patch" if the version exists. To this point maybe its worth having a bool argument - `is_patch` or something like that. Without this specified, we throw error on version collision, while with it, we patch.

And I still think we should do away with overwriting as a possibility

dcpy/connectors/edm/publishing.py
```diff
@@ -202,22 +204,53 @@ def publish(
...
     if latest:
-        s3.copy_folder(
-            BUCKET,
-            source,
-            f"{draft_key.product}/publish/latest/",
-            acl,
-            max_files=max_files,
-        )
+        try:
+            latest_version = get_latest_version(draft_key.product)
```
sf-dcp, line 231: Note to self: return `None` from `get_latest_version`

dcpy/connectors/edm/publishing.py
sf-dcp: Per discussion, keep as is for now.

.github/workflows/publish.yml
sf-dcp: Per discussion, remove `overwrite` option.

dcpy/connectors/edm/publishing.py
sf-dcp: Add `is_patch` option: 

- if a version exists and `is_patch = True`, create a patched version
- if it doesn't and `is_patch = True or False`, create the normal version
-  if a version exists and `is_patch = False`, throw an error

sf-dcp:
Something not shown here, feature to be added is revision of version in the `metadata.json` file when bumping a version.

sf-dcp:
PR is ready for re-review. Changes: 
1. Add `is_patch` arg. 
2. Remove `overwrite_if_exists` flag. With this change, if an input version already exists AND `is_patch=False`,  throw an error. Otherwise, patch the version
3. Update version in `build_metadata.json` file in s3 if patching a version.
4. Update function `publishing.get_version()` fn to return `None` if metadata file doesn't exist. Previously we would get a Client error. Update downstream functions as well.
5. Update/add tests to reflect these code changes

dcpy/utils/versions.py
```diff
@@ -157,27 +174,80 @@ def parse(v: str) -> Version:
...
+
+def is_newer(version_1: str, version_2: str) -> bool:
+    """
+    Compares `version_1` to `version_2`. Returns True if `version_1` is newer than `version_2`.
+    Both versions are expected to be of same Version subtype.
+    """
+    try:
+        version_1_obj = parse(version_1)
+        version_2_obj = parse(version_2)
+        versions_sorted = sort([version_1_obj, version_2_obj])
```
fvankrieken, line 242: nit - can just return `version_1_obj > version_2_obj` here, no?

This is also maybe a sign that we should just use `Version` objects and serialize to strs rather than deal with strs throughout the codebase... but again lets get our version logic a little more mature

dcpy/utils/versions.py
```diff
@@ -157,27 +174,80 @@ def parse(v: str) -> Version:
...
+    except TypeError:
+        raise TypeError(
+            f"Can't compare mixed types of dataset versions: {version_1_obj, version_2_obj}"
```
fvankrieken, line 245: Should we maybe just return the type error thats raised?

dcpy/utils/versions.py
sf-dcp: makes sense!

dcpy/connectors/edm/publishing.py
```diff
@@ -37,12 +37,18 @@ def get_build_metadata(product_key: ProductKey) -> BuildMetadata:
     return BuildMetadata(**yaml.safe_load(file_content))
 
 
-def get_version(product_key: ProductKey) -> str:
-    """Given product key, gets version"""
-    return get_build_metadata(product_key).version
+def get_version(product_key: ProductKey) -> str | None:
```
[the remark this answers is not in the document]
fvankrieken: And then [this](https://github.com/NYCPlanning/data-engineering/pull/975/commits/dec89121d821ad4271e9062ad002a68b0fcfb0ff#diff-c39cbd5c9784cd5e96cca1ad8aa91e9b20a8e76b0d6b3ad43d47a0d67e07c717R219) isn't needed - which makes sense, since we fully expect to get a version there

dcpy/utils/versions.py
sf-dcp: something like this? 
```python
except TypeError:
    raise
```

fvankrieken: Just no try/except clause at all

dcpy/connectors/edm/publishing.py
```diff
@@ -202,22 +208,82 @@ def publish(
     keep_draft: bool = True,
     max_files: int = s3.MAX_FILE_COUNT,
     latest: bool = False,
+    is_patch: bool = False,
 ) -> None:
     """Publishes a specific draft build of a data product
     By default, keeps draft output folder"""
     if version is None:
         version = get_version(draft_key)
+        assert version is not None
```
fvankrieken, line 217: this line can be dropped now as well

damonmcc (approved):

sf-dcp:
Ran the `publish` GHA on Template DB, everything works as expected: 

image

`build_metadata.json` in `latest` folder:

image

PR #977, opened 2024-07-05
Pluto zoning tweaks

fvankrieken:
A couple things were discovered in #972 for ztl that are the same for pluto
1. Bug in special districts - sorting wrong order. This changes about 60 lots - outside of the lots that have their ordering manually set for the specific combo of special districts, there are luckily very few that are in multiple
2. Use grouping logic in commercial overlay just like we do for zoning districts. This ends up only affecting a few rows as well

Running build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9942202430) to compare to nightly_qa. Differences are
- 40 sp dist flips (expected). I need to confirm with GIS about these (and ordering in ztl)
- 12 commercial dist changes (expected) - due to grouping by commercial overlay (so if there are two smaller C-2 overlays, that's captured)
- 1 splitzone change (due to a commercial overlay change)
- 25 lottype changes - unexpected. See #1012

### DIFF
--- a/products/pluto/pluto_build/02_build.sh
+++ b/products/pluto/pluto_build/02_build.sh
@@ -74,8 +74,6 @@ run_sql_file sql/zoning_specialdistrict.sql
 run_sql_file sql/zoning_limitedheight.sql
 run_sql_file sql/zoning_zonemap.sql
 run_sql_file sql/zoning_parks.sql
-run_sql_file sql/zoning_correctdups.sql
-run_sql_file sql/zoning_correctgaps.sql
 run_sql_file sql/zoning_splitzone.sql
 run_sql_command "VACUUM ANALYZE pluto;"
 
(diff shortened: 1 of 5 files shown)


fvankrieken:
Got go-ahead from GIS on these changes

products/pluto/pluto_build/02_build.sh
```diff
@@ -74,8 +74,6 @@ run_sql_file sql/zoning_specialdistrict.sql
 run_sql_file sql/zoning_limitedheight.sql
 run_sql_file sql/zoning_zonemap.sql
 run_sql_file sql/zoning_parks.sql
-run_sql_file sql/zoning_correctdups.sql
-run_sql_file sql/zoning_correctgaps.sql
```
sf-dcp, line 78: is this change related to the new group logic mentioned in the second bullet point?

fvankrieken: Yup - this deduplication happens via the `GROUP BY` clause now

sf-dcp (approved):
LGTM! Curious, did you end up using the pandas test util for comparing data frames to identify the data changes?

fvankrieken:
No, I used Pluto's QA queries. So 
- checking total row counts
- checking total number of changes per column (changes identified by bbl)
- investigating changed rows for rows with changes in columns found above (1 column at a time)

sf-dcp:
merge?

fvankrieken:
Going over one more time with Casey today (who was out previously), so figured I'd hold off. Will get this merged as-is or with tweaks today!

PR #978, opened 2024-07-05
`dcpy`: Add capability to scrape input datasets from recipe

sf-dcp:
It would be nice to automatically ingest input datasets for product recipes. Not all inputs can be automated: it depends on its source type. But we can create a fn that will try to extract as many as possible. 

For now, it's meant to be a draft PR until we fully migrate to `ingest` code. 

`TODO`: write tests

### DIFF
--- a/dcpy/lifecycle/ingest/extract.py
+++ b/dcpy/lifecycle/ingest/extract.py
@@ -8,14 +8,19 @@ from dcpy.models.lifecycle.ingest import (
     S3Source,
     ScriptSource,
     Source,
+    Template,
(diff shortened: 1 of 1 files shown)


dcpy/lifecycle/ingest/extract.py
fvankrieken, line 1: Love it. I do think this code should live somewhere within `builds` - even if it's doing ingest logic, it's a logical step that happens at the beginning of the build cycle and obviously is quite coupled to recipes which are really `build` entities

PR #979, opened 2024-07-05
Package from Bytes

alexrichey:
Adds the ability to assemble a dataset's package (or all datasets in a product) straight from Bytes. To do this, I've introduced a few new types and some templating so that we can describe the locations of files on Bytes. 

For now, we just use the new `package.zip_files` entries to download the file from Bytes, and extract the relevant datasets and attachments. But in the future we can flip this around - these are basically the instructions for what we need to zip up and send to the webteam. 

Also, this switches to treat attachments like other file types, which have a `name` and `filename` e.g `name=main_readme` and `filename=readme.pdf.` It was a real pain to treat them differently from other files.

The commits are relatively atomic, but not entirely. (e.g. sometimes I introduce a bug in a commit, then fix it in the next) So it's possible to read through commit-by-commit, but also perhaps a little difficult.

### A few things I'm dissatisfied with: 
Some keys in the metadata are in a frustratingly inconsistent state. For files, what we've generally called the `name` should really be the `key` or the `id`. We can figure that out later after our whiteboarding. 

Also, if we want to make this code applicable to other agencies, we need to generalize the Destination types.


dcpy/models/product/dataset/metadata.py
```diff
@@ -40,23 +43,18 @@ class DatasetOverrides(BaseModel, extra="forbid"):
     display_name: str | None = None
     description: str | None = None
     tags: list[str] = []
-    destination_file_name: str | None = None
+    destination_file_name: str | None = None  # TODO filename
```
alexrichey, line 46: looks like most of our code uses `filename` with no underscores, so we should standardize

dcpy/models/product/dataset/metadata.py
```diff
@@ -40,23 +43,18 @@ class DatasetOverrides(BaseModel, extra="forbid"):
...
 
-    def get_dataset_destination_name(self, version) -> str:
-        """Get filename for the destination dataset. Should only be invoked for unparsed datasets."""
-        return (
-            self.destination_file_name.replace(self._VERSION_TEMPLATE_TOKEN, version)
-            if self.destination_file_name is not None
-            else ""
-        )
+class RemoteFile(BaseModel, extra="forbid"):
+    id: str
```
alexrichey, line 50: introducing the concept of a `file.id`, to refer to a file in the package

dcpy/models/product/dataset/metadata.py
```diff
@@ -142,16 +147,51 @@ def get_columns(self, metadata: Metadata) -> list[Column]:
         return cols
 
 
+class ZipFile(File, extra="forbid"):
+    type: Literal["Zip"] = "Zip"
+    contains: list[File]
+
+
 class Package(BaseModel, extra="forbid"):
+    # TODO: should these just be combined into one flat list of files? Why have three distinctions
```
alexrichey, line 156: specifically, why have these three different attributes on the class, rather than just files, e.g. `List[File]`

dcpy/models/product/dataset/metadata.py
```diff
@@ -142,16 +147,51 @@ def get_columns(self, metadata: Metadata) -> list[Column]:
...
     dataset_files: list[DatasetFile]
-    attachments: list[str]
+    attachments: list[File]
+    zip_files: list[ZipFile] = []
+
+    _ASSET_TYPES = Literal["dataset_files", "attachments", "zip_files"]
+
+    def __init__(self, **data):
+        file_attachments = []
+        for a in data["attachments"]:
```
alexrichey, line 165: previously we had attachments as a list of strings, which was a real pain to deal with. This warns on import, and ports them to the new format.

dcpy/models/product/dataset/metadata.py
```diff
@@ -161,13 +201,35 @@ class Metadata(BaseModel, extra="forbid"):
...
 
+    # Hold onto this for serialization, to avoid Pydantics reformatting of the metadata.yml
+    _templated_source_metadata: str
+
+    def __init__(self, templated_source_metadata: str = "", **data):
+        super().__init__(**data)
+        self._templated_source_metadata = templated_source_metadata
+
     @staticmethod
-    def from_yaml(path: Path):
```
alexrichey, line 169: this wasn't really "from_yaml"

dcpy/models/product/dataset/metadata.py
```diff
@@ -135,16 +143,27 @@ def get_columns(self, metadata: Metadata) -> list[Column]:
...
 
+class ZipFile(File, extra="forbid"):
+    type: Literal["Zip"] = "Zip"
+    contains: list[str]
+
+
 class Package(BaseModel, extra="forbid"):
     dataset_files: list[DatasetFile]
     attachments: list[str]
+    zip_files: list[ZipFile] = []
```
fvankrieken, line 159: I'm still a tad unsure of the placement here - not that it doesn't work, but that it's a little unintuitive given that the rest of the `Package` is a very literal description of what is present and this is more an instruction.

Could you maybe at least add a comment here in code describing what this field is?

dcpy/models/product/dataset/metadata.py
```diff
@@ -135,16 +143,27 @@ def get_columns(self, metadata: Metadata) -> list[Column]:
...
 
     def get_dataset(self, ds_id: str) -> DatasetFile:
         ds = [d for d in self.dataset_files if d.name == ds_id]
         if len(ds) == 1:
             return ds[0]
         raise Exception(f"No dataset named {ds_id}")
 
+    def _files_by_name(self):
+        return {ds.name: ds for ds in self.dataset_files + self.zip_files} | {
```
fvankrieken: Are there going to be entries in `self.zip_files` that aren't also in `dataset_files`? Or maybe am I misunderstanding zip_files a bit

dcpy/test/lifecycle/package/resources/colp_single_feature_package/metadata.yml
```diff
@@ -299,3 +311,11 @@ destinations:
     overrides:
       display_name: "display_name overridden at the destination level"
       destination_file_name: shapefile_blob_{{ version }}
+  # BYTES
+  - id: bytes
+    type: bytes
```
fvankrieken, line 322: nit since this really isn't being used. But is BYTES a "bytes" type destination? I would think of it more as "manual" given that the update process is currently out of our hands and involves sending data manually to web team (or maybe it's a `web_team` type then).

dcpy/lifecycle/package/assemble.py
```diff
@@ -0,0 +1,40 @@
...
+):
+    out_dir = ASSEMBLY_DIR / product / version / dataset
+    out_dir.mkdir(exist_ok=True, parents=True)
+
+    for bytes_dest in [d for d in product_metadata.destinations if d.type == "bytes"]:
+        package_files = product_metadata.package._files_by_name()
+
+        for dest_file in bytes_dest.files:
+            f = package_files[dest_file.name]
+            file_name: str = f.filename if issubclass(type(f), models.File) else f  # type: ignore
```
fvankrieken: nit - you can just use `isinstance` here I believe. And for line 31, I think `isinstance` is slightly preferred over checking the returned value of `type(f)`

dcpy/test/lifecycle/package/resources/colp_single_feature_package/metadata.yml
```diff
@@ -255,9 +255,12 @@ package:
   - name: csv_package
     filename: csv_package.zip
     contains:
-      - primary_csv
-      - colp_readme.pdf
-      - colp_readme.pdf
+      - name: primary_csv
+        filename: colp_{{ version }}.csv
```
fvankrieken, line 259: This is specifying renaming the file for the zip archive right?

dcpy/models/product/dataset/metadata.py
```diff
@@ -177,15 +177,29 @@ class Metadata(BaseModel, extra="forbid"):
...
+        self._templated_source_metadata = templated_source_metadata
+
     @staticmethod
-    def from_path(path: Path, *, template_vars=None):
+    def from_yaml(yaml_str: str, *, template_vars=None):
+        logger.info(f"Templating the metadata with vars: {template_vars}")
+        templated = jinja2.Template(yaml_str, undefined=jinja2.StrictUndefined).render(
+            template_vars or {}
+        )
+        return Metadata(templated_source_metadata=templated, **yaml.safe_load(templated))  # type: ignore
```
fvankrieken, line 223: What does mypy not like about this?

dcpy/models/product/dataset/metadata.py
```diff
@@ -177,15 +177,29 @@ class Metadata(BaseModel, extra="forbid"):
     package: Package
     columns: list[Column]
 
+    # Hold onto this for serialization, to avoid Pydantics reformatting of the metadata.yml
+    _templated_source_metadata: str
```
fvankrieken, line 211: Just curious because I've thoughts on this too - what mainly here? None values in the output, reordering, loss of comments, etc

fvankrieken: At some point it would be worth making our own subclass of BaseClass to have a little more fine-grained control over some of this, and have things like having all datetimes serialize the way we want, enums serializing to values by default, etc

dcpy/models/product/dataset/metadata.py
fvankrieken: Yeah this makes a lot of sense if we're having the same file with different filenames in different places

dcpy/models/product/dataset/metadata.py
fvankrieken: Or sorry - This is the actual zip file and not the contents of it, right? In which case, makes sense, carry on

dcpy/models/product/dataset/metadata.py
```diff
@@ -201,7 +201,7 @@ class Metadata(BaseModel, extra="forbid"):
     # Hold onto this for serialization, to avoid Pydantics reformatting of the metadata.yml
     _templated_source_metadata: str
 
-    def __init__(self, templated_source_metadata: str, **data):
+    def __init__(self, templated_source_metadata: str = "", **data):
```
fvankrieken, line 213: If we do this, should `write_source_metadata_to_file` below model dump in case of empty `self.templated_source_metadata`?

dcpy/models/product/dataset/metadata.py
alexrichey: Yeah, I agree - this does feel a little misplaced. Package has so far been used to describe outputs of a build, whereas these are more like instructions for... err, "packaging". I almost feel like this class should just be `BuildArtifacts`, and then we'd want a separate class for `Package`, which is where `zip_files` as above makes sense. thoughts @fvankrieken

alexrichey: although... attachments aren't BuildArtifacts. Clearly need to think harder about this, but good fodder for whiteboarding tomorrow.

dcpy/test/lifecycle/package/resources/colp_single_feature_package/metadata.yml
alexrichey: I do think Bytes is a destination, even if we don't push to it ourselves. I've been conceiving of it as the noun describing where the dataset lives, either where we'd push to or pull from. (Right now we're just using it pull, but honestly I don't see why, in the future, we wouldn't just use it to automate the push as well.)

Framing it another way, I'm envisioning a future where `destination.type` is just used to to dispatch to a given connector, to either push or pull, and the heuristic I'm using for what to call the `destination.type` is: "what would we call the connector." So in this case, Bytes.

dcpy/models/product/dataset/metadata.py
alexrichey: Maybe what we're looking for is `Assembly` which would have a zip_files entry? 🤔

dcpy/models/product/dataset/metadata.py
alexrichey: Yeah, it's all of those things, esp reordering. I'm using this to get the templated metadata into the package, but that said, my broader concern here is actually whether we _should_ be doing that at all. Or does it just introduce potential new problems, e.g. what if we change a Socrata destination four-four... do we really want stale product metadata hanging around. I'm unsure.

dcpy/models/product/dataset/metadata.py
alexrichey: yeah, good call. Will make that change.

dcpy/models/product/dataset/metadata.py
fvankrieken: I think it makes sense that a specific packaged or distributed version of a dataset has version-specific metadata. Ideally it doesn't become stale but it seems unavoidable.

That being said, it also doesn't seem super wrong to have the untemplated metadata. Depends what's being templated I suppose. There's an argument `version` field (or some field in general that is focused on any metadata specific to a given version/build, or a separate file with this information that gets templated) and then leave fields untemplated to keep the mutable and immutable fields a little more separate. I don't know 🤷 

If we keep templating, I think this also ties back into a custom class - would be nice to allow fairly simple validation around what fields can be templated or not.

dcpy/models/product/dataset/metadata.py
fvankrieken: Just in case you didn't know - you can exclude defaults and nones in a model_dump. For exclude_defaults, not sure what it'd do if you instantiated something with the default manually specified, I assume it would be excluded as well but I hope not

dcpy/models/product/dataset/metadata.py
alexrichey: writing this down for my records. I'm thinking that manually preserving yaml formtting is a losing game, and we can do most of what we (or just I) need by 
- We can order top-level keys in the yaml like this `yaml.dump(data, default_flow_style=False, sort_keys=False)`
- I like having a newline between top-sections (makes it easier for me to traverse) which we can do with something like [this](https://github.com/yaml/pyyaml/issues/127#issuecomment-525800484)

fvankrieken (approved):

PR #985, opened 2024-07-08
move to single date version class

fvankrieken:
Just a little tinkering. I do think this is an improvement on the classes we have now. It assumes that "24Q4" corresponds to Oct 1, 2024, and that a "month" version refers to first of month. Maybe sorting should be pickier - not sort "24-10" and "24-10-02" since they're somewhat incompatible. But currently we don't have mixed date formats for versions, and aside from a future switchover I don't think we intend to

Does not resolve issue in #980 

Tests which involve bumping quarter versions ("24Q4") are working properly, I think this should be okay to go. Just in case, ran build action for all products https://github.com/NYCPlanning/data-engineering/actions/runs/9893720433. All bumps/version settings went fine, build failed due to db running out of storage

### DIFF
--- a/dcpy/test/utils/test_versions.py
+++ b/dcpy/test/utils/test_versions.py
@@ -99,4 +99,4 @@ class TestVersions(TestCase):
             [None, 7, "2023-01", "2023-08"],
             [None, 1, "2023-12", "2024-01"],
         ]:
-            self.assertEqual(v_expected, versions.bump(v, bumped_part, bump_by))
+            self.assertEqual(v_expected, versions.bump(v, bumped_part, bump_by).label)
--- a/dcpy/test/utils/test_versions.py
+++ b/dcpy/test/utils/test_versions.py
@@ -1,3 +1,4 @@
+from datetime import date
 from unittest import TestCase
 
 from dcpy.utils import versions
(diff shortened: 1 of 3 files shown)


dcpy/test/utils/test_versions.py
```diff
@@ -27,57 +39,51 @@ def test_sort_valid_versions(self):
...
-                    versions.FirstOfMonth(year=2023, month=1),
-                    versions.Date(year=2023, month=1, day=2),
-                ],
-            ],
-            [
-                [
-                    versions.FirstOfMonth(year=2023, month=3),
-                    versions.FirstOfMonth(year=2023, month=2),
-                    versions.Date(year=2023, month=2, day=2),
+                    "23v2",
```
fvankrieken, line 42: Seemed nicer to work with strings here because they're more easily human-identifiable as "real" versions

alexrichey (approved):
Looks great. One potential thought: if you want to remove the `versions.sort` function altogether, could you just override the `Version.__lt__` to throw an exception when one isn't descended from the other. Might be too much of a pain (and maybe would slow down sort)

damonmcc (approved):

PR #986, opened 2024-07-08
Unpin sqlfluff, clean up some dbt stuff

fvankrieken:
sqlfluff put out a version with my feature! So we're unpinning and finally getting off my fork.

On top of that, updated python requirements. We're back up to numpy 2.0.0 @damonmcc, looks like leafmap was the culprit?

## Smaller commits
1. for devcontainer, put workspace inside the home directory. This allows sqlfluff to look between the directory it's run from and the home directory for config files. Meaning, we can have pyproject.yaml at the top-level specify configs, .sqlfluff files in projects to have project-specific configs, run from the project folder, and it will take the root level, then override with anything it finds closer to the cwd. The one exception is `templater` configs... which is the main thing we'd want to override. Not sure why that one isn't supported. But anyways, at least we can run sqlfluff from within product folders and it will look upwards. This is new in 3.1.0. Running outside a dev container should be fine without this tweak
2. self-explanatory
3. a few mypy tweaks from python package bumps
4. latest sqlfluff is a little pickier I guess - these were failing

## Commit 5 - tweaks to gha, sqlfluff dbt
We do a couple dbt-related tests: sqlfluff, and dbt-checkpoint. They both rely on dbt being configured and use an active connection to the db. I thought it would make sense to isolate these tests a bit, and clean up the funky logic around using sqlfluff-dbt (issues with project/profiles-dir, etc).

So for the test github action, sqlfluff for dbt is broken out from the rest of the sqlfluff calls, and now we have a job that runs dbt tests for each dbt project. For each project, it
- spins up a temporary db as a service within the action. All tests use this rather than our persisted db
- sets up dbt (dbt deps, build seeds)
- runs sqlfluff
- runs dbt-checkpoint (currently just our pre-commit file). It runs the checks defined for all dbt projects. I know we could instead define hooks per project, but I figure we'll likely want the same checks on our dbt projects

I'd like to move away from dbt-checkpoint, both because the "official" dbt tests that @damonmcc linked seem like the right way to go and because we don't use pre-commit for anything else and it feels a little awkward to use it just for these github tests. And for now it was a little awkward to configure - these tests only can run from the root of the project, and cannot really be provided with arguments to format the checks, so to get run properly I had to 
- dynamically create a `.dbt-checkpoints.yaml` to point it at the right directory in the matrix
- have a somewhat duplicate `.profiles.yml` in the root of our repo for it to use, because while we can specify a project-dir in the checkpoins yml, there's no option to specify profiles-dir. Why? I have no idea.
  - side note - since we currently have the same profiles.yml for each of our projects, I don't hate this. Unfortunately, it seems tough to get dbt to look for this file in the root of our repo - it'll look in `~/.dbt`, but that's tougher to keep consistent across people's systems and in ghas
  so for now, we have an extra `profiles.yml` laying around

~Gha still uses a custom script to run `sqlfluff --templater=dbt`. But locally,~ I'd say with this PR the ~best practice~ **only way to run sqlfluff for dbt projects** would be to not run using this script, but to just run `sqlfluff fix/lint models --templater=dbt` **from the project folder**. That will leave building seeds and setting of env vars (for db connection in profiles) in the hands of the developer, but I figure you'll generally be running this when you're already developing, so it's not a huge pain. And gets away from that incredibly awkward implementation in the script

Very open to thoughts on this whole approach.


damonmcc:
betrayed by my beloved `leafmap` 😔

bash/sqlfluff_dbt.sh
```diff
@@ -11,22 +11,9 @@ then
...
-dbt build --select config.materialized:seed --indirect-selection=cautious --full-refresh --profiles-dir $dir --project-dir $dir
-set -e
-sqlfluff $command $dir/$folder --templater=dbt
-
-echo "$(head -n -5 pyproject.toml)" > pyproject.toml
+(
+    cd $dir
+    dbt deps
+    sqlfluff $command $dir/$folder --templater=dbt
+    dbt build --select config.materialized:seed --indirect-selection=cautious --full-refresh
```
damonmcc, line 18: should this be seeds before sqfluff?

.github/workflows/test_helper.yml
```diff
@@ -58,10 +40,58 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        project:
+          - green_fast_track
+          - zoningtaxlots
+    steps:
+      - uses: actions/checkout@v4
+      - name: setup
+        run: |
+          git config --global --add safe.directory /__w/data-engineering/data-engineering
```
damonmcc, line 85: is this needed for `pre-commit` to work

if it doesn't add much time, could probably be `./bash/docker_container_setup.sh` like the pytest section in this file

fvankrieken: Yeah, but that also sets up mc connection to DO spaces, which I'd prefer to not do in this step. I guess I could tweak that command to default to empty strings rather

fvankrieken: mc command is not happy with it

https://github.com/NYCPlanning/data-engineering/actions/runs/9875059942/job/27270890880#step:4:14

I'd rather leave this git command than supply a bunch of env vars that aren't needed otherwise

bash/sqlfluff_dbt.sh
fvankrieken: Oops - messed up my writeup. I'm dropping this script entirely

damonmcc (approved):

sf-dcp:
Unassigned myself from the PR review

PR #990, opened 2024-07-09
Fix `dcpy` util function `s3.get_subfolders()`

sf-dcp:
### What
While working on DO directory restructure stuff, I tried to use existing `s3.get_subfolders()` function but realized it doesn't work correctly when input param `index` isn't the default value (default value is set to be the top level of directory). For example, let's say we have the following objects in s3:

* `folder1/folder2/file.txt`
* `folder3/`

**The expected behavior of the function is this:**

```bash
>>> s3.get_subfolders(bucket='my_bucket', prefix='', index=1)
>>> ['folder1', 'folder3']
>>>
>>> s3.get_subfolders(bucket='my_bucket', prefix='', index=2)
>>> ['folder1/folder2']
```

**The current behavior:**
```bash
>>> s3.get_subfolders(bucket='my_bucket', prefix='', index=1)
>>> ['folder1', 'folder3']
>>>
>>> s3.get_subfolders(bucket='my_bucket', prefix='', index=2)
>>> []
```

### Why
The reason why the current code doesn't work is because supplying the `Delimiter` param below automatically returns directories ONLY from the top level: 
```python
paginator = client().get_paginator("list_objects_v2")
paginator.paginate(Bucket=bucket, Prefix=prefix, Delimiter="/")
```  

We probably didn't notice the code issue before because all references use this function with the default index value.

### DIFF
--- a/dcpy/utils/s3.py
+++ b/dcpy/utils/s3.py
@@ -351,19 +351,53 @@ def get_filenames(bucket: str, prefix: str) -> set[str]:
 
 
 def get_subfolders(bucket: str, prefix: str, index=1) -> list[str]:
+    """
+    List subfolders in an S3 bucket at a specific depth.
+
+    Parameters:
+    bucket (str): The name of the S3 bucket.
+    prefix (str): The prefix (directory path) to search within the bucket.
+    index (int): The depth level of subfolders to list (default is 1 for top-level subfolders).
+
+    Returns:
+    list[str]: A list of subfolder names at the specified depth.
+
+    Examples:
+    >>> # Given the following objects in the S3 bucket:
+    >>> # - 'root/folder1/file1.txt'
+    >>> # - 'root/folder1/folder2/file2.txt'
+    >>> # - 'root/folder3/'
+    >>>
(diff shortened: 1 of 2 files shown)


dcpy/utils/s3.py
```diff
@@ -351,19 +351,53 @@ def get_filenames(bucket: str, prefix: str) -> set[str]:
...
+            if len(relative_parts) >= index:
+                partial_path = Path(*relative_parts[:index])
+                subfolders.add(str(partial_path))
+
     except Exception as exc:
-        print(f"get_subfolders(bucket={bucket}, prefix={prefix}) failed")
+        print(f"get_subfolders(bucket={bucket}, prefix={prefix}, index={index}) failed")
         raise exc
+
     return list(subfolders)
```
damonmcc, line 401: I think the order of strings in the list will be non-deterministic because it's starting as a set. maybe we should do `return sorted(list(subfolders))`

sf-dcp: Good point! Will fix now

damonmcc (approved):

PR #991, opened 2024-07-09
replace `\x1a` with `'` in facdb source data

### LINKED ISSUE #500: Unknown Character in FacDB output
# What:
When reviewing different versions of FacDB outputs (23v2 from Bytes, 24v1 draft), @jackrosacker noticed that the total count of records differed between csv and geodabase formats. Specifically, there were fewer records in geodabase. After further digging, Jack found that some records in csv output included a weird character which caused ArcGIS to omit all records after its first occurrence: 
* Opened via VSCode: 
   Screenshot 2024-01-09 at 2 18 33 PM
* Opened via Excel: 
   MicrosoftTeams-image
* Opened in ArcGIS: 
MicrosoftTeams-image (1). 

As seen in the last 2 screenshots, ArcGIS shows a total of 5,935 records, stopping at the first occurrence of the unknown character. 

# Next steps: 
* Investigate all data products that may have a similar issue

damonmcc:
resolves #500 

successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/9864520948) (temporarily used changes from #992 because FacDB is broken on main)

### query showing no records with the bad character:

Screenshot 2024-07-09 at 5 48 44 PM

### DIFF
--- a/products/facilities/facdb/sql/dohmh_daycare.sql
+++ b/products/facilities/facdb/sql/dohmh_daycare.sql
@@ -28,11 +28,13 @@ _dohmh_daycare_tmp AS(
     SELECT
         uid,
         source,
-        (CASE
-            WHEN center_name LIKE '%SBCC%' THEN initcap(legal_name)
-            WHEN center_name LIKE '%SCHOOL BASED CHILD CARE%' THEN initcap(legal_name)
-            ELSE initcap(center_name)
-        END) as facname,
+        REGEXP_REPLACE(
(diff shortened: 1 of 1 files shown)


damonmcc:
this branch incorporates changes from https://github.com/NYCPlanning/data-engineering/pull/992 so that the build will work

**if this PR is approved first, I'll rebase to ensure only these changes are merged**

sf-dcp (approved):

PR #993, opened 2024-07-09
Modify ZTL build logic, github action to use new dbt and qa code

### LINKED ISSUE #927: DBT ZTL: Modify ZTL build logic, github action to use new dbt and qa code
The final step!

Following #925, #926, we must put it into operation. This will include
- [x] deleting old sql files. Just delete the `sql` folder entirely. Also delete the reference to this folder in the sqlfluff section of `.github/workflows/test_helper.yml`
- [x] modifying bash files to use dbt commands instead
- [x] modifying export commands as needed. We need to recreate the same exports that currently exist. This should largely be taken care of - our main product table and qa tables, but could potentially need a tweak
- [x] updating github action yml file if necessary

HengJiang0206:
resolves #927 

Outputs uploaded to `db-zoningtaxlots/draft/hj-ztl-action` with ACL "public-read"

Latest Action #1174 https://github.com/NYCPlanning/data-engineering/actions/runs/9893044604

#### `hj-ztl-action.int__zoningtaxlots` VS `main.dcp_zoning_taxlot`

[qa_freq_changes.csv](https://github.com/user-attachments/files/16196757/qa_freq_changes.csv)
- 2 new `commercial_overlay` count 

[qaqc_bbl.csv](https://github.com/user-attachments/files/16196758/qaqc_bbl.csv) 
- 0 added, 0 removed 

[qaqc_bbldiffs.csv](https://github.com/user-attachments/files/16196759/qaqc_bbldiffs.csv) 
- 60 different rows (58 rows of `special_district`s being in different orders, 2 rows of additional `special_district_2` slight change in threshold resulting from calculation syntax changes)

[qaqc_frequency.csv](https://github.com/user-attachments/files/16196760/qaqc_frequency.csv) 

[qaqc_mismatch.csv](https://github.com/user-attachments/files/16196761/qaqc_mismatch.csv) 
- 4 `tax_lot`, 58 `special_district_1` and 58 `special_district_2` mismatches 

[qaqc_null.csv](https://github.com/user-attachments/files/16196762/qaqc_null.csv)
- 2 null to values for `commercial_overlay_2`

[qc_versioncomparison.csv](https://github.com/user-attachments/files/16196763/qc_versioncomparison.csv)
- 58 `special_district_1` and 58 `special_district_2` mismatches


.github/workflows/zoningtaxlots_build.yml
```diff
@@ -65,8 +69,8 @@ jobs:
       - name: Dataloading
         run: python -m dcpy.lifecycle.builds.load load
 
-      - name: 2. build ...
-        run: ./ztl.sh build
```
fvankrieken, line 69: if we're scrapping calls `ztl.sh` let's also remove it. I'm good with that, it's not much of a wrapper and doesn't add much simplicity

products/zoningtaxlots/bash/export.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+source bash/config.sh
```
fvankrieken, line 2: let's scrap this (and source ../../bash/utils.sh instead) - I see you had temporarily. It's not adding much and we reference `utils` directly in `build.sh` so let's be consistent here

products/zoningtaxlots/bash/export.sh
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Generate csvs and shapefiles ..."
+rm -rf output && mkdir -p output
+(
+    cd output
+
+    cp ../source_data_versions.csv ./
+    cp ../build_metadata.json ./
+    
+    csv_export zoningtaxlots zoningtaxlot_db &
```
fvankrieken, line 13: Let's get rid of the extra newlines in between these `csv_export` calls

products/zoningtaxlots/bash/export.sh
```diff
@@ -0,0 +1,47 @@
...
+
+    csv_export qa_freq qaqc_bbl &
+
+    csv_export qa_freq qaqc_bbl &
+
+    csv_export qa_vers_comparison qc_versioncomparison &
+
+    csv_export qa_null qaqc_null &
+
+    run_sql_command "\copy (
```
fvankrieken, line 29: Seems like this can also be a `csv_export` - if there's a difference between this and the table (column is dropped, etc), let's modify the sql so that a simple `csv_export` gives us the right output

products/zoningtaxlots/bash/export.sh
```diff
@@ -0,0 +1,47 @@
...
+        zmcnew , area, inzonechange , 
+        bblprev, zd1prev, zd2prev, zd3prev, zd4prev, 
+        co1prev, co2prev, sd1prev, sd2prev, sd3prev, 
+        lhdprev, zmnprev, zmcprev
+    FROM ${BUILD_ENGINE_SCHEMA}.qa_bbldiffs
+    ) TO STDOUT DELIMITER ',' CSV HEADER;" > qc_bbldiffs.csv &
+
+    shp_export qa_bbldiffs MULTIPOLYGON -t_srs "EPSG:2263"
+
+    echo "${DATE}" > version.txt
```
fvankrieken, line 42: This should be `"${VERSION}"` now

fvankrieken (changes requested):
A couple tweaks, but almost there! Hope you enjoyed your bash adventures, that's a real engineer rite of passage.

Can you (in the PR description or in a comment) link the latest run action?

products/zoningtaxlots/bash/export.sh
```diff
@@ -0,0 +1,28 @@
...
+    
+    csv_export zoningtaxlots zoningtaxlot_db &
+    csv_export qa_freq qaqc_frequency &
+    csv_export qa_bbl qaqc_bbl &
+    csv_export qa_mismatch qaqc_mismatch &
+    csv_export qa_vers_comparison qc_versioncomparison &
+    csv_export qa_null qaqc_null &
+    csv_export qa_bbldiffs qc_bbldiffs &
+
+    shp_export qa_bbldiffs MULTIPOLYGON -t_srs "EPSG:2263"
```
fvankrieken, line 21: Last little thing - looks like the output typically is named `qc_bbldiffs`. You can do this with `shp_export qa_bbldiffs MULTIPOLYGON -f qc_bbldiffs -t_srs "EPSG:2263"` (just telling you instead of letting you check it out because the bash function is a little funky).

We should probably clean up these filenames (make them consistent, with each other and with the dbt models) but that can wait

HengJiang0206: @HengJiang0206 I made some modifications to the qa model file names so they are consistent with the output file names

fvankrieken: Ah I had wanted to leave the model names. But let's keep this. 

BUT then we need one more change - the renaming in the calls to csv_export and shp_export don't need to happen anymore. So we can just have

`csv_export qa_null &`

for the csv exports, and drop the `-f "qc_bbldiffs"` from the shp_export

fvankrieken:
Can you run a build with the code in it's current state?

HengJiang0206:
> Can you run a build with the code in it's current state?

Just did, link updated to #1174

fvankrieken (approved):
Looks great. Thank you for generating these QA outputs!

PR #996, opened 2024-07-10
add linguist gitattributes

fvankrieken:
Silly little PR

See https://github.com/github-linguist/linguist/blob/master/docs/overrides.md for documentation on these "gitattributes"

This bugs me
image

Especially because when you click on "Jupyter Notebooks", nothing shows up - see [here](https://github.com/search?q=repo%3ANYCPlanning%2Fdata-engineering++language%3A%22Jupyter+Notebook%22&type=code)

We do actually have a single notebook - https://github.com/NYCPlanning/data-engineering/blob/main/products/template/python/map_data.ipynb. We could maybe just get rid of this, but I'd still rather not see ipynbs dominate the language dist for one small file.

I've also added sql as a language which is not included by default, though this ignores the couple GIGANTIC files in `data` folders. We should probably get these into library/recipes (cpdb has a monster one. Not entirely sure at first glance why its needed. It's called "sprints"). 

Now, we see a much more reasonable distribution
image

We could specify HTML as "generated" since we don't actually write any as well

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+*.sql linguist-detectable=true
+# for sql files that show up as "PLpgSQL"
+*.sql linguist-language=SQL
+**/data/** linguist-vendored
+*.ipynb linguist-vendored
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.sql linguist-language=SQL
 **/data/** linguist-vendored
 *.ipynb linguist-vendored
+**/docs/** linguist-documentation


fvankrieken:
We could also add logic to distinguish dbt sql from postgres, though that seems a little silly maybe. Could add yml if we really want to flex our yml engineering chops

fvankrieken:
Without ignoring `data` folders, we're 92% sql

damonmcc:
I'm down to specify HTML as "generated" or ignore it

damonmcc:
just curious, are [these the docs](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) ur using?

fvankrieken:
> I'm down to specify HTML as "generated" or ignore it

To be clear, specifying it as generated will ignore it in the stats

fvankrieken:
> just curious, are [these the docs](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) ur using?

That's linked at the beginning of the PR description!

damonmcc (approved):

fvankrieken:
Seemed more appropriate to mark anything in a `docs` folder as documentation (which also gets ignored in stats). This includes all of our html files now. So if we eventually write some, they'll still get counted.

damonmcc (approved):

damonmcc:
instant change!

Screenshot 2024-07-10 at 3 25 53 PM

sf-dcp:
Love it!

PR #997, opened 2024-07-11
Add proposed metadata changes

alexrichey:
This is the metadata we were looking at it our meeting. There are two files
- `as-is-metadata.yml`: represents current state of one our slimmed down metadata files
- `proposed-metadata.yml`: my restructured metadata, with comments to indicated intent.

Nitpicking encouraged!

### DIFF
--- a/dcpy/test/lifecycle/package/resources/proposed_md/proposed-metadata.yml
+++ b/dcpy/test/lifecycle/package/resources/proposed_md/proposed-metadata.yml
@@ -6,7 +6,8 @@ each_row_is_a: ""
 tags: ["a", "b"]
 
 columns:
-  - name: uid # key? id? or leave it?
+  - id: myreallylongcolumnuid # Use id
+    # TODO: api_name'ish stuff
     display_name: Unique ID
     data_source: Department of City Planning
     description: ""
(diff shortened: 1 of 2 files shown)


dcpy/test/lifecycle/package/resources/proposed_md/proposed-metadata.yml
```diff
@@ -0,0 +1,96 @@
+name: db-colp
+display_name: City Owned and Leased Property (COLP)
+summary: "Raw Summary"
+description: "Raw Description"
+each_row_is_a: ""
+tags: ["a", "b"]
+
+columns:
+  - name: uid # key? id? or leave it?
```
alexrichey, line 9: consensus: id

dcpy/test/lifecycle/package/resources/proposed_md/proposed-metadata.yml
```diff
@@ -0,0 +1,96 @@
...
+    type: socrata
+    custom:
+      four_four: "fn4k-qyk2"
+    files:
+      - id: primary_shapefile
+        overrides:
+          omit_columns: [dcpedited, uid]
+        custom:
+          display_name: "display_name overridden at the destination level"
+          destination_use: dataset_file # better name than type? destination_use
```
alexrichey, line 68: I'm having some trouble naming this concept... but what I'm going for is, suppose we have a CSV in a dataset. That csv might be a `dataset_file` in one destination and an attachment on Socrata. My gut was to use `destination_type` but is that overloading the word `type`?

PR #998, opened 2024-07-12
remove references to $EDM_DATA, unused bash

fvankrieken:
Small one - just wanted to clean up some junk laying around. Most of these were commented out already

Only remaining references to $EDM_DATA are in `ceqr`, we'll need a little project to restructure that product


damonmcc (approved):
🧹

PR #1000, opened 2024-07-12
DBT ZTL Cleanup

### LINKED ISSUE #999: DBT ZTL: Cleanup
Just a few things I missed in review
- final model is "zoningtaxlots-db" - dash should be an underscore. Apparently our bash utils don't catch a non-existent table, simply export it as an empty csv
- int__limited_height and int__inwoodrezoning should pull from int__valid_dtm rather than stg__dof_dtm for consistency
- drop unused dcp_specialpurposesubdistrict both from _sources.yml and recipe.yml

HengJiang0206:
resolves #999 

Latest Run #1191 https://github.com/NYCPlanning/data-engineering/actions/runs/9942452828/job/27463929446


products/zoningtaxlots/models/intermediate/int__inwoodrezooning.sql
```diff
@@ -11,7 +11,7 @@ rezone_bbl AS (
         b.bbl,
         b.id AS dtm_id,
         '1' AS notes
-    FROM dof_dtm AS b, dcp_zoningmapamendments AS c
+    FROM validdtm AS b, dcp_zoningmapamendments AS c
     WHERE
```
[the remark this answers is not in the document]
fvankrieken: JOIN on the intersects.

`geom IS NOT NULL` can be dropped

The rest of the bool values can stay in the `WHERE` clause

products/zoningtaxlots/models/intermediate/int__zoningtaxlots.sql
```diff
@@ -39,33 +39,34 @@ zonechange AS (
 insertion AS (
     SELECT
         a.id AS dtm_id,
-        (CASE
```
fvankrieken, line 42: Ah - you do need to rebuild your dev container to sqlfluff properly, sorry. This isn't using our config, and we differ from standard sqlfluff a bit. In this case. This is an int table so it's not necesssarily the worst thing in the world, but I don't want sqlfluff setting the query order

fvankrieken: The rule here that sqlfluff applies that we ignore is that calculated columns are selected after non-calculated columns. This is an insane rule to me

fvankrieken: So I think the easiest thing to do would be
- rebuild dev container (I believe this can be done solely by clicking the button in the bottom left of vscode to bring up dev container options, then selecting "rebuild container"
- drop the sqlfluff commit (easiest to do with a rebase - I can show you on a call)
- re-run sqlfluff fix and add the new commit

Then we should be good to go

HengJiang0206: @fvankrieken I reverted the previous `sqlfluff fix` and re-ran it with the rebuilt dev container. It looks like some of the fixes I reverted, such as splitting long lines, aren't re-applied by the newly configured sqlfluff. Is that also ignored?

fvankrieken: Yup, we made the line length limit just a tad longer than default

fvankrieken: You can look through `pyproject.toml` for our sqlfluff custom config

PR #1007, opened 2024-07-15
Fix ztl in repeat build

fvankrieken:


.github/workflows/repeat_build.yml
```diff
@@ -20,7 +20,7 @@ on:
           - green_fast_track
           #- knownprojects
           - pluto
-          - ztl
```
fvankrieken, line 23: the `plan` command created in this job uses this string input, so needed to tweak it

damonmcc (approved):
nice. just noting that the only other DE dataset we archive form our publishing bucket is DevDB/HousingDB, but (from a IRL chat today) we'd rather archive from our bucket than from bytes

PR #513, opened 2024-01-12
Move to forked gdal 3.8.3

fvankrieken:
resolves #508 

I forked gdal, branched off of release/3.8 (which seemed to be most up-to-date branch for 3.8.3, a few commits ahead of 3.8.3 tag with some bash cleanup). Added [this commit](https://github.com/fvankrieken/gdal/commit/fe3bd65448a8e15bcdb870d2a15e374c9a6e1bf4), undoing a [change from 3.8.0 to 3.8.1](https://github.com/OSGeo/gdal/compare/v3.8.0...v3.8.1#diff-a13ee344527a8c597d816c3e7527be0a03a58c0a941a0ff397f3065cc5d36caeL1449). It seems like maybe [this](https://github.com/OSGeo/gdal/compare/v3.8.0...v3.8.1#diff-3c2d0a796b98eed8a1f7c6f1fc29d010012c40bd5b4149945ed35751b93e4582R1168) was supposed to catch this case, given the similar comment?

[Here's the troublesome commit](https://github.com/OSGeo/gdal/commit/813cca6d783cda46538778c8d038c59d3e5122c1)

Anyways, this branch now downloads gdal from my forked repo with the fix. I built `base` image off of it, let PR tests here build dev images, then ran dof_shoreline archival from this branch, and we're back in business

https://nyc3.digitaloceanspaces.com/edm-recipes/datasets/dof_shoreline/gdalcustom/dof_shoreline.sql

PR #1014, opened 2024-07-16
WIP/On Hold: upgrade gdal to 3.9.2

### LINKED ISSUE #752: Upgrade GDAL
GDAL is on 3.8.5

We are currently on my fork of 3.8.3, which was forked due to a bug in null date exporting (see my commit [here](https://github.com/fvankrieken/gdal/commit/fe3bd65448a8e15bcdb870d2a15e374c9a6e1bf4))

Need to check if this issue is resolved in 3.8.5, and upgrade if so. Would be nice to be getting patches. So todos are roughly
- [ ] see what's new in 3.8.4, 3.8.5 and if we care about these updates
- [ ] see if null date export issue is fixed
- [ ] potentially either just start using 3.8.5, or update my fork if we still need this manual fix

fvankrieken:
Closes #752 

We moved to my fork of gdal in #513 to take advantage of the new geoparquet driver but fix a bug that was introduced in handling null dates. I verified that the null date issue was resolved in 3.9.0. Now that we have a little more tooling around testing our base docker image, I figured we should get off my fork


fvankrieken:
Seems that there are some breaking changes in the python api

https://github.com/NYCPlanning/data-engineering/actions/runs/9976507340/job/27568859329

Sad that this wasn't caught by `library` tests. Maybe this PR waits until we are off of library and in ingest (where gdal-related functionality is better tested). At a minimum, on hold for now because I don't want to get into the weeds of the gdal python package in `library`

fvankrieken:
Maybe a relatively easy fix - we're not doing this the most accepted way currently

https://gdal.readthedocs.io/en/latest/development/rfc/rfc97_feature_and_fielddefn_sealing.html

PR #1018, opened 2024-07-19
Create streamlit page for "local file" ingestion 

### LINKED ISSUE #1017: Create streamlit page for "local file" library uploads
Some of our library datasets that end up in `edm-recipes` come from a manual upload - someone has emailed us a csv, we've grabbed something via an ftp on NYCNET, etc. Their templates look roughly like [this](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/templates/dcp_pops.yml#L6) for source. For these, we
- download the file from wherever it comes from
- place the file in the correct location on our computer and/or edit path in template yml
- run library locally to archive

This isn't TOO painful, but for a lot of datasets it certainly is tedious to make sure these paths align when something a little more drag-and-drop would make this process much simpler.

This is why a basic streamlit page would be very helpful, basically with
- text input (or selectbox input) to enter dataset name
- text input to enter version
- file upload dialog

We'd then want to
- archive this raw file to `edm-recipes/inbox/{dataset}/{version}/{filename}` so that we have the raw file archived
- call library for the given dataset, but pointing to this file in `inbox` for the `path`. This is probably the trickiest part, because we'll need to figure out the simplest way to feed this into `library` code which is a little opaque at times

So to summarize what needs to be done
- [x] add new page "ingest dataset" to apps/qa
- [x] add inputs for this new page
- [x] write helper function within QA app to upload file to `edm-recipes/inbox/{dataset}/{version}/{filename}`
- [ ] write helper function within QA app to read in `template` file (using `library` utilities) but then replace the `path` with the new s3 path pushed to in the last step
- [ ] call appropriate `library` function using this new "config" object

HengJiang0206:
resolves #1017

This PR creates a page in the QA `streamlit` app to process drag-and-drop local data archives into the `edm-recipe` bucket on S3. 

The interface prompts for dataset name, version, and a local file upload: 

Screenshot 2024-07-23 at 11 47 00

Upon hitting the `ingest` button, the uploaded file will be saved to a temporary local directory `.library` (in `gitignore`). Then the `s3.upload_file` function will be called to upload the file to `"inbox" / dataset_name / version / file_name` in `edm-recipes`. 

Screenshot 2024-07-23 at 11 53 44

Screenshot 2024-07-23 at 11 54 48


fvankrieken:
A few notes for your first python contributions to code - 

Two tools/commands that will be useful/necessary are `black` and `mypy`. `black` is a linter which formats python code - at any time, you can run `black apps/qa/...` (as specific as you want - it's pretty fast though so no need to get too specific) and it'll format python files in the provided path.

Mypy you run similarly, but it's doing type-checking - so if you define a function in one spot and declare that it takes a `str` arg and elsewhere call it with an `int` arg, `mypy` will throw an error. Mypy takes a tad longer to run so it's definitely worth just running on what you need

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -0,0 +1,20 @@
...
+    from pathlib import Path
+
+    st.title("Ingest Dataset")
+
+    dataset_name = st.text_input("Dataset Name")
+    version = st.text_input("Version")
+    uploaded_file = st.file_uploader("Choose a file")
+    s3_path = Path("inbox") / dataset_name / version
+
+    if st.button("Ingest"):
```
fvankrieken, line 13: This does work, but I think I would prefer setting a variable equal to the call to `st.button`. I think it just "flows" a little better. "if button" is a slightly confusing boolean value. Something like `result = st.button()` or `ingest_button_pressed = st...` makes it clear what value is being returned by the input, and helps separate out the __visual__ code of calling `st` components and the more logical/control flow code of checking its boolean value

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -0,0 +1,41 @@
...
+from pathlib import Path
+import streamlit as st
+from dcpy.utils import s3
+
+BUCKET = "edm-recipes"
+
+
+def archive_raw_data(dataset_name: str, version: str, uploaded_file, file_name: str):
+
+    base_path = Path(dataset_name)
```
fvankrieken, line 11: let's maybe make this `Path(".library") / "upload" / dataset_name`? That's a convention we've followed to some extent, `.library` is in our `.gitignore`, and it's nice to not just have things dumped in the current active folder

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -0,0 +1,41 @@
+import shutil
+from pathlib import Path
+import streamlit as st
+from dcpy.utils import s3
+
+BUCKET = "edm-recipes"
+
+
+def archive_raw_data(dataset_name: str, version: str, uploaded_file, file_name: str):
```
fvankrieken, line 9: can you add typing for uploaded_file and provide a return type?

fvankrieken:
Let's test it out?

Given that we're not calling library yet, let's just make up a dummy dataset - call it for dataset "test" with say an empty text file or one line csv

fvankrieken:
I'll test locally to look at the interface, but would you mind putting some screenshots in the PR description? As well as describe the scope of the PR

HengJiang0206:
> I'll test locally to look at the interface, but would you mind putting some screenshots in the PR description? As well as describe the scope of the PR

Just did!

fvankrieken:
This looks great. Getting to this point has me thinking a bit about improving user experience, I have some ideas for how we can iterate (just in terms of keeping the UI clean while reporting the "state" of what's going on. But let's get this squashed and merged, and we can brainstorm a little bit about next steps (outside of the library connection)

PR #1021, opened 2024-07-20
DO restructure: Implement `build` & new version `draft` folders

sf-dcp:
Related to https://github.com/NYCPlanning/data-engineering/issues/825 and https://github.com/NYCPlanning/data-engineering/issues/960.

## Changes 

First 3 commits are of QOL type + existing bugs with comparing bumped versions. The rest of the commits address the following (it's probably easiest to go commit by commit):

- [x] **Add `build` functionality**
    - [x] create `BuildKey`
    - [x] redirect `upload()` fn to `build` dir
- [x] **Implement new `draft`**
    - [x] modify `DraftKey` model
    - [x] implement versioning for `DraftKey` via `dcpy.utils.versions.DraftVersionRevision` (ex: `1-initial`)
    - [x] create `promote_to_draft` fn and CLI
    - [x] GH action to promote to draft
- [x] **Fix downstream code impacted by new `build` & `draft`**
    - [x] fix `publish()` fn 
    - [x] fix `download_file` CLI 
    - [x] fix `repeat_build` CLI & GH action
    - [x] automatic promotion of ZTL when runninng `Zoning Tax lots - 📁 DataLoading` GH actiton
    - [x] appease mypy
    - [x] fix streamlit

Tasks outside of this PR: 
- s3: move current drafts to `build` folder and promote to `draft` via new GH action. Delete old drafts.
- s3 clean up GH action: delete builds 
-  enhance code setup for `test_publishing.py` by adding more flexibility
-  increase test coverage for `publishing.py`

## Tests

<details><summary><h3>Build action</h3></summary>

- Successful templateDB build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10125279741)
    image
    image
</details>

<details><summary><h3>Promote to Draft action</h3></summary>

- Successful promotion (first one) [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10125298534/job/28000616806)
- Successful promotion (second one) [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10125321543/job/28000665958)
    image
</details>

<details><summary><h3>Publish action</h3></summary>

1. When running this action with the following setup and the following draft folder: 
    image
    image

    * Successful publishing of latest (2nd) draft when draft number is **not** provided [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10126053035/job/28002261165#step:7:20)
    image

    * As expected, `latest` folder doesn't get updated because it has a later version (`24Q3.1`) [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10146746310/job/28055534461#step:7:79)    
    image

2. When running publish action with draft revision number, it grabs the correct [draft number](https://github.com/NYCPlanning/data-engineering/actions/runs/10146922673/job/28056129960#step:7:19): 
image

image

</details>
    
<details><summary><h3>ZTL Data loading action (automatic promotion of ZTL)</h3></summary>

- ZTL being promoted to draft when ZTL being built: [action](https://github.com/NYCPlanning/data-engineering/actions/runs/10124989828/job/28000060900)
    image

- no promotion when ZTL isn't being built: [action](https://github.com/NYCPlanning/data-engineering/actions/runs/10125109841)
</details>

<details><summary><h3>Repeat Product Build action</h3></summary>

- Successfully repeating `build` product type with the setup below [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10147450837/job/28057854615):  

    image
    image

    Result: 

    image

- Successfully repeating `draft` product type with the setup below [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10147604669/job/28058393371):  

    image

    image

    Result:
    
    image

</details>

If you want to test any of the revised GH actions, use my branch `sf-s3-restructuring` as the action input.

<details><summary><h3>Streamlit page (draft)</h3></summary>

`draft` folder looks like this:
image

Selecting a draft version:
image

Selecting a draft revision:
image

Result:
image

</details>


.github/workflows/repeat_build.yml
```diff
@@ -27,9 +27,10 @@ on:
         options:
           - publish
           - draft
+          - build
```
fvankrieken, line 30: Ooh question on draft - would we want to supply the full string for the draft revision here? As in `24v1/1`? I think that's probably simplest. Sorry if I just haven't gotten to where you deal with this

.github/workflows/promote_to_draft.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Set version and build sha
+        run: |
+          echo "version=$(cat build_metadata.json | jq -r .version)" >> $GITHUB_ENV
+          echo "commit=$(cat build_metadata.json | jq -r .commit)" >> $GITHUB_ENV
+
+      - name: create tag
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GHP_TOKEN }}
```
fvankrieken, line 93: Maybe you ran into issue because perhaps a standard action token can _create_ a tag but not _overwrite_ one? If the tag you were creating existed?

That being said, I'm not 100% sure we need to tag when promoting to draft. I guess maybe there's not much downside, but I think maybe keeping tags specifically for datasets that we've published might make more sense

.github/workflows/repeat_build.yml
sf-dcp: I think it might be annoying for consecutive drafts... Say something like `24v1/3-fix-zoning-data-and-update-sql-logic`. It shouldn't be too bad to have a separate logic for drafts: select version and then select draft

sf-dcp:
I did a thing. I noticed that one of the existing files `dcpy.test.lifecycle.builds.build.py` appeared to be incorrectly named, so I fixed it in my second commit by renaming to `test_plan.py`. 

A bunch of tests started failing even though the tested code wasn't impacted by my code changes. See [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10148260881/job/28060604549) or go to failed tests under `fix mypy/pytest errors` commit. Once I reverted the file name back, tests started passing again. 

My theory is that this file was never tested before since its name didn't start with word "test".

dcpy/test/utils/test_versions.py
```diff
@@ -132,3 +134,32 @@ def test_bumping_versions(self):
             [None, 2, "23Q4.1", "24Q2"],
         ]:
             self.assertEqual(v_expected, versions.bump(v, bumped_part, bump_by).label)
+
+    def test_parse_draft_version_valid_versions(self):
```
damonmcc, line 138: love it

.github/workflows/zoningtaxlots_dataloading.yml
```diff
@@ -11,6 +11,11 @@ on:
         description: "Build Pluto (minor) after dataloading is complete"
         type: boolean
         default: false
+      promote_ztl:
+        description: "Promote  Zoning Tax Lots to Draft folder after build?"
```
damonmcc, line 15: nit: `Promote  Zoning` should be `Promote Zoning`

damonmcc (approved):
no notes. great work, testing, and screenshots!

.github/workflows/zoningtaxlots_dataloading.yml
fvankrieken: I also think this should default to `true` - just one less click for GIS

apps/qa/src/shared/components/sidebar.py
fvankrieken, line 1: Nice

dcpy/connectors/edm/publishing.py
```diff
@@ -200,50 +228,104 @@ def legacy_upload(
...
+
+    logger.info(
+        f"Promoted {build_key.product} to drafts as {version}/{draft_revision_label}"
+    )
+
+    if not keep_build:
+        s3.delete(BUCKET, source)
+
+
+def determine_publish_version(
```
fvankrieken, line 276: Bit of a nit, but just pondering the name of this function. It is determining, but I wouldn't call it the most intuitive name. Like maybe `validate_or_patch_version` or something like that? Not that we should go with that specifically, just food for thought

.github/workflows/publish.yml
```diff
@@ -71,31 +75,42 @@ jobs:
         working-directory: ./
         run: ./bash/docker_container_setup.sh
 
-      - name: Download Metadata
+      - name: Determine Published Version
         run: |
-          python3 -m dcpy.connectors.edm.publishing download_file \
-            --filepath build_metadata.json \
+          published_version=$(python3 -m dcpy.connectors.edm.publishing determine_publish_version \
```
fvankrieken, line 80: I don't love this - I know that you you wrote this function as part of this PR so you know what went into it, but if anything else gets printed by python during a command like this (logging, etc) it ends up in the variable

fvankrieken: Maybe the publish cli command could have a flag to save the metadata locally? That would sort of make sense, since that's where we potentially "lose" the version (when we patch it at least), so that would be a sensible time to make a local record of where the draft is being pushed to. And then we can just grab the version from the saved metadata rather than havin the additional call to download it

fvankrieken: Curious on your thoughts

fvankrieken:
Two minor notes and one slightly more open question. Looks great

sf-dcp:
> I did a thing. I noticed that one of the existing files `dcpy.test.lifecycle.builds.build.py` appeared to be incorrectly named, so I fixed it in my second commit by renaming to `test_plan.py`.
> 
> A bunch of tests started failing even though the tested code wasn't impacted by my code changes. See [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10148260881/job/28060604549) or go to failed tests under `fix mypy/pytest errors` commit. Once I reverted the file name back, tests started passing again.
> 
> My theory is that this file was never tested before since its name didn't start with word "test".

@damonmcc @fvankrieken thoughts what to do with this file outside of the PR?

fvankrieken:
> > I did a thing. I noticed that one of the existing files `dcpy.test.lifecycle.builds.build.py` appeared to be incorrectly named, so I fixed it in my second commit by renaming to `test_plan.py`.
> > A bunch of tests started failing even though the tested code wasn't impacted by my code changes. See [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10148260881/job/28060604549) or go to failed tests under `fix mypy/pytest errors` commit. Once I reverted the file name back, tests started passing again.
> > My theory is that this file was never tested before since its name didn't start with word "test".
> 
> @damonmcc @fvankrieken thoughts what to do with this file outside of the PR?

Want to make an issue for it? We should fix this

.github/workflows/publish.yml
alexrichey: yeah, I think I steered @sf-dcp a little wrong here. Writing to a file (either metadata) or just a textfile specifically for this seems like the way to go.

dcpy/connectors/edm/publishing.py
```diff
@@ -57,10 +58,37 @@ def get_published_versions(product: str) -> list[str]:
...
+
+def get_draft_revision_label(product: str, version: str, revision_num: int) -> str:
+    """Given a draft revision number, return draft revision label in s3."""
+    draft_revision_label = None
+    draft_revision_objects = [
+        versions.parse_draft_version(version)
+        for version in get_draft_version_revisions(product, version)
+    ]
+    for obj in draft_revision_objects:
+        if obj.revision_num == revision_num:
```
alexrichey, line 79: just FYI, you can do this filtering in the list comprehension above. Ie
```
    draft_revision_objects = [
        versions.parse_draft_version(version)
        for version in get_draft_version_revisions(product, version)
        if versions.parse_draft_version(version).revision_num == revision_num
    ]
```

dcpy/utils/versions.py
```diff
@@ -67,10 +67,11 @@ def label(self) -> str:
     def __lt__(self, other) -> bool:
         match other:
             case MajorMinor():
-                return (self.year, self.major, self.minor) < (
+                return (self.year, self.major, self.minor, self.patch) < (
```
alexrichey, line 70: nit: if we just add @dataclass(order=True), we can turn this into `return self < other`. I think.

dcpy/utils/versions.py
```diff
@@ -125,7 +127,11 @@ def label(self) -> str:
     def __lt__(self, other) -> bool:
         match other:
             case Date():
-                return (self.date, self.format) < (other.date, other.format)
+                return (self.date, self.format, self.patch) < (
```
alexrichey, line 130: question/nit: why is the format included in the ordering?

alexrichey: not strictly about this PR, but glancing through... it looks like there's one case where we allow comparisons between classes that aren't the same, and that case (Date < MajorMinor) is unintuitive. I think we could chop about ~100 lines and a lot of complexity by just relying on dataclass sort orders. Thoughts? cc @fvankrieken @damonmcc

fvankrieken: To the nit - just to have something deterministic in the case of a "tie". Maybe a bit silly but I figured might as well

To the other question - so that code doesn't break when we moved FacDB from 2023-XX-XX to 24v1. I agree we generally don't want to be making those sorts of comparisons, but occasionally we do when we're changing versions to better align. I think any changes to this should happen more in context of our versioning discussion and outside of Sasha's PR

dcpy/utils/versions.py
fvankrieken: Seems a little odd to use the function that's being overriden within here? Doesn't < make a call to __lt__ for these classes? Just seemed cleaner to not use `self <` in `self.__lt__`. though to be honest not entirely sure the way these things are handled under the hood

dcpy/utils/versions.py
```diff
@@ -306,3 +316,47 @@ def bump(
...
+            return f"{self.revision_num}"
+        else:
+            return f"{self.revision_num}-{self.revision_summary}"
+
+
+def parse_draft_version(v: str) -> DraftVersionRevision:
+    """Takes a version string and attempts to parse it into DraftVersionRevision object."""
+    split_index = v.find("-")
+
+    if split_index != -1:
```
alexrichey, line 350: nit: if the intent is just to not "oversplit" (e.g. in the case of "1-my-dashy-draft-version", you can just add a max_split, e.g. v.split("-", 1). Then this becomes:
```
num_str, *maybe_summary = v.split("-", 1)
summary = maybe_summary[0] if len(maybe_summary) == 1 else ""
return DraftVersionRevision(int(num_str), summary)
```

alexrichey (approved):
Looks good! A few non-blocking nits, conversations

Issue #1016, opened 2024-07-17
Rework PFF data flow


fvankrieken:
PFF needs some reworking in its data flow, which presents some questions for us both in terms of how we approach what datasets get pulled into builds and what preprocessing we support in `library`/`ingest`

I'm going to stay specific to ACS - decennial is another dataset with essentially the same constraints

In its original flow (2023 edition), PFF was maybe fairly simple, largely because the data sources were simple. Pop sent us an excel file with four tabs (housing, econ, social, dem). This file was checked into the repo

```mermaid
flowchart LR

subgraph ide1 [PFF Build]

direction LR

acs.xlsx--pd.read_df-->housing
acs.xlsx--pd.read_df-->econ
acs.xlsx--pd.read_df-->social
acs.xlsx--pd.read_df-->dem

housing--pivot-->housing_pivoted
econ--pivot-->econ_pivoted
social--pivot-->social_pivoted
dem--pivot-->dem_pivoted

housing_pivoted--concat-->pff_acs
econ_pivoted--concat-->pff_acs
social_pivoted--concat-->pff_acs
dem_pivoted--concat-->pff_acs

end
```

Then, we had the issue of ONE of these four datasets being updated, but not the other three. At that point, we made a fairly necessary tweak to separate these into four `recipe` datasets. This makes sense given that for the rest of recipes, each template/dataset/entry is in fact a dataset, and not a collection of them

```mermaid
flowchart LR

subgraph ide1 [edm-recipes]

direction LR

acs.xlsx--library_archive-->housing
acs.xlsx--library_archive-->econ
acs.xlsx--library_archive-->social
acs.xlsx--library_archive-->dem

end

subgraph ide2 [PFF Build]

direction LR

housing-->housing_df
econ-->econ_df
social-->social_df
dem-->dem_df

housing_df--pivot-->housing_pivoted
econ_df--pivot-->econ_pivoted
social_df--pivot-->social_pivoted
dem_df--pivot-->dem_pivoted

housing_pivoted--concat-->pff_acs
econ_pivoted--concat-->pff_acs
social_pivoted--concat-->pff_acs
dem_pivoted--concat-->pff_acs

end
```

This still take relatively manual archiving of the source datasets, but at least it's simple to archive them separately. 

Now though, we have a few more use cases
[...]

PR #1023, opened 2024-07-22
Ingest preprocessing "modes" for ACS inputs

fvankrieken:
Couple main things here

## data util to upsert columns in a df
I could imagine many cases where we want to upsert a column, with a variety of cases. This assumes that we want to upsert based on some key. So not just blindly `df["column"] = my_series_to_upsert` but rather taking in an "upsert_df" where we have keys, say `geotype` and `geoid` where we want to update values in our original df, so ensuring that geotype=CCD2023 and geoid=01 gets the right updated value as opposed to assuming order is consistent. I wanted this especially in cases where there's not a guarantee of the same number of rows.

There are a bunch of flags for this - how to handle "new" columns (i.e. only allow update rather than inserts as well), how to handle keys missing in the upsert df provided, etc. I think it's thoroughly tested

## "modes" for preprocessing in ingest
We have the case of PFF of having different types of dataset provided that necessitate different modes of updating. See #1016 for more context
- provide a brand new dataset (standard - no interaction with previous versions of dataset)
- provide additional rows (i.e. a new geotype - CCD2023 in this case, for 2010 acs data. In this case, we append to prev)
- provide updated columns (what the latest update is, hence the creation of the upsert util above)

To handle this, I've added a field to the processing step definitions in yml, they can have `mode` defined. Any steps without a `mode` are **always** run, while steps with `mode` defined are only run when the entire process is run with said `mode` (provided when running ingest from command line, option `-m` or `--mode`.

So for example, `python3 -m dcpy.cli lifecycle ingest dcp_pop_acs2010_social` will run ingest, looking for the dataset source as it usually does, and fairly naively process and dump it, without running any processing steps that have a `mode`. But running `python3 -m dcpy.cli lifecycle ingest dcp_pop_acs2010_social -m append` will instead append it to the previous, and running `python3 -m dcpy.cli lifecycle ingest dcp_pop_acs2010_social -m update_column` will run the step to update columns of the previous dataset

## Operational steps

Done as part of updating PFF with latest 2010 ACS data addendum

From my comp, I archived all 4 2010 acs datasets with mode `append` (which was previously being taken care of with that "manual" addendum logic in pff code) and then also ran housing with `update` to update the two problematic columns

[Passing ingest gha](https://github.com/NYCPlanning/data-engineering/actions/runs/10100406894)

[Passing acs build](https://github.com/NYCPlanning/data-engineering/actions/runs/10100183743)

I've also done some qa to ensure that only the desired changes occurred - read csvs into tables in duckdb, then looking at row counts as well as any missing combo of geoid and var
```python
con.sql("""
    select *
    from new_2 as new
    left join old 
        on new.labs_geotype = old.labs_geotype
        and new.labs_geoid = old.labs_geoid
        and new.pff_variable = old.pff_variable
    where new.labs_geotype is null
""")
```
and changes in values
```python
con.sql("""
    select new.labs_geotype, new.labs_geoid, old.m, new.m
    from new 
    left join old 
        on new.labs_geotype = old.labs_geotype
        and new.labs_geoid = old.labs_geoid
        and new.pff_variable = old.pff_variable
    where old.labs_geotype <> 'fb2000ltr'
    and old.m <> new.m
""")
```

### DIFF
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -64,13 +64,17 @@ def archive_dataset(config: ingest.Config, file_path: Path):
     _archive_dataset(config, file_path, config.dataset_key.s3_path("ingest_datasets"))
 
 
-def get_config(name: str, version="latest") -> library.Config:
+def get_config(name: str, version="latest") -> library.Config | ingest.Config:
     """Retrieve a recipe config from s3."""
     obj = s3.client().get_object(
         Bucket=BUCKET, Key=f"{DATASET_FOLDER}/{name}/{version}/config.json"
     )
     file_content = str(obj["Body"].read(), "utf-8")
-    return library.Config(**yaml.safe_load(file_content))
+    config = yaml.safe_load(file_content)
+    if "dataset" in config:
+        return library.Config(**config)
+    else:
+        return ingest.Config(**config)
 
 
 def get_latest_version(name: str) -> str:
@@ -258,9 +262,14 @@ def get_archival_metadata(
         version = get_latest_version(name)
     logger.info(f"looking up metadata for {name}/{version}")
     config = get_config(name, version)
-    if config.execution_details:
-        timestamp = config.execution_details.timestamp
-        runner = config.execution_details.runner_string
+    match config:
+        case library.Config():
+            execution_details = config.execution_details
+        case ingest.Config():
+            execution_details = config.run_details
+    if execution_details:
+        timestamp = execution_details.timestamp
+        runner = execution_details.runner_string
     else:
         s3metadata = s3.get_metadata(
             BUCKET, f"{DATASET_FOLDER}/{name}/{version}/config.json"
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -38,7 +38,7 @@ def _archive_dataset(config: ingest.Config, file_path: Path, s3_path: Path):
         tmp_dir_path = Path(tmp_dir)
         shutil.copy(file_path, tmp_dir_path)
         with open(tmp_dir_path / "config.json", "w") as f:
-            f.write(json.dumps(config.model_dump(), indent=4))
+            f.write(json.dumps(config.model_dump(exclude_none=True), indent=4))
         s3.upload_folder(
             BUCKET,
             tmp_dir_path,
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -56,12 +56,20 @@ def archive_raw_dataset(config: ingest.Config, file_path: Path):
     _archive_dataset(config, file_path, config.raw_dataset_key.s3_path(RAW_FOLDER))
 
 
-def archive_dataset(config: ingest.Config, file_path: Path):
+def archive_dataset(config: ingest.Config, file_path: Path, latest: bool = False):
     """
     Given a config and a path to a processed parquet file, archive it in edm-recipes
     Unique identifier of a raw dataset is its name and its version
     """
-    _archive_dataset(config, file_path, config.dataset_key.s3_path("ingest_datasets"))
+    s3_path = config.dataset_key.s3_path("ingest_datasets")
+    _archive_dataset(config, file_path, s3_path)
+    if latest:
+        s3.copy_folder(
+            BUCKET,
+            f"{s3_path}/",
+            f"ingest_datasets/{config.name}/latest/",
+            acl=config.acl,
+        )
 
 
 def get_config(name: str, version="latest") -> library.Config | ingest.Config:
--- a/dcpy/lifecycle/ingest/configure.py
+++ b/dcpy/lifecycle/ingest/configure.py
@@ -11,7 +11,7 @@ from dcpy.models.lifecycle.ingest import (
     S3Source,
     ScriptSource,
     Source,
-    FunctionCall,
+    PreprocessingStep,
     Template,
     Config,
 )
@@ -103,7 +103,9 @@ def get_filename(source: Source, ds_name: str) -> str:
             )
 
 
-def get_config(dataset: str, version: str | None = None) -> Config:
+def get_config(
+    dataset: str, version: str | None = None, mode: str | None = None
+) -> Config:
     """Generate config object for dataset and optional version"""
     run_details = metadata.get_run_details()
     template = read_template(dataset, version=version)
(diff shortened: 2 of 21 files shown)


dcpy/lifecycle/ingest/transform.py
```diff
@@ -267,4 +290,8 @@ def preprocess(
 
         for step in compiled_steps:
             df = step(df)
+
+        if output_csv:
```
fvankrieken, line 294: This just seemed like a nice little QOL for troubleshooting - cli flag to output a csv at end of processing

fvankrieken: I kept having errors at time of parquet generation due to pandas weirdness

dcpy/utils/data.py
```diff
@@ -11,6 +12,21 @@
 import zipfile
 
 
+def _get_dtype(dtype: str | dict | None) -> str | dict | defaultdict | None:
```
fvankrieken, line 15: didn't end up using the defaultdict feature outside of a little testing but this seemed nice. When specifying a csv you can do
```yml
file_format:
    type: csv
    dtype:
        geoid: str
        __default__: float
```
To specify dtype for one column and a default for the rest

products/factfinder/pipelines/utils.py
```diff
@@ -142,6 +142,12 @@ def pivot_table_with_suffixes(
         raise Exception(f"Some field names missing certain suffixes: {field_names}")
 
     expected_columns = set(pivot_columns + suffix_columns)
+    for c in df.columns:
```
fvankrieken, line 145: Anything being printed here means an error will be thrown in that `assert` below

alexrichey (approved):
Cool stuff!

dcpy/lifecycle/ingest/configure.py
```diff
@@ -113,11 +115,18 @@ def get_config(dataset: str, version: str | None = None) -> Config:
...
     if template.target_crs:
-        reprojection = FunctionCall(
+        reprojection = PreprocessingStep(
             name="reproject", args={"target_crs": template.target_crs}
         )
         processing_steps = [reprojection] + processing_steps
 
+    if mode:
+        modes = {s.mode for s in processing_steps}
+        if mode not in modes:
```
fvankrieken, line 125: worth noting here that filtering processing steps happen at time of config generation. Meaning, the config.json that gets dumped with a dataset ONLY has the steps that were actually run.

writing this now, it might be worth keeping `mode` as a saved field in the config? In that case, we could include all preprocessing steps (and it can be infered by `mode` which one was run), though I sort of like the fact that steps are only included if they were run. But then it doesn't quite reflect the definition in the original template anymore. @alexrichey would be curious if you have any strong feelings

alexrichey: @fvankrieken Including all the fields + mode seems sensible. Or you could add a flag called `skipped`, or something like, that to each step that was omitted?

dcpy/utils/data.py
sf-dcp: Given your code snippet above, geoid column will be casted to a `string` and the rest will be casted to a `float`, correct?

dcpy/utils/data.py
```diff
@@ -179,3 +179,78 @@ def serialize_value(value):
...
+    ):
+        raise ValueError("Cannot upsert: upsert_df keys are not unique")
+
+    if insert_behavior == "error" and not upsert_cols.issubset(cols):
+        raise ValueError(
+            f"Unexpected columns found in upsert_df: {upsert_cols.difference(cols)}"
+        )
+    if insert_behavior == "ignore":
+        upsert_df = upsert_df[
+            [c for c in upsert_df.columns if c in on or c in df.columns]
```
sf-dcp: I think `if c in on` is not needed here because you check this condition in the beginning of the fn

sf-dcp:
Could you please explain how the mode logic works when you have steps with multiple modes for a given recipe?

Using your example: 
```yaml
name: dcp_pop_acs2010_demographic
 acl: public-read
 source:
   type: local_file
   path: .library/upload/CCD2023_ACS0610Data_for1822Update.xlsx
 file_format:
   type: xlsx
   sheet_name: CCD2023_Dem0610
   dtype:
     GeoID: str
 processing_steps:
 - name: clean_column_names
   args:
     lower: true
 - name: append_prev
   mode: append
 - name: upsert_column_of_previous_version
   args:
     key: [geotype, geoid]
     insert_behavior: error
     missing_key_behavior: error
   mode: update_column
```

Would running `python3 -m dcpy.cli lifecycle ingest dcp_pop_acs2010_social -m append` skip `upsert_column_of_previous_version ` step?

dcpy/connectors/edm/recipes.py
```diff
@@ -64,13 +64,17 @@ def archive_dataset(config: ingest.Config, file_path: Path):
     _archive_dataset(config, file_path, config.dataset_key.s3_path("ingest_datasets"))
 
 
-def get_config(name: str, version="latest") -> library.Config:
+def get_config(name: str, version="latest") -> library.Config | ingest.Config:
```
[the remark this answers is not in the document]
fvankrieken: It would error out in that case, since these models forbid extra arguments

dcpy/utils/data.py
fvankrieken: Precisely!

fvankrieken:
> Could you please explain how the mode logic works when you have steps with multiple modes for a given recipe?
> 
> Using your example:
> 
> ```yaml
> name: dcp_pop_acs2010_demographic
>  acl: public-read
>  source:
>    type: local_file
>    path: .library/upload/CCD2023_ACS0610Data_for1822Update.xlsx
>  file_format:
>    type: xlsx
>    sheet_name: CCD2023_Dem0610
>    dtype:
>      GeoID: str
>  processing_steps:
>  - name: clean_column_names
>    args:
>      lower: true
>  - name: append_prev
>    mode: append
>  - name: upsert_column_of_previous_version
>    args:
>      key: [geotype, geoid]
>      insert_behavior: error
>      missing_key_behavior: error
>    mode: update_column
> ```
> 
> Would running `python3 -m dcpy.cli lifecycle ingest dcp_pop_acs2010_social -m append` skip `upsert_column_of_previous_version ` step?

Yeah, as mentioned above

> Any steps without a mode are always run, while steps with mode defined are only run when the entire process is run with said mode (provided when running ingest from command line, option -m or --mode.

So for this yml template, `upsert_column_of_previous_version` is only defined for mode `update`, and when run in mode `append` it will not be run. So that file can be run in 3 modes
- no mode: only `clean_column_names` run
- append: `clean_column_names`, `append_prev`
- update_column: `clean_column_names`, `upsert_column_of_previous_version`

dcpy/connectors/edm/recipes.py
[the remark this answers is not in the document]
sf-dcp: perfect

sf-dcp:
Gotcha. So these modes are exclusive (only 1 will run at a time). LGTM!

Side note: the code is quite self explanatory + docstrings make a huge difference. It would be nice to add documentation on yaml syntax for new ingest code in the future (similar to DO docs on their various yaml files)

sf-dcp (approved):

PR #1024, opened 2024-07-23
Add proposed product.metadata_v2 class

alexrichey:
This defines the updated (v2) Metadata format for `Product.Dataset.Metadata` and adds an upgrade method to the existing class to go from v1->v2. (or really, v0 to v1). 

Here's a [Draft PR](https://github.com/NYCPlanning/product-metadata/pull/8/files?diff=split&w=0) in the product metadata-repo with the result of running this upgrade on all files. 

The plan in my mind is:
- We review this code, make any changes required
  - After merge, refresh/regenerate files in the linked PR
- Update dependencies on the existing metadata format. (mostly the Socrata Publishing Connector)
- Merge the linked PR
- Delete old models, all references to v1, v2, etc.

There is a lot here - __Let's chat about how best to review on Monday before folks dive in__. In particular, the logic in `Metadata.upgrade_to_v2` is a little gnarly, and we'll just end up deleting it, while time might be better spent spot-checking the transformed metadata files to ensure that information isn't lost. 

#### Issues (Add any new ones here)

- [ ] Description fields are outputting with extra lines and special characters (e.g. `comma\u2013\
    separated values (CSV)` in the PLUTO metadata


dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,100 @@
...
+class PackageFile(CustomizableBase):
+    id: str
+    filename: str
+
+
+class Package(CustomizableBase):
+    id: str
+    type: str
+    filename: str
+    contains: List[PackageFile]
```
fvankrieken, line 46: nit - maybe `contents` instead of `contains`? Unless there's a specific reason you prefer a verb

alexrichey: yeah, `contents` feels nice.

dcpy/models/product/dataset/metadata_v2.py
fvankrieken, line 1: looks like the concept of somethin like "SocrataMetadata" or "SocrataDestination" are
1. generalized slightly
2. handled with "custom" fields? Not sure if something like a "four_four" really belongs there as opposed to bein an expected value for something going to Socrata (would rather fail at validation time rather than run time)

alexrichey: yeah, this was for the purposes of generalization and extensibility. I want the ability to add new Destination types without editing these models. I've got some ideas for that, but would be beneficial to chat/brainstorm how to generalize these without losing type safety entirely. (e.g. prior, we had a `SocrataDestination` that required a four-four, and now we just have a dictionary) I'm sure there's a tolerable middle-ground.

PR #1025, opened 2024-07-23
fix: UP GH per-page for branches/pulls

alexrichey:
Fixes the failing github get-branches test. I don't think that realistically we should hit this limit, unless our team grows massively and our branch hygiene deteriorates.


fvankrieken (approved):
Ha. Yeah this seems like a good case to just take the easy way out for now!

PR #1026, opened 2024-07-23
Product metadata format transformation

### LINKED ISSUE #1013: Product metadata format transformation
Explore packages and approaches to transform metadata formats from YAML to PDF, potentially with custom formatting options (fonts and mages). This could involve a mix of packages (readme/text builder or other combinations), from intermediate Python objects to PDF. 

Criteria for package selection 
- Functionality (do what we need)
- Decent user base and well-regarded
- Good documentation 

Once there is a shortlist of options, compare them based on ease of use, reference code snippets, and examples of their API.   

Sample Metadata: https://github.com/NYCPlanning/product-metadata/blob/main/products/pluto/pluto/metadata.yml

HengJiang0206:
resolves #1013

The goal of this PR is to automate the process of transforming YAML metadata into formatted PDFs. 

Step 1: Transform the YAML file into an HTML file using Jinja templating. 
- The YAML file is parsed using utils from `dcpy.models.product.dataset.metadata`. 
- It is then rendered into an output HTML using the Jinja template `metadata_template.html`

Step 2: Transform the metadata HTML into PDF using Pandoc 
- Install the `Pandoc` package and customize the output by setting fonts, line width, page margins, etc.
- Run Pandoc commands to input the HTML and output a PDF. 

<del> How to run: 
<del>- cd to `products/metadata`
 Assume the input YAML file and the HTML template are saved under the current directory, and we want to output the rendered HTML and the PDF to the current directory: 
<del>-  Run `python3 save_html.py <PATH TO YAML> <PATH TO HTML TEMPLATE> <PATH TO OUTPUT HTML>`. Example: `python3 save_html.py metadata.yml metadata_template.html output.html`
<del>- After installing Pandoc packages, run `pandoc output.html -o output.pdf -V geometry:margin=1.5cm`
<del>- The above command contains an optional arg to set the PDF page margin to 1.5 cm. Ideally, we could use CSS files or YAML headers to set up consistent PDF templates. </del>

***functions moved to `dcpy/lifecycle/package` and added Typer 

Sample Input: 
https://github.com/NYCPlanning/product-metadata/blob/main/products/pluto/pluto/metadata.yml
Sample Output: 
[output.pdf](https://github.com/user-attachments/files/16383179/output.pdf)


products/metadata/metadata_template.html
```diff
@@ -0,0 +1,239 @@
...
+    <ul>
+        {% for tag in metadata.tags %}
+        <li>{{ tag }}</li>
+        {% endfor %}
+    </ul>
+
+    <h2><strong>Each row is a: </strong> </h2>
+    <p>{{ metadata.each_row_is_a }}</p>
+
+    <h2><strong>Destinations: </strong> </h2>
```
alexrichey, line 24: Let's remove destination info for now. It might be interested to include something like that in the future, but I'll have to think about it.

products/metadata/metadata_template.html
```diff
@@ -0,0 +1,239 @@
...
+                        <ul>
+                            {% for tag in destination.overrides.tags %}
+                            <li>{{ tag }}</li>
+                            {% endfor %}
+                        </ul>
+        {% endif %} </li>
+        {% endfor %}
+    </ul>
+
+    <h2><strong>Package:</strong></h2>
```
alexrichey, line 143: let's also leave out package info

products/metadata/metadata_template.html
alexrichey, line 1: Let's move this to dcpy/lifecycle/package/resources/data_dictionary_template.jinja (new folder)

products/metadata/save_html.py
alexrichey, line 1: Let's move this to dcpy/lifecycle/package/generate_metadata_assets.py (or something like that. I'm open to a better filename)

products/metadata/save_html.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_metadata(yaml_file_path):
+    metadata = Metadata.from_path(
+        Path(yaml_file_path), template_vars={"var1": "value1"}
+    )
+    return metadata
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Convert a YAML file to a HTML file.")
```
alexrichey, line 29: When you move this into dcpy, you'll notice that we use Typer to handle CLI commands. Would be nice to convert that that.

alexrichey (changes requested):
Great Start! Let's move this into dcpy and add Typer for commands, but then we should be good to merge and iterate.

alexrichey (approved):
LGTM!

dcpy/lifecycle/package/generate_metadata_assets.py
```diff
@@ -0,0 +1,29 @@
...
+import argparse
+from pathlib import Path
+from jinja2 import Environment, FileSystemLoader
+import typer
+
+app = typer.Typer()
+
+
+@app.command()
+def generate_metadata_assets(yaml_file_path, html_template_path, output_html_path):
```
sf-dcp, line 11: I wood add types in fn signature here and in the remaining functions. 

For example, this fn signature would look like this: 

```python
def generate_metadata_assets(yaml_file_path: str, html_template_path: str, output_html_path: str) -> Path:
...
```

HengJiang0206: Quick question: For the get_metadata function that returns the SocrataMetada, do I just say object? Do I need to add the SocrataMetada class into the module?

dcpy/lifecycle/package/generate_metadata_assets.py
```diff
@@ -0,0 +1,31 @@
+from dcpy.models.product.dataset.metadata import Metadata, SocrataMetada
```
fvankrieken, line 1: Hakuna metada (@alexrichey I assume the Metada class will be retired or renamed by your PR?)

sf-dcp (approved):
Good job!

dcpy/lifecycle/package/generate_metadata_assets.py
alexrichey: Ha, yes, something like that. It'll probably just be moved to a DCP-specific Socrata model class. Sort of an extension of the base metadata

PR #1028, opened 2024-07-25
[STUB]: `promote_to_draft.yml` GH workflow

sf-dcp:
Need to create a stub of new workflow on `main` in order to be able to test on my other dev brunch.

### DIFF
--- a/.github/workflows/promote_to_draft.yml
+++ b/.github/workflows/promote_to_draft.yml
@@ -1,0 +1,36 @@
+name: 🛫  Promote a Dataset to Draft
+run-name: "🛫 Promote a Dataset to Draft: ${{ inputs.product }}"
+
+on:
+  workflow_dispatch:
+    inputs:
+      product:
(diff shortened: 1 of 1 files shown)


.github/workflows/promote_to_draft.yml
fvankrieken, line 1: I think without a job it might fail too... Give it a shot though. Sorry!!

fvankrieken: Oh great I was wrong. I think what I did wrong once was a stub that had a job but no steps, which prevented anything from working even on a dev branch. Or something like that. Carry on!

fvankrieken: Oh of course now I remember - non-dev branches fail on push (or other dev branches with this commit from main)

https://github.com/NYCPlanning/data-engineering/actions/runs/10098654226

Not terrible, just annoying 🙂

sf-dcp: oh wow. Good to know!

PR #1029, opened 2024-07-26
Ingest streamlit page - introduce state

### LINKED ISSUE #1027: Ingest streamlit page - introduce state
Using the ingest streamlit page, there's room to improve user experience, specifically when thinking about the "state" of the page.

Logically, we might have a couple states, or things that we want to do, or results we should see
- start: text input, file selection. Everything up till hitting the button
- raw archival - the app is actively uploading to s3. We shouldn't be allowed to interact with the page while this is happening
- archival failed - pop up error, page unlocks
- archival succeeded - library logic begins. The user should be identified, and page should remain locked
- library fails - open question to what should happen here, since we're in an intermediate state.
- library succeeds - maybe a pop-up message, click to dismiss and "reset" page to empty inputs

The "library fails" step is an interesting one to think about. It wouldn't be crazy to say that maybe we also want the ability to JUST run library pointed at a specific file in s3. Maybe we have a selectbox on the left side bar, choosing whether we want to upload a file or just run library aimed at a custom s3 path. Then, when library fails, we can auto-populate the page so that it could be retried (or a user could push code changes, and rerun this step supplying the s3 path as text input).

This is a bit complicated - there are a bunch of ways to track state in a streamlit app, and I'll leave it up to choice as exactly how. But essentially we will need to track what part of the process the page is in, and render different views depending on the state.

HengJiang0206:
resolves #1027

*** The pushed code uses dummy ingest and library functions to test failures. 
- When input dataset as `error`, ingestion fails 
- when input version as  `error`, library fails 

Use the drop-down sidebar to select a process:
Screenshot 2024-07-31 at 11 56 32

Ingest Success:
Screenshot 2024-07-31 at 11 59 36

Ingest Fails:
Screenshot 2024-07-31 at 11 59 59

Library Success:
Screenshot 2024-07-31 at 12 00 31

Library Fails:
Screenshot 2024-07-31 at 12 00 49

Ingest and Library Success:
Screenshot 2024-07-31 at 12 01 35

Library Fails after Successful Ingest:
Screenshot 2024-07-31 at 12 02 24

Retry Library Calls (Pressed button `Call Library with File s3 Path`):
`s3 Path` auto-filled:
Screenshot 2024-07-31 at 12 03 52

Retry Success:
Screenshot 2024-07-31 at 12 04 41

Retry Fails - Can continue to retry unless click on `End Retry`: 
Screenshot 2024-07-31 at 12 07 14

`End Retry`: Resets Process 

*** Update: Add validation for dataset name - dropdown selection 
Screenshot 2024-08-02 at 10 58 32


apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,41 @@ def ingest():
...
+    disabled = st.session_state.get("running", False)
+    ingest_button_pressed = st.button("Ingest", disabled=st.session_state.running)
     if ingest_button_pressed == True:
         if dataset_name and version and uploaded_file:
-            file_name = uploaded_file.name
-            file_path = archive_raw_data(
-                dataset_name, version, uploaded_file, file_name
-            )
+            st.session_state.running = True
+            st.rerun()
```
fvankrieken, line 22: So I don't think we should need to use `rerun`. I'm guessing you drew inspiration from the GRU page, but that's an unusual case in that it needs to track status of external orchestration - it's refreshing to see what the status of actions are on github. We _should_ (I think) be able to get by without any `sleep` or `rerun`. At least for this initial part. In general in the streamlit model, when a widget is interacted with, the screen automatically reloads. So the button press is all we need.

For our button, we might want to use an `on_click` function rather than just treating it as a bool - this way, we can set state, and this state can be used at the beginning to determine what we show.

Let's also get more specific with state - one, let's use notation `key["value"]` rather than `key.value` just because that's generally the way we've gone in our codebase. But let's have a `st.session_state["ingest"]` as a dict in which we can store these, to make sure this page wouldn't interfere with another page.

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,41 @@ def ingest():
...
         else:
             st.warning("Please input all fields.")
+
+    if st.session_state.running == True:
```
fvankrieken, line 26: For this (and other cases), you can drop the ` == True`! Just `if st.session_state["running"]`

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,41 @@ def ingest():
...
+        with st.spinner("Ingesting"):
+            time.sleep(5)
```
fvankrieken, line 28: In this case, we definitely don't want to sleep here - it's sleeping before we actually run the archival code

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,41 @@ def ingest():
...
+                )
+            except Exception as e:
+                st.error("Ingestion Failed: {e}")
+                time.sleep(5)
+                st.session_state.running = False
+                st.rerun()
+        st.session_state.running = False
+        st.success("Ingest Successful")
+        time.sleep(5)
+        st.rerun()
```
fvankrieken, line 42: This case we might still want to rerun mainly in the case where we also have the library call). But again, we don't need any `sleep`s

apps/qa/src/pages/ingest/ingest.py
HengJiang0206: Got it. Since I have the reruns, the pop-up messages disappear right away, so I figured to put it to sleep for a few seconds for the message to be displayed longer 😂. I'll try to get rid of reruns. I was referencing here
https://docs.streamlit.io/develop/api-reference/status/st.spinner
https://docs.streamlit.io/develop/api-reference/execution-flow/st.rerun

fvankrieken: Gotcha. I think in those cases, we could use state too though - `if state["archive_status"] == "failed":` then display a failure message, or something like that

fvankrieken: And in that first link, they're really just using `sleep` as an example for a function that takes some time!

fvankrieken:
Also - don't worry about the weird failed jobs on pushes - that'll be resolved by @sf-dcp 's pr

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,81 @@ def ingest():
...
     from pathlib import Path
+    import time
 
-    st.title("Ingest Dataset")
-
-    dataset_name = st.text_input("Dataset Name")
-    version = st.text_input("Version")
-    uploaded_file = st.file_uploader("Choose a file")
-    s3_path = Path("inbox") / dataset_name / version
+    if "ingest" not in st.session_state:
```
fvankrieken, line 7: Nice - I like having an `init` of the model/state.

fvankrieken: Though I don't think things like dataset_name, version, s3_path, or uploaded_file need to be in here - we really want to be capturing things that need to be tracked between resets that aren't handled by a simple input in `session_state`, but if we can get by with just the return value of an input we should stick with that

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -2,20 +2,81 @@ def ingest():
...
+        st.session_state["ingest"]["dataset_name"]
+        and st.session_state["ingest"]["version"]
+    ):
+        st.session_state["ingest"]["s3_path"] = (
+            Path("inbox")
+            / st.session_state["ingest"]["dataset_name"]
+            / st.session_state["ingest"]["version"]
+        )
+    if (
+        st.button(
```
fvankrieken, line 71: Again, let's keep the call to the component out of the `if` clause, just for clarity - keep python logic and calls to render/view functions a little separate

fvankrieken:
This is looking pretty good. I'm going to think a little more about how we maybe can make things look a little clearer (in terms of just the layout of the code), but I'm really happy with this.

Can you go ahead and
1. keep the call to the dummy archive function for now
2. add logic to call the next ingest step, as outlined in #1027 ?

sf-dcp:
I would recommend adding validation logic for `dataset_name` (ex: replace text field with a dropdown menu) and maybe `version`

HengJiang0206:
## Validation Logic for Input Fields 

Following up on @sf-dcp's recommendation, validation logics are likely needed for the input fields (dataset name and version). This could hardly be handled by a dropdown menu, given the very large list of raw datasets we would process. I'm wondering if we should include backend validation logic and if they should be part of this PR? @fvankrieken

fvankrieken:
There's a handy function dcpy.library.utils.get_all_templates() which returns a list of all dataset names.

I would go ahead and try a dropdown for the dataset name (`selectbox` in streamlit I think) - streamlit dropdowns are text-searchable. If it feels way too unwieldy, then we could just validate the input in the text box.

Version is a little more complicated, and I'm not sure every version format we use follows the most consistent of formats - I would leave unvalidated for now.

fvankrieken:
Small request - can the dropdown default to s3 + library?

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -1,21 +1,206 @@
...
+            "library_status": None,
+            "error_message": None,
+            "retry": False,
+            "s3_path": None,
+        }
 
-    ingest_button_pressed = st.button("Ingest")
-    if ingest_button_pressed == True:
+    def lock_for_ingest():
         if dataset_name and version and uploaded_file:
```
fvankrieken, line 18: not sure how mypy isn't erroring here - these aren't defined at this point, so this is relying on these variables happening to exist. They should be provided as arguments to this function (and similar functions below)

fvankrieken: Can you add type signatures for the functions as well? And return as well (even though it's just `-> None`)

fvankrieken: Ah that's why mypy wasn't catching this - mypy doesn't lint untyped functions

fvankrieken: In that case, add a type signature for the top-level function as well - this is a pattern we should probably enforce for other pages, but that's obviously out of scope here

fvankrieken (approved):

PR #1036, opened 2024-08-01
improve reporting of `pytest` coverage

damonmcc:
## changes

- upload `dcpy` test coverage report to DCP's Codecov account
  - only happens when `dcpy` tests are run
- updates badge in the README:
  - make Nightly QA badge link to the action runs
  - add badge for test coverage percentage on `main` (it won't have valid results until this is merged to and tests are run on `main`)

## outcomes

PRs will now have:
- an auto-generated comment related to code coverage ([docs](https://docs.codecov.com/docs/pull-request-comments))
- status checks related to code coverage ([docs](https://docs.codecov.com/docs/commit-status))

Codecov report for this branch is [here](https://app.codecov.io/gh/NYCPlanning/data-engineering/tree/dm-test-coverage-report) (you should be prompted to login via github):

Screenshot 2024-08-05 at 8 47 10 PM

## background

we've found it's difficult to have confidence in code changes (especially to `dcpy`) without visibility of test coverage of the changes and/or code related to the changes

the two most popular/established options for code coverage reporting are [Codecov](https://about.codecov.io/) and [Coveralls](https://coveralls.io/). both seem great and are free for our type of usage

I chose Codecov because DCP has a history of using it (Labs aka AE), it seems more lightweight, and seems more user-friendly. if we ever want to switch to Coveralls, it would be easy to

`dcpy` tests are run on `main` as part of Nightly QA and on all PRs with changes to [these files](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/test.yml#L31-L35)

### DIFF
--- a/.github/workflows/data/pytest.yml
+++ b/.github/workflows/data/pytest.yml
@@ -1,5 +1,5 @@
 - name: dcpy
-  calls: ["dcpy/test -s --verbose --verbose --cov-config=pyproject.toml --cov=dcpy"]
+  calls: ["dcpy/test -s --verbose --verbose --cov-config=pyproject.toml --cov=dcpy --cov-report=xml"]
 - name: checkbook
   db: db-checkbook
   calls: ["./products/checkbook -s --verbose --verbose"]
--- a/.github/workflows/test_helper.yml
+++ b/.github/workflows/test_helper.yml
@@ -167,6 +167,7 @@ jobs:
         export-env: true
       env:
         OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+        CODECOV_TOKEN: "op://Data Engineering/Codecov/CODECOV_TOKEN"
         AWS_S3_ENDPOINT: "op://Data Engineering/DO_keys/AWS_S3_ENDPOINT"
         AWS_SECRET_ACCESS_KEY: "op://Data Engineering/DO_keys/AWS_SECRET_ACCESS_KEY"
         AWS_ACCESS_KEY_ID: "op://Data Engineering/DO_keys/AWS_ACCESS_KEY_ID"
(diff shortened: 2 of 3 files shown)


.github/workflows/test_helper.yml
```diff
@@ -165,6 +167,7 @@ jobs:
         export-env: true
       env:
         OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+        CODECOV_TOKEN: "op://Data Engineering/Codecov/CODECOV_TOKEN"
```
damonmcc, line 170: I got this token from our org account in Codecov and added it to our 1password vault

.github/workflows/data/pytest.yml
```diff
@@ -1,5 +1,5 @@
 - name: dcpy
-  calls: ["dcpy/test -s --verbose --verbose --cov-config=pyproject.toml --cov=dcpy"]
+  calls: ["dcpy/test -s --verbose --verbose --cov-config=pyproject.toml --cov=dcpy --cov-report=xml"]
```
damonmcc, line 2: this change creates `./coverge.xml` after the tests are run. if we expand coverage reporting to other tests (like the QA app), we can use more a specific filename or directory to upload multiple reports

sf-dcp:
This is so cool. I love that you are able to see percentage coverage of your modified/new code on PRs like below:

image

Q: will every push to a branch be generating codecov reports? If that's true and I push new commits (say 10-20), it will spam the comments section...

sf-dcp (approved):

.github/workflows/data/pytest.yml
damonmcc: and xml isn't the only format (html locally is cool), but it seems to be the default format for uploading to Codecov

alexrichey (approved):
cool! thanks

.github/workflows/test_helper.yml
```diff
@@ -129,8 +129,10 @@ jobs:
         run: |
           echo "path filters with changed files: ${{ inputs.path_filter }}"
           if [[ -z "${{ inputs.path_filter }}" ]]; then
+            echo "No files changed"
```
fvankrieken, line 132: Can we actually make this more specific? Maybe

"No path filters provided, running all pytest tests"

Could use some language other than "path filter". Maybe even "No file changes provided" - just to specify that this case occurs when the detect file change step isn't even run (and then passed in to this job in workflow call)

PR #1038, opened 2024-08-01
DevDB 24Q2 build and enhancements

fvankrieken:
I do not want to add coverage for this one fixed line of a library script. I was thinking about having codecov ignore library script sources, but they really should keep our coverage down until we move over to ingest (and have any scripts there tested).

Commits are atomic - just some general maintenance first and then commits relating to adding parking spaces. Happy to chat about anything.

Build run today [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10526534508)

We're still waiting on DOB data, but I'm going to get this to Housing for review of the parking space data


dcpy/library/templates/dob_cofos.yml
```diff
@@ -4,6 +4,9 @@ dataset:
   source:
     script:
       path: .library/upload/dob_cofos.csv
+      ## This field must be updated manually when archiving. Should eventually be refactored out of here
+      ## Do not check in a specific version, leave blank version check in - this has led to issues in the past
+      previous_version: 20240105
```
fvankrieken, line 9: Seemed to make more sense to have this be an input to the script rather than tucked away in "info"

dcpy/library/templates/dob_jobapplications_parkingspaces.yml
```diff
@@ -0,0 +1,55 @@
...
+        )
+      )
+      SELECT 
+        jobnumber,
+        dobrundate,
+        enclosed_spaces,
+        existing_parking_spaces,
+        proposed_parking_spaces
+      FROM numbered
+      WHERE row_id = 1
```
fvankrieken, line 46: This script is largely copied from the template `dob_jobapplications` - this is basically the same dataset, just with additional columns for parking. I figured it doesn't make sense to store it all, given that `jobapplications` is on opendata and theoretically could have updates (follow ups to old bis jobs), so will just keep this as a separate dataset with just keys (jobnum, rundate) and parking space info

.github/workflows/build.yml
damonmcc, line 1: nice. to align with the folder name?

damonmcc (approved):

.github/workflows/build.yml
fvankrieken: yup

PR #1040, opened 2024-08-01
Pandoc implementation for metadata pdf 

### LINKED ISSUE #1013: Product metadata format transformation

HengJiang0206:
resolves #1013

1. Add a new function into `dopy/lifecycle/package/generate_metadata_assets`, which transforms yaml into pdf. 
2. Add a sample pdf output metadata yaml file, which could be used to config the format of the output pdf. For more instructions, read [https://pandoc.org/chunkedhtml-demo/6.2-variables.html#variables-for-latex](url)

## Sample pdf outputs: 
standard 
[output.pdf](https://github.com/user-attachments/files/16461921/output.pdf)
Landscape layout with two columns
[output.pdf](https://github.com/user-attachments/files/16461969/output.pdf)

### ***To Install Pandoc 
With the container running, enter from your local terminal as root host: 
Follow the instructions to install pandoc on Linux 
[https://github.com/jgm/pandoc/blob/main/INSTALL.md#linux](url)

We also need to install Latex and Font packages:
`apt-get install -y texlive`
`apt-get install -y texlive-fonts-extra`

***TBD
The files are messy at this point, and the function requires too many args, which are probably not necessary. For example, for 

def generate_pdf_from_yml(
    yaml_file_path: str,
    html_template_path: str,
    output_html_path: str,
    output_pdf_path: str,
    pdf_metadata_path,
) -> Path:

We could omit one or two arguments if we have fixed directories for those intermediate HTML or templates.


alexrichey (approved):
Cool! I think this a good place check this in, and regroup about next steps.

HengJiang0206:
*** CSS vs. YAML
It seems like we could use a css to control elements such as font and margins, but the pdf metadata yaml is best capable of handling more advanced things like table of contents, hyperlinks, etc.. We could pass in both the CSS and YAML though.

damonmcc (approved):
lgtm!

it'd be nice to add the install stuff to (probably) [`DEV_APT_PACKAGES`](https://github.com/NYCPlanning/data-engineering/blob/main/docker/config.sh#L6), but we can wait until we try out pandoc more

alexrichey:
@HengJiang0206 Let's merge this, and call it good on any further Data Dictionary work. For this week, we'll likely be turning our focus to READMEs, which we'll discuss at our meeting later today. Thanks!

PR #1047, opened 2024-08-06
Fix DevDB QA endpoint, run tests on pushes to main

### LINKED ISSUE #1045: Scheduled Action Failure - ⚙️ Nightly QA
[Failed action](https://github.com/NYCPlanning/data-engineering/actions/runs/10260590945)

fvankrieken:
fixes #1045 - this logic should really be reworked ideally to not be grabbing a raw file in bash (from latest folder) but this will at least get nightly_qa running. Will take care of slightly better in #1038 

Building [here](https://github.com/NYCPlanning/data-engineering/actions/runs/10274030552)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,8 +5,9 @@ on:
   workflow_dispatch:
   workflow_call:
   pull_request:
-    branches:
-      - main
+    branches: [main]
+  push:
+    branches: [main]
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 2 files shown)


fvankrieken:
Added a commit to run tests on pushes to main. Kinda handy in case we somehow introduce breaking changes but also to have codecov run after a pr merges

fvankrieken:
Just rebase since starting that build

damonmcc (approved):
I was gonna say let's drop running tests during the Nightly QA job, but maybe it's worth
1. the redundancy in case we don't push to `main` in a few days
2. running tests at the same time as all the builds (a build and test failing at the same time would speed up fixing them)

PR #1050, opened 2024-08-07
Add path override option to library

fvankrieken:
For work on the streamlit ingest page, we need the ability to feed in a `path` to grab a dataset from. This can be done for datasets with a `url` source

Writing this now, I debated if this would better be covered by simply editing relevant templates to come from `path: s3://edm-recipes/inbox/{{ name }}/{{ version }}/filename.csv`, then no override is needed. I do think in principle it makes sense to have path to dataset be a parameterized thing - if a filename changes between times that its emailed to us, I'm fine with preserving the original filename (and not have to manually edit it - that was sort of the whole point of this exercise). 

So I do think in principle this is the right approach, even if it's a bit clunky in `library`. Though maybe for these templates we'd want to essentially make it a required arg, rather than a possible override. Anyways, very open to feedback.

### DIFF
--- a/dcpy/library/archive.py
+++ b/dcpy/library/archive.py
@@ -24,7 +24,7 @@ class Archive:
         clean: bool = False,
         latest: bool = False,
         name: str | None = None,
-        path_override: str | None = None,
+        source_path_override: str | None = None,
         *args,
         **kwargs,
     ) -> Config:
@@ -38,10 +38,11 @@ class Archive:
         push: if `True` then push to s3
         clean: if `True`, the temporary files created under `.library` will be removed
         latest: if `True` then tag this current version we are processing to be the `latest`
+        name: name of the dataset, if you would like to use templates already included in this package
+        source_path_override: if applicable, a str url/path to override the source path in the template
 
         Optional Parameters
         ----------
-        name: name of the dataset, if you would like to use templates already included in this package
         compress: if compression is needed, this is passed into the `ingestor`
         inplace: if compressed zip file will replace the original file, this is passed into the `ingestor`
         postgres_url: Please specify if `output_format=='postgres'`
@@ -92,7 +93,7 @@ class Archive:
 
         # Initiate ingestion
         output_files, config = ingestor_of_format(
-            path, *args, path_override=path_override, **kwargs  # type: ignore
+            path, *args, source_path_override=source_path_override, **kwargs  # type: ignore
         )
         version = config.dataset.version
         acl = config.dataset.acl
--- a/dcpy/library/archive.py
+++ b/dcpy/library/archive.py
@@ -24,6 +24,7 @@ class Archive:
         clean: bool = False,
         latest: bool = False,
         name: str | None = None,
+        path_override: str | None = None,
         *args,
         **kwargs,
     ) -> Config:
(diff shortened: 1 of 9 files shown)


damonmcc (commented):
> I debated if this would better be covered by simply editing relevant templates to come from `path: s3://edm-recipes/inbox/{{ name }}/{{ version }}/filename.csv`

in this case, the upload from the strealit page would always go to `edm-recipes/inbox/`? sorry if that's already the case and I didn't realize

fvankrieken:
I guess a little more context would be useful on the streamlit page - 

Really, it does two things
- takes a dataset name, a supplied version string, and a raw file. and dumps it in `edm-inbox/{dataset}/{version}/{filename}` (this function call returns the s3 path)
- calls library to archive this dataset

Theoretically, we could have a file locally on the streamlit server and make it match the "path" given for these manual uploads. I would much rather the config file saved in recipes is actually reflective of reality - just point at the raw file in s3 so that it's actually meaningful (pointing at something that exists and is persisted rather than a log of where something happened to be on the comp that ran it.

dcpy/library/archive.py
```diff
@@ -24,6 +24,7 @@ def __call__(
         clean: bool = False,
         latest: bool = False,
         name: str | None = None,
+        path_override: str | None = None,
```
alexrichey, line 27: do you want to add this to the function docstring below? And is it worth specifying that this is a `source_path_override`, to distinguish from the `path` arg above?

alexrichey: and just to check that I understand, `path` is where the config file lives, and this `path_override` sets the location for where the dataset is?

fvankrieken: Yes to all of the above

alexrichey (approved):
Looks good! Had the one question/suggestion, but otherwise 👍

dcpy/test/library/test_config.py
```diff
@@ -53,3 +57,32 @@ def test_backwards_compatility_with_jinja_version():
...
+
+def test_url_with_override_path():
+    config = Config(get_config_file("url"), path_override=FAKE_PATH)
+    dataset = config.compute
+    assert dataset.source.url
+    assert dataset.source.url.path == FAKE_PATH
+    assert dataset.source.gdalpath == f"{FAKE_PATH}/{dataset.source.url.subpath}"
+
+
+@patch("dcpy.library.script.bpl_libraries.Scriptor.runner")
```
sf-dcp, line 77: I'm quite intrigued by using `patch`. Any rec how to better understand its scope?

sf-dcp (approved):
LGTM!

dcpy/test/library/test_config.py
fvankrieken: patch is my new favorite `mock` function.

It uses MagicMock, but the MagicMock that existed in this file before actually affected other tests in unrelated files (edm-recipes tests were failing because their calls were mocked)

This annotation (and then argument expected for the test function is equivalent to
```
def test_script_with_override_path():
    with patch("dcpy.library.script.bpl_libraries.Scriptor.runner") as runner:
        runner.return_value = FAKE_PATH
        config = Config(get_config_file("bpl_libraries_sql"), path_override=FAKE_PATH)
        dataset = config.compute
        assert dataset.source.script
        assert dataset.source.script.path == FAKE_PATH
        assert dataset.source.gdalpath == FAKE_PATH
```

So the scope is just within this test. Within that `with` block, all calls to the patched function are mocked

fvankrieken: There's some argument to using the latter form - a little clearer what the function is rather than having this "assumed" argument for the test

sf-dcp: thanks for the example! Makes sense. Do you know if you can use patch with pytest parametrize feature?

fvankrieken: I would assume so, just keep in mind that annotations sort of go "inside-out". So if you had

```python
@patch(myfunc)
@pytest.mark.paramaterize("a, b, c", [...])
test_something(a, b, c, myfunc):
    ...
```

the order is important - paramaterization is going to be the first wrapper, so its dedicated args go first, and then after that the parametrized functions are patched, so the patch implied argument goes second. If you patched and parametrized in opposite order, it'd be `myfunc, a, b, c` for args.

Issue #1004, opened 2024-07-15
GFT 24v3


damonmcc:
### Product Name

OTHER

### Build Version

24v3

### Status of Update

- [ ] source data extracted
- [ ] draft build succeeded
- [ ] draft build passed QA
- [ ] data packaged and distributed

jackrosacker:
Hi! Wondering if you all have a goal date or timeline for these updates so we can plan around them. Generally speaking, we're ready when the data is, but have some app updates we're coordinating around this as well.

fvankrieken:
### Pluto
- [ ] dcp_mappluto_wi

- [ ] dcp_boroboundaries_wi
- [ ] dcp_lion

### "ceqr app" data
- [ ] dep_cats_permits
- [ ] nysdec_state_facility_permits
- [ ] nysdec_title_v_facility_permits

### ArcGIS Feature Service
These can have version programmatically determined, so maybe should be pulled on weekly basis

- [ ] set up recurring job

Datasets here are 
- [ ] dcp_cscl_commonplace
- [ ] dcp_cscl_complex
- [ ] dcp_waterfront_access_map_wpaa
- [ ] dcp_waterfront_access_map_pow
- [ ] dcp_wrp_recognized_ecological_complexes
- [ ] dcp_wrp_special_natural_waterfront_areas
- [ ] nysdec_freshwater_wetlands_checkzones
- [ ] nysdec_freshwater_wetlands
- [ ] nysdec_tidal_wetlands
- [ ] nysdec_priority_lakes
- [ ] nysdec_priority_estuaries
- [ ] nysdec_priority_streams
- [ ] nysdec_natural_heritage_communities
- [ ] nysparks_historicplaces_esri
- [ ] nysshpo_historic_buildings_points
- [ ] nysshpo_historic_buildings_polygons
- [ ] nysshpo_archaeological_buffer_areas
- [ ] nysparks_parks_polygons
- [ ] usnps_parks    

### Socrata
add these to weekly socrata pull if they're not there already

- [ ] dpr_forever_wild
- [ ] lpc_scenic_landmarks
- [ ] lpc_historic_district_areas
- [ ] lpc_landmarks
- [ ] dpr_parksproperties
- [ ] dpr_schoolyard_to_playgrounds
- [ ] dcp_edesignation_csv

### Script source
- [ ] usfws_nyc_wetlands

### Manual updates
- [ ] dcp_air_quality_vent_towers
- [ ] dcm_arterial_highways
[...]

Issue #923, opened 2024-06-17
Operationalizing GFT


fvankrieken:
Not finished yet, will edit and add more

A couple parts to this, for now going to focus on process for keeping source data up-to-date.

## Source Data

Unfortunately ingest is not rolled out yet. Maybe part of operationalizing gft will be moving from library -> ingest for the source data. But for now, it's easiest to think about how we'll update these based on either their ingest source type or their current gha/way we update them

### Pluto
- dcp_mappluto_wi - self-explanatory

### Bytes quarterly updates
These are pretty straightforward. Lion has one issue, its parquet doesn't get created without error at the moment, so I've run manually when it's needed
- dcp_boroboundaries_wi
- dcp_lion

### "ceqr app" data
Each of these is something else a bit under the hood. Needs some investigation. For now, we can "build" ceqr data 

- dep_cats_permits
- nysdec_state_facility_permits
- nysdec_title_v_facility_permits

### ArcGIS Feature Service
These can have version programmatically determined, so maybe should be pulled on weekly basis

- [ ] set up recurring job

- dcp_cscl_commonplace
- dcp_cscl_complex
- nysdec_freshwater_wetlands_checkzones
- nysdec_freshwater_wetlands
- nysdec_tidal_wetlands
- nysdec_priority_lakes
- nysdec_priority_estuaries
- nysdec_priority_streams
- nysdec_natural_heritage_communities
- 

### Bytes - unknown frequency of update

- [ ] dcp_wrp_rec
- [ ] dcp_wrp_snwa

### Socrata
add these to weekly socrata pull if they're not there already

- [ ] dpr_forever_wild
- [ ] lpc_scenic_landmarks
- [ ] lpc_historic_district_areas
- [ ] lpc_landmarks

### Script source
- [ ] usfws_nyc_wetlands - need to investigate update frequency

### Manual updates
For each of these, we need to figure out both update frequency and if we think that we maybe can pull it ourselves instead
- [ ] dcp_air_quality_vent_towers
- [ ] dcm_arterial_highways
- [ ] panynj_jfk_65db
- [ ] panynj_lga_65db
- [ ] dcp_beaches
- [ ] dob_natural_resource_check_flags

## Versioning
- [ ] add logic to dcpy plan to determine version of product from one of the sources (in this case, pluto)

## Cleanup
[...]

PR #1054, opened 2024-08-08
Add flexibility to ArcGIS Feature Services connector

fvankrieken:
Part of #1004, and touches a bit on #923 

Thought updating ArcGIS Feature Service datasets for GFT was going to be simple. Had a bit of a wrench thrown in when [multiple attempts failed](https://github.com/NYCPlanning/data-engineering/actions/runs/10308316965)

I broke out the commits so that the first commit has a lot of the test data (big part of the added lines in this PR), so commit-by-commit will be easier.

## Context

As a refresher for terms
- **Feature Service** is what this general platform is called
- a **Feature Server** is essentially a dataset collection (a data product if you will), which generally supports a map view
- a **Feature Server Layer** is a classic ESRI layer - one dataset

So our library datasets refer to a single layer of a given Feature Server - in some cases, we have a couple datasets per feature server. We identified these in our templates with a `layer_id` (or assumed it to be `0`.) Turns out layer ids aren't necessarily constant for a given feature server. I think ideally they are supposed to be, but for a handful of layers, the maintainers seemed to delete the previous layer and simply create a new one (with a new id).

Given that, I thought it made sense to
1. not specify id when possible, and grab the only layer there (and throw an error if there's an unexpected layer on the server). Seems to be cleanest for these Feature Servers that are tightly coupled to a single dataset
2. specify name otherwise, and attempt to look up layer by name.

There's a strong argument that generally, it's safer to specify id rather than name... but I don't know if that's something we can rely on here. I would hope so, but it's hard to tell.

## Summary of changes

Separated out logic of models a little bit. Before, we just had a not quite accurately named `FeatureServer` object, shared across library and the connector. I separated things out so we have
- FeatureServer (not just a layer)
- FeatureServerLayer (has name and id, and could contain more metadata if we want)
- FeatureServerLayerDefinition (library-specific - optional name and id)

Separating these out made the different utilities a lot cleaner - not having to worry about if the layer object had its id and name resolved yet.

Beyond that, added and expanded FeatureService utils and added tests, which hadn't existed before

## Test run

[Passing run of calling library for all datasets](https://github.com/NYCPlanning/data-engineering/actions/runs/10325253970)


fvankrieken:
Just rebased to fix clunky language in a docstring, no code changes since the linked action running all of these

dcpy/library/templates/dcp_waterfront_access_map_pow.yml
fvankrieken, line 1: These I looked at the feature server linked in the info section to make sure the changes being made were accurate

fvankrieken:
Build is failing due to duplicate rows in a source dataset for templatedb (failing a dbt check). This is still ready to go!

dcpy/connectors/esri/arcgis_feature_service.py
```diff
@@ -12,41 +12,142 @@
...
+            name=feature_server.name,
+            layer_name=layer_name,
+            layer_id=layer_id,
+        )
+        if layer not in layers:
+            raise LookupError(
+                f"Layer '{layer}' not found in feature server {feature_server}"
+            )
+        return layer
+    layers = get_feature_server_layers(feature_server)
```
alexrichey, line 72: ✂️?

fvankrieken: Oops. Yup!

dcpy/connectors/esri/arcgis_feature_service.py
```diff
@@ -12,41 +12,142 @@
...
+            server=feature_server.server,
+            name=feature_server.name,
+            layer_name=l["name"],
+            layer_id=l["id"],
+        )
+        for l in resp["layers"]
+    ]
+
+
+def resolve_layer(
```
fvankrieken, line 46: I refactored this just slightly - seemed to be a bit cleaner with a `match ... case`. See latest commit

alexrichey (approved):
LGTM!

damonmcc (approved):

fvankrieken:
Annoyed at this

https://github.com/python/mypy/issues/12364

(see final comments especially)

damonmcc:
@fvankrieken got some builds failing and seems related to this

GFT and FacDB have the same [error](https://github.com/NYCPlanning/data-engineering/actions/runs/10354811836/job/28661025717?pr=1059#step:6:422) while planning the build:
```bash
ValidationError: 1 validation error for Config
dataset.source.arcgis_feature_server.layer
  Extra inputs are not permitted [type=extra_forbidden, input_value=0, 
input_type=int]
```

damonmcc:
(note from in-person chat) [this action run](https://github.com/NYCPlanning/data-engineering/actions/runs/10355168669) to archive all ArcGIS Feature Server datasets may fix it

PR #1056, opened 2024-08-09
update NYS Parks data library sources

### LINKED ISSUE #690: FacDB: pick recipe
While getting Shadow category recipes for the Green Fast Track product, I realized that one of the needed datasets, NYS Parks Property, may be already present in data library (`nysparks_parks`). I did end up pulling the requested data from a different source chosen by the GIS into a new recipe `nysparks_parks_polygons` as these datasets had different geometries and different row count. 

### Motivation
FacDB uses `nysparks_parks` recipe and we should re-evaluate whether we continue using this recipe or should switch to `nysparks_parks_polygons ` instead. 

### Info on the recipes: 
* `nysparks_parks `
    * Source: https://data.ny.gov/Recreation/State-Park-Facility-Points/9uuk-x7vh
    * Geometry: point
    * Row count: 254
    * Last updated (as of 3/15/24): 12/7/2020

* `nysparks_parks_polygons`
    * Source: https://data.gis.ny.gov/datasets/nysparks::ny-state-parks-property/about
    * Geometry: polygon
    * Row count: 860
    * Last updated (as of 3/15/24): 2/1/2024

### Timeline
Until next FacDB build (May 2024?)

damonmcc:
resolves #690

## changes
- update the `source` details in two data library templates: `nysparks_parks` and `nysparks_historicplaces`
- delete two data library templates: `nysparks_parks_polygons` and `nysparks_historicplaces_esri`

## follow-up work
1. run data library using the updated templates
2. run and fix all builds which used any of these 4 datasets (FacDB and GFT)

## background

The current sources of `nysparks_parks` and `nysparks_historicplaces` are NY State Open Data pages and they are no longer being updated.

During the creation of GFT, the GIS team found new sources for this data on the [NYS GIS Clearinghouse](https://data.gis.ny.gov/). We chose to create new datasets and called them `nysparks_parks_polygons` and `nysparks_historicplaces_esri`.

It seems clear that NYS Parks has replaced their use of the Open Data pages with the GIS Clearinghouse pages. We should update the source of the original recipes rather than maintain two.

While building FacDB, I switched to `nysparks_parks_polygons`/`nysparks_historicplaces_esri` in order to use updates source data and we end up with some misleading QA diffs like these:

Screenshot 2024-08-09 at 9 52 14 AM

Screenshot 2024-08-09 at 10 03 01 AM

Screenshot 2024-08-09 at 9 52 26 AM

Screenshot 2024-08-09 at 9 51 29 AM


dcpy/library/templates/nysparks_historicplaces_esri.yml
fvankrieken, line 1: This is the active version of historicplaces (at least for gft)

fvankrieken: Oh oops. Carry on!

fvankrieken: It seems like this will change things for FacDB - getting pts instead of polygons, not sure how that will be handled downstream. I think that was part of the keeping these separate, just to not have to change anything in FacDB

fvankrieken: See https://github.com/NYCPlanning/data-engineering/issues/690

fvankrieken: I am moving especially slow this morning apparently

fvankrieken (approved):

dcpy/library/templates/nysparks_historicplaces_esri.yml
damonmcc: lol no worries

the change in geometry type definitely could have an impact, but I think we already end up using the centroid of every FacDB source [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/facilities/facdb/sql/_create_facdb_geom.sql#L66) so it might be ok

dcpy/library/templates/nysparks_historicplaces.yml
```diff
@@ -2,21 +2,21 @@ dataset:
   name: nysparks_historicplaces
   acl: public-read
   source:
-    url:
-      path: https://data.ny.gov/api/views/iisn-hnyv/rows.csv
+    arcgis_feature_server:
+      server: nys_parks
+      name: National_Register_Building_Listings
+      layer: 12
```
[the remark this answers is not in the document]
damonmcc: wow I didn't even notice this was the dataset you showed me yesterday

PR #1057, opened 2024-08-09
Ingest streamlit page - finalize library connection

### LINKED ISSUE #1053: Ingest streamlit page - finalize library connection
With #1050 done, we can get this page working. 

Primarily, it's time to switch over from dummy functions to the real deal. There's also a small grab-bag of tasks to do
- [x] see if we can skip the local dump of the `UploadedFile` on the server and just upload the object that's already in memory to `inbox`
- [x] in the archive_raw_data function, throw an error if our upload path already exists (and maybe add a checkbox to allow overwrite, but stop accidental overwrites)
- [x] alter call to library.Archive to have `clean=True` (remove local files after library runs)
- [ ] handle errors in the call to library helper

This should all be tested with the call to `library.Archive` as is (without `push=True`) while testing. Once that's running

HengJiang0206:
resolves #1053


fvankrieken:
So you're adding a util to `dcpy`... which means that it's time to write a test!

Appropriate file is `dcpy/tests/utils/test_s3.py`. You should go ahead and take a look, but then we actually should have a broader discussion about testing

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -5,46 +5,52 @@
...
-            "public-read",
-        )
-    except Exception as e:
-        st.error(
-            "Failed to archive Dataset {dataset_name} version {version} to {s3_path}: {e}"
-        )
-    return file_path
+    exists = s3.exists(BUCKET, str(s3_path))
+    if exists == True:
+        st.warning("Warning: File path already exists")
```
fvankrieken, line 26: I think we can drop this warning - just either error or succeed

dcpy/test/utils/test_s3.py
```diff
@@ -87,3 +89,27 @@ def test_get_subfolders(create_buckets, prefix, index, expected_num_folders):
...
+    RESOURCES_DIR = Path(__file__).parent / "resources"
+    with open(RESOURCES_DIR / "data_wkb.csv", "rb") as f:
+        file_obj = BytesIO(f.read())
+    s3.upload_file_obj(
+        file_obj=file_obj,
+        bucket=TEST_BUCKET,
+        key=test_file_key,
+        acl="public-read",
+    )
+    assert s3.exists(TEST_BUCKET, test_file_key) == True
```
fvankrieken, line 115: Drop any `== True`s in this PR - redundant!

`if value == True` is equivalent to `if value`

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -5,46 +5,50 @@
...
-        st.error("Failed to save temporary file: {e}")
-
-    try:
-        s3.upload_file(
-            BUCKET,
-            file_path,
-            f"inbox/{dataset_name}/{version}/{file_name}",
-            "public-read",
+    exists = s3.exists(BUCKET, str(s3_path))
+    if exists == True and allow_override == False:
```
fvankrieken, line 25: I would get rid of line 24, and just have
```python
if s3.exists(BUCKET, str(s3_path) and not allow_override:
```

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -5,46 +5,50 @@
...
-    file_path = base_path / file_name
-    base_path.mkdir(parents=True, exist_ok=True)
+    dataset_name: str,
+    version: str,
+    uploaded_file: UploadedFile,
+    file_name: str,
+    allow_override: bool,
+) -> None:
 
     s3_path = Path("inbox") / dataset_name / version / f"{file_name}"
```
fvankrieken, line 22: let's just create this as a string (format string - `f"..."`)

fvankrieken: So that below we don't need to coerce it back to `str`

apps/qa/src/pages/ingest/helpers.py
```diff
@@ -5,46 +5,50 @@
...
-            "Failed to archive Dataset {dataset_name} version {version} to {s3_path}: {e}"
-        )
-    return file_path
+
+    file_obj = BytesIO(uploaded_file.read())
+
+    s3.upload_file_obj(
+        file_obj,
+        BUCKET,
+        f"inbox/{dataset_name}/{version}/{file_name}",
```
fvankrieken, line 35: this is just s3_path, no?

apps/qa/src/pages/ingest/ingest.py
```diff
@@ -1,6 +1,11 @@
 def ingest() -> None:
     import streamlit as st
-    from .helpers import archive_raw_data, dummy_library_call, dummy_archive_raw_data
+    from .helpers import (
+        archive_raw_data,
+        library_archive,
+        dummy_library_call,
```
fvankrieken, line 6: let's remove these, both in helpers and from this import line

fvankrieken (approved):
Nice tests!

PR #1067, opened 2024-08-12
add ability to pin version to source dataset

fvankrieken:
We've discussed basing GFT's version on PLUTO with GIS, so included here is the logic to do that.

Also figured I'd go ahead and get rid of the silly `StrEnum` deserializers and specify to call `model_dump(mode="json")` whenever we're actually dumping to a file instead

Final commit reworks a bit in plan, including adding a bit more test coverage (mainly to repeat build logic). They're getting a little hard to read with all the patching... but I think they largely make sense. 

I would go by commit

[gft build](https://github.com/NYCPlanning/data-engineering/actions/runs/10359663405) - will verify outputs are properly deserialized (build_metadata.json)

### DIFF
--- a/dcpy/lifecycle/builds/plan.py
+++ b/dcpy/lifecycle/builds/plan.py
@@ -23,6 +23,48 @@ RECIPE_FILE_TYPE_PREFERENCE = [
 ]
 
 
+def resolve_version(recipe: Recipe) -> str:
+    match recipe.version_strategy:
+        case None:
+            raise Exception("No version or version_strategy provided")
+        case versions.SimpleVersionStrategy.first_of_month:
+            return versions.generate_first_of_month().label
+        case versions.SimpleVersionStrategy.bump_latest_release:
+            previous_version = publishing.get_latest_version(recipe.product)
+            assert previous_version is not None
+            return versions.bump(
+                previous_version=previous_version,
+                bump_type=recipe.version_type,
+            ).label
+        case versions.BumpLatestRelease() as bump:
+            previous_version = publishing.get_latest_version(recipe.product)
+            assert previous_version is not None
+            return versions.bump(
+                previous_version=previous_version,
+                bump_type=recipe.version_type,
+                bump_by=bump.bump_latest_release,
+            ).label
+        case versions.PinToSourceDataset():
+            dataset = recipe.version_strategy.pin_to_source_dataset
+            inputs_by_name = {d.name: d for d in recipe.inputs.datasets}
+            if dataset not in inputs_by_name:
+                raise ValueError(
+                    f"Cannot pin build version to dataset '{dataset}' as it is not an input dataset"
+                )
+            input = inputs_by_name[dataset]
+            if not input.version and (
+                input.version_env_var
+                or (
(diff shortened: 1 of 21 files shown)


dcpy/lifecycle/builds/plan.py
```diff
@@ -23,6 +23,48 @@
...
+            ).label
+        case versions.BumpLatestRelease() as bump:
+            previous_version = publishing.get_latest_version(recipe.product)
+            assert previous_version is not None
+            return versions.bump(
+                previous_version=previous_version,
+                bump_type=recipe.version_type,
+                bump_by=bump.bump_latest_release,
+            ).label
+        case versions.PinToSourceDataset():
```
fvankrieken, line 47: codecov thinks this line is only partially covered - I think by their syntax its really the `match` statement. As in, it isn't covered when none of these match. Unfortunately for codecov, this matching is exhaustive (which mypy would catch if not). Don't really want to add an unreachable `case _:` just to make codecov happier

dcpy/lifecycle/builds/plan.py
```diff
@@ -55,6 +55,25 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
...
+            case versions.PinToSourceDataset():
+                dataset = recipe.version_strategy.pin_to_source_dataset
+                inputs_by_name = {d.name: d for d in recipe.inputs.datasets}
+                if dataset not in inputs_by_name:
+                    raise ValueError(
+                        f"Cannot pin build version to dataset '{dataset}' as it is not an input dataset"
+                    )
+                input = inputs_by_name[dataset]
+                if not input.version and (
+                    input.version_env_var
```
sf-dcp: could you please explain the line above? From my understanding, if `input.version` is not specified AND env var is present, then you get an error...

dcpy/lifecycle/builds/plan.py
sf-dcp, line 1: looks so much cleaner :)

dcpy/test/lifecycle/builds/test_plan.py
sf-dcp, line 1: 👏👏👏

dcpy/lifecycle/builds/plan.py
fvankrieken: Yeah sorry, it definitely is not intuitive to read.

Basically, I decided that it made sense to only allow pinning to this version if either
1. version is explicitly defined explicitly (great - it's used)
2. version is not defined explicitly and "latest" is being used

All other cases of not defining it explicitly should error - so either no explicit version, but `version_env_var ` being defined, or no explicit version and the missing version strategy is something other than "find_latest"

fvankrieken: The only env var we currently use is `VERSION_PREV`, which makes no sense to pin to. In general, thought it made sense to just limit this a little - only allow explicit definition or "latest", and be overly cautious throwing errors

sf-dcp: Got it, thank you

sf-dcp (approved):
LGTM! Nice feature

PR #1069, opened 2024-08-13
Add Metadata v2 models and tests

alexrichey:
### Part 1 of the Metadata upgrade to v2. 

[Metadata Output Here](https://github.com/NYCPlanning/product-metadata/pull/8)

Adds v2 of the product metadata and tests. I'd recommend starting with the test for overrides, as that makes a lot of the rationale clear. But the general idea is that we have three core entities:
1. dataset (composed of columns and attributes)
2. files (can override dataset attributes)
3. destinations (can override files and datasets)

When we calculate eventual metadata for a destination, we start with the basic dataset, then apply file overrides, then destination overrides. 

Part 2 (incoming, mostly finished, different PR) 
upgrading the connectors and lifecycle modules to use v2.

### The Plan
The plan overall is to get this reviewed, upgrade all dependencies (e.g. connectors, lifecycle modules) to md_v2, delete v1 metadata, then merge the PR in the product metadata repo.

### DIFF
--- a/dcpy/models/product/dataset/metadata.py
+++ b/dcpy/models/product/dataset/metadata.py
@@ -6,6 +6,7 @@ from typing import Any, Literal
 import yaml
 
 from dcpy.utils.logging import logger
+from . import metadata_v2 as md_v2
 
 # Putting this here so we can have a constant value in all connectors
 # and throw an exception when we attempt to deserialize files that contain it.
@@ -216,13 +217,15 @@ class Metadata(BaseModel, extra="forbid"):
 
     @staticmethod
     def from_yaml(yaml_str: str, *, template_vars=None):
-        logger.info(f"Templating the metadata with vars: {template_vars}")
-        templated = jinja2.Template(yaml_str, undefined=jinja2.StrictUndefined).render(
-            template_vars or {}
-        )
-        return Metadata(
-            templated_source_metadata=templated, **yaml.safe_load(templated)
-        )
+        if template_vars:
+            logger.info(f"Templating metadata with vars: {template_vars}")
+            templated = jinja2.Template(
+                yaml_str, undefined=jinja2.StrictUndefined
+            ).render(template_vars or {})
+            return Metadata(**yaml.safe_load(templated))
+        else:
+            logger.info("No Template vars supplied. Skipping templating.")
+        return Metadata(**yaml.safe_load(yaml_str))
 
     @classmethod
     def from_path(cls, path: Path, *, template_vars=None):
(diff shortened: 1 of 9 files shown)


dcpy/models/product/dataset/metadata.py
```diff
@@ -245,3 +248,183 @@ def validate(self):
             set(col_names)
         ), "There should be no duplicate column names"
         # TODO: all the rest
+
+    def upgrade_to_v2(self):
```
alexrichey, line 252: This is the fn that will convert v1 metadata to v2. We'll eventually just delete this entire file, so it's maybe less important to review every line of this, as opposed to just spot-checking the upgraded metadata

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+
+    Note: This is put in its own class because it might be useful for other
+    classes unrelated to product metadata in the future.
+    """
+
+    _exclude_falsey_values: bool = True
+    _head_sort_order: list[str] = PrivateAttr(default=["id"])
+    _tail_sort_order: list[str] = PrivateAttr(default=["custom"])
+
+    @model_serializer(mode="wrap")
```
fvankrieken, line 56: just checking my/your understanding of a "wrap" serializer - this gets called on the pydantic object, returning some sort of dict/pydantic like object that we've configured as per this function, and _then_ its children are serialized? Or vice-versa -> handler gets us the "native" serialized dict, and we modify it there.

I guess those work out to roughly the same. Just find the pydantic docs a tad uncertain

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+    def _model_dump_ordered(self, handler):
+        unordered = handler(self)
+
+        ordered_items_head = []
+        ordered_items_tail = []
+        simple_type_items = []
+        other_items = []
+
+        for model_field in list(unordered.items()):
+            if not model_field[1] and self._exclude_falsey_values:
```
fvankrieken, line 66: Per convo - `0` is falsey, and a value that we would actually want to serialize

fvankrieken: I don't think there are any other falsey values we'd want to keep, after some googling

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+    @model_serializer(mode="wrap")
+    def _model_dump_ordered(self, handler):
+        unordered = handler(self)
+
+        ordered_items_head = []
+        ordered_items_tail = []
+        simple_type_items = []
+        other_items = []
+
+        for model_field in list(unordered.items()):
```
fvankrieken, line 65: model_field is a tuple, right? Might be just a little more readable with something like
```python
for key, value in list(unordered.items()):
```

Just to avoid all the `model_field[0]`s

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+
+# BASE
+class SortedSerializedBase(BaseModel):
+    """A Pydantic BaseModel that will allow for sensible (and overrideable) deserialization order.
+
+    The serialization order is as follows:
+    - model attributes defined in the head sort order
+    - simple (and nullable simple) types: strings, ints, bools, literals
+    - complex types
+    - model attributes defined in the tail sort order
```
fvankrieken, line 46: I think I like how this works out, in terms what the output looks like. 

Is there a reason you went this way rather than, say, something like trying to serialize in the order the model is actually defined?

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+    like data-dictionary generation (e.g. the `readme_data_type` field on the columns)
+
+    It's also important that custom be preserved in the
+    course of overriding, specifically with a deep merge of the dictionary elements.
+    """
+
+    custom: dict[str, Any] = {}
+
+
+class YamlWriter(BaseModel):
```
fvankrieken, line 121: Lovely 🙂

fvankrieken:
This is a meaty one, holy cow.

Looks great - I think I need a little more coffee and then another look to just digest it a bit better

sf-dcp:
Conceptual question: do you still need to override a column name in `destinations` if: 
- it's different from column name in the `columns` section
- it's the same as in the `files` section?

For example, do I need to add anything under `destinations` below if it should be the same as in `files` (BOROUGH)?
 
```yaml
columns:
- id: borough
  name: Borough
  data_type: text

files:
- file:
    id: primary_csv
    filename: colp.csv
    type: csv
  dataset_overrides:
    overridden_columns:
    - id: borough
      name: BOROUGH

destinations:
- id: socrata
  type: socrata
  files:
  - id: primary_shapefile
     dataset_overrides:
```

dcpy/models/product/dataset/metadata_v2.py
sf-dcp, line 121: It would be nice to move this to "util" models or something like that as it doesn't feel to be specific to metadata model exclusively

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+    translator = str.maketrans(char_map)
+    # Normalize Unicode characters
+    normalized = unicodedata.normalize("NFKD", s)
+    # Apply the translation
+    cleaned = normalized.translate(translator)
+    return cleaned.strip()
+
+
+# BASE
+class SortedSerializedBase(BaseModel):
```
sf-dcp, line 39: I'm a bit confused by this class and what's happening inside of it. Could we chat about it on Monday?

dcpy/models/product/dataset/metadata.py
fvankrieken: another note for the possibility of `v` being 0 for the falsey function below

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -0,0 +1,424 @@
...
+
+# MISC UTILS
+def normalize_text(s):
+    """
+    Normalize the text we may receive from the various metadata sources.
+    Primarily useful for cleaning long-text like descriptions.
+    """
+    char_map = {
+        "–": "-",  # en dash to hyphen
+        "—": "-",  # em dash to hyphen
```
damonmcc, line 25: duplicated?

dcpy/models/product/dataset/metadata_v2.py
alexrichey: Yeah, the docs aren't great here. But to your question, the serialization is recursive - the model will descend through it's children attributes calling `model_dump` on those. That bubbles up to, eventually, just returning a top-level dictionary. Since all our metadata models inherit from this class, they'll all use this machinery to deserialize.

dcpy/models/product/dataset/metadata_v2.py
alexrichey: This was a great catch. It would have been a painful thing to track down. Fixed [here](https://github.com/NYCPlanning/data-engineering/pull/1069/commits/b30714297a2337e057f44e66f60b506cd09ad24c)

alexrichey:
> Conceptual question: do you still need to override a column name in `destinations` if:

Per convo, thankfully not! The overrides are applied sequentially as we go from:
1. dataset (attributes + columns)
2. files
3. destination

so any overrides carry forward. And in the case of the `custom` dict, these overrides are deep merged together. There's a test case around this, but in short if I have a column:
``` 
COLUMN
id: my_uid
custom:
  - overriden:
      greeting: hello

FILE
id: my_uid
custom:
  - overriden:
      greeting: hi there
      other_val: I get carried through

DESTINATION
id: my_uid
custom:
  - overriden:
      greeting: hola

Then the final calculated value is:
id: my_uid
custom:
  - overriden:
      greeting: hola
      other_val: I get carried through
```

dcpy/models/product/dataset/metadata_v2.py
alexrichey: I think the font isn't doing us any favors here, but one is U+2013 and the other is U+2014. Lots of this stuff came from copying in from pdfs.

dcpy/models/product/dataset/metadata_v2.py
alexrichey: def agree. Moved [here](https://github.com/NYCPlanning/data-engineering/pull/1069/commits/62f2acb68d82645249223110a8e80827aef599ca) to a file called `models/base.py` - thinking we'll eventually have a few top-level base models we'd like to mix in. Open to other names though!

damonmcc (approved):

sf-dcp (approved):
nice!

alexrichey:
Merging this for now, but still happy for feedback - let's continue any conversations on [part 2](https://github.com/NYCPlanning/data-engineering/pull/1073)

PR #1070, opened 2024-08-13
first draft of dev recipes, publishing

fvankrieken:
## Main changes

Ability to supply a env var flag `DEV_FLAG` to show that dev bucket should be used

One thing that isn't super well defined in code right now is write vs read (some functions are implicitly read functions, others are write). Behavior as stands
- recipes
  - write - if `DEV_FLAG`, then write to dev bucket
  - read - if `DEV_FLAG`, look for data in dev bucket, fall back to production if not found (for use in dev builds)
 - publishing
   - write - if `DEV_FLAG`, then write to dev bucket
   - read - only read from production 

It might be better to codify this slightly (and categorize functions). Curious on people's thoughts. We currently have many different types or read functions - get all versions, get latest versions, fetch dataset, read_df, etc.

I didn't include any publishing read functions to use this dev logic because it didn't seem entirely necessary at the moment.

## commits
1. just adds more tests for recipes
2. the main meat of this PR. I would suggest looking at `configuration.py`, then `recipes` and `publishing`, then tests. `configuration` was an effort to start moving interactions with `os.environ` to a more centralized place so their easier to keep track of. I didn't do it here, but we could do something similar with env vars for pg connection, s3 connection, etc if we like this approach. Would love thoughts/feedback.
3. downstream effects from 2 - just figured it was easier to review 2 without this included
4. two additional publishing tests
5. move a function from template db to publishing and write a test, then clean up the way templatedb interacts with publishing
6. gha stuff to be able to run dev_flag workflows

#### library
[archival of bpl_libraries](https://github.com/NYCPlanning/data-engineering/actions/runs/10602322901)
image

#### build
[build template](https://github.com/NYCPlanning/data-engineering/actions/runs/10602708367)
image

(for facilities) confirmed that dev version was found and used
image

#### ingest
[archival of bpl_libraries](https://github.com/NYCPlanning/data-engineering/actions/runs/10603276471)
image


fvankrieken:
I was going to add cleaning up dev buckets to the `clean_artifacts` file... but I think I'd prefer we don't programmatically delete any buckets. Just seems like a risky line to cross. I think these will be happening infrequently enough that we can trust ourselves to delete them manually

fvankrieken:
Sorry - had to clean up some git stuff (and rename the dev bucket to not have "recipes" in it since we're using it for multiple things). No more pushes

dcpy/configuration.py
```diff
@@ -0,0 +1,16 @@
...
+
+CI = "CI" in env
+
+BUILD_NAME = env.get("BUILD_NAME")
+if DEV_FLAG and not BUILD_NAME:
+    raise ValueError("'BUILD_NAME' env var needed with 'DEV_FLAG'")
+
+RECIPES_BUCKET = "edm-recipes"
+PUBLISHING_BUCKET = "edm-publishing"
+DEV_BUCKET = f"de-dev-{BUILD_NAME}" if DEV_FLAG else None
```
fvankrieken, line 16: Maybe this `DEV_BUCKET` should just be the env var instead of `DEV_FLAG`. Would allow flexibility without adding an additional env var to check. But currently, it's nice just having a checkbox in a GHA

dcpy/test/connectors/edm/test_recipes.py
```diff
@@ -112,21 +156,63 @@ def test_parse_config():
...
-    config = recipes.get_config(TEST_LIBRARY_DATASET, load_library_parquet.version)
+def test_get_library_config(load_library_parquet: library.Config):
+    config = recipes.get_config(TEST_DATASET, load_library_parquet.version)
     assert config == load_library_parquet
 
 
-def test_get_latest_version(load_library_parquet: library.Config):
-    assert (
-        recipes.get_latest_version(TEST_LIBRARY_DATASET) == load_library_parquet.version
+def test_get_ingest_config(load_ingest: library.Config):
```
sf-dcp, line 161: does this test use the pytest fixture above, `load_ingest`? If so, I'm a bit confused what exactly we are comparing here

dcpy/test/connectors/edm/test_recipes.py
```diff
@@ -47,11 +60,28 @@ def load_library_all(create_buckets):
...
 
 
-def test_archive_dataset(create_buckets, create_temp_filesystem: Path):
+def _mock_template(name, version, path=None):
+    configure.read_template(name, version, RESOURCE_DIR)
+
+
+@pytest.fixture(scope="function")
+def load_ingest(create_buckets):
+    with mock.patch.object(read_template, "__defaults__", (None, None, RESOURCE_DIR)):
```
sf-dcp, line 66: Q about the generated mock object above. From my understanding:
* `read_template` is the method we are mocking
* `"__defaults__"` are `read_template` input arg names (ie `dataset`, `version`, and `template_dir`)
* `(None, None, RESOURCE_DIR)` are the values to call `read_template` with. 

My question is: how do we call with `dataset` = `None` if `dataset` is a required arg?

dcpy/configuration.py
damonmcc: yea a boolean like `DEV_FLAG` at the top-level to drive downstream variables seems nice

dcpy/test/connectors/edm/test_recipes.py
fvankrieken: This is patching the defaults to the function `read_template` - when it's called in a test, any supplied args at run time will overwrite these defaults. This function when called within `run` doesn't supply an argument for the `template_dir`, so it will use the default that I've defined here. The other two args are supplied, so it's fine that I didn't set a default here (or set a default of `None`)

dcpy/test/connectors/edm/test_recipes.py
fvankrieken: It does.

Yeah, this is maybe a bit sloppy by me - I figured easiest way to mock having a config in s3 was to just run the ingest process, but this is not great practice. recipe tests really shouldn't depend on `ingest` code having run successfully, since that same ingest code is relying on other recipe functions. 

There's some argument that 
- "write" recipe code is tested on its own
- "write" ingest code tested on already tested recipe code
- "read" recipe code is tested using tested ingest code

Isn't horribly improper, since it's not circular... But probably best to keep them separate entirely. I'll rework this to not use ingest code

fvankrieken: So the idea with this function was 
- use ingest to dump a config in s3
- test that this config is read successfully

dcpy/configuration.py
damonmcc, line 1: great to move interactions with `os.environ` to a more centralized place

could we do something like this in `configuration.py` to avoid stuff like `configuration.DEV_BUCKET or configuration.PUBLISHING_BUCKET` later:

```python
RECIPES_BUCKET = DEV_BUCKET if DEV_FLAG else "edm-recipes"
PUBLISHING_BUCKET = DEV_BUCKET if DEV_FLAG else "edm-publishing"
```

maybe it'd be more readable/maintainable do:

```python
PROD_RECIPES_BUCKET = "edm-recipes"
RECIPES_BUCKET = DEV_BUCKET if DEV_FLAG else PROD_RECIPES_BUCKET
```

damonmcc: but maybe this is at odds with the variety of read/write behavior you described

dcpy/connectors/edm/publishing.py
```diff
@@ -227,7 +239,7 @@ def legacy_upload(
         s3.upload_file("edm-publishing", output, str(key), "public-read", metadata=meta)
```
sf-dcp, line 237: change `"dem-publishing"` to `bucket` here?

dcpy/configuration.py
fvankrieken: Yeah - the specific problem is for building

Say we archive `bpl_libraries` to a dev bucket

Then we want to build template db using this dev input dataset. Ideally, we specify dev - then for each dataset, it will look at the dev bucket. If we have a global bucket, it will fail, because we're missing the other datasets.

Similarly, say we want to increment the version from the previous build. If it looks at dev publishing, there are not previous builds, so it fails - needs to read from prod publishing

dcpy/connectors/edm/recipes.py
```diff
@@ -246,6 +267,9 @@ def purge_recipe_cache():
 
 
 def log_metadata(config: library.Config):
+    if configuration.DEV_FLAG:
+        logger.info("DEV_FLAG found, not library run metadata")
```
sf-dcp, line 271: a bit confused by the meaning of this log message...

fvankrieken: Yeah that is fair

dcpy/library/__init__.py
sf-dcp, line 1: curious about the differences between `__init__.py` vs `configuration.py` because `configuration.py` seems to be similar to the init file in defining global variables...

fvankrieken:
Per discussion today, two things to add/change
1. for this, have dev bucket just take the place of publishing/recipe buckets. None of the current "try to get from dev, fall back to prod" behavior
2. add a utility to set up a dev bucket. With some args/flags, create/re-create a bucket and copy in some amount of data from our production buckets - likely latest version of every `recipe` dataset and latest `publish` version of each `publishing` dataset. Maybe a flag for `edm-publishing/datasets` as well? Will probably leave out for now. We'll each get a dev bucket, `de-dev-{first_initial}`, these will be dropdown options for these actions

dcpy/library/__init__.py
fvankrieken: This is for `library`, this file will eventually be deleted

dcpy/configuration.py
alexrichey, line 1: Love the move to extract these into the env. What do you think about putting our default values in a .env file outside of dcpy, and maybe just loading them with python-dotenv?

fvankrieken:
Some complications with our hope to just quickly spin up dev buckets.

Copy commands are not allowed across clusters within a region for spaces buckets. We can choose what region a bucket is in, but cluster seems entirely internal to DO. This means that, with some tool (mc/rclone/boto3), we need to actually download/upload to get files from edm-recipes or publishing into a new bucket.

This has some serious drawbacks, primarily
- speed. On my personal comp, creating a dev bucket with latest of all recipe datasets needed for `pluto` took about an hour. Not the end of the world, not as lightweight and easy as I had hoped
- downloads count towards our monthly read bandwidth, which is 1TB. My run right now of `latest` of all pluto-related datasets and latest `publish` folder for pluto was ~5gb. I think cpdb and GFT would be larger. So maybe not a huge issue - I don't think anyone would make more than one of these a month, and it's only a bit bulkier than running a few builds (which we already do all the time)
These are both made worse by the fact that we currently have multiple file types for each source dataset, and different builds sometimes use different file types (pluto uses pg_dump, facdb uses csv) so it's a bit tough to limit this. I suppose I could use the recipe to determine which file types are needed and only download those. This could reduce this size (and also then the time to setup) to close to 1/3 - right now it just grabs the whole `datasets/{ds_id}/latest` folder

Tangent, but maybe that's also something to think about a bit with our nightly_qa job - I wouldn't be surprised if we're reading 10-15 gb per day with that, which would get us (monthy) to somewhere between 1/3 and 1/2 of our limit right there. Persisting `latest` of each recipe dataset in a db (at time of library archival) and only grabbing from s3 when needed might be a more sustainable (and performant) long term solution. That could be part of 1 db per build environment

Anyways - would love to chat about how we see ourselves using this just as a sanity check - I've gone in circles a bit in my vision here

dcpy/configuration.py
fvankrieken: Something like that would make sense - I think I shied away from python-dotenv early on just because of having both bash utilities that used env vars as well as the python parts of the build that need them, so it seems easier (especially locally) to just use bash to set env vars and have python just expect them to be present. I'm down to re-evaluate a little though

dcpy/connectors/edm/publishing.py
```diff
@@ -24,14 +24,18 @@
 from dcpy.utils import s3, git, versions
 from dcpy.utils.logging import logger
 
+assert (
+    PUBLISHING_BUCKET
+), "'PUBLISHING_BUCKET' must be defined to use edm.publishing connector"
+BUCKET = PUBLISHING_BUCKET
```
fvankrieken, line 30: This is a funny little thing here - mypy is fine with this, but not okay without line 30 (and just using the asserted variable later on)

PR #1073, opened 2024-08-14
Metadata Refactor p2

alexrichey:
### Commits 1 and 2
Commit 1 fixes a bug, and Commit 2 adds some validation machinery. After that...

### The Rest
This PR updates references to prior metadata format for connector and lifecycle modules, and adds unit tests. The commits are atomic, so I'd recommend going one-by-one.

I think that the trickiest commit to review will be the update to `connectors.socrata.publish`.`push_dataset`. Nothing has functionally changed, but the diff is pretty rough. However, it is now covered by unit tests, and I've deployed datasets using it. (Locally only, though. So unfortunately I don't have a GHA to point you toward as proof). 

Anyways, happy to chat about this!


dcpy/connectors/socrata/publish.py
```diff
@@ -102,73 +136,100 @@ def __init__(self, data):
...
+            column_update_endpoint=f"https://{SOCRATA_DOMAIN}"
+            + socrata_source.input_schemas[0].links["transform"],
+        )
+
+    @property
+    def column_names(self) -> list[str]:
+        return [c["field_name"] for c in self.uploaded_columns]
+
+    def calculate_pushed_col_metadata(self, our_columns: list[md.DatasetColumn]):
+        # TODO: using c.id or c.name?
```
alexrichey, line 175: I'm not entirely sure how to handle this. Most of the our metadata.column ids are 1-1 with the socrata column, while the column `name` is used for the socrata display name. 

The "correct" way to handle this is to add an "api_name" to the destination overrides, but that just creates a lot of boilerplate. We could potentially add an `external_id` field to the columns, that defaults to the `id` when not present.

dcpy/test/lifecycle/distribute/test_bytes.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def test_plan_errors_for_socrata():
+    with pytest.raises(Exception, match=bytes.NON_BYTES_DEST_ERROR):
+        plan = bytes.pull_destination_files(Path(""), make_metadata(), SOCRATA_DEST_ID)
+
+
+@patch("dcpy.lifecycle.package.assemble.unpackage")
+@patch("urllib.request.urlretrieve")
+def test_pull_destination_files_mocked(mock_urlretrieve, mock_unpackage, tmp_path):
```
damonmcc, line 107: this test failed with
```bash
AttributeError: <module 'dcpy.lifecycle.package.assemble' from
'/__w/data-engineering/data-engineering/dcpy/lifecycle/package/assemble.py'> does not have the attribute 'unpackage'
```

maybe line 105 should be `@patch("dcpy.lifecycle.package.assemble.unzip_into_package")` ?

alexrichey: yeah, had just fixed that! Not sure how let that one slip through.

damonmcc (approved):
thanks for walkin us through this last week!

PR #1077, opened 2024-08-14
fix pandoc commands 

HengJiang0206:
Fix errors in Pandoc commands where there is an extra space between `--metadata-file=` and the path provided, resulting in Pandoc being unable to find the metadata file.


damonmcc:
nice! can you try rebasing locally? the PR is showing 22 commits even thought that [last commit](https://github.com/NYCPlanning/data-engineering/pull/1077/commits/82773987db0193a21f7ca87f14fec7431c1023bd) is really the only one here

if it's a pain to rebase, might be easier to start a new branch

PR #1081, opened 2024-08-15
fix pandoc command extra space

HengJiang0206:
Fix errors in Pandoc commands where there is an extra space between --metadata-file= and the path provided, resulting in Pandoc being unable to find the metadata file.


HengJiang0206:
@damonmcc Hey Damon, I just created a new branch for this - would need to resolve a lot of conflicts for rebase.

damonmcc (approved):

Issue #944, opened 2024-06-24
Opendata: Packaging - Data Dictionary / README generation from metadata


alexrichey:
### Tasks

These are the bare minimum things to do before potentially handing off to fellows:

- [ ] Research what we need from OTI to meet compliance
- [ ] List common elements in our READMEs and Data Dictionarys. (e.g. tables, images?, etc.)
- [ ] Find DCP style guide, or reference examples? Maybe AE has something?
- [ ] Find a few reference examples that we like, from out in the wide world
- [ ] Research the technical side of things (e.g. generating pdfs)

alexrichey:
Running comment...

### Common README/DD elements

#### Changelog
- should point at the Github tag/release?

damonmcc:
for reviewing the current state of our metadata out in the world, I added links to the [Data Catalog](https://github.com/NYCPlanning/data-engineering/wiki/Data-Catalog) in the DE wiki for each data product's destinations

on NYC Open Data, we usually have a data dictionary excel file. on DCP Bytes, we usually have a data dictionary and a readme PDF (this is a generalization of what we have at both destinations)

damonmcc:
I think the only example of a templated markdown file we have is for CEQR: [data-engineering/products/ceqr/recipes/dep_cats_permits/README.md](https://github.com/NYCPlanning/data-engineering/blob/main/products/ceqr/recipes/dep_cats_permits/README.md)

alexrichey:
@damonmcc also, here's the [AE Tokens/Typography](https://streetscape.nycplanningdigital.com/?path=/docs/tokens-typography--docs) that Tyler mentioned two weeks'ish ago

PR #1085, opened 2024-08-16
revise metadata packaging

damonmcc:
related to #944

before adding code to generate product READMEs, these changes test and refactor the code to generate product data dictionaries

the function added to `connectors/edm/product_metadata.py` is untested (although it's used in the test Template DB build), but all of those functions are untested


dcpy/connectors/edm/product_metadata.py
```diff
@@ -21,3 +22,11 @@ def fetch_raw(product: str, *, dataset: str | None = None) -> str:
 
 def fetch(product: str, *, dataset: str | None = None) -> md.Metadata:
     return md.Metadata(**yaml.safe_load(fetch_raw(product, dataset=dataset)))
+
+
+def download(product: str, output_path: Path, *, dataset: str | None = None) -> Path:
+    if not output_path.parent.exists():
```
alexrichey, line 28: just fyi, you can one-line this with `mkdir(parents=True, exists_ok=True)`

dcpy/lifecycle/package/generate_metadata_assets.py
```diff
@@ -1,54 +1,64 @@
...
     yaml_file_path: str,
     html_template_path: str,
     output_html_path: str,
     output_pdf_path: str,
-    pdf_metadata_path,
-) -> Path:
-    generate_html_from_yaml(yaml_file_path, html_template_path, output_html_path)
+) -> None:
+    generate_html_from_yaml(
+        Path(yaml_file_path),
```
alexrichey, line 22: Any reason not to make these `Path` objects in the function args? Typer will take care of handling converting them.

dcpy/lifecycle/package/generate_metadata_assets.py
```diff
@@ -1,54 +1,64 @@
...
 
+    return output_html_path
 
-def get_metadata(yaml_file_path: str) -> SocrataMetada:
-    metadata = Metadata.from_path(
-        Path(yaml_file_path), template_vars={"var1": "value1"}
-    )
-    return metadata
+
+if __name__ == "__main__":
```
alexrichey, line 63: should we ✂️ this and add to the dcpy.cli?

pyproject.toml
```diff
@@ -149,4 +145,7 @@ exclude = ["dcpy.test*"]
 library = "dcpy.library.cli:run"
 
 [tool.setuptools.package-data]
-dcpy = ["library/templates/*.yml"]
+dcpy = [
+    "library/templates/*.yml",
+    "lifecycle/package/resources/data_dictionary_template.jinja",
```
alexrichey, line 150: is it possible to specify any jinja templates, so we don't continually stub our toe on this?

dcpy/lifecycle/package/generate_metadata_assets.py
damonmcc: ~~that sounds good, I'll change the args. but I didn't wanna worry about optimizing the CLI stuff here yet~~

per the use in `template/build_scripts.export.py` of the non-CLI functions, a datasets's metadata has to be downloaded to be converted to a PDF. so these CLI functions aren't very usable in a build

I just deleted the CLI part of this so that we can invoke this sub-step of packaging with higher-level CLI command later

dcpy/lifecycle/package/generate_metadata_assets.py
damonmcc: see [comment](https://github.com/NYCPlanning/data-engineering/pull/1085#discussion_r1729396643)

pyproject.toml
damonmcc: should be! but my first few tries didn't work and I was gonna generalize this in the follow-up work with READMEs. that work will likely add a `readme_template.jinja`

dcpy/connectors/edm/product_metadata.py
damonmcc: done!

alexrichey (approved):
Good stuff!

PR #1089, opened 2024-08-19
PLUTO:  add lots to `pluto_input_research.csv`

sf-dcp:
While working on the PLUTO QA project, several lots failed a new data check for residential units. I manually researched these lots and they do have incorrect res unit count coming from raw data. 

Per conversation w/Amanda, we were planning to update them in PLUTO 24v4 if the first QA of current PLUTO passed. But it didn't so fixing it in the current version.

### DIFF
--- a/products/pluto/pluto_build/data/pluto_input_research.csv
+++ b/products/pluto/pluto_build/data/pluto_input_research.csv
@@ -27372,16 +27372,29 @@ bbl,field,old_value,new_value,Type,reason,version
 "3020580050",unitsres,229,0,2,PTS error - vacant lot with units,"24v2"
 "3020580050",unitstotal,231,0,2,PTS error - vacant lot with units,"24v2"
 3052497501,unitsres,2025,45,2,PTS error - units,24v3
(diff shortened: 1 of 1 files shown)


damonmcc (approved):

products/pluto/pluto_build/data/pluto_input_research.csv
fvankrieken, line 1: If we're going to correct unitsres for these, we should correct unitstotal as well

sf-dcp: my next step in the PLUTO project is researching units total field (second data check). I'm not sure if we should do both types of corrections all together or separately

fvankrieken: Since they're so tightly coupled, I think it makes sense to do them together. We typically don't really units total separate from corrections to units res (if you look at the csv, they're always in pairs)

sf-dcp: makes sense. will correct

PR #1091, opened 2024-08-22
Fix patching in publishing

sf-dcp:
As mentioned in #1092, the current logic doesn't do correct version bumping when publishing. 

### This PR:
1. adds a function to `versions.py` that takes in a version and a list of versions and then returns a list of same versions (patched or non-patched). This sentence is too much so here is an example:
   * input version: `24Q1`
   * input version list: [`24Q1`, `24Q1.1`, `24Q2`, `23v1`, `22v3`] (could be mixed types of versions)

   * The new function returns: [`24Q1`, `24Q1.1`]
2. modifies existing `publishing.py` fn that validates or bumps a version, using the function from step 1.

### Testing
1. I wrote additional tests and they all pass
2. Additionally, I ran [this](https://github.com/NYCPlanning/data-engineering/actions/runs/10599912048/job/29376064105) GHA to patch template db and it works:
    * GHA settings:    
    image

    * **Before** GHA:  
    image

    * **After** GHA:     
    image

### DIFF
--- a/apps/qa/src/shared/components/sidebar.py
+++ b/apps/qa/src/shared/components/sidebar.py
@@ -43,7 +43,9 @@ def data_selection(
                     )
         case "Published":
             label = "Select a version"
-            options = publishing.get_published_versions(product)
+            options = publishing.get_published_versions(
+                product=product, exclude_latest=False
+            )
             select = st.sidebar.selectbox(label, options, key=f"{section_label}_output")
             if select:
                 return publishing.PublishKey(product, select)
--- a/dcpy/utils/versions.py
+++ b/dcpy/utils/versions.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from abc import abstractmethod
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, fields
 from datetime import date
 from dateutil.relativedelta import relativedelta
 from enum import StrEnum
@@ -237,7 +237,7 @@ def sort(versions: list[Version]) -> list[Version]:
     """
     version_types = set([type(v) for v in versions])
     # can't compare different types
-    if len(version_types) != 1:
+    if len(version_types) > 1:
         # only handle date-like versions
         raise TypeError(
             f"Can't sort mixed types of dataset versions: {[v.__name__ for v in version_types]}"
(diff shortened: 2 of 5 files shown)


dcpy/utils/versions.py
```diff
@@ -237,7 +237,7 @@ def sort(versions: list[Version]) -> list[Version]:
     """
     version_types = set([type(v) for v in versions])
     # can't compare different types
-    if len(version_types) != 1:
+    if len(version_types) > 1:
```
sf-dcp, line 240: changed this because it errors when input list is empty.

dcpy/utils/versions.py
```diff
@@ -255,6 +255,56 @@ def is_newer(version_1: str, version_2: str) -> bool:
...
+    print(matching_versions)  # Output: ['24Q1', '24Q1.1']
+    ```
+    """
+    parsed_version = parse(version)
+    version_list_wo_latest = [v for v in versions_list if v != "latest"]
+    same_version_type_lst = [
+        v for v in version_list_wo_latest if isinstance(parse(v), type(parsed_version))
+    ]
+
+    def is_matching_version(base_version: Version, compare_version: Version) -> bool:
```
fvankrieken, line 279: Should we maybe just have `Version` have `patch` explicitly? Then we could do something a little more naive, like set version to zero for each and then just use `==`.

That being said, maybe it sort of complicates other things (need to ensure that dataclasses with inherited attributes still sort/compare properly). Seems like more a headache to change. So more just food for thought. Let's leave this as is!

fvankrieken: Or maybe version could have a property called `base_fields`, which returns the dict put out by the helper within this helper? Just spitballing

fvankrieken: I feel like maybe that's better taken care of when we overhaul versions a bit. Which we should probably do sometime soon 😬

apps/qa/src/shared/components/sidebar.py
```diff
@@ -43,7 +43,9 @@ def data_selection(
                     )
         case "Published":
             label = "Select a version"
-            options = publishing.get_published_versions(product)
+            options = publishing.get_published_versions(
```
fvankrieken, line 46: Ahh - makes sense that the QA app would have some cases where we'd want latest. Thank you for catching!

fvankrieken (approved):

PR #1092, opened 2024-08-22
Publishing: temp fix for patch

sf-dcp:
I noticed that current patching doesn't work as expected and I missed a test for this case:

Regular and patched versions already exist in the published folder. For example, `24Q1` & `24Q1.1`. 
Expected patched version is `24Q1.2`; however, the current code will publish `24Q1.1`, overwriting the initial existing patch. 

See comment in the code below where this is happening exactly. 

In this PR, I'm adding a temp fix to prevent overwriting of existing patches (it will simply fail if patch already exists). I'm working on the actual fix in #1091, but it may take a bit.

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -283,11 +283,15 @@ def validate_or_patch_version(
 
     if version_already_published:
         if is_patch:
-            return versions.bump(
+            patched_version = versions.bump(
                 previous_version=version,
                 bump_type=versions.VersionSubType.patch,
                 bump_by=1,
             ).label
+            assert patched_version not in get_published_versions(
+                product=product
+            )  # sanity check
+            return patched_version
         else:
             raise ValueError(
                 f"Version '{version}' already exists in published folder and patch wasn't selected"


dcpy/connectors/edm/publishing.py
```diff
@@ -283,11 +283,15 @@ def validate_or_patch_version(
 
     if version_already_published:
         if is_patch:
-            return versions.bump(
+            patched_version = versions.bump(
                 previous_version=version,
```
sf-dcp, line 287: This is the culprit of the bug: this line takes the input version, ex: `24Q1`, and patches to `24Q1` even though it may exist.

dcpy/connectors/edm/publishing.py
```diff
@@ -283,11 +283,15 @@ def validate_or_patch_version(
...
                 bump_type=versions.VersionSubType.patch,
                 bump_by=1,
             ).label
+            assert patched_version not in get_published_versions(
```
alexrichey, line 291: nit: should we raise a ValueError here? (just to maintain consistency in what Exceptions this fn raises)

dcpy/connectors/edm/publishing.py
fvankrieken: A little unclear to me - how does this fix it? It seems to offer validation, but I don't see how this actually changes behavior

fvankrieken: Oops

fvankrieken (approved):

PR #1095, opened 2024-08-26
export unzipped csvs to analysis folder

damonmcc:
related to https://github.com/NYCPlanning/ae-cp-map/issues/44

these changes allow AE to link to each csv in the new Capital Projects app. we've mentioned two more ideal long-term approaches (our package folder `product-dataset/cpdb/`, using signed links which expire) but this seems like a fair short-term approach for their MVP app

builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-cpdb-geographies-files) and repeat builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/repeat_build.yml?query=branch%3Adm-cpdb-geographies-files)

## files in draft build (a repeat of 24prelim)

Screenshot 2024-08-29 at 3 48 40 PM

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -590,6 +590,9 @@ def _cli_wrapper_upload(
     ),
     build: str = typer.Option(None, "-b", "--build", help="Label of build"),
     acl: str = typer.Option(None, "-a", "--acl", help="Access level of file in s3"),
+    max_files: int = typer.Option(
+        s3.MAX_FILE_COUNT, "--max_files", help="Max number of files to upload"
+    ),
 ):
     acl_literal = s3.string_as_acl(acl)
     if not output_path.exists():
@@ -602,7 +605,12 @@ def _cli_wrapper_upload(
     logger.info(
         f'Uploading {output_path} to {product}/build/{build_name} with ACL "{acl}"'
     )
-    upload(output_path, BuildKey(product, build_name), acl=acl_literal)
+    upload(
+        output_path,
+        BuildKey(product, build_name),
+        acl=acl_literal,
+        max_files=max_files,
+    )
 
 
 @app.command("publish")
--- a/products/cpdb/python/projects_in_geographies.py
+++ b/products/cpdb/python/projects_in_geographies.py
@@ -64,11 +64,3 @@ if __name__ == "__main__":
             geography.geography_name,
         )
         export_table(geography.geography_type, geography.table_name)
-
-    # zip folder with all csv files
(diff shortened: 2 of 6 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -602,7 +605,12 @@ def _cli_wrapper_upload(
     logger.info(
         f'Uploading {output_path} to {product}/build/{build_name} with ACL "{acl}"'
     )
-    upload(output_path, BuildKey(product, build_name), acl=acl_literal)
+    upload(
+        output_path,
+        BuildKey(product, build_name),
+        acl=acl_literal,
+        max_files=int(max_files),
```
alexrichey, line 612: does just specifying this as an int in the signature cause any issues?

products/cpdb/python/projects_in_geographies.py
```diff
@@ -64,11 +64,3 @@ def export_table(geography_type: str, table_name: str) -> None:
             geography.geography_name,
         )
         export_table(geography.geography_type, geography.table_name)
-
-    # zip folder with all csv files
-    print(f"Zipping folder\n\t{OUTPUT_DIR}")
```
alexrichey, line 69: does this affect the zipfile on BYTES?

alexrichey (approved):
Just one nitpicky Q, and a question about the zipfile. Otherwise good!

products/cpdb/python/projects_in_geographies.py
damonmcc: since this was creating a zip within the primary zip, it shouldn't impact metadata/distribution

we haven't been distributing these csvs so I don't think the file `projects_in_geographies.zip` (now left unzipped) was specified in metadata

dcpy/connectors/edm/publishing.py
damonmcc: not that I can tell. seems like the logical branches would be
- default value `s3.MAX_FILE_COUNT` which is already an `int`
- a number passed in via CLI like `300` which would I think come in as a `str` so `int("300")`
- a non-number passed in via CLI like `number` which would throw a valid error when doing `int("number")`

alexrichey: ah, so what I'm asking is whether we should just let `typer` handle coercion/validation. Which would give us the benefit that the CLI would show the right type in the --help section for the command

damonmcc: oops I misunderstood what you meant by "signature". I'll change this to let `typer` handle it

damonmcc:
@alexrichey I made that CLI signature change

PR #1096, opened 2024-08-26
Ar p2 validation

alexrichey:


dcpy/test/lifecycle/package/test_column_validation.py
```diff
@@ -105,72 +107,33 @@ def _fake_row(columns: list[models.Column]):
...
 
 
 # AR note: I mostly use this to conveniently get a fake dataset
 # in a Jupyter Notebook.
 def _generate_fake_dataset_from_test_md(row_count: int):
-    dataset = metadata.package.get_dataset("primary_csv")
-    return generate_fake_dataset(row_count, columns=dataset.get_columns(metadata))
-
-
-def test_socrata_column_overrides_wkb_geom():
```
alexrichey, line 119: This is tested elsewhere pretty extensively, so removing it.

dcpy/test/lifecycle/package/test_packages_validation.py
```diff
@@ -19,67 +18,37 @@ def _get_colp_md():
...
 
-    validation = validate.validate_package(COLP_PACKAGE_PATH, overridden_md)
-    assert (
-        len(validation.errors) == 1
-    ), f"An error should have been found for the missing attachment. Found: {validation.errors}"
-
-    assert fake_attachment_name in validation.errors[0].message
-
-
-def test_destination_overrides():
```
alexrichey, line 47: This is tested elsewhere pretty extensively, so removing it.

dcpy/lifecycle/package/validate.py
```diff
@@ -106,12 +106,12 @@ def _is_geom_poly(s):
...
 col_validators = {
     "bbl": lambda df, col_name: df[~df[col_name].str.match(r"^\d{10}$")],
-    "integer": lambda df, col_name: df[~df[col_name].apply(_is_int)],
-    "double": lambda df, col_name: df[~df[col_name].apply(_is_float_or_double)],
     "wkb": lambda df, col_name: df[~df[col_name].apply(_is_valid_wkb)],
     "geom_point": lambda df, col_name: df[~df[col_name].apply(_is_geom_point)],
     "geom_poly": lambda df, col_name: df[~df[col_name].apply(_is_geom_poly)],
+    "integer": lambda df, col_name: df[~df[col_name].apply(_is_int)],
     "decimal": lambda df, col_name: df[~df[col_name].str.match(r"^-?\d*\.\d*$")],
+    "number": lambda df, col_name: df[~df[col_name].str.match(r"^-?\d*")],
```
alexrichey: Adding a generic number type, for when any type of number will do.

dcpy/lifecycle/package/validate.py
```diff
@@ -103,12 +110,14 @@ def _is_geom_poly(s):
 
 col_validators = {
     "bbl": lambda df, col_name: df[~df[col_name].str.match(r"^\d{10}$")],
-    "integer": lambda df, col_name: df[~df[col_name].apply(_is_int)],
-    "double": lambda df, col_name: df[~df[col_name].apply(_is_float_or_double)],
     "wkb": lambda df, col_name: df[~df[col_name].apply(_is_valid_wkb)],
```
alexrichey, line 113: We should probably remove the geometry types - leaving that til we've figured out what we actually want to do.

dcpy/test/lifecycle/package/test_column_validation.py
```diff
@@ -58,24 +58,26 @@ def bbl(boro_code, block, lot):
...
     for c in columns:
         if c.data_type == "bbl":
-            found_bbl_name = c.name
+            found_bbl_name = c.name or ""
         elif c.values:
-            row[c.name] = random.choice(c.values)[0]
+            row[c.name] = random.choice(c.values).value
         else:
-            val = fakes[c.data_type]()
+            val = fakes[c.data_type or ""]()
```
alexrichey, line 80: appeasing type checker. `data_type` is nullable because of how inheritance works in the DatasetColumn. I'm not entirely happy with that...

dcpy/lifecycle/package/validate.py
```diff
@@ -211,95 +203,110 @@ def validate_df(
...
-) -> DatasetFileValidation:
+    csv_path: Path, columns: list[dataset_md.DatasetColumn]
+) -> list[ValidationError]:
     df = pd.read_csv(csv_path, dtype=str)
-    return validate_df(df, dataset, metadata)
+    return validate_df(df, columns)
 
 
 def validate_shapefile(
-    shp_path: Path, dataset: models.DatasetFile, metadata: models.Metadata
```
fvankrieken, line 242: If these are becoming the same, we could use one of our utils to read in df/gdf from filepath - then this just becomes one "validate_file" function.

I don't actually think we're quite there though - `read_data_to_df` in `utils/data.py` is a little more tailored to ingest. But it'd be nice to align that sort of logic with this metadata logic

dcpy/lifecycle/package/validate.py
```diff
@@ -126,14 +126,17 @@ def _is_geom_poly(s):
...
+    *,
+    ignore_columns: list[str] | None = None,
 ) -> list[ValidationError]:
     """Validate a dataframe against a metadata file."""
     df_stringified_nulls = df.fillna("")
 
     errors = []
 
-    column_names = {c.name for c in columns}
+    column_names = {c.name or "" for c in columns}
```
fvankrieken, line 143: There a reason we don't want `{c.name for c in columns if c}` here?

dcpy/lifecycle/package/validate.py
```diff
@@ -76,20 +77,23 @@ def _is_valid_wkb(g):
...
 
 
+_decimal_regex = re.compile(r"^-?\d*\.\d*$")
+
+
+def _is_decimal(s):
+    return bool(_decimal_regex.match(s)) if s else False
+
+
+def _is_number(s):
```
fvankrieken, line 93: What's the rational for having ints, decimals, and numbers?

dcpy/lifecycle/package/validate.py
alexrichey: That's very true. Added a note about [follow-on work](https://github.com/orgs/NYCPlanning/projects/27/views/4?filterQuery=is%3Aissue++-project%3A*%5Bbacklog%5D*&pane=issue&itemId=68576094) for when we swap in Pandera

dcpy/lifecycle/package/validate.py
alexrichey: Yeah, I've added a comment about that in the last commit. It was mostly about appeasing the type-checker around c.name's nullability. But it points to the need to change the Column model.

dcpy/lifecycle/package/validate.py
alexrichey: So IIRC we'd settled on `ints` and `decimans` in our talks about ingest validation. I added number just for the case when we don't know or care whether there's a decimal in the number. IIRC it was due to laziness on my part about validating something in the test case. I'm kind of ambivalent about keeping it - open to nixing it. Also looping in @damonmcc and @sf-dcp for thoughts.

fvankrieken: Gotcha - makes sense on the validation side, though in terms of the actual data I think `decimal` is fine as a catch-all for numbers. Counter argument might still be that we want our metadata to be as descriptive as possible of the actual structure of the data, but as a user I would prefer to just see "decimal" even if it's a mix of floats and ints

alexrichey: That's interesting... one of the reasons for the current implementation is to distinguish between, say, "10" and "10.0", which touches on both formatting and underlying implementations. I'm wondering if it makes sense to distinguish between those, or allow them both to be decimals. It seemed vaguely important when I wrote this, but now I'm not so sure.

damonmcc: I'm down to drop `number` so that while validating we're forced to learn (and then accurately declare) the type, but I'd understand if you wanna keep the flexibility for now

and since columns usually can't (and shouldn't?) have mixed types, if a column has both "9.5" and "10" in it it'd expect to see/use them as `["9.5", "10.0"]`

sf-dcp: agree w/@damonmcc

PR #1097, opened 2024-08-27
add checked/freshness timestamps to ingest, various tweaks

fvankrieken:
Definitely just go through by commit. 6 is the meat of this one.

### Commits
1. rename ingest dataset `name` to `id`. This is more in line with our product metadata - "dob_job_applications" is more an identifier than a human-readable name. It'd be nice to also have names (that could end up in outputs, like how GIS actually uses the source_data_versions for GFT) as to make them more informative to a wider audience.
2. remove custom datetime serialization. By default, when calling `model_dump(mode="json")` this is the behavior anyways. Since we've gone in that direction when dumping to files, this is now unnecessary
3. remove the `s3.exists` util and replace it with two more specific utils, to check folders or objects existences
4. expand tests for `lifecycle.build.load`. Just trying to bump up coverage.
5. cleanup logic for recipe paths - I'd had a really ugly compromise of models not having knowledge of a constant in `recipes.py` that required always providing a "prefix" whenever trying to get a path out of these objects. I've made it more functional to get around this
6. Main point of this PR. Adds a `check_timestamps` field to ingest configs, which are a log of when there have been attempts to update this file which did not happen because it already exists and data have not changed. There are some tweaks to recipes/publishing that support the new behavior in ingest
    - recipes: add a few utils - `set_latest` to copy something to the latest folder, and `update_freshness`, which takes a timestamp and uses it to append to the `check_timestamps`. For `archive_dataset`, throw error if it already exists
    - publishing: add a new `exists` util and expand the errors that are returned by `get_build_metadata` (this latter change seems a bit unrelated to this commit - must have accidentally rebased it in.)
    - ingest: if the version of a dataset being archived already exists, compare the just-processed parquet with the previous parquet. If there is a difference, an error is thrown (NOTE - extra logic in this commit that I'm leaving in this commit for posterity if I feel like coming back to it. Removed in next commit). If they are identical, update the config to log this action but do not overwrite.
7. remove patching behavior in previous commit
8. a couple little test coverage bumps. Specific edge cases that aren't covered by other tests.


sf-dcp:
Commit 6 logic looks great to me. A couple of thoughts: possibly using the `Version` object to deal with recipe versions (at least would expect Socrata versions to align with our defined Version syntax); just throw an error when data from of supposedly same version, from old and new pulls, differ instead of patching

alexrichey:
@fvankrieken Approach in commit #6 makes sense. Maybe there are more immutable ways to do it, but this seems more convenient.

fvankrieken:
> @fvankrieken Approach in commit #6 makes sense. Maybe there are more immutable ways to do it, but this seems more convenient.

Yeah in some ways logging would be a more appropriate approach here, but given our current workflows there's a really nice simplicity to just updating this field in the config file, at least on the more "read" side of things

fvankrieken:
WRT to the "patching" of datasets where the version is constant but data has diverged - @sf-dcp had the thought that it might make more sense right now just punt the issue (paraphrasing). It's currently not a very well-scoped problem, and it might make more sense to address it as it comes up - if it does occur, 1. just throw an error, and 2. maybe look into what's actually changing.

I think this maybe makes more sense than the awkward code that's here now for a problem that we don't really know how much it will occur

dcpy/lifecycle/ingest/run.py
```diff
@@ -8,6 +9,53 @@
...
+            )
+        if "." in config.dataset.version:
+            raise ValueError(
+                f"Archived dataset '{config.dataset_key}' already exists and has different data. "
+                "Due to period in version name, cannot patch automatically. "
+                "Specify version manually."
+            )
+        patch = 0
+        patched = config.dataset
+        while recipes.exists(patched):
```
alexrichey: seems like a pretty sensible and reusable function to add to recipes?

fvankrieken: This part is actually just scrapped for now in the next commit, sorry

fvankrieken: But I agree - if we end up going this route, this should be a recipe function called `patch` or something like that

dcpy/test/lifecycle/builds/test_load.py
```diff
@@ -83,15 +106,174 @@ def test_import_parquet(self, _patched):
...
+
+    def test_unresolved_filetype(self):
+        ds = InputDataset(id="test", version="1")
+        with pytest.raises(
+            Exception, match="Cannot import a dataset without a resolved file type"
+        ):
+            load.import_dataset(ds, None)
+
+
+@patch("dcpy.connectors.edm.recipes.fetch_dataset", return_value=RESOURCES_DIR)
```
alexrichey, line 161: I don't think I've ever seen so many patches 😳

Maybe time to think about integration tests, vs unit test?

fvankrieken: Mayyyybe

alexrichey (approved):
Sweet!

PR #1098, opened 2024-08-28
update generate_metadata_assets to use v2

fvankrieken:


alexrichey (approved):
Apologies for leaving it in a broken state, and thank you for fixing!

Issue #976, opened 2024-07-05
Pluto 24v3


sf-dcp:
## Main tasks

- [x] source data extracted
- [x] draft build succeeded
- [x] draft build passed QA
- [x] data packaged and distributed

#### Updated 2x a year typically in June and December
- [x] **dcp_colp** (check [here](https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-colp.page)) 

#### Open data automated pull

> Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/data_library_open_data.yml) to see the latest run

- [x] **dsny_frequencies**
- [x] **dpr_greenthumb**
- [x] **lpc_historic_districts**
- [x] **lpc_landmarks**

#### DOF Automated Pull and Number of Buildings

- [x] **pluto_pts** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_pts.yml))
- [x] **pluto_input_geocodes** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_pts.yml))
- [x] **pluto_input_cama_dof** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_cama.yml))
- [x] **pluto_input_numbldgs** (Check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/pluto_input_numbldgs.yml))

### Updated with Quarterly updates (check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/data_library_quarterly.yml))

- [x] **dcp_cdboundaries_wi**
- [x] **dcp_ct2010**
- [x] **dcp_cb2010**
- [x] **dcp_ct2020**
- [x] **dcp_cb2020**
- [x] **dcp_school_districts**  
- [x] **dcp_councildistricts_wi**  
- [x] **dcp_firecompanies**  
- [x] **dcp_policeprecincts**
- [x] **dcp_healthareas**  
- [x] **dcp_healthcenters**

### Updated with Zoning Taxlots 

(check [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/zoningtaxlots_dataloading.yml) for latest run).

These are all produced by GIS, who typically update them sometime in the first week of each month.
Check in with them before archiving with data library

- [x] **dof_dtm**
- [x] **dof_shoreline**
- [x] **dof_condo**
- [x] **dcp_commercialoverlay**
- [x] **dcp_limitedheight**
[...]

PR #1099, opened 2024-08-28
PLUTO: manually correct 2020 census tract 

sf-dcp:
During #976, GIS identified a few BBLs whose 2020 census tract values changed in PLUTO 24v3. Amanda advised to correct BBL 4142600001 (JFK airport lot) for PLUTO 24v3 version; the rest of BBLs are under investigation. 

### 🚨 Feedback needed: 
we haven't corrected for census tract and I'm using a brand new reason for the correction `Geosupport error - CensusTract`. We have only corrected geosupport errors with a reason `Geosupport error - CD`. Please let me know if it should be changed to something else.


fvankrieken:
Looks like the right approach. An unfortunate amount of boilerplate at the moment

damonmcc (approved):

PR #1104, opened 2024-09-03
Data Lifecycle Monitoring table

### LINKED ISSUE #958: logging of data promotions
Related [discussion](https://github.com/NYCPlanning/data-engineering/discussions/1083)

sf-dcp:
Related discussion on implementation [here](https://github.com/NYCPlanning/data-engineering/discussions/1083).

Closes #958. 

## What
1. Refactor `upload` CLI which is exclusively responsible for uploading product builds to S3. I moved all CLI logic into a separate function and added tests. This is done to increase test coverage.

2. Update "event" functions (upload build, promote, publish) to return a `ProductKey`. Previously, these functions didn't return anything. The returned value can be used for logging, specifically identifying old and new stages. For example, the function `publish` now returns `PublishKey`

3. Implement logging of data lifecycle events (upload build to s3, promote to draft folder, and publish) in db
    * Create pydentic `EventLog` model to represent an event. All the model fields go into monitoring table. I chose the model instead of event individual fields for validating data and keeping myself sane. 
    * Create a helper function `publishing.log_event_in_db()` to enter `EventLog` to database
    * Update build, promote to draft, publish functions to log an event using `publishing.log_event_in_db()`
    * The db table for lifecycle monitoring was manually created. There is no code creating the table - don't see a need for it. 
    * Create **unit** tests for `publishing.log_event_in_db()`. I couldn't find something similar to `moto` package but for `posgres` (or any database really) to mock interactions with a database.

## Testing functions interacting with a database
* As noted in # 3, I implemented unit tests for logging an event in db; these tests really just check whether a call was made to execute a query given various conditions. It doesn't check whether a table exists or a record is added correctly. To test the latter, we need integration tests. Since we don't have db integration tests yet, it would be nice to first decide on an approach and implement these tests uniformly in a separate PR.

* Some thoughts on how we could implement integration tests for postgres  code. 
   * use a local host db inside of a container or on our local machines to create test databases
   * another option is to create a test db in our production db and run tests there. Don't love this option personally because then results may not be deterministic due to connection issues
   * use ORM models to interact with a database. These models can also be used to set up test tables.

## Table
I ran [`build`](https://github.com/NYCPlanning/data-engineering/actions/runs/10799515995/job/29955436907), [`promote_to_draft`](https://github.com/NYCPlanning/data-engineering/actions/runs/10799695340/job/29956026749), and [`publish`](https://github.com/NYCPlanning/data-engineering/actions/runs/10799737146/job/29956165579) GH actions on `db-template` product. Resulting table (you can find this table under `edm-qaqc` db, `product_data.event_logging` table) below: 

image

image

One of the records came from this PR tests.

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -273,13 +273,16 @@ def upload_build(
 
     if build_key.build not in IGNORED_LOGGING_BUILDS:
         version = get_version(build_key)
+        run_details = metadata.get_run_details()
         event_metadata = EventLog(
-            event_type=EventType.BUILD,
+            event=EventType.BUILD,
             product=build_key.product,
-            release_version=version,
-            new_path=build_key.path,
+            version=version,
+            path=build_key.path,
             old_path=None,
-            run_details=metadata.get_run_details(),
+            timestamp=run_details.timestamp,
+            runner_type=run_details.type,
+            runner=run_details.runner_string,
         )
         log_event_in_db(event_metadata)
 
@@ -330,13 +333,16 @@ def promote_to_draft(
     if not keep_build:
         s3.delete(BUCKET, source)
 
+    run_details = metadata.get_run_details()
     event_metadata = EventLog(
-        event_type=EventType.PROMOTE_TO_DRAFT,
+        event=EventType.PROMOTE_TO_DRAFT,
         product=draft_key.product,
-        release_version=draft_key.version,
-        new_path=draft_key.path,
+        version=draft_key.version,
+        path=draft_key.path,
         old_path=build_key.path,
-        run_details=metadata.get_run_details(),
+        timestamp=run_details.timestamp,
+        runner_type=run_details.type,
+        runner=run_details.runner_string,
     )
     log_event_in_db(event_metadata)
 
@@ -454,13 +460,16 @@ def publish(
         )
         logger.info(f"Downloaded build_metadata.json from {publish_key.path}")
 
+    run_details = metadata.get_run_details()
     event_metadata = EventLog(
-        event_type=EventType.PUBLISH,
+        event=EventType.PUBLISH,
         product=publish_key.product,
-        release_version=draft_key.version,  # this is release version, not patched
-        new_path=publish_key.path,
+        version=draft_key.version,  # this is release version, not patched
+        path=publish_key.path,
         old_path=draft_key.path,
-        run_details=metadata.get_run_details(),
+        timestamp=run_details.timestamp,
+        runner_type=run_details.type,
+        runner=run_details.runner_string,
     )
     log_event_in_db(event_metadata)
 
@@ -670,7 +679,7 @@ def log_event_in_db(event_details: EventLog) -> None:
         logger.info("DEV_FLAG env var found, skipping event metadata logging")
         return
     logger.info(
-        f"Logging event '{event_details.event_type}' metadata for product {event_details.product} in db..."
+        f"Logging event '{event_details.event}' metadata for product {event_details.product} in db..."
     )
     pg_client = postgres.PostgresClient(database=LOGGING_DB, schema=LOGGING_SCHEMA)
     query = f"""
@@ -682,13 +691,13 @@ def log_event_in_db(event_details: EventLog) -> None:
     pg_client.execute_query(
         query,
         product=event_details.product,
-        version=event_details.release_version,
-        event=event_details.event_type.value,
-        path=event_details.new_path,
+        version=event_details.version,
+        event=event_details.event.value,
+        path=event_details.path,
         old_path=event_details.old_path,
-        timestamp=event_details.run_details.timestamp,
-        runner_type=event_details.run_details.type,
-        runner=event_details.run_details.runner_string,
+        timestamp=event_details.timestamp,
+        runner_type=event_details.runner_type,
+        runner=event_details.runner,
         custom_fields=json.dumps(event_details.custom_fields),
     )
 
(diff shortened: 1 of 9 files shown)


dcpy/connectors/edm/publishing.py
```diff
@@ -30,6 +30,22 @@
 BUCKET = PUBLISHING_BUCKET
 
 BASE_DO_URL = f"https://cloud.digitalocean.com/spaces/{BUCKET}"
+LOGGING_DB = "edm-qaqc"
+LOGGING_SCHEMA = "product_data"
+LOGGING_TABLE_NAME = "event_logging"
+PRODUCTS_TO_LOG = [
```
damonmcc, line 36: this seems like a good idea to have a "whitelist" of products to log, rather than a list of products to ignore. should prevent surprise over-logging

dcpy/connectors/edm/publishing.py
```diff
@@ -230,13 +246,59 @@ def legacy_upload(
...
+    event_metadata = EventLog(
+        event_type=EventType.BUILD,
+        product=build_key.product,
+        release_version=version,
+        new_path=build_key.path,
+        old_path=None,
+        run_details=metadata.get_run_details(),
+    )
+    if build_key.build not in IGNORED_LOGGING_BUILDS:
+        log_event_in_db(event_metadata)
```
damonmcc, line 290: (nits) I think this could all go under the `if build_key.build not in IGNORED_LOGGING_BUILDS:` clause

I was gonna suggest `EventLog` take a `ProductKey` as an argument to reduce the number of args and having to do `version = get_version(build_key)`. but maybe we don't want models to be dependent on other models?

damonmcc: although, I do see now that `dcpy/models/lifecycle/builds.py` imports other models

dcpy/test/connectors/edm/test_publishing.py
```diff
@@ -112,6 +114,90 @@ def test_legacy_upload(
         assert not s3.exists(TEST_BUCKET_NAME, f"{TEST_PRODUCT_NAME}/{prefix}/{path}")
 
 
+@patch("dcpy.connectors.edm.publishing.BUILD_NAME", TEST_BUILD)
```
damonmcc, line 117: interesting use of `patch`, this effectively patches an envar?

dcpy/connectors/edm/publishing.py
alexrichey: I do think the general move has been towards making dcpy _more_ agnostic about the products, so I'm hesitant about this living here. I'd prefer that it were defined in a configuration, similar to how @fvankrieken configured buckets in that recent PR.

dcpy/test/connectors/edm/test_publishing.py
sf-dcp: no, this line here overrides `BUILD_NAME` variable in `publishing.py`. I tried to do an envar but it wasn't working

alexrichey:
@sf-dcp Do you have a SQL statement, or a python script to create the logging table? We should store that in a database migrations folder.

sf-dcp:
> @sf-dcp Do you have a SQL statement, or a python script to create the logging table? We should store that in a database migrations folder.

no, I created the table manually in our db. The SQL statement is stored under the table properties, DDL section. Where can I find the database migration folder?

dcpy/connectors/edm/publishing.py
sf-dcp: @alexrichey, do you think `LOGGING_DB`, `LOGGING_SCHEMA`, `LOGGING_TABLE_NAME`, and `IGNORED_LOGGING_BUILDS` variables should be moved to the config file as well?

dcpy/test/connectors/edm/test_publishing.py
fvankrieken: There's some weirdness with env vars and reloading the configuration module - I think this approach makes sense, though also happy to chat about in which cases reloading the configuration module works and in which cases it doesn't actually reload the env var

dcpy/models/lifecycle/builds.py
```diff
@@ -101,6 +102,22 @@ class LoadResult(BaseModel, extra="forbid"):
     datasets: dict[str, ImportedDataset]
 
 
+class EventType(StrEnum):
+    BUILD = "build"
+    PROMOTE_TO_DRAFT = "promote_to_draft"
+    PUBLISH = "publish"
+
+
+class EventLog(BaseModel, extra="forbid"):
```
alexrichey, line 111: I think we want this class to mirror the structure of the postgres table exactly, for both insertion and querying, and maybe even creating the table itself. So I think we should probably either have the `run_details` be a json field in the db, or add columns to the model for `timestamp`, `runner_type`, and `runner`.

dcpy/connectors/edm/publishing.py
sf-dcp: Yeah, I was thinking of this approach initially. But I ultimately decided to set up `EventLog` like this because: 1) `promote_to_draft` and `publish` steps need both "old" and "new" Product Keys and 2) having external interactions with s3 or env variables (i.e. getting a version from s3 for `build` step or getting `run_details` from envars) would add more complexity.

alexrichey:
> > @sf-dcp Do you have a SQL statement, or a python script to create the logging table? We should store that in a database migrations folder.
> 
> no, I created the table manually in our db. The SQL statement is stored under the table properties, DDL section. Where can I find the database migration folder?

It's our first foray into all this, so we gotta create one. Maybe in dcpy/migrations, for now? We can figure out a long-term structure down the road, I think.

dcpy/connectors/edm/publishing.py
alexrichey: Yeah, I think we should move those.

alexrichey (changes requested):
good stuff! Have some requested changes, but this is really cool.

alexrichey (approved):

PR #1107, opened 2024-09-04
Utilities to create dev buckets

fvankrieken:
Decided to spin this out from #1070 to reduce clutter - that one can stay focused on using buckets that exist already

Some complications with our hope to just quickly spin up dev buckets.

Copy commands are not allowed across clusters within a region for spaces buckets. We can choose what region a bucket is in, but cluster seems entirely internal to DO. This means that, with some tool (mc/rclone/boto3), we need to actually download/upload to get files from edm-recipes or publishing into a new bucket.

This has some drawbacks, primarily

- speed. On my personal comp, creating a dev bucket with latest of all recipe datasets needed for pluto took about an hour. Not the end of the world, not as lightweight and easy as I had hoped
- downloads count towards our monthly read bandwidth, which is 1TB. My run right now of latest of all pluto-related datasets and latest publish folder for pluto was ~5gb. I think cpdb and GFT would be larger. So maybe not a huge issue - I don't think anyone would make more than one of these a month, and it's only a bit bulkier than running a few builds (which we already do all the time)

These are both made worse by the fact that we currently have multiple file types for each source dataset, and different builds sometimes use different file types (pluto uses pg_dump, facdb uses csv) so it's a bit tough to limit this. I suppose I could use the recipe to determine which file types are needed and only download those. This could reduce this size (and also then the time to setup) to close to 1/3 - right now it just grabs the whole datasets/{ds_id}/latest folder

Tangent on reducing read bandwidth:
-  maybe that's also something to think about a bit with our nightly_qa job - I wouldn't be surprised if we're reading 10-15 gb per day with that, which would get us (monthy) to somewhere between 1/3 and 1/2 of our limit right there. Persisting latest of each recipe dataset in a db (at time of library archival) and only grabbing from s3 when an older version is needed might be a more sustainable (and performant) long term solution. That could be part of 1 db per build environment

Anyways - would love to chat about how we see ourselves using this just as a sanity check - I've gone in circles a bit in my vision here

### DIFF
--- a/admin/ops/setup_dev_bucket.py
+++ b/admin/ops/setup_dev_bucket.py
@@ -3,6 +3,7 @@ import importlib
 import os
 from pathlib import Path
 import typer
+from typing import Literal
 
 PROD_RECIPES_BUCKET = "edm-recipes"
 PROD_PUBLISHING_BUCKET = "edm-publishing"
@@ -24,6 +25,7 @@ from dcpy.lifecycle.builds import plan
 
 ROOT_PATH = Path(__file__).parent.parent.parent
 BUCKET_PREFIX = "de-dev-"
+ACL: Literal["private"] = "private"
 PRODUCTS = [
     "db-cbbr",
     "db-colp",
@@ -57,7 +59,7 @@ def resolve_latest_recipe(
         bucket=target_bucket,
         source_path=input.s3_folder_key("datasets") + "/",
         target_path=resolved.s3_folder_key("datasets") + "/",
-        acl="private",
+        acl=ACL,
     )
 
 
@@ -79,7 +81,7 @@ def resolve_latest_publish(target_bucket: str, data_product: str):
         bucket=target_bucket,
         source_path=input.path + "/",
         target_path=resolved.path + "/",
-        acl="private",
+        acl=ACL,
     )
 
 
@@ -98,7 +100,7 @@ def clone_recipe(
         target_bucket=target_bucket,
         source_path=f"{key.s3_path("datasets")}/",
         target_path=f"{key.s3_path("datasets")}/",
-        acl="private",
+        acl=ACL,
     )
     if version == "latest":
         resolve_latest_recipe(target_bucket, dataset_id)
@@ -155,7 +157,7 @@ def clone_data_products(
             target_bucket=target_bucket,
             source_path=key.path + "/",
             target_path=key.path + "/",
-            acl="private",
+            acl=ACL,
         )
 
         resolve_latest_publish(target_bucket, product)
@@ -176,7 +178,7 @@ def clone_data_product_by_key(
         target_bucket=target_bucket,
         source_path=key.path + "/",
         target_path=key.path + "/",
-        acl="private",
+        acl=ACL,
     )
     if include_recipe_datasets:
         for _index, row in publishing.get_source_data_versions(key).iterrows():
@@ -193,7 +195,7 @@ def setup(id: str, clean: bool = False) -> str:
     if bucket not in [
         bucket["Name"] for bucket in s3.client().list_buckets()["Buckets"]
     ]:
-        s3.client().create_bucket(Bucket=bucket, ACL="private")
+        s3.client().create_bucket(Bucket=bucket, ACL=ACL)
     else:
         if not clean:
             raise Exception(
--- a/admin/ops/setup_dev_bucket.py
+++ b/admin/ops/setup_dev_bucket.py
@@ -18,7 +18,6 @@ from dcpy.models.connectors.edm.publishing import (
     PublishKey,
 )
 from dcpy.utils import s3
-from dcpy.utils.logging import logger
 from dcpy.connectors.edm import recipes, publishing
 from dcpy.lifecycle.builds import plan
 
--- a/admin/ops/setup_dev_bucket.py
+++ b/admin/ops/setup_dev_bucket.py
@@ -138,8 +138,8 @@ def clone_recipes_latest(
     datasets = get_subfolders(bucket=PROD_RECIPES_BUCKET, prefix="datasets/")
     if include:
         datasets = include
-    if exclude == "exclude":
-        datasets = [ds for ds in datasets if ds not in datasets]
+    if exclude:
+        datasets = [ds for ds in datasets if ds not in exclude]
 
     for ds in datasets:
         clone_recipe(target_bucket, ds)
--- a/admin/ops/setup_dev_bucket.py
+++ b/admin/ops/setup_dev_bucket.py
@@ -24,7 +24,7 @@ from dcpy.lifecycle.builds import plan
 
 
 ROOT_PATH = Path(__file__).parent.parent.parent
-BUCKET_PREFIX = "de-dev-"
+DEV_BUCKET_PREFIX = "de-dev-"
 ACL: Literal["private"] = "private"
 PRODUCTS = [
     "db-cbbr",
@@ -54,7 +54,7 @@ def resolve_latest_recipe(
     resolved = recipes.Dataset(name=ds, version=recipes.get_latest_version(ds))
     os.environ["RECIPES_BUCKET"] = PROD_RECIPES_BUCKET
     importlib.reload(configuration)
-    assert target_bucket.startswith(BUCKET_PREFIX)
+    assert target_bucket.startswith(DEV_BUCKET_PREFIX)
     s3.copy_folder(
         bucket=target_bucket,
         source_path=input.s3_folder_key("datasets") + "/",
@@ -191,7 +191,7 @@ def setup(id: str, clean: bool = False) -> str:
     Creates bucket if it doesn't exist.
     If clean flag provided, deletes all objects within bucket
     """
-    bucket = f"{BUCKET_PREFIX}{id}"
+    bucket = f"{DEV_BUCKET_PREFIX}{id}"
     if bucket not in [
         bucket["Name"] for bucket in s3.client().list_buckets()["Buckets"]
     ]:
@@ -204,14 +204,14 @@ def setup(id: str, clean: bool = False) -> str:
         aws_s3_endpoint = os.environ.get("AWS_S3_ENDPOINT")
         aws_access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
         aws_secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
-        assert bucket.startswith(BUCKET_PREFIX)
+        assert bucket.startswith(DEV_BUCKET_PREFIX)
         bucket_obj = boto3.resource(
             "s3",
             aws_access_key_id=aws_access_key_id,
             aws_secret_access_key=aws_secret_access_key,
             endpoint_url=aws_s3_endpoint,
         ).Bucket(bucket)
-        assert bucket_obj.name.startswith(BUCKET_PREFIX)
+        assert bucket_obj.name.startswith(DEV_BUCKET_PREFIX)
         bucket_obj.objects.all().delete()
     return bucket
 
@@ -225,7 +225,7 @@ def _recipe_single(
     dataset_id: str = typer.Option(None, "--dataset", "-d"),
     version: str = typer.Option(None, "--version", "-v"),
 ):
-    clone_recipe(f"{BUCKET_PREFIX}{target_bucket_id}", dataset_id, version)
+    clone_recipe(f"{DEV_BUCKET_PREFIX}{target_bucket_id}", dataset_id, version)
 
 
 @app.command("recipes_all")
@@ -234,7 +234,7 @@ def _recipes(
     include: list[str] | None = typer.Option(None, "--include", "-i"),
     exclude: list[str] | None = typer.Option(None, "--exclude", "-e"),
 ):
-    clone_recipes_latest(f"{BUCKET_PREFIX}{target_bucket_id}", include, exclude)
+    clone_recipes_latest(f"{DEV_BUCKET_PREFIX}{target_bucket_id}", include, exclude)
 
 
 @app.command("product")
@@ -242,7 +242,7 @@ def _publishing_single(
     target_bucket_id: str = typer.Option(None, "--bucket-id"),
     product: str = typer.Option(None, "--product", "-p"),
 ):
-    clone_data_products(f"{BUCKET_PREFIX}{target_bucket_id}", [product])
+    clone_data_products(f"{DEV_BUCKET_PREFIX}{target_bucket_id}", [product])
 
 
 @app.command("publishing_build_key")
@@ -253,7 +253,7 @@ def _build_key(
     include_sources: bool = typer.Option(False, "--include-sources"),
 ):
     clone_data_product_by_key(
-        f"{BUCKET_PREFIX}{target_bucket_id}",
+        f"{DEV_BUCKET_PREFIX}{target_bucket_id}",
         BuildKey(product=product, build=build),
         include_recipe_datasets=include_sources,
     )
@@ -268,7 +268,7 @@ def _draft_key(
     include_sources: bool = typer.Option(False, "--include-sources"),
 ):
     clone_data_product_by_key(
-        f"{BUCKET_PREFIX}{target_bucket_id}",
+        f"{DEV_BUCKET_PREFIX}{target_bucket_id}",
         DraftKey(product=product, version=version, revision=revision),
         include_recipe_datasets=include_sources,
     )
@@ -282,7 +282,7 @@ def _publish_key(
     include_sources: bool = typer.Option(False, "--include-sources"),
 ):
     clone_data_product_by_key(
-        f"{BUCKET_PREFIX}{target_bucket_id}",
+        f"{DEV_BUCKET_PREFIX}{target_bucket_id}",
         PublishKey(product=product, version=version),
         include_recipe_datasets=include_sources,
     )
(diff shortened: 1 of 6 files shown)


sf-dcp:
👋 is this ready for review? noticed it's not a draft but no one has been assigned

fvankrieken:
Not quite

utilities/setup_dev_bucket.py
fvankrieken, line 1: Put this in its own folder for now because it really feels like this doesn't belong in `dcpy`. I know we currently have some stuff in there that's a little more focused on our cloud env than more pure code, but this felt like it was less of a gray area for me. Would love thoughts on where we want stuff like this to live

fvankrieken:
Some cli tests

### setup
creates empty dev bucket when doesn't exist
image
image

fails on existing bucket without specifying clean option
image
image

cleans bucket successfully
image
image

fvankrieken:
Requesting reviews but definitely in a slightly brainstorming spot. Mainly thinking about
- do we feel this is safe?
- where does this code live, and what sort of tests do we want?
- do we want this in a gha? or just run locally

And just in general what other thoughts folks have

utilities/setup_dev_bucket.py
```diff
@@ -0,0 +1,351 @@
...
+    "db-developments",
+    "db-facilities",
+    "db-green_fast_track",
+    "db-template",
+    "db-pluto",
+    "db-zoningtaxlots",
+]
+
+
+def copy_folder(
```
alexrichey, line 42: seems like a nice s3 util? Also, should you use a `tempfile` folder locally, rather than `./tmp/dev_bucket_setup`?

utilities/setup_dev_bucket.py
alexrichey: Had that same thought/question when I glanced through the previous iteration of the PR. It might be nice to have a spot in `dcpy` for environment/test-environment infrastructure setup, maybe colocated with the new database migrations folder? Basically just one spot for infra setup.

utilities/setup_dev_bucket.py
```diff
@@ -0,0 +1,351 @@
...
+        path=f"{key.s3_path("datasets")}/",
+    )
+    if version == "latest":
+        resolve_latest_recipe(target_bucket, dataset_id)
+
+
+def clone_recipes_latest(
+    target_bucket: str,
+    datasets: list[str] | None = None,
+    dataset_mode: Literal["include", "exclude"] | None = None,
```
alexrichey, line 131: I think a kwargs for `exclude_datasets` and `include_datasets` would be a little more straightforward.

fvankrieken: That is far more sensible

utilities/setup_dev_bucket.py
fvankrieken: Well there is an s3 for copying a folder, but it does it without download/uploading. And while it offers ability to copy across buckets, this apparently isn't a guarantee that it could work (given the issues I ran into)

So I should probably
- edit existing function in `s3` to not have `target_bucket` as argument
- add this function to `s3` as `copy_folder_across_buckets` or something like that. This could even attempt to copy just using the s3 api to avoid downloading, and fallback to downloading.

And definitely use a tempfile folder, yes

Thoughts?

utilities/setup_dev_bucket.py
fvankrieken: This seems like a good thing to actually discuss as a group

utilities/setup_dev_bucket.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def resolve_latest_publish(target_bucket: str, data_product: str):
+    """
+    Within a dev bucket, resolves/copies the data product currently in "publish/latest"
+    into its versioned folder within "{product}/publish"
+    """
+    os.environ["PUBLISHING_BUCKET"] = target_bucket
+    importlib.reload(configuration)
+    assert PROD_PUBLISHING_BUCKET != PROD_PUBLISHING_BUCKET
```
sf-dcp, line 93: is the line 93 correct?

fvankrieken: Lol. Oops!

utilities/setup_dev_bucket.py
alexrichey: Totally makes sense.

sf-dcp:
Q about possibly persisting latest recipes to db: would we read from db only when working a dev bucket or in our prod as well? 

I really like the idea. Also would love to declutter recipes to one format

fvankrieken:
> Q about possibly persisting latest recipes to db: would we read from db only when working a dev bucket or in our prod as well?
> 
> I really like the idea. Also would love to declutter recipes to one format
> 
> Re: nightly QA using dev buckets. I think the cost of DO outweighs the benefit of building as we do now. I think testing in dev buckets during development will mainly occur on our own branches anyways.

So thoughts nightly qa and latest recipes in db were unrelated to this pr and dev buckets, sorry. Was just on a tangent thinking about trying to not go too high on our read bandwidth for DO spaces

sf-dcp:
To your other questions: personally, I don't think there is a need for a GHA rn. In terms of safety: I wonder if there is a way to modify access to read only for `publishing` bucket. Otherwise looks good!

dcpy/utils/s3.py
```diff
@@ -367,14 +378,19 @@ def delete(bucket: str, path: str) -> None:
     client_.delete_object(Bucket=bucket, Key=path)
 
 
-def get_filenames(bucket: str, prefix: str) -> set[str]:
```
fvankrieken, line 370: Bit of a change in behavior - before, we were getting all filenames without any directories included regardless of level. So if we had
```
/folder1
    /folder2
        file1.txt
    file2.txt
```
And we called this on `folder1` we'd get `{ "file1.txt", "file2.txt" }`. More confusingly, if they were named the same, we'd only get back one of them.

Now, we'd get back `{ "folder2/file1.txt", "file2.txt" }` which seems much more intuitive

dcpy/utils/s3.py
```diff
@@ -367,14 +378,19 @@ def delete(bucket: str, path: str) -> None:
...
     return {
-        obj["Key"].split("/")[-1]
+        obj["Key"].removeprefix(prefix)
         for obj in list_objects(bucket, prefix)
-        if obj["Key"].split("/")[-1] != ""
+        if obj["Key"].removeprefix(prefix) != ""
     }
 
 
+def get_filenames(bucket: str, prefix: str) -> set[str]:
```
fvankrieken, line 417: Bit of a change in behavior from the old `get_filenames` (which I've replaced with this and the above `get_suffixes`, which is a bit more general). Before, we were getting all filenames without any directories included regardless of level. So if we had
```
/folder1
    /folder2
        file1.txt
    file2.txt
```
And we called this on `folder1` we'd get `{ "file1.txt", "file2.txt" }`. More confusingly, if they were named the same, we'd only get back one of them.

Now, we'd get back `{ "folder2/file1.txt", "file2.txt" }` which seems much more intuitive

dcpy/utils/s3.py
```diff
@@ -101,6 +102,22 @@ def client() -> S3Client:
     )
 
 
+def list_buckets() -> list[str]:
+    """Get list of buckets for the defined connection variables"""
+    return [b["Name"] for b in client().list_buckets()["Buckets"]]
+
+
+def get_bucket_region(bucket: str) -> str:
```
fvankrieken, line 110: Very glad that this can be determined programmatically, though see docstring. recipes and publishing are both "nyc3b" while the new dev buckets that get spun up are in "nyc3d"

fvankrieken:
Ready for re-review.

Also can't tell if I mentioned it anywhere - idea here is that we have "id"d dev buckets, single letter identifiers. So I have "de-dev-f", Damon can have "de-dev-d", etc. 

We will never have two engineers whose first names start with the same letter so this will never be an issue

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,340 @@
...
+    logger.info(f"Copying {source_bucket}/{path} to ./tmp")
+    s3.download_folder(
+        source_bucket, prefix=path, export_path=Path("./tmp/dev_bucket_setup")
+    )
+    logger.info(f"Copying ./tmp to {target_bucket}/{path}")
+    s3.upload_folder(
+        target_bucket,
+        Path(f"./tmp/dev_bucket_setup/{path}"),
+        Path(path),
+        "private",
```
sf-dcp: private acl... who does this exclude/include?

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,340 @@
...
+    into its versioned folder under "datasets/{dataset_id}"
+    """
+    os.environ["RECIPES_BUCKET"] = target_bucket
+    importlib.reload(configuration)
+    assert PROD_RECIPES_BUCKET != PROD_RECIPES_BUCKET
+    input = recipes.Dataset(name=ds, version="latest")
+    resolved = recipes.Dataset(name=ds, version=recipes.get_latest_version(ds))
+    os.environ["RECIPES_BUCKET"] = PROD_RECIPES_BUCKET
+    importlib.reload(configuration)
+    assert target_bucket.startswith(BUCKET_PREFIX)
```
sf-dcp, line 55: tiiiny nit: can we please rename as `DEV_BUCKET_PREFIX`

fvankrieken: Sure thing

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,340 @@
...
+    target_bucket: str,
+    ds: str,
+):
+    """
+    Within a dev bucket, resolves/copies the dataset currently in "latest"
+    into its versioned folder under "datasets/{dataset_id}"
+    """
+    os.environ["RECIPES_BUCKET"] = target_bucket
+    importlib.reload(configuration)
+    assert PROD_RECIPES_BUCKET != PROD_RECIPES_BUCKET
```
sf-dcp: I'm guessing this should be configuration.RECIPES_BUCKET

admin/ops/setup_dev_bucket.py
fvankrieken: Just means that you need credentials to access - public read means anyone with url can access. Figured might as well make it private. Maybe at least I'll pull that out to a constant at the top of the script

admin/ops/setup_dev_bucket.py
fvankrieken: Fixed in a later commit - I would just look at the whole PR

sf-dcp: oh sorry. I was going one by one

sf-dcp:
looks great! I feel a bit uneasy in terms of bucket safety. Tests would take a while to write, so perhaps we could add the assert statement "bucket starts with dev bucket prefix" in either CLI commands or across all functions that copy objects to target bucket? I've seen it in your code

fvankrieken:
> looks great! I feel a bit uneasy in terms of bucket safety. Tests would take a while to write, so perhaps we could instead add the assert statement "bucket starts with dev bucket prefix" in either CLI commands or across all functions that copy objects to target bucket? I've seen it in your code

My approach with the cli was that you can never provide the whole bucket name - whatever you provide will get the prefix added to it. So to that end, none of the CLI commands _should_ be able to target a production bucket

fvankrieken:
> looks great! I feel a bit uneasy in terms of bucket safety. Tests would take a while to write, so perhaps we could instead add the assert statement "bucket starts with dev bucket prefix" in either CLI commands or across all functions that copy objects to target bucket? I've seen it in your code

@sf-dcp - check the latest commit for safety. I've removed the s3 module import and only import it within the two functions that use it (other than importing one "read" function generally). Now calls to copy_folder only use a little wrapper function that verifies that the target bucket is a dev bucket.

sf-dcp (approved):
🎉

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,342 @@
...
+    Specific datasets can be excluded or included with `include` or 'exclude_dataset`
+    If `include` is specified, only the specified datasets are cloned
+    """
+    if include and exclude:
+        raise ValueError("Only one of `include` and `exclude` may be provided")
+    datasets = get_subfolders(bucket=PROD_RECIPES_BUCKET, prefix="datasets/")
+    if include:
+        datasets = include
+    if exclude == "exclude":
+        datasets = [ds for ds in datasets if ds not in datasets]
```
alexrichey, line 142: also, `if ds not in exclude`?

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,342 @@
...
+    By default, targets all datasets
+    Specific datasets can be excluded or included with `include` or 'exclude_dataset`
+    If `include` is specified, only the specified datasets are cloned
+    """
+    if include and exclude:
+        raise ValueError("Only one of `include` and `exclude` may be provided")
+    datasets = get_subfolders(bucket=PROD_RECIPES_BUCKET, prefix="datasets/")
+    if include:
+        datasets = include
+    if exclude == "exclude":
```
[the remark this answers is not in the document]
fvankrieken: Oh oops. Thank you

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,342 @@
...
+):
+    """
+    Clone latest versions of various recipe datasets from production bucket to dev bucket
+    By default, targets all datasets
+    Specific datasets can be excluded or included with `include` or 'exclude_dataset`
+    If `include` is specified, only the specified datasets are cloned
+    """
+    if include and exclude:
+        raise ValueError("Only one of `include` and `exclude` may be provided")
+    datasets = get_subfolders(bucket=PROD_RECIPES_BUCKET, prefix="datasets/")
```
alexrichey, line 138: should we use `recipes.get_all_versions`, to encapsulate folder structure (`datasets/` specifically) within the recipes module?

fvankrieken: That would be something like `get_all_datasets` which currently doesn't exist, but I could make that function in `recipes`

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,342 @@
...
+def resolve_latest_recipe(
+    target_bucket: str,
+    ds: str,
+):
+    """
+    Within a dev bucket, resolves/copies the dataset currently in "latest"
+    into its versioned folder under "datasets/{dataset_id}"
+    """
+    os.environ["RECIPES_BUCKET"] = target_bucket
+    importlib.reload(configuration)
```
alexrichey, line 68: What's the reason for reloading these? This does make me think that we need a more comprehensive way to load env vars. And maybe that module can have an `ensure_dev` function, to error out if detects prod variables. Not something we need to address now, but something to think about.

admin/ops/setup_dev_bucket.py
alexrichey: probably ideal, but also super minor

admin/ops/setup_dev_bucket.py
```diff
@@ -0,0 +1,342 @@
...
+    if bucket not in [
+        bucket["Name"] for bucket in s3.client().list_buckets()["Buckets"]
+    ]:
+        s3.client().create_bucket(Bucket=bucket, ACL=ACL)
+    else:
+        if not clean:
+            raise Exception(
+                f"bucket {bucket} already exists. specify '-c' to clean contents of bucket"
+            )
+        aws_s3_endpoint = os.environ.get("AWS_S3_ENDPOINT")
```
alexrichey, line 220: one alternative possibility, curious your thoughts: What if in our s3 utils, we add delete functionality that throws an exception if pointed at the prod bucket. Would allow us to unit test it.

alexrichey:
Just one lingering thought... I think we should spend some time on comprehensive environment configuration, ideally from a .env file. Then maybe for dev stuff like this, we can make use of a context manager, e.g. [this](https://stackoverflow.com/questions/2059482/temporarily-modify-the-current-processs-environment) to say something like:

```
with dev_environment():
    s3.my_dangerous_action()

s3.back_to_business_as_usual()
```

fvankrieken:
> this

I like this idea a lot. I think we then also need to get into env vars that declare the overall scope - ability to label a run as `PROD` or `DEV` separate from any variables for specific buckets.

Because in this case, this script would always run in a "DEV" environment, but still needs to read heavily from prod buckets (ideally using our API, not custom functions). So in addition to what you're suggesting for a context manager to wrap a dangerous action, we could also have annotations, say `@write` or something like that for any s3 functions that put or delete. Then these would make sure that either we're writing to a dev bucket (always allowed), or that run env and bucket are BOTH `PROD`. Meanwhile, `@read` functions could run regardless of environment (so maybe don't need an annotation) so that scripts like this can still grab data from prod

Sounds like another roadmapping discussion to me

alexrichey:
Also, I somehow didn't realize it until this morning, but looks like you can just make a DO user with the ability to read/write, but not delete. (Unless I'm missing it?)

fvankrieken:
> Also, I somehow didn't realize it until this morning, but looks like you can just make a DO user with the ability to read/write, but not delete. (Unless I'm missing it?)

Can you share a link?

fvankrieken:
> > Also, I somehow didn't realize it until this morning, but looks like you can just make a DO user with the ability to read/write, but not delete. (Unless I'm missing it?)
> 
> Can you share a link?

Generally, delete is just part of write I think. Also, DO spaces are very much based on permissions per user, but across all buckets I think. 

Going to merge this for now, but agree that we should re-evaluate `env` stuff holistically

PR #1108, opened 2024-09-04
Pull buckets into env/configuration, expand edm connector testing

fvankrieken:
Cleaning up #1070 and moving to a new pr for tidiness

Since I'm decoupling from #1107 , behavior for dev buckets here just assumes that a different bucket exists, and uses that bucket if specified. There's no logic to setup a dev bucket, or have dev buckets interact with prod buckets in any way

## commits
1. just adds more tests for recipes
2. the main meat of this PR. Much simplified from #1070 - now we just define recipe/publishing buckets with env vars and those get used.
3. downstream effects from 2 - just figured it was easier to review 2 without this included
4. two additional publishing tests
5. move a function from template db to publishing and write a test, then clean up the way templatedb interacts with publishing
6. gha stuff to be able to run dev_flag workflows

#### library
https://github.com/NYCPlanning/data-engineering/actions/runs/10725945882
image

#### ingest
https://github.com/NYCPlanning/data-engineering/actions/runs/10726165124
image

#### build
prod build (just make sure I didn't royally screw anything up) - [template build](https://github.com/NYCPlanning/data-engineering/actions/runs/10725640475)
image

dev build - [pluto](https://github.com/NYCPlanning/data-engineering/actions/runs/10726169696)

[pluto build (dev bucket)](https://github.com/NYCPlanning/data-engineering/actions/runs/10725719087)
image


dcpy/lifecycle/ingest/run.py
```diff
@@ -49,6 +50,8 @@ def run(
     if not skip_archival:
         recipes.archive_dataset(config, staging_dir / config.filename, latest=latest)
 
+    return config
```
fvankrieken, line 53: This change isn't used in recipes test after latest refactor but still is a reasonable thing - after archival, return the config

dcpy/connectors/edm/publishing.py
```diff
@@ -1,32 +1,34 @@
...
 )
 from dcpy.models.lifecycle.builds import BuildMetadata
+from dcpy.utils import s3, git, versions
+from dcpy.utils.logging import logger
 
-from botocore.exceptions import ClientError
+assert (
+    PUBLISHING_BUCKET
+), "'PUBLISHING_BUCKET' must be defined to use edm.publishing connector"
+BUCKET = PUBLISHING_BUCKET
```
fvankrieken, line 30: mypy weirdness - just asserting it doesn't resolve the type (from `Optional`), but asserting and binding to a new variable does

.github/workflows/ingest_single.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     env:
       BUILD_NAME: ${{ github.head_ref || github.ref_name }}
       RECIPES_BUCKET: ${{ inputs.dev_bucket && format('de-dev-{0}', inputs.dev_bucket) || 'edm-recipes' }}
+      PUBLISHING_BUCKET: ${{ inputs.dev_bucket && format('de-dev-{0}', inputs.dev_bucket) || 'edm-publishing' }}
```
fvankrieken, line 43: Needed because GIS datasets come from the publishing bucket

apps/qa/src/pages/edde/helpers.py
```diff
@@ -3,12 +3,15 @@
...
 import streamlit as st
 
+from dcpy.configuration import PUBLISHING_BUCKET
 from dcpy.utils import s3
 from dcpy.connectors import github
 from dcpy.connectors.edm import publishing
 
 REPO_NAME = "db-equitable-development-tool"
 PRODUCT = "db-eddt"
+assert PUBLISHING_BUCKET, "PUBLISHING_BUCKET must be defined"
```
damonmcc, line 13: (nit/idea) would doing `os.environ["PUBLISHING_BUCKET"]` instead of `os.environ.get("PUBLISHING_BUCKET")` in `configuration.py` effectively be the same thing as all the downstream assert statments?

I often can't decide between environ[] and environ.get() and maybe `mypy` is a significant factor here

fvankrieken: My thought was that we only need to throw an error if we're actually trying to use that part of dcpy - we could have a job which calls our geocoding utilities (which totally exist) with cli that doesn't need to interact with s3 buckets, so it's fine if those env vars aren't set

fvankrieken: There's probably a slightly more elegant approach than what I have done

dcpy/connectors/edm/recipes.py
```diff
@@ -245,6 +248,11 @@ def purge_recipe_cache():
 
 
 def log_metadata(config: library.Config):
+    # long term, it probably makes more sense to have db interactions isolated by environment as well
```
sf-dcp, line 251: oh I see.

sf-dcp: Btw, are we running `log_metadata` when using `ingest`, or only under `library`?

dcpy/test/connectors/edm/test_publishing.py
sf-dcp, line 1: are we now using our own actual buckets or fake buckets for tests?

sf-dcp (approved):
Nice job! Also want to compliment on great structure of PR, super easy to review!

dcpy/connectors/edm/recipes.py
fvankrieken: Currently `library`, though I should change that. About to work more on `ingest` so I'll incorporate logging there

dcpy/test/connectors/edm/test_publishing.py
fvankrieken: fake buckets

PR #1119, opened 2024-09-09
use `RECIPES_BUCKET` envar in all scheduled actions

damonmcc:
follow-up to https://github.com/NYCPlanning/data-engineering/pull/1108/

fixes all scheduled action failures [listed here](https://github.com/NYCPlanning/data-engineering/issues?q=is%3Aissue+is%3Aopen+scheduled+action+in%3Atitle)


alexrichey (approved):
Thanks. I think that we should put our default env vars (e.g. `edm-recipes`) into a dotenv file and load them that way, to avoid cases like this. I think it'd be pretty trivial - doesn't need to happen now, but I think we should schedule it for this sprint.

PR #1128, opened 2024-09-11
group earlier in ztl cm logic, keep per geoms

fvankrieken:
## Motivation

Overall changes here are nicely summarized by this lot. @caseysmithpgh fyi

image

lot 2039437501, though in the newest dtm it is now 2039430100

Our cutoffs for commercial overlays are
- 10% or more of bbl is covered by an overlay type (such as "C1-2" - not specific overlay/shape)
- 50% or more of an overlay is on the lot

For the former, we obviously need to group by bbl and sum the percentages

The latter depends on what we mean. I checked in with @croswell81 and we agreed that it should be 50% or more of **a specific overlay** (or "instance" of a commercial overlay type).

So for this lot, nothing reaches the 10% cutoff, but due to the two C1-2 overlays and one C2-2 that are fully contained within the lot, it should be assigned both.

## Code changes and results

Main changes for this are in the first two commits. I've kept them as two because
1. aligns ztl with pluto
2. fixes logic in both ztl and pluto

Making these changes in ztl results in 9 rows changing - luckily it seems commercial overlays aren't too weird
|bblnew|co1new|co2new|co1prev|co2prev|
|------|------|------|-------|-------|
|2039430100|C2-2|C1-2|C1-2|C2-2|
|3035260021|C2-3|C1-3|C2-3||
|3070490015|C1-2||||
|3070720005|C2-4||||
|4000060045|C2-5||||
|4014710012|C1-3||||
|4024720015|C1-2||||
|5004880078|C2-3||||
|5046240020|C1-1||||

## Other commits
Some cleanup/performance tweaks. Builds are down 3ish minutes!
https://github.com/NYCPlanning/data-engineering/actions/runs/10817948380


products/zoningtaxlots/models/intermediate/int__validdtm.sql
fvankrieken, line 1: This really wasn't doing anything - staging dtm already makes zones valid, and we only have polygons. Null geometries (which currently don't exist but I think we could foresee) simply won't get spatially joined. I ran a build and verified no changes to final table

products/zoningtaxlots/models/intermediate/int__zoningmapindex.sql
```diff
@@ -20,14 +20,14 @@ zoningmapper AS (
         n.zoning_map,
         ST_AREA(
             CASE
-                WHEN ST_COVEREDBY(ST_MAKEVALID(p.geom), n.geom) THEN p.geom
```
fvankrieken, line 27: Already made valid in the staging table

alexrichey:
@fvankrieken curious what spurred you changing this? (it's not jumping out at me from `2039430100`) Was it that lots like `3070490015` weren't having the correct Commercial Overlays applied?

PR #1132, opened 2024-09-13
downgrade gdal python package

fvankrieken:
Small tweaks due to recompiling python packages


python/requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.12
 # by the following command:
 #
-#    ./bash/compile_python_packages.sh python
+#    pip-compile --output-file=python/requirements.txt python/requirements.in
```
damonmcc, line 5: can u compile via `./bash/compile_python_packages.sh python`? especially because it deletes `requirements.txt` before compiling which is needed for the `.txt` to fully reflect the resolved `.in`

fvankrieken: Ah didn't realize that was default behavior. But if you run with `--upgrade` you don't need to delete requirements.txt

fvankrieken: realpath doesn't work properly on my mac, so the script doesn't run properly. Do you have no issues with it?

fvankrieken: Or at least the realpath command in the script doesn't work for me

damonmcc: damn i forgot about macbook realpath pain

bash/compile_python_packages.sh
```diff
@@ -2,24 +2,20 @@
...
+path_to_requirements="python"
 
 # Set GDAL verion
-sed -i -e "s/GDAL==.*$/GDAL==$(gdal-config --version)/g" ${path_to_requirements}/requirements.in
+sed -i "" -e "s/GDAL==.*$/GDAL==$(gdal-config --version)/g" ${path_to_requirements}/requirements.in
 
 # Compile requirements
 echo -e "🛠 compiling from ${path_to_requirements}/requirements.in"
-CUSTOM_COMPILE_COMMAND="./${RELATIVE_SCRIPTPATH} ${path_to_requirements}" python3 -m piptools compile --output-file=${path_to_requirements}/requirements.txt ${path_to_requirements}/requirements.in
+CUSTOM_COMPILE_COMMAND=$0 python3 -m piptools compile --upgrade --output-file=${path_to_requirements}/requirements.txt ${path_to_requirements}/requirements.in
```
fvankrieken, line 18: @damonmcc any reason to not just drop realpath and do this?

damonmcc (approved):

bash/compile_python_packages.sh
```diff
@@ -2,24 +2,20 @@
...
 python3 -m pip install --upgrade pip-tools wheel
 
-# Delete existing requirements file to ensure full dependency resolution
-echo -e "🛠 deleting ${path_to_requirements}/requirements.txt"
-rm --force ${path_to_requirements}/requirements.txt
+path_to_requirements="python"
 
 # Set GDAL verion
-sed -i -e "s/GDAL==.*$/GDAL==$(gdal-config --version)/g" ${path_to_requirements}/requirements.in
+sed -i "" -e "s/GDAL==.*$/GDAL==$(gdal-config --version)/g" ${path_to_requirements}/requirements.in
```
fvankrieken, line 14: this didn't have issue in debian for me but in zsh on my mac was leaving a temp file behind

bash/compile_python_packages.sh
damonmcc: not I don't mind dropping realpath. nice to not have to specify `python` at the end when running the script

PR #1139, opened 2024-09-17
Metadata Only Deployments

alexrichey:
Enable metadata (dataset attributes + attachments) only deployments to Socrata. An example use case is LION, for which we'll be redeploying metadata post-review, but not the datasets themselves.

### DIFF
--- a/dcpy/connectors/socrata/publish.py
+++ b/dcpy/connectors/socrata/publish.py
@@ -42,6 +42,10 @@ def _simple_auth():
     return (SOCRATA_USER, SOCRATA_PASSWORD)
 
 
+def _socratapy_client():
+    return SocrataPy(Authorization(SOCRATA_DOMAIN, SOCRATA_USER, SOCRATA_PASSWORD))
+
+
 def _socrata_request(
     url,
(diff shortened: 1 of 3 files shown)


dcpy/connectors/socrata/publish.py
```diff
@@ -42,6 +42,10 @@ def _simple_auth():
     return (SOCRATA_USER, SOCRATA_PASSWORD)
 
 
+def _socratapy_client():
```
alexrichey, line 45: pulled this out so I could use it in notebooks.

dcpy/connectors/socrata/publish.py
```diff
@@ -613,36 +615,37 @@ def push_dataset(
...
-            data_source.push_socrata_column_metadata(
-                overridden_dataset_md.dataset.columns
-            )
-        except Exception as e:
-            # Upating column Metadata is tricky, and there's still some work to be done
-            logger.error(
-                f"""Error Updating Column Metadata! However,
-                the Dataset File was uploaded and the revision can still be applied manually, here: {rev.page_url}
-                Error: {e}"""
+    if not metadata_only:
```
alexrichey, line 618: code under this branch wasn't modified at all, fyi.

PR #1146, opened 2024-09-19
Bytes Packaging With OTI_XLSX Generation

alexrichey:
Adds ability to assemble from Bytes and generate an OTI Data Dictionary if a file of type `oti_data_dictionary` is included in the package files. 

This is very much a first pass at package assembly, and it's not ideal for a few reasons, but primarily because we're hardcoding in knowledge of BYTES, OTI, and a specific file format (the OTI Data Dictionary). I imagine down the line, the metadata will contain a set of assembly steps. But I think we can figure when we stop pulling from BYTES, and begin _actually assembling_ packages. 

Lingering thoughts: 
- I'm not sure if `File.type` is the correct place to denote that it's an OTI_XLSX. (Though I suppose there is an analog to how we treat shapefiles. A `shapefile` is to a `zipfile` as an `oti_data_dictionary` is to an XSLX.) It just means `File.type` refers not to a raw file type, but something above that.

Only the second commit _really_ changes anything, so commit-by-commit is recommended.


dcpy/test/lifecycle/assemble/test_assemble.py
```diff
@@ -31,3 +35,41 @@ def test_unpack_package(metadata, tmp_path, package_path):
...
+    metadata = md.Metadata.from_path(
+        colp_package_path / "metadata.yml", template_vars={"version": "24b"}
+    )
+    metadata.files.append(
+        md.FileAndOverrides(
+            file=md.File(
+                id="oti_xlsx", filename=OTI_DD_FILENAME, type=OTI_METADATA_FILE_TYPE
+            )
+        )
+    )
```
damonmcc: would adding a file with type `oti_data_dictionary` to `colp_package_path / "metadata.yml"` effectively do the same thing as this append?

I'd understand not wanting to modify it to reflect every possible test though

dcpy/test/lifecycle/assemble/test_assemble.py
```diff
@@ -31,3 +35,41 @@ def test_unpack_package(metadata, tmp_path, package_path):
...
+
+    assemble.assemble_dataset_from_bytes(
+        metadata, destination_id="socrata", out_path=tmp_path
+    )
+
+    attachments_path = MOCK_PULLED_PACKAGE_PATH / "attachments"
+    assert attachments_path.exists(), "Sanity check that the mock side_effect works"
```
damonmcc: is this assertion moot after `assemble.assemble_dataset_from_bytes` succeeds and covered by the next assertion?

dcpy/lifecycle/package/assemble.py
```diff
@@ -96,4 +98,48 @@ def unzip_into_package(
...
+        out_path, dataset_metadata, destination_id, unpackage_zips=True
+    )
+
+    oti_data_dictionaries = [
+        f.file
+        for f in dataset_metadata.files
+        if f.file.type == oti_xlsx.OTI_METADATA_FILE_TYPE
+    ]
+    if len(oti_data_dictionaries) > 0:
+        # minor TODO: what to do if there are multiple oti data dictionaries specified?
```
fvankrieken, line 193: This is where maybe it'd be nice to expand some validation utils - just a function called `list.exactly_one` that gives you the single element of a list or throws an error.

In this case, do we want to throw an error? in the strange case that we query bytes and somehow get two

fvankrieken: Or sorry - this is based on what's defined in our metadata, not what's found on bytes, correct?

Might also be worth adding pydantic validation to the class if that's a real constraint. Happy to leave that for now though

damonmcc (approved):

dcpy/test/lifecycle/assemble/test_assemble.py
alexrichey: yeah, that's a good (and open) question... I'd initially started by adding the OTI file there, but `dcpy/test/lifecycle/package/test_packages_validation.py` is using the COLP metadata package, and it tests validating the full package. So since that file is missing (ie not generated yet) it'll cause another test to fail. (TLDR: wanted to leave that as a valid/complete package)

dcpy/test/lifecycle/assemble/test_assemble.py
alexrichey: yeah, I left it in because that's probably the first thing that's gonna fail when we mess with these tests again. Could effectively remove it, but I tend to like those little checkpoints

dcpy/lifecycle/package/assemble.py
alexrichey: > Or sorry - this is based on what's defined in our metadata, not what's found on bytes, correct?

Yeah, what's in the metadata. I've changed this to just generate each OTI file listed, because it seems more consistent. (I don't think there's a real use case for it, but this is all going to change once we have other assets to generate)

PR #1149, opened 2024-09-20
Fix QA app

sf-dcp:
First, the QAQC app wouldn't load pages because it didn't have `PUBLISHING_BUCKET` and `RECIPES_BUCKET` env variables. 

After fixing the env bug, this bug showed up:

image

Adding setup tools package resolves the issue. [Link](https://de-qaqc.nycplanningdigital.com/?page=PLUTO) to sample page in the QA app.

### DIFF
--- a/admin/run_environment/requirements.in
+++ b/admin/run_environment/requirements.in
@@ -42,6 +42,7 @@ PyYAML>=6.0.1
 types-PyYAML>=6.0.1
 rich>=13.4.2
 ruff
+setuptools
 socrata-py==1.1.13
 sqlalchemy
 sqlalchemy-stubs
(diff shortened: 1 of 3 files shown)


sf-dcp:
I'm not sure why sqlfluff is failing for cbbr and, presumably, for ztl...

fvankrieken:
> I'm not sure why sqlfluff is failing for cbbr and, presumably, for ztl...
> 
> edit: I tried removing quotes from column names in cbbr that were failing the test, and I got like 40 more formatting errors.

Version got bumped in python requirements - if you want, you can pin it for now. Or you can bump the version of sqlfluff locally and run it use the latest tweaks

fvankrieken:
You know, maybe it just makes sense to drop the commit for compiling packages - for whatever reason, setuptools still isn't even in the compiled requirements - maybe because it's sort of a "meta" package in some way.

fvankrieken (approved):

PR #1158, opened 2024-09-24
Product Metadata + LION tweaks

alexrichey:
Adds product-level metadata!

Product-level md allows for defaults for datasets attributes, as well as the ability to instantiate metadata for all the listed products. This also adds the ability to find tagged destinations for a products datasets, as well as a script to package from bytes, and push to tagged destinations. 

All told, this allows us to run something like 
```
python -m dcpy.cli lifecycle scripts package_and_dist from_bytes_to_tagged_socrata ../product-metadata/products/lion 24c -t my-tag
```

which will find all dataset-destinations tagged as `my-tag` in the metadata, packaged them up, and distribute them. The impetus is that for LION, we have 30 or so new unpublished datasets. (The others are either old, or in production) 

I've tagged the new destinations as `socrata_unpublished`. As @jackrosacker updates the metadata, we can deploy them all with a single click. 🎉

The result will look something like this, if `publish` is specified. 
image

in the case of errors, it will look like this:
image
(incidentally, we'll need to write some looping code to wait for columns to process in the socrata publish connector)

Commit-by-Commit should mostly be good, as it's a bit of a grab-bag of tweaks.


dcpy/connectors/socrata/publish.py
```diff
@@ -542,11 +542,12 @@ class DestinationUses:
 
 class SocrataDestination:
     dataset_file_id: str
-    attachment_ids: set[str] = set()
```
alexrichey, line 545: 🐛

damonmcc:
one of the commit messages is "True up Dataset attributes and overrides"

what's "True up" mean?

dcpy/test/resources/mini_lion_prod_with_datasets/metadata.yml
```diff
@@ -0,0 +1,11 @@
...
+
+attributes:
+  display_name: LION Product Display Name
+  description: The lion dataset blah blah
+  publishing_frequency: "{{ lion_prod_level_pub_freq }}"
+  publishing_purpose: "legal compliance."
+
+datasets:
+  - pseudo_lots
+  - school_districts
```
damonmcc, line 11: in product metadata, is `datasets` a list of all datasets or just those that should inherit attributes from the product?

as I type the question though, it does feels like all datasets would inherit attributes from the product lol

dcpy/models/product/metadata.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class DefaultDatasetAttributes(SortedSerializedBase):
+    """Anything overrideable at the dataset level"""
+
+    contains_address: bool | None = None
+    date_made_public: str | None = None
+    publishing_purpose: str | None = None
+    potential_uses: str | None = None
+    publishing_frequency: str | None = None  # TODO: picklist values
```
damonmcc, line 16: would love to help with the list of publishing frequency values later

dcpy/models/product/metadata.py
```diff
@@ -0,0 +1,74 @@
+from collections import defaultdict
+from pathlib import Path
+
+from dcpy.models.base import SortedSerializedBase, YamlWriter, TemplatedYamlReader
+from dcpy.models.product.dataset.metadata_v2 import Metadata as DatasetMetadata
+# from dcpy.models.product.dataset.metadata_v2 import DatasetAttributes
```
damonmcc, line 6: ✂️?

dcpy/test/resources/mini_lion_prod_with_datasets/metadata.yml
alexrichey: yeah, that's a good question. It should be all datasets... which also implies that it's not enough to just have a folder with metadata. The Dataset has to be listed here, or it's like it doesn't exist to the product

PR #1162, opened 2024-09-26
product metadata snippets

fvankrieken:
Reworked it a bit. Go by commit.

1. reworked basic logic that we'd talked about. Not quite implemented though - `DatasetColumnTemplate` class not actually used
2. consolidate the (now) 3 `DatasetColumn` types to 1, with no validation on id or name. This logic is moved up to the dataset validation call. I think the other clean approach here would be having separate classes for something that's just read in (and validates any issues with the actual raw file) and then also having a "computed" class that we can access, say, `column.name` in a type-safe way and knowing that it's `not None`.
3. (and onwards) pytest


alexrichey:
@fvankrieken I think this is a great start. I do agree that the logic, specifically around when we apply the defaults, feels not-quite-right. 

My main thought is that parsing/reading of the snippets and column defaults probably should happen outside the Metadata class, maybe a level up at the product metadata level. Then when we instantiate the `dataset metadata`, we can just pass along the snippets/column default classes. 

So maybe instead of using an env var, we just create class that represents the product metadata repo, that that class knows where the snippets live. (maybe we just hardcode that it's a folder called /snippets or something). 

Then the process of instantiating metadata looks like:
```
pmd_repo = ProductMetadataRepo(Path("./product-metadata"))
column_defaults = pmd_repo.shared_columns()
snippets = pmd_repo.snippets()

# instantiating metadata the tedious way
dataset_metadata = Metadata(..., column_defaults=column_defaults, template_vars = template_vars | snippets)

# the better way, which we'd have to implement, but which should do all the defaulting for you
pmd_repo.product("my_product").dataset("my_dataset")
```

In general I think it probably makes sense to instantiating dataset metadata within the context of the whole metadata repo, rather than individual yaml files. 

Anyhow, thanks, and let's chat tomorrow!

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -94,21 +93,7 @@ def _validate_colum_types(cls, v):
             assert v in get_args(COLUMN_TYPES)
         return v
 
-
-class DatasetColumn(DatasetColumnOverrides):
-    """Like DatasetColumnOverrides, but with constraints for non-null fields"""
-
-    @field_validator("name")
```
fvankrieken, line 101: I think these were never running before, based on the docs maybe it's something to do with these not being annotated as classmethods?

https://docs.pydantic.dev/latest/concepts/validators/#field-validators

fvankrieken:
@alexrichey - back atcha. Tests are failing for reasons noted, want to check in with you on the code changes before deciding what to do about them

dcpy/configuration.py
```diff
@@ -24,3 +24,5 @@
     "db-zoningtaxlots",
 ]
 IGNORED_LOGGING_BUILDS = ["nightly_qa", "compile_python_reqs"]
+
+PRODUCT_METADATA_SNIPPETS = env.get("PRODUCT_METADATA_SNIPPITS_PATH")
```
alexrichey, line 28: is this used anywhere now?

alexrichey:
> @alexrichey - back atcha. Tests are failing for reasons noted, want to check in with you on the code changes before deciding what to do about them

Yikes! Fixed the data_type issue in the product metadata repo, and now the tests are fine. Good validation for your changes!

fvankrieken:
Oops. Made an ill-advised commit last night that I didn't push and forgot about. It's fixed now, and still good for review

dcpy/models/product/dataset/metadata_v2.py
alexrichey: Thanks for wading into this. I was never quite sure how best implement nullability for the overrides. Ideally there's a shared set of fields for the class and overrides class, which suggest inheritance... but that's got some serious drawbacks. Probably has to be somewhat dynamic, which makes anything `mypy` related kind of a pain

dcpy/test/resources/test_product_metadata_repo_with_snippets/products/test_product/metadata.yml
```diff
@@ -0,0 +1,10 @@
+id: test_product
+
+attributes:
```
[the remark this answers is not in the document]
fvankrieken: Sure, I can add that to the tests

I also realize I'm also not testing the functions to read in the columns/snippets directly - I'll add that in too

dcpy/models/product/metadata.py
```diff
@@ -113,19 +122,43 @@ class OrgMetadata(SortedSerializedBase, extra="forbid"):
     root_path: Path
     template_vars: dict = Field(default_factory=dict)
     metadata: OrgMetadataFile
+    column_defaults: dict[tuple[str, str], DatasetColumn]
+
+    @classmethod
+    def get_yml(cls, path: Path):
```
alexrichey, line 128: I think I'd probably move this out to the module, and I'd definitely make it throw an exception if the path doesn't exist. Seems like a potential for masked errors otherwise?

dcpy/models/product/metadata.py
```diff
@@ -113,19 +122,43 @@ class OrgMetadata(SortedSerializedBase, extra="forbid"):
...
+    def get_string_snippets(cls, path: Path) -> dict:
+        yml = cls.get_yml(path / "snippets" / "strings.yml") or {}
+        if not isinstance(yml, dict):
+            raise ValueError("snippets must be valid yml dict, not array")
+        return yml
+
+    @classmethod
+    def get_column_defaults(cls, path: Path) -> dict[tuple[str, str], DatasetColumn]:
+        yml = cls.get_yml(path / "snippets" / "column_defaults.yml")
+        columns = TypeAdapter(list[DatasetColumn]).validate_python(yml or [])
```
alexrichey, line 145: This is a neat approach. I could see wanting a JSON schema for the snippets file, which might mean an actual SnippetsFile class. But this seems good for now

dcpy/test/models/product/test_metadata.py
```diff
@@ -122,3 +122,40 @@ def test_query_product_dataset_tags(test_metadata_repo: Path):
...
+    return resources_path / "test_product_metadata_repo_with_snippets"
+
+
+@pytest.fixture
+def dataset_with_snippets(test_metadata_repo_snippets: Path):
+    repo = md.OrgMetadata.from_path(
+        test_metadata_repo_snippets, template_vars={"version": "VERSION"}
+    )
+    product = repo.product("test_product")
+    return product.get_product_dataset("test_dataset", product.get_product_metadata())
```
alexrichey, line 138: This pattern of passing around the full Metadata is one that I introduced, and I wince every time I see it. Maybe an unfair ask, but is it possible to tweak things in the OrgMetadata and ProductFolder so that we can just do 
```
product = repo.product("test_product").dataset("test_dataset") 
```
and have that dataset be fully snippeted?

alexrichey (approved):
This is great. I had a few suggestions which can implement if you'd like. Would also be cool to add some snippets to the product-metadata repo.

dcpy/models/product/metadata.py
fvankrieken: In this case I made this just to not write the exact same codes for the string snippets/column defaults, though it obviously is a little out of place. And it should be valid to not have snippets defined. I'll add an underscore to the name for now if that sounds okay? Or could skip this helper and just have slightly repeated code

fvankrieken: In this case I made this just to not write the exact same codes for the string snippets/column defaults, though it obviously is a little out of place. And it should be valid to not have snippets defined. I'll add an underscore to the name for now if that sounds okay? Or could skip this helper and just have slightly repeated code

alexrichey:
Some lingering thoughts bigger thoughts and future enhancements. 

### Tests Metadata
The metadata resource are starting to sprawl a little bit. I also find it really frustrating when writing/reading tests that when I assert on something in the yaml, I either have to write a value out literally, e.g. `assert md.attributes.description == 'thing I put in yaml' or I have to do yaml parsing to grab that value. Neither approach really evolves very well. So I think we may have to work to consolidate.


### Templates by type and id
I think it works to reference them that way for now. I think there may be value in an approach like we talked about, where  we define templated columns as in a separate class, and in the metadata they look something like:
```
- id: my_templated_bbl
  template=true
```
Which we can deserialize easily, and JSON schemas will be happy.

dcpy/models/product/metadata.py
alexrichey: for sure, should be fine to not have snippets defined. This is more of a nit about having simple helper functions do one thing only (in this case, trying to read a yaml file), and having them complain when called incorrectly, e.g. when called with a nonexistent path. So I'd generally prefer to see something like:
```
snippets = read_yaml(snippets_path) if snippets_path.exists() else {}
```

it reminds me of a line I wrote, which I changed because of codecov. It went something like this:
```
def my_str_fn(s: str):
    if not s:
        return None
    else return s.some_str_fun(s)
```
Which I changed because it's up to the caller of `my_str_fn` to make sure they're passing in a valid string. (Well, I actually changed it because I was too lazy to write the unit test for the None case.)

dcpy/test/models/product/test_metadata.py
fvankrieken: See https://github.com/NYCPlanning/data-engineering/pull/1162/commits/7280c5aab84c676acc3af85a3da047781245a713

dcpy/models/product/metadata.py
fvankrieken: Yeah you're right. I just got rid of it... While it might be nice to have a `safe_load_path` throughout our repo, I didn't feel like making it just for these two lines.

dcpy/test/resources/test_product_metadata_repo_with_snippets/products/test_product/metadata.yml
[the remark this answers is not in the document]
fvankrieken: See tweaks here https://github.com/NYCPlanning/data-engineering/pull/1162/commits/873578c6a223bde8897805188c152753726ba40d

alexrichey (approved):
Looks great!

PR #1164, opened 2024-09-28
Compile python requirements

fvankrieken:


damonmcc:
only failing test is PLUTO (I think) because we haven't archived 24v3 from bytes yet

PR #1155, opened 2024-09-23
Tools to compare datasets, random ingest fixes

fvankrieken:
Definitely easiest to go commit-by-commit! **And read the whole PR description first**. For commits 5 and 6, might be good to look at some sample outputs below for big-picture context

Minor commits
1. recipe tweaks to import parquet files with geometry when possible. Otherwise, wkb ends up as a simple text column
2. a couple needed postgres utils
3. a default column cleaning step for ingest to align with library
4. make the directory provided as the ingest staging/processing dir if it doesn't exit

Bigger stuff
5. Create a pydantic model to support printing of little reports, which was helpful within jupyter notebooks. See tests for behavior or sample outputs below in the body of this PR
6. Meat of the comparison code. Largely two frameworks, one where we assume that we have some sort of unique key by which to examine the data, and one where we just compare all rows in a dataset.

Comparing bpl_libraries ingest vs library using the key column of "title" gives the following

```python
compare_ingest_and_library(
    "bpl_libraries",
    key_columns=["title"],
    build_schema="fvk_compare")
).report()
```
giving
```
________________________________________________________________________________
Row count:
    Left: 62
    Right: 62
________________________________________________________________________________
Column comparison:
    Column Comparison
    Columns in both:
        access
        address
        branchid
        branchlist
        bus
        closingmsg
        data_library_version
        hours
        ogc_fid
        path
        phone
        pickuploc
        subway
        tags
        title
    Column found in left only:
        friday
        latitude
        longitude
        monday
        position
        saturday
        sunday
        thursday
        tuesday
        wednesday
        wkb_geometry
    Column found in right only:
        Friday
        Monday
        Saturday
        Sunday
        Thursday
        Tuesday
        Wednesday
        geom
    Columns in both with different types:
        path
        left='character varying' right='text'
        pickuploc
        left='character varying' right='text'
        access
        left='character varying' right='text'
        closingmsg
        left='character varying' right='text'
        phone
        left='character varying' right='text'
        title
        left='character varying' right='text'
        tags
        left='character varying' right='text'
        address
        left='character varying' right='text'
        bus
        left='character varying' right='text'
        hours
        left='character varying' right='text'
        branchid
        left='character varying' right='text'
        subway
        left='character varying' right='text'
        branchlist
        left='character varying' right='text'
________________________________________________________________________________
Data comparison:
    Keyed Table Comparison
    Key columns:
        title
    Keys found in left only: None
    Keys found in right only: None
    Columns with changed values for specific keys: None
[...]

PR #1166, opened 2024-09-30
Finalize/clean up ingest models, add additional preprocessing

fvankrieken:
I'd say the __biggest__ thing for review here is still the models. See changes here https://github.com/NYCPlanning/data-engineering/pull/1166/files#diff-6fffbad770f67b00cdc414260a7922b1de4258230b2dab380fefbb7843401fd0

For model changes, easier to look at that file with overall changes than by commit. Other than that, easier to go by commit (see below the yml/json for commit summaries)

The new templates look like this
```yml
id: dcp_commercialoverlay
acl: public-read

attributes:
  name: DCP NYC Commercial Overlay Districts
  description: |
    Polygon features representing the within-tax-block limits for commercial overlay districts,
    as shown on the DCP zoning maps. Commercial overlay district designations are indicated in the OVERLAY attribute.
  url: https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-gis-zoning.page#metadata

ingestion:
  target_crs: EPSG:4326
  source:
    type: edm_publishing_gis_dataset
    name: dcp_commercial_overlays
  file_format:
    type: shapefile
    crs: EPSG:2263
  processing_steps:
  - name: rename_columns
    args:
      map: {"geom": "wkb_geometry"}

columns:
- id: overlay
  data_type: text
- id: shape_leng
  data_type: decimal
- id: shape_area
  data_type: decimal
- id: wkb_geometry
  data_type: geometry

```

And the computed config then looks like this
```json
{
    "id": "dcp_commercialoverlay",
    "version": "20240831",
    "crs": "EPSG:4326",
    "attributes": {
        "name": "DCP NYC Commercial Overlay Districts",
        "description": "Polygon features representing the within-tax-block limits for commercial overlay districts,\nas shown on the DCP zoning maps. Commercial overlay district designations are indicated in the OVERLAY attribute.\n",
        "url": "https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-gis-zoning.page#metadata"
    },
    "archival": {
        "acl": "public-read",
        "raw_filename": "dcp_commercial_overlays.zip",
        "archival_timestamp": "2024-09-30T16:23:51.693603-04:00"
    },
    "ingestion": {
        "target_crs": "EPSG:4326",
        "file_format": {
            "crs": "EPSG:2263",
            "encoding": "utf-8",
            "type": "shapefile"
        },
        "processing_steps": [
            {
                "name": "reproject",
                "args": {
                    "target_crs": "EPSG:4326"
                }
            },
            {
                "name": "rename_columns",
                "args": {
                    "map": {
                        "geom": "wkb_geometry"
                    }
                }
            }
        ],
        "source": {
            "type": "edm_publishing_gis_dataset",
            "name": "dcp_commercial_overlays"
        }
    },
    "columns": [
        {
            "id": "overlay",
            "data_type": "text"
        },
        {
            "id": "shape_leng",
            "data_type": "decimal"
        },
        {
            "id": "shape_area",
            "data_type": "decimal"
        },
        {
            "id": "wkb_geometry",
            "data_type": "geometry"
        }
    ],
    "run_details": {
        "type": "manual",
        "runner": {
            "username": "Finn van Krieken"
        },
        "timestamp": "2024-09-30T16:23:51.693603-04:00",
        "logged": false
    }
}
```

Other than that, it's helpful to go by commit

1. Some cleanup from #1155 - getting test coverage up and fixing a random literal
2. New ingest template that's been my guinea pig
3. Move the current "ingest" templates to a "dev" folder - in next PR, want to add logic to start using any found ingest templates, then fall back to library for our cli calls, but don't want to throw these ones which are sort of "in-progress" away (and didn't want to update them all right now without validating the data)
4. add `Attributes` section to `Template` and `Config` models, falling in line a little bit with our product metadata
5. use `SortedBase` across the models that get dumped to `config.json` for nice serialization. Make a few more submodels of config to simplify what's at the top level
6. adjustments for tests from commit 5 (just seemed easier to have these separate). Also tweaked `run.run` to make it easier to test (and aim it somewhere other than the "production" template folder)
7. For the four datasets that I've actually archived with ingest, move them to the production folder (and adjust to new structure)
8. Add columns to ingest templates. Include sthe most basic validation step, that all defined columns exist in the dataset. This is also where eventually column-specific checks could be specified
9. Fix a bug - gdfs get unhappy if you try to naively rename the geometry column (as in, just treating it as any other pandas column)
10. Add a default step that any dataset with geometry converts its geom to multi. Effectively same as calling `ST_MULTI` in postgis. See test cases for the utility for expected behavior. This seems like maybe a necessary evil for now - I'd prefer to keep geometry as they are, but this keeps things more consistent with library outputs (and makes sure things are consistent when we import to postgis, which can be more finicky/specific with types)


dcpy/lifecycle/scripts/ingest_validation.py
fvankrieken, line 1: Just made more sense to have these in `dcpy` - made it easier to access them in a notebook

dcpy/models/file.py
```diff
@@ -2,10 +2,11 @@
 from pydantic import BaseModel
```
sf-dcp, line 2: ✂️?

dcpy/models/lifecycle/ingest.py
```diff
@@ -38,55 +39,92 @@ class ScriptSource(BaseModel, extra="forbid"):
...
     processing_steps: list[PreprocessingStep] = []
 
+
+class Column(SortedSerializedBase):
+    id: str
+    data_type: Literal["text", "integer", "decimal", "geometry", "bool", "datetime"]
+    description: str | None = None
+
+
+class Template(BaseModel, extra="forbid"):
```
sf-dcp, line 79: this one will be eventually phased out, correct?

fvankrieken: No - this is the class that maps to the dataset template definitions

dcpy/models/lifecycle/ingest.py
```diff
@@ -38,55 +39,92 @@ class ScriptSource(BaseModel, extra="forbid"):
...
-    raw_filename: str
-    acl: recipes.ValidAclValues
-
-    target_crs: str | None = None
-
-    source: Source
-    file_format: file.Format
-    processing_mode: str | None = None
-    processing_steps: list[PreprocessingStep] = []
+    crs: str | None = None
```
sf-dcp, line 109: is this attribute here to accommodate the old template?

dcpy/models/file.py
fvankrieken: Well, `PointColumns` is technically still using it... but there's no reason that shouldn't just use sorted serialized base too. Thanks!

sf-dcp: oh I missed that

dcpy/models/lifecycle/ingest.py
fvankrieken: Template exists to simply read in and validate the dataset definition

Config is the "computed" config of a dataset. It has the template definition (attributes, instructions to ingest, etc) as well as the details of archival (where to find the archived raw file, timestamp, etc) and any other fields that need to be determined at runtime.

dcpy/models/lifecycle/ingest.py
fvankrieken: Nothing in this PR is really to accommodate anything old. These config objects are meant for reading data about an archived dataset, so with that I thought that `crs` was worthy of being a top-level field since it's quite relevant to how we use the data. Given that it's also stored in the parquet metadata, this is maybe up for debate

dcpy/models/lifecycle/ingest.py
```diff
@@ -45,20 +45,24 @@ class PreprocessingStep(BaseModel):
...
+
+
 class Template(BaseModel, extra="forbid", arbitrary_types_allowed=True):
     """Definition of a dataset for ingestion/processing/archiving in edm-recipes"""
 
     id: str
     acl: recipes.ValidAclValues
 
-    target_crs: str | None = None
+    attributes: DatasetAttributes | None = None
```
sf-dcp: maybe make it a required attribute? To avoid creation of new templates without relevant dataset info

fvankrieken: That seems like the right choice

dcpy/models/lifecycle/ingest.py
sf-dcp: I see. My thought is that crs should "live" in one place under ingestion to avoid 1) duplication and 2) any confusion whether`crs` should or shouldn't be specified in both places.

dcpy/test/lifecycle/ingest/__init__.py
sf-dcp, line 1: I haven't see this file. Nice!

dcpy/models/lifecycle/ingest.py
sf-dcp: I thought It's the `Config` class that's computed, not `Template`?

sf-dcp:
Ok, finished my review! Love how the new dataset templates look like and the models representing them make so much more sense. Great work. 

The first commit is a bit obscure to me since I missed the previous related PR - can't provide feedback

Questions:
* ongoing convo above about `Template` vs `Config`
* Could you please explain the need for the last commit creating a multi geometry? What is the benefit

alexrichey:
@fvankrieken One immediate though: would it make sense to publish JSON schemas for these models? (similar to `admin.ops.update_product_metadata_schema`)

dcpy/lifecycle/ingest/run.py
```diff
@@ -49,34 +50,46 @@ def run(
...
     skip_archival: bool = False,
     output_csv: bool = False,
+    template_dir: Path = configure.TEMPLATE_DIR,
 ) -> Config:
-    config = configure.get_config(dataset_id, version=version, mode=mode)
-    transform.validate_processing_steps(config.id, config.processing_steps)
+    config = configure.get_config(
+        dataset_id, version=version, mode=mode, template_dir=template_dir
+    )
+    transform.validate_processing_steps(config.id, config.ingestion.processing_steps)
```
damonmcc, line 58: is this redundant since `transform.validate_process_steps` is called during `transform.preprocess`?

and if we do wanna validate early, would it make sense to do it during `configure.get_config`?

damonmcc: talked about it and we'll keep it here. it's nicer to assemble these pieces at a high-level rather than have ingest modules importing each other

sf-dcp (approved):
addressed my question during a code walk-through. Great work!

damonmcc (approved):

fvankrieken:
Notes from group chat
- separate out default processing step logic into helper
- clean up language "processing" vs "preprocessing" (make consistent)
- type signature for reprojection func
- raise not implemented for custom script
- ability for steps to log information about themselves

I'll do these in a follow-up pr

alexrichey (approved):
🎉

PR #1170, opened 2024-10-01
Add action to publish JSON Schemas

alexrichey:
Adds CI for JSON Schemas and Product Metadata Validation.

For JSON Schemas, add the following to your VS Code settings for autocomplete and validation. Note: I can't get the globbing to work correctly for `org_metadata`, which is either a good argument to move it into the products folder, or change the name of the file from `metadata.yml` to `org.yml` (and same for product and datasets metadata). 
```
    "yaml.schemas": {
        "https://edm-publishing.nyc3.digitaloceanspaces.com/data-engineering-devops/schemas/product/product_metadata.schema.json": "/products/*/metadata.yml",
        "https://edm-publishing.nyc3.digitaloceanspaces.com/data-engineering-devops/schemas/product/dataset/dataset_metadata.schema.json": "/products/*/*/metadata.yml",
        // "https://edm-publishing.nyc3.digitaloceanspaces.com/data-engineering-devops/schemas/product/org_metadata.schema.json": "/metadata.yml",
    },
```

## Job Runs
Successful JSON Schemas push run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11133787284/job/30940620138)

Successful Metadata Validation with no errors [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11133854964/job/30940828640)
Successful Metadata Validation with introduced errors [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11133945489/job/30941119185?pr=1170) (I made a few fields newly required, which breaks every product)

### DIFF
--- a/.github/workflows/product_metadata_checks.yml
+++ b/.github/workflows/product_metadata_checks.yml
@@ -1,4 +1,0 @@
-name: Product_metadata_checks
-
-on:
-  workflow_dispatch:
--- a/admin/ops/update_product_metadata_schemas.py
+++ b/admin/ops/update_product_metadata_schemas.py
@@ -1,0 +1,39 @@
+# type: ignore
+from dcpy.utils import s3
+from dcpy.models.product import metadata as product_metadata
+from dcpy.models.product.dataset import metadata_v2 as dataset_metadata
+
+import json
+from tempfile import TemporaryDirectory
+from pathlib import Path
+
+DO_SCHEMA_FOLDER = "data-engineering-devops/schemas/"
+
+schemas = [
+    {
+        "name": "org_metadata.schema.json",
+        "folder": "product/",
+        "schema": product_metadata.OrgMetadataFile.model_json_schema(),
+    },
+    {
+        "name": "product_metadata.schema.json",
+        "folder": "product/",
+        "schema": product_metadata.ProductMetadataFile.model_json_schema(),
+    },
+    {
+        "name": "dataset_metadata.schema.json",
+        "folder": "product/dataset/",
+        "schema": dataset_metadata.Metadata.model_json_schema(),
+    },
+]
+
+for schema in schemas:
+    with TemporaryDirectory() as _dir:
+        p = Path(_dir) / schema["name"]
+        open(p, "w").write(json.dumps(schema["schema"]))
+        s3.upload_file(
+            bucket="edm-publishing",
+            path=p,
+            key=DO_SCHEMA_FOLDER + schema["folder"] + schema["name"],
+            acl="public-read",
+        )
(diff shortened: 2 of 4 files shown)


alexrichey:
@fvankrieken There's definitely an open question about how these schemas work when we introduce reusable components... we may have to add a literal to the column field indicating it's a template. But these are still super useful in the meantime.

.github/workflows/product_metadata_checks.yml
```diff
@@ -1,4 +1,33 @@
-name: Product_metadata_checks
+name: 🧪 Validate Product Metadata
+run-name: 🧪 Validate Product Metadata (${{ github.event_name }})
 
 on:
+  pull_request:
```
fvankrieken, line 5: I was going to write - maybe make this part of the pr test action?

But thinking about it... maybe legibility of having some separate actions is actually a good thing, rather than continuing to add to the bloat in our pr test job.

One downside of this is that it won't make use of the logic to create dev images for running tests in. However this specific test doesn't seem like that would present an issue.

Thoughts?

admin/ops/update_product_metadata_schemas.py
```diff
@@ -0,0 +1,39 @@
...
+        "schema": dataset_metadata.Metadata.model_json_schema(),
+    },
+]
+
+for schema in schemas:
+    with TemporaryDirectory() as _dir:
+        p = Path(_dir) / schema["name"]
+        open(p, "w").write(json.dumps(schema["schema"]))
+        s3.upload_file(
+            bucket="edm-publishing",
```
fvankrieken, line 35: There's an argument for a new bucket? More in that we could also have stuff specific to other buckets in here as well (ingest/library schemas, other devopsy stuff). But we can also just have those in their own buckets. Actually writing it out, I think this makes more sense. Carry on

fvankrieken (approved):
One question where I'm more just curious on your thoughts but looks great

.github/workflows/product_metadata_checks.yml
alexrichey: yeah, I think you're right. Can't recall why I didn't just add it there. Will change.

fvankrieken: That'd be `test.yml` for the path filter logic and test_helper then for the actual tests - top level `test.yml` takes care of the docker image logic and calls test_helper with image tag and path_filters output as args. These were separated to make it easier to call the tests within `test_helper` from the nightly_qa job

Individual jobs in test_helper look like this
```yml  
template-db:
    if: (inputs.path_filter == '') || contains(inputs.path_filter, 'template')
```
With the first clause in the `if` there being in the case of nightly_qa (where path_filters job wasn't run, we just run all tests)

fvankrieken: So you could also just leave this as its own yml file and just call it from one of the other actions

PR #1172, opened 2024-10-02
Add functionality to query dest tags at prod level

alexrichey:
In order to track which new datasets are unpublished, I've been going through and tagging the destinations in the metadata with "socrata_unpublished". Adding the ability to query them here. 

The long-term intent is to eventually have a one-click option to deploy them all.


damonmcc (approved):
> The long-term intent is to eventually have a one-click option to deploy them all.

and then remove the tag?

PR #1174, opened 2024-10-02
Generate XLSX's with file/destination Overrides

alexrichey:
Certain datasets have multiple files or destinations with different attributes, for example the description. We currently only generate metadata for the base case. Ie we don't apply any overrides.

For example, a few LION datasets have a base dataset and a Water Included type. We only generate metadata for the base. This adds the ability to specify the dataset_id and file_id when generating the asset.

### DIFF
--- a/dcpy/test/resources/product_metadata/colp_single_feature_package/metadata.yml
+++ b/dcpy/test/resources/product_metadata/colp_single_feature_package/metadata.yml
@@ -5,348 +5,367 @@ attributes:
   display_name: City Owned and Leased Property (COLP)
   each_row_is_a: City Owned Property
   tags:
-  - a
-  - b
+    - a
+    - b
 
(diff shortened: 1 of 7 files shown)


dcpy/test/resources/product_metadata/colp_single_feature_package/metadata.yml
```diff
@@ -5,348 +5,367 @@ attributes:
...
-  files:
-  - id: colp_readme.pdf
+  - id: socrata_prod
+    type: socrata
+    files:
+      - id: colp_readme
+        custom:
+          destination_use: attachment
+      - id: primary_shapefile
+        dataset_overrides:
```
alexrichey, line 33: This is the only functional change in the file. The rest is accidental auto-formatting 🤦‍♂️

PR #1176, opened 2024-10-04
promote PLUTO along with ZTL if specified, use dev buckets

fvankrieken:
Lot of testing for a pretty simple feature! Ended up spending most of time working on kinks around my data setup in the dev bucket. Overall changes are simple enough to follow! Plus I fix some things in sequential commits, so just look at overall

Two main things
- when we both build pluto as part of ztl dataloading action and promote ztl to draft in the same action, also promote pluto
- make use of dev buckets an option for ztl dataloading, promote to draft, and publish actions

Ztl dataloading action runs
- [just dataloading](https://github.com/NYCPlanning/data-engineering/actions/runs/11237942638)
- [just build ztl](https://github.com/NYCPlanning/data-engineering/actions/runs/11238057598)
- [just build ztl and promote](https://github.com/NYCPlanning/data-engineering/actions/runs/11238725997)
- [just build pluto and promote](https://github.com/NYCPlanning/data-engineering/actions/runs/11238358518)
- [build both and promote](https://github.com/NYCPlanning/data-engineering/actions/runs/11240152576)

Dev buckets working for dataloading
image

Automated promotion to draft
image

Job promoting to publish in dev bucket [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11223980881). Important note that I added logic to not create a github tag for dev runs of this. Below, 2024-10-01 was promoted from draft folder in dev bucket, not copied from edm-publsihing.
image


fvankrieken:
Oops sorry - need to let some data finish being loaded to my dev bucket before I can rerun the the ztl dataloading job one more time

sf-dcp:
I think there is an issue with the logging table when you run this code (found the issue by accident while investing the logging table)...

This is the latest [run](https://github.com/NYCPlanning/data-engineering/actions/runs/11240152576/job/31249230259) I see for today. 

And here is the logging table: 

image

It seems like `build` events with dev flag don't get logged while `promote_to_draft` do (not expected behavior). Not sure whether logging happens with the `publish` event, with dev flag set to true.

fvankrieken:
> I think there is an issue with the logging table when you run this code (found the issue by accident while investing the logging table)...
> 
> This is the latest [run](https://github.com/NYCPlanning/data-engineering/actions/runs/11240152576/job/31249230259) I see for today.
> 
> And here is the logging table for ztl (looks same for PLUTO):
> 
> image
> 
> It seems like `build` events with dev flag don't get logged while `promote_to_draft` do (not expected behavior). Not sure whether logging happens with the `publish` event, with dev flag set to true.
> 
> I don't think we want to log dev stuff; unless we indicate a presence of dev flag in the `custom_fields column`

Ooh good catch. I'll look into that

fvankrieken:
Do we want to be logging db-template events? I would lean no

.github/workflows/promote_to_draft.yml
```diff
@@ -69,6 +69,7 @@ jobs:
       image: nycplanning/build-base:latest
     env:
       PUBLISHING_BUCKET: ${{ inputs.dev_bucket && format('de-dev-{0}', inputs.dev_bucket) || 'edm-publishing' }}
+      DEV_FLAG: ${{ inputs.dev_bucket && 'true' || 'false' }}
```
fvankrieken, line 72: Long term this doesn't feel the __most__ sustainable way of handling this, ideally we have a single value somewhere setting the environment and we have a script or something handle translating that to buckets, db connection strings (if we had different db envs), etc... but this is what we currently have in other places and I think it's simplest for now

sf-dcp:
> Do we want to be logging db-template events? I would lean no
> 
> Also, I deleted the rows I generated through this in the logging table
> 
> fixed publish action [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11256760633)

I think we originally decided to log db-template events. I'm personally indifferent. Perhaps I'm leaning towards not logging them because every triggered test logs db-template, resulting in many records in the logging table

.github/workflows/promote_to_draft.yml
sf-dcp: maybe we could also add a condition, not necessarily here, to only run a job if one of the following conditions is true? 

(`DEV_FLAG = false` and `dev_bucket` is an empty string)
or 
(`DEV_FLAG = true` and `dev_bucket` is not empty)

.github/workflows/publish.yml
```diff
@@ -55,7 +59,8 @@ jobs:
     container:
       image: nycplanning/build-base:latest
     env:
-      PUBLISHING_BUCKET: edm-publishing
+      PUBLISHING_BUCKET: ${{ inputs.dev_bucket && format('de-dev-{0}', inputs.dev_bucket) || 'edm-publishing' }}
+      DEV_FLAG: ${{ inputs.dev_bucket && 'true' || 'false' }}
```
sf-dcp, line 63: same comment as for `promote_to_draft.yml`

.github/workflows/promote_to_draft.yml
sf-dcp: or maybe I'm tripping and we don't need it?

dcpy/utils/versions.py
```diff
@@ -121,7 +121,7 @@ def label(self) -> str:
                 if self.patch == 0:
                     return self.date.strftime("%Y-%m-%d")
                 else:
-                    return f"{self.date.strftime('%Y-%m')}.{self.patch}"
+                    return f"{self.date.strftime('%Y-%m-%d')}.{self.patch}"
```
sf-dcp, line 124: uh oh!

.github/workflows/promote_to_draft.yml
fvankrieken: I would lean with no - it's just adding another thing that needs to be manually specified. Right now, the dev bucket is really the only flag we have, and I think it makes sense to set a dev flag this way. And I'd rather put effort into consolidating logic to generate the various variables related to some sort of overall env flag than to try to add logic to validate that they're consistent

sf-dcp (approved):
LGTM! Thanks for fixing a silly error with patching.


I tripped on GH syntax around setting up a dev flag. I thought dev flag is a user input variable in addition to the bucket

PR #1180, opened 2024-10-06
More Metadata Changes

alexrichey:
This ties up a bunch of lose ends:
- and adds quality of life improvements for the CLI
- changes CLI to point at the entire product metadata repo, rather than individual dataset md files
- implements Org-Level defaults
- Implements those defaults in the Socrata publish connector
- Implements metadata-only package_from_bytes+deployment. 

Commits are very atomic and tidy (the only leakage is in the last commit). Rationale for changes is included in the commit message for each commit.

Product Metadata Repo PR with related changes [here](https://github.com/NYCPlanning/product-metadata/pull/20)
Note: the `Tests / test / Validate Product Metadata` will fail (both here and in the product metadata repo) until we merge both of these. So they should go together.

### DIFF
--- a/dcpy/connectors/socrata/publish.py
+++ b/dcpy/connectors/socrata/publish.py
@@ -124,23 +124,28 @@ class Socrata:
 
             @classmethod
             def from_dataset_attributes(cls, attrs: md.DatasetAttributes):
+                if not (attrs.category and attrs.agency and attrs.publishing_frequency):
+                    raise Exception(
+                        f"Required metadata fields are missing. Found category: {attrs.category}, agency: {attrs.agency} or publishing_frequency: {attrs.publishing_frequency}"
+                    )
+
                 return cls(
                     name=attrs.display_name,
                     description=attrs.description,
-                    category=attrs.category or "",
+                    category=attrs.category,
                     attribution=attrs.attribution or "",
                     attributionLink=attrs.attributionLink or "",
                     tags=attrs.tags or [],
                     metadata={
                         "rowLabel": attrs.each_row_is_a,
                         "custom_fields": {
-                            "Dataset Information": {"Agency": attrs.agency or ""},
+                            "Dataset Information": {"Agency": attrs.agency},
                             "Update": {
                                 "Data Change Frequency": attrs.publishing_frequency,
                                 "Date Made Public": attrs.date_made_public,
                                 "Update Frequency Details": attrs.publishing_frequency_details,
                                 "Update Frequency": translate_legislative_freq_to_update_freq(
-                                    attrs.publishing_frequency or ""
+                                    attrs.publishing_frequency
                                 )
                                 or attrs.publishing_frequency,
                                 "Automation": "Yes",
@@ -148,7 +153,7 @@ class Socrata:
                         },
                     },
                     privateMetadata={
-                        # "contactEmail": "opendata@planning.nyc.gov", # what's the right email?
+                        # "contactEmail": "", # Leaving this here in case we want to add it, so we don't have to remember what the field is called
                         "custom_fields": {
                             "Legislative Compliance": {
                                 "Removed Records?": "Yes",  # refers to row removal at time of push to Socrata. Always true since we overwrite the existing dataset.
--- a/dcpy/connectors/socrata/publish.py
+++ b/dcpy/connectors/socrata/publish.py
@@ -114,7 +114,11 @@ class Socrata:
         class DatasetMetadata(BaseModel):
             name: str
             description: str
+            category: str
+            attribution: str
+            attributionLink: str
             tags: list[str]
+            # licenseId: str
             metadata: dict[str, Any]
             privateMetadata: dict[str, Any]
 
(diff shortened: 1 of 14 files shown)


dcpy/connectors/socrata/publish.py
```diff
@@ -123,20 +127,28 @@ def from_dataset_attributes(cls, attrs: md.DatasetAttributes):
...
                                     attrs.publishing_frequency or ""
                                 )
                                 or attrs.publishing_frequency,
                                 "Automation": "Yes",
-                            }
+                            },
                         },
                     },
                     privateMetadata={
+                        # "contactEmail": "opendata@planning.nyc.gov", # what's the right email?
```
alexrichey, line 151: OTI docs actually just say to ignore this field, so this is removed later.

dcpy/connectors/socrata/publish.py
```diff
@@ -123,20 +127,28 @@ def from_dataset_attributes(cls, attrs: md.DatasetAttributes):
                 return cls(
                     name=attrs.display_name,
                     description=attrs.description,
+                    category="City Government",
+                    attribution="Department of City Planning (DCP)",
+                    attributionLink="https://www.nyc.gov/site/planning/data-maps/open-data.page",
+                    # licenseId="ODC_BY", # TODO: what's the right license?
```
alexrichey: OTI docs actually just say to ignore this field, so this is removed later.

dcpy/connectors/socrata/publish.py
```diff
@@ -114,7 +114,11 @@ class Inputs:
         class DatasetMetadata(BaseModel):
             name: str
             description: str
+            category: str
+            attribution: str
+            attributionLink: str
             tags: list[str]
+            # licenseId: str
```
alexrichey, line 121: OTI docs actually just say to ignore this field, so this is removed later.

dcpy/lifecycle/distribute/socrata.py
```diff
@@ -53,27 +62,31 @@ def dist_from_local(
...
+            metadata_only=bool(metadata_only),
         )
     except Exception as e:
         return f"Error pushing {md.attributes.display_name}, destination: {dest.id}: {str(e)}"
 
 
 def dist_from_local_all_socrata(
     package_path: Path,
-    **pub_kwargs,
+    **pub_kwargs: Unpack[PublishKwargs],
```
fvankrieken, line 74: Ooh this is fun

dcpy/lifecycle/distribute/socrata.py
```diff
@@ -53,27 +62,31 @@ def dist_from_local(
...
 ):
     """Distributes all Socrata destinations within a given metadata"""
     md = m.Metadata.from_path(package_path / "metadata.yml")
+    local_pub_kwargs = pub_kwargs.copy()
+    local_pub_kwargs.pop(
```
fvankrieken, line 79: nit, but is there a reason to not just do a standard if clause here? Since the value is not being set. Unless this is meant to be a variable and that's fixed in another commit

dcpy/lifecycle/distribute/socrata.py
fvankrieken: So do the required ones throw a runtime error if missing?

fvankrieken (approved):
One nit. This is all lovely

dcpy/connectors/socrata/publish.py
```diff
@@ -123,20 +127,28 @@ def from_dataset_attributes(cls, attrs: md.DatasetAttributes):
                 return cls(
                     name=attrs.display_name,
                     description=attrs.description,
-                    tags=attrs.tags,
+                    category=attrs.category or "",
```
fvankrieken, line 130: curious if you have any thoughts on this - should this be nullable? Throw an error?

dcpy/lifecycle/package/assemble.py
```diff
@@ -141,6 +144,12 @@ def pull_destination_files(
     package_ids = {p.id for p in product_metadata.assembly}
     for f in dest.files:
         paths_and_dests = ids_to_paths_and_dests[f.id]
+        f_is_metadata = (
+            f.id in product_metadata.get_file_ids()
+            and product_metadata.get_file_and_overrides(f.id).file.is_metadata
+        )
+        if metadata_only and not f_is_metadata:
+            continue
```
fvankrieken, line 152: Codecov flagging - this might actually be a worthwhile test (not so much for this pr specifically, but just ensure that metadata_only updates only update metadata

dcpy/lifecycle/package/assemble.py
```diff
@@ -219,44 +233,79 @@ def assemble_dataset_from_bytes(
...
+        False,
+        "--output-path",
+        "-m",
+        help="Only Assemble Metadata.",
+    ),
+    source_destination_id: str = typer.Option(
+        BYTES_DEST_TYPE,
+        "--source-destination-id",
+        "-s",
+        help="Only Assemble Metadata.",
```
fvankrieken, line 268: Wrong help!

fvankrieken (changes requested):
Okay actually at least one tiny nit that should definitely be fixed. Couple other small questions/nits

dcpy/lifecycle/distribute/socrata.py
alexrichey: I thought you'd like this. And yes, it behaves just as if they were literally written out kwargs.

dcpy/lifecycle/distribute/socrata.py
alexrichey: Yeah, this was a little uncomfortable, due to the unpacking of the kwargs. Basically, I wanted to not modify the kwargs to never pass along the `metadata_path`, so what I would have wanted to do is, for immutability sake:
```
local_pub_kwargs = {k:v for k,v in kwagrs if k != "metadata_path"}
```
but `mypy` loses type of `local_pub_kwargs` in the above. (Could just slap a type annotation on it, but that feels like cheating). Def open to alternatives!

dcpy/lifecycle/package/assemble.py
alexrichey: very good call! added.

dcpy/lifecycle/package/assemble.py
alexrichey: 🤦‍♂️ thank you. Fixed

dcpy/lifecycle/distribute/socrata.py
fvankrieken: Well i just meant

```python
if "metadata_path" in local_pub_kwargs:
    local_pub_kwargs.pop("metadata_path")
```

To get rid of the `else None`. But I think actually just
```python
local_pub_kwargs.pop("metadata_path", None)
```

Would work too

dcpy/connectors/socrata/publish.py
alexrichey: yeah, this should throw an exception in the connector if we don't have those fields. Gonna fix that real quick.

alexrichey:
@fvankrieken Fixed the broken test - all good to go now

PR #1187, opened 2024-10-07
Create a GHA file to run dbt (qaqc)

sf-dcp:


sf-dcp:
No longer needed. May reopen if we add dbt testing for other products

PR #1189, opened 2024-10-08
PLUTO dbt QA: update test to check version of ignored bbls

sf-dcp:
### What
Update logic of newly implemented test for PLUTO build to only ignore "ignored" bbls in the seed file if their version is the same as of PLUTO being build. For example, a seed file with "ignored bbls" can look like this:

```csv
bbl,pluto_version
3001497501,24v3
3038627501,24v3
4155377501,24v1
```

If we are building PLUTO 24v3, we shouldn't ignore the last bbl that was added during 24v1 because the raw data for this bbl could change. With the new logic, we will only be ingoring the first 2 bbls because their version equals to the current PLUTO version.

### Why not to overwrite this file for each PLUTO build? 
Because we need a deterministic way of building any PLUTO version. If we 'ignored' some bbls during past PLUTO, we want to be able to rebuild same output for the past PLUTO.

### Receipts it works
* PLUTO build with no records to ignore (empty seed file) [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11225072798/job/31203060304#step:16:189). There are 12 failing records
* PLUTO build with 12 records to ignore where 1 record has a different version [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11241736679/job/31253925842#step:16:189). As expected, only 1 record failed.

### DIFF
--- a/products/pluto/dbt_project.yml
+++ b/products/pluto/dbt_project.yml
@@ -8,3 +8,6 @@ test-paths: ["tests"]
 tests:
   +store_failures: true
   schema: "_tests"
+
+vars:
+  version: "{{ env_var('VERSION') }}"
--- a/products/pluto/seeds/_seeds.yml
+++ b/products/pluto/seeds/_seeds.yml
@@ -9,3 +9,4 @@ seeds:
     config:
       column_types:
         bbl: text
+        pluto_version: text
--- a/products/pluto/tests/assert_condo_bbl_unit_count_research_required.sql
+++ b/products/pluto/tests/assert_condo_bbl_unit_count_research_required.sql
@@ -54,6 +54,7 @@ not_ignored_primebbls AS (
     WHERE primebbl NOT IN (
         SELECT bbl
         FROM {{ ref('ignored_bbls_for_unit_count_test') }}
+        WHERE '{{ var('version') }}' = pluto_version
     )
 ),
 


fvankrieken:
Food for thought - it'd maybe be consistent to handle this like current corrections. Maybe not match on version, but instead match on bbl and some calculated value to ignore if match? As in, ignored "number of rows found in table" or "number of units", etc. Not sure if that's actually possible for these specific checks, but it'd give that specificity. And then, we don't always have to see that one "bad" bbl that we can't correct for whatever reason

sf-dcp:
> Food for thought - it'd maybe be consistent to handle this like current corrections. Maybe not match on version, but instead match on bbl and some calculated value to ignore if match? As in, ignored "number of rows found in table" or "number of units", etc. Not sure if that's actually possible for these specific checks, but it'd give that specificity. And then, we don't always have to see that one "bad" bbl that we can't correct for whatever reason

That's definitely an option. I can't think of drawbacks of either approach. @damonmcc, thoughts?

sf-dcp:
@fvankrieken, what if, for some bbl, a unit number changes but the bbl still should be ignored? Are you suggesting to append a new record for same bbl or update the value in-place?

fvankrieken:
> @fvankrieken, what if, for some bbl, a unit number changes but the bbl still should be ignored? Are you suggesting to append a new record for same bbl or update the value in-place?

Either new record (and join to table only using max version per bbl) or updating bbl would work. I would prefer updating.

If we think about how a bbl could change in the context of this seed table, if a bbl was flagged and ignored one build, in the next build it could either 
- be unchanged, and still will be flagged and need to be ignored
- be changed, but be incorrect and still need to be ignored
- be changed, but now be a correct value (so no longer flagged hopefully?)
- be unchanged but now be correct value? Given how logic is working, don't think this is an issue

So if we update by bbl and by version
- unchanged, flagged -> we'd need to manually qa this bbl every time
- changed, flagged -> we'd need to manually qa this bbl
- changed, no longer flagged -> No issue

If we update by bbl and value
- unchanged, flagged -> this would just get ignored every time
- changed, flagged -> we'd need to manually qa this, same as other scheme. But this is I think a good thing in general - if values are unstable but still getting flagged, it's worth needing to qa them again
- changed, no longer flagged, -> No issue

So I think it just makes it a little more specific to match by value instead of version, and would save work of not having to qa the same bbls every time (and maybe start keeping a list of bbls that we always ignore - that would seem redundant with this seed file).

Not entirely sure about potential pitfalls. I need to think a little more about that

sf-dcp:
Per standup discussion, we will keep version logic for now and may tweak it later. This test should be run only for PLUTO major, so I need to create `major` and `minor` tags for the test. Will do it in a separate PR.

damonmcc (approved):

sf-dcp:
Deleted the second commit that was adding test records to the seed file (the second GHA run used that commit).

PR #1192, opened 2024-10-10
PLUTO: tag tests based on PLUTO version type (major vs minor)

sf-dcp:
### What 

* Add `version_type` env variable to PLUTO minor and major recipe files.
* Create `major` and `minor` tags in `dbt` tests. Depending on the `version_type` value (`major` or `minor`), corresponding tests will run. If a test is tagged with both `major` and `minor`, then it will run for both PLUTO versions. 
* Create new singular test for PLUTO minor version, checking pluto total record count is the same as in previous PLUTO version. 

### GHA runs:
* PLUTO major [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11296716119/job/31422129835#step:16:183). As expected, it only ran the `assert_condo_bbl...` test (testing condo bbl unit counts), which has a `major` tag
* PLUTO minor [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11294916986/job/31416489905#step:16:184). As expected, it only ran the new test tagged as `minor`

### DIFF
--- a/dcpy/lifecycle/builds/plan.py
+++ b/dcpy/lifecycle/builds/plan.py
@@ -84,6 +84,9 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
     recipe.vars = recipe.vars or {}
     recipe.vars["VERSION"] = recipe.version
 
+    if recipe.version_type is not None:
+        recipe.vars["VERSION_TYPE"] = recipe.version_type.value
+
     # Determine previous version
     try:
         previous_recipe = publishing.get_previous_version(
--- a/dcpy/test/lifecycle/builds/test_plan.py
+++ b/dcpy/test/lifecycle/builds/test_plan.py
@@ -14,6 +14,7 @@ from dcpy.test.lifecycle.builds.conftest import REQUIRED_VERSION_ENV_VAR, RESOUR
 RECIPE_PATH = RESOURCES_DIR / "recipe.yml"
 RECIPE_NO_DEFAULTS_PATH = RESOURCES_DIR / "recipe_no_defaults.yml"
 RECIPE_NO_VERSION_PATH = RESOURCES_DIR / "recipe_no_version.yml"
+RECIPE_W_VERSION_TYPE = RESOURCES_DIR / "recipe_w_version_type.yml"
 BUILD_METADATA_PATH = RESOURCES_DIR / "build_metadata.json"
 SOURCE_VERSIONS_PATH = RESOURCES_DIR / "source_data_versions.csv"
 
@@ -179,6 +180,30 @@ class TestRecipesNoDefaults(TestCase):
         }
         lock_file = plan.plan(RECIPE_NO_DEFAULTS_PATH)
         plan.recipe_from_yaml(lock_file)
+
+
+@pytest.mark.usefixtures("file_setup_teardown")
+@pytest.mark.usefixtures("create_buckets")
+@patch("dcpy.connectors.edm.recipes.get_latest_version")
+class TestRecipeVars(TestCase):
+    def test_version_type_var_is_absent(self, get_latest_version):
+        """Ensures VERSION_TYPE is absent in recipe 'vars' attribute when version_type is None."""
+        setup()
+        version = "test_version"
+        planned = plan.plan_recipe(RECIPE_PATH, version=version)
+        assert planned.version_type is None  # sanity check
+        assert "VERSION_TYPE" not in planned.vars
+
+    def test_version_type_is_present(self, get_latest_version):
+        """Test that the version_type is set correctly and matches env 'VERSION_TYPE' variable."""
+        planned = plan.plan_recipe(RECIPE_W_VERSION_TYPE)
+        assert planned.version_type is not None  # sanity check
+        assert (
+            planned.version_type == planned.vars["VERSION_TYPE"]
+        ), "version_type mismatch with recipe.vars"
+        assert (
+            planned.vars["VERSION_TYPE"] == os.environ["VERSION_TYPE"]
+        ), "'version_type' recipe variable mismatch with 'VERSION_TYPE' env variable"
 
 
 def build_metadata_exists(key, file):
--- a/dcpy/test/lifecycle/builds/test_plan.py
+++ b/dcpy/test/lifecycle/builds/test_plan.py
@@ -182,8 +182,6 @@ class TestRecipesNoDefaults(TestCase):
         plan.recipe_from_yaml(lock_file)
 
 
-@pytest.mark.usefixtures("file_setup_teardown")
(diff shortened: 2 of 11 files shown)


products/pluto/dbt_project.yml
```diff
@@ -11,3 +11,4 @@ tests:
 
 vars:
   version: "{{ env_var('VERSION') }}"
+  version_type: "{{ env_var('VERSION_TYPE') }}"
```
sf-dcp, line 14: This `dbt` var `version_type` isn't actually used in code.

I define it here nonetheless because this env var is used in a dbt CLI command (see last line of `products/pluto/pluto_build/07_custom_qaqc.sh` file), and `dbt` checks that this env variable is present before running any code - very convenient.

products/pluto/pluto_build/07_custom_qaqc.sh
```diff
@@ -16,7 +16,7 @@ echo "Build staging tables"
 dbt build --select staging --exclude tag:de_check
 
 echo "Build intermediate QAQC tables"
-dbt run --select qaqc --exclude tag:de_check  # need to use 'dbt run' because otherwise it fails if downstream test fails
+dbt build --select qaqc --exclude tag:de_check
 
 echo "🔥 Run DE aka important tests 🔥"
-dbt test --select tag:de_check
+dbt test --select tag:de_check,tag:$VERSION_TYPE    # this will only run tests that have both tags, not just one of them
```
sf-dcp, line 22: This line will run tests that have both `de_check` and `<VERSION_TYPE>` tags. 

Why not just label tests with `<VERSION_TYPE>` tag? I was thinking about this for a while. As we create more tests, I imagine that there will be version type specific tests but not necessarily for the DE to address in case of a failure. `de_check` tag serves as a classifier: if a test fails, DE must address it. Then there will be tests/reports for GIS or tests that yield a warning. 

Would love to get some thoughts on this

sf-dcp: Note to self: [link](https://discourse.getdbt.com/t/exclude-multiple-tags/6538) showing how to run tests having or not having both tags (not just one of them).

products/pluto/recipe-minor.yml
```diff
@@ -3,6 +3,8 @@ base_recipe: recipe.yml
 version_type: minor
 version_strategy: bump_latest_release
 product: db-pluto
+vars:
+  VERSION_TYPE: minor
```
damonmcc, line 7: since `version_type` is already declared above this, I think it can just be added to `recipe.vars` during `plan_recipe` [here](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/lifecycle/builds/plan.py#L84) so that it becomes an envar [here](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/lifecycle/builds/plan.py#L105)

maybe something like:
```python
if recipe.version_type is not None: recipe.vars["VERSION_TYPE"] = recipe.version_type
```

products/pluto/tests/assert_bbl_count_is_the_same.sql
```diff
@@ -0,0 +1,34 @@
...
+        }
+    ) 
+}}
+
+WITH current_pluto AS (
+    SELECT bbl::bigint, 'current' AS source
+    FROM {{ source('build_sources', 'export_pluto') }}
+),
+previous_pluto AS (
+    SELECT bbl::decimal::bigint, 'previous' AS source
```
alexrichey, line 22: Curious about this cast to decimal here. What necessitates that?

alexrichey (approved):
Had one question, but otherwise looks great. I agree with Damon's suggestion to set the version type in `plan.py`

products/pluto/tests/assert_bbl_count_is_the_same.sql
sf-dcp: Previous Pluto table has bbl as a decimal  text 🙃

products/pluto/recipe-minor.yml
sf-dcp: wow. I didn't see this var being already declared. Thanks!

products/pluto/tests/assert_bbl_count_is_the_same.sql
fvankrieken: Maybe something to work on when we move the suite of `pluto` templates over to ingest

dcpy/lifecycle/builds/plan.py
```diff
@@ -84,6 +84,9 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
     recipe.vars = recipe.vars or {}
     recipe.vars["VERSION"] = recipe.version
 
+    if recipe.version_type is not None:
```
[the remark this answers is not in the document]
sf-dcp: @alexrichey, yep, working on it today

fvankrieken:
So from my reading of the code - a major build on nightly qa will still fail with this as written, yes? Are you still planning on implementing this? https://github.com/NYCPlanning/data-engineering/issues/1190#issuecomment-2416951750

dcpy/test/lifecycle/builds/test_plan.py
```diff
@@ -181,6 +182,30 @@ def test_serializing_and_deserializing(self, get_file_types):
         plan.recipe_from_yaml(lock_file)
 
 
+@pytest.mark.usefixtures("file_setup_teardown")
+@pytest.mark.usefixtures("create_buckets")
+@patch("dcpy.connectors.edm.recipes.get_latest_version")
+class TestRecipeVars(TestCase):
+    def test_version_type_var_is_absent(self, get_latest_version):
+        """Ensures VERSION_TYPE is absent in recipe 'vars' attribute when version_type is None."""
+        setup()
```
alexrichey, line 191: I personally find one-liner `setup` methods like this make the code harder to read. Thoughts about just setting the env var here instead of calling setup?

sf-dcp: I used the `setup()` fn instead of explicitly setting up an env variable because it already exists in the file and is used across many other tests. 

If it's a preference, I can replace rename the function to something more telling in another PR but I would rather not do it here

alexrichey: (and by fix it, I mean either rename the function, or just inline the env var setting.)

sf-dcp: I prefer to keep PRs focused on their main objective, but I'm happy to address this for you here

sf-dcp:
Waiting for nightly qa to confirm that tests are failing with `warn` severity level [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11484352461/job/31961931429).

alexrichey (approved):

PR #1196, opened 2024-10-15
GFT tweaks for 24v3 draft 1 review

fvankrieken:
Tweaks for https://github.com/NYCPlanning/data-engineering/issues/1004#issuecomment-2414555945

rebuild [here](https://github.com/NYCPlanning/data-engineering/actions/runs/11351451569/job/31572066514)

Next draft is in QA, but I think it makes sense to get this merged in the meantime and follow-up as needed.

Changes are
1. making sure that when we export both `points` and `lots` source dataset outputs that the `lots` only contain rows where the geom is not null
2. make sure that for open spaces/shadow polys, we're exporting both poly raw geoms and points that were joined to lots.

### DIFF
--- a/dcpy/lifecycle/builds/plan.py
+++ b/dcpy/lifecycle/builds/plan.py
@@ -341,11 +341,13 @@ def _cli_wrapper_repeat_recipe(
     ),
 ):
     product_key: publishing.BuildKey | publishing.DraftKey | publishing.PublishKey
-
+    product_label = (
+        "db-green-fast-track" if product == "green_fast_track" else f"db-{product}"
+    )
     match product_type:
         case "build":
             product_key = publishing.BuildKey(
-                product=f"db-{product}", build=version_or_build
+                product=product_label, build=version_or_build
             )
         case "draft":
             if draft_revision_number is None:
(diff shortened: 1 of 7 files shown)


dcpy/lifecycle/builds/plan.py
```diff
@@ -341,30 +341,32 @@ def _cli_wrapper_repeat_recipe(
     ),
 ):
     product_key: publishing.BuildKey | publishing.DraftKey | publishing.PublishKey
-
+    product_label = (
```
alexrichey, line 344: Thoughts on just passing along a `--product-label` CLI option?

fvankrieken:
> @fvankrieken Do we need to do this for all `lot` types? e.g. `air_quality/source__industrial_lots`? or is it just that these were problematic somehow?

We need to do this for every dataset that starts out as points and attempts to join to lots (so final geom column contains both points and polys). Industrial lots has pluto as a source, so every row is a lot from the start

dcpy/lifecycle/builds/plan.py
fvankrieken: Something like that would make sense, though then we'd need to have this sort of logic in the `repeat` gha - still need to refer to both formattings in our code at the moment

alexrichey (approved):
Open question about hard-coding overrides. Seems less than ideal, but I'm fine if we to address that later.

dcpy/lifecycle/builds/plan.py
fvankrieken: Cool. Yeah I feel like a CLI option isn't really great either, in that it still adds some unneeded complexity. But it's 100% a lesser evil than what's in this PR

Maybe we should just have some sort of minimal `ProductPaths` object? And then dcpy can rely on pyproject.toml (or `.dcpy` file, etc) that defines the path to product folder, DO folder name, db build name, etc. OR for each product, we have a standard spot to look, (i.e. the product folder) and store the additional metadata there. This seems like this could belong in a recipe actually. Though then you would need the recipe to run an upload step. 

Long way of saying I think we should punt the issue. I can make an issue for this

PR #1202, opened 2024-10-16
Ingest - tweaks following group demo

fvankrieken:
Follow up to #1166. Primarily,

- separate out default processing step logic into helper
- clean up language "processing" vs "preprocessing" (make consistent)
- raise not implemented for custom script
- publish json schema


admin/ops/update_json_schemas.py
```diff
@@ -2,37 +2,55 @@
 from dcpy.utils import s3
 from dcpy.models.product import metadata as product_metadata
 from dcpy.models.product.dataset import metadata_v2 as dataset_metadata
+from dcpy.models.lifecycle import ingest as ingest_models
 
 import json
 from tempfile import TemporaryDirectory
 from pathlib import Path
 
+RECIPES_BUCKET = "edm-recipes"
```
alexrichey, line 11: Curious your rationale on adding a new bucket vs having all our schemas colocated. Just thinking that if we keep adding schemas, not sure they'll be 1-1 with buckets.

fvankrieken: Yeah, I agree with that, it just on one level made sense to have them in the same place as all the objects that refer to them. But I'm with you - this should really just live in one place

fvankrieken: And maybe we move it somewhere more "neutral" but that place can be `publishing` for now

fvankrieken: fixed in new commit

alexrichey (approved):
Looks great

PR #1216, opened 2024-10-26
Compile python requirements

fvankrieken:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #1217, opened 2024-10-28
Fix test severity to `warn` for PLUTO when running compile_python_reqs

sf-dcp:
PLUTO build fails when compiling python reqs (example #1216) because of incorrect dbt test severity. This PR changes severity from `error` to `warn` for python reqs builds.

[Reference](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-matching-an-array-of-strings) for GHA syntax used in the fix.

### DIFF
--- a/.github/workflows/pluto_build.yml
+++ b/.github/workflows/pluto_build.yml
@@ -38,7 +38,7 @@ jobs:
       RECIPES_BUCKET: ${{ inputs.dev_bucket || 'edm-recipes' }}
       PUBLISHING_BUCKET: ${{ inputs.dev_bucket || 'edm-publishing' }}
       DEV_FLAG: ${{ inputs.dev_bucket && 'true' || 'false' }}
-      TEST_SEVERITY: ${{ contains(inputs.build_name, 'nightly_qa') && 'warn' || 'error' }}
+      TEST_SEVERITY: ${{ contains(fromJson('["nightly_qa", "compile_python_reqs"]'), inputs.build_name) && 'warn' || 'error' }}
     steps:
       - uses: actions/checkout@v4
 


alexrichey:
@sf-dcp Just confirming my understanding: This is the case where we've written a test that detected an issue, but not yet fixed the underlying issue? (e.g. incorrect BBL counts) 

If so, I'm wondering if we should set the TEST_SEVERITY at the level of the nightly_qa/compile_python_reqs jobs. 

Or maybe we could set a different severity for tests that are expected to fail in DBT via use of a different env var. I'm thinking that if we had some override env var in our builds like `IGNORE_EXPECTED_FAILING_DBT_DE_CHECKS`, we could engineer things in DBT so that:
1. Manual Builds of PLUTO fail until the issue is resolved.
2. Integration tests (e.g. nightly_qa/compile_python_reqs) can just ignore those expected failures, but not ALL dbt test failures.

Which would be the best of both worlds. Curious @damonmcc 's thoughts

sf-dcp:
Yes, your understanding is correct: this env variable is to set severity level when running dbt tests. 

Note that this variable sets severity level for the whole dbt project; i.e. it applies to all tests. For instances where we want some tests to fail regardless, we can override their severity level by modifying the test's cofig object. 

The tests I wrote so far don't need to fail actual builds in nightly qa and compile python requirements. We discussed it [here](https://github.com/NYCPlanning/data-engineering/issues/1190#issuecomment-2416951750). 

I would prefer to keep this env var at pluto build job as it so PLUTO test config can be found in one place.

alexrichey (approved):
Cool, I think that makes sense for now I think. In the long term, I could see this happening pretty often, for projects other than PLUTO as we add tests to them.

I do think it would be nice to have two tiers of tests - one level that should absolutely never fail, and one that we're content to ignore until we need to actually build the product for real. It seems like dbt overrides can handle that. But for now I think you're right... probably simplest to keep this here.

PR #1194, opened 2024-10-11
Product Logging: Add sql scripts used for creating views in db

sf-dcp:
I created a couple of views built on top of the logging table. Adding the scripts used for the views. 

What these views look like: 

image

image

Issue #854, opened 2024-05-17
PLUTO QAQC Enhancements


fvankrieken:
For 24v3

Inspired by work during corrections #831. Slight overlap with #847 though this has more focus on QA and will hopefully be less rambling

### Unreasonable increase in building area
Currently (and since this table's creation in 2021) this actually is showing changes in LOT area, not building area. This maybe should just be split into two different checks
- lotarea - since we have geometries, maybe more informative than a change in lot area would simply be comparing reported lot area to calculated area (with `st_area` in postgis). Most of the ones flagged this past time were lots that in the past had had far too low lot areas that were now fixed, and so they're unhelpful to be reported.
- bldgarea - since we can't calculate this, we could use the current intended check (in addition to the bldgarea/lotarea vs floors check)

### Unreasonably small apartments
In #538, I added ownername as a field and ability to filter out NYCHA records. It would be great to be able to filter out or flag other records that have simple explanations
- During QA, I found many records with class b units, that might actually have accurate areas. It would be great if we could somehow get this in the table, though I don't know if that info is currently available in PLUTO's build. Some investigation could be done can we easily join DOB records? Are there owners other than NYCHA that correspond to class b records?
- There were multiple vacant lots with new PTS entries that were clearly wrong. Can these be identified upstream? Maybe this belongs more in #847, but flagging is lower risk than trying to implement something automated

## New and Vanished BBLs
It would be nice to have these as two separate tables. That could be handled on QA app side rather than the data export, either way

## Other
A handy report for us would be corrections corresponding to the current build version that do not get applied, as presumably that should never happen

fvankrieken:
The "Unreasonably small apartments" section is largely handled by @sf-dcp 's ongoing work, and is out of scope for this issue

PR #1224, opened 2024-11-04
Migrations Follow-up tiny PR

sf-dcp:
Follow up to @alexrichey's [comment](https://github.com/NYCPlanning/data-engineering/pull/1194#discussion_r1828235859)


fvankrieken:
For building and lot area, I had written about it in https://github.com/NYCPlanning/data-engineering/issues/854. Per that issue, I think we should
1. Make this check actually be building area, as in change the code that generates the output. That's what GIS expects and is used to researching (and makes sense as a check)
2. instead of checking for changes in lot area as reported in PTS, compare lot area as reported in PTS to calculated lot area from geom, and flag if there's a significant deviation.

The second maybe is out of scope for now - would require a bit of research into how prevalent mismatches are, what a threshold is, when and why we fix it, etc. But for the former, I think we should correct the check itself, not the name.

apps/qa/src/pages/pluto/components/source_data_versions_report.py
```diff
@@ -12,6 +12,8 @@
     "pluto_pts": "Department of Finance - Property Tax System (PTS)",
     "lpc_historic_districts": "Landmarks Preservation Commission - Historic Districts",
     "lpc_landmarks": "Landmarks Preservation Commission - Individual Landmarks",
+    "pluto_input_numbldgs": "Office of Technology & Innovation – Building Footprint Centroids",  # this dataset is based on OTI building footprints
+    "dpr_greenthumb": "Department of Parks and Recreation – GreenThumb Garden Info",
```
fvankrieken, line 16: Looking forward to moving to ingest so we can have both "dataset_id" and "dataset_name" in source_data_versions.csv

For this - I can foresee a future where another source dataset gets added or something and we forget to add this. Instead of filtering in line 52, should we just rename the datasets if they're found in source name mappings? Something like

```python
df["dataset"] = df["dataset"].apply(lambda d: SOURCE_NAME_MAPPINGS.get(d, d))
```

So if we forgot to have "dof_dtm" in the name mappings, it'll still show up in the table as "dof_dtm" instead of being missing entirely

sf-dcp:
> For building and lot area, I had written about it in #854. Per that issue, I think we should
> 
> 1. Make this check actually be building area, as in change the code that generates the output. That's what GIS expects and is used to researching (and makes sense as a check)
> 2. instead of checking for changes in lot area as reported in PTS, compare lot area as reported in PTS to calculated lot area from geom, and flag if there's a significant deviation.
> 
> The second maybe is out of scope for now - would require a bit of research into how prevalent mismatches are, what a threshold is, when and why we fix it, etc. But for the former, I think we should correct the check itself, not the name.

@fvankrieken, for 1, are you suggesting to update the query on the last screenshot with the correct field (building instead of lot) and keep everything else the same? Agree for 2, this can be added to the list of PLUTO data checks for Q4 (I created a master GH issue for it)

apps/qa/src/pages/pluto/components/source_data_versions_report.py
sf-dcp: In the QA page, we have a note about this source table that only a subset of sources are listed, which GIS copies intro pluto readme... So not sure about a case where we would be adding sources implicitly

fvankrieken: Gotcha. If this serves a more specific purpose, then that makes sense

fvankrieken:
> > For building and lot area, I had written about it in #854. Per that issue, I think we should
> > 
> > 1. Make this check actually be building area, as in change the code that generates the output. That's what GIS expects and is used to researching (and makes sense as a check)
> > 2. instead of checking for changes in lot area as reported in PTS, compare lot area as reported in PTS to calculated lot area from geom, and flag if there's a significant deviation.
> > 
> > The second maybe is out of scope for now - would require a bit of research into how prevalent mismatches are, what a threshold is, when and why we fix it, etc. But for the former, I think we should correct the check itself, not the name.
> 
> @fvankrieken, for 1, are you suggesting to update the query on the last screenshot with the correct field (building instead of lot) and keep everything else the same? Agree for 2, this can be added to the list of PLUTO data checks for Q4 (I created a master GH issue for it)

Precisely

sf-dcp:
@fvankrieken, something I just thought of... Starting 24v4, we will be calculating the actual building area, while in previous versions we still have **lot** area as the qa reports are generated at the time of a build. 

Meaning, for 24v4 qa review, the comparison will be between LOT (23v3) and BUILDING areas. It will be an invalid comparison. How should we go about 24v4? Do a manual comparison report for bldg area for GIS or rebuild 24v3 after merging this PR?

fvankrieken:
For outlier analysis, it's just comparing to the past version, right? The mismatch section compares to last version, then compares this comparison to the comparison from the previous build (love that sentence). So in a case in there, I think we'd run into issue. But from my understanding, for this, it's querying "previous_pluto" and making all calculations on the fly, so as long as we query the right column in previous pluto we should be good, right?

sf-dcp:
> For outlier analysis, it's just comparing to the past version, right? The mismatch section compares to last version, then compares this comparison to the comparison from the previous build (love that sentence). So in a case in there, I think we'd run into issue. But from my understanding, for this, it's querying "previous_pluto" (which is just the last version, as imported from edm-recipes) and making all calculations on the fly, so as long as we query the right column in previous pluto we should be good, right?

oh yes, you are right :)

sf-dcp:
@fvankrieken, I added this disclaimer to the description: 

image

fvankrieken (approved):

PR #1236, opened 2024-11-08
Ingest Validation - developer QOL

### LINKED ISSUE #1233: Ingest Validation Framework - compare geometries
In https://github.com/NYCPlanning/data-engineering/pull/1191, I've run into issues where geometries are functionally identical but different from basic sql perspective (seemingly `(1, 0), (0, 1), (0, 0), (1,0)` vs `(0,1), (0,0), (1,0), (0,1)` for a polygon. Same shape, different order of vertices). 

Data comparison tooling should be expanded to evaluate differences in geometry columns. I'd still find it slightly troubling if our two processes somehow ended up with changed ordering of vertices in our polygons, though functionally in our pipelines it shouldn't matter. Regardless, all we have right now in the outputs of `compare` logic is tables of "different" wkb, which is entirely unhelpful

fvankrieken:
Closes #1233. Doesn't get its complete original scope (flush out how we compare geometries to something relatively exhaustive), keeps it a little more simple for the sake of what seems needed/practical for the types of geom differences I'm seeing. Will write up some of it in a new issue and come back to if needed.

## Big picture things
`ingest` was not fixing invalid geoms before, while library was. Per brief convo with @alexrichey , we thought this makes sense more upstream in our pipeline - not a proper preprocessing step, but as a first part of generating a valid geoparquet file (immediately before the "processing" portion of ingest code). Unlike other "default" preprocessing steps, I can't imagine a world where we don't want this.

The utils in `lifecycle.scripts.ingest_validation` were incredibly clunky and user-unfriendly before. Now you just need to run either a one-liner, or a few commands in a terminal and you'll have spit out a report, with simple command to re-run either library or ingest (if iterating on an ingest template), and another to reprint report. This will go in the wiki, but basically it's

```python3 -m dcpy.cli lifecycle scripts validate_ingest run_and_compare dcp_specialpurpose -k sdlbl```

`-k` is specifying a key column. No keys specified -> simple comparison. `-i` can specify columns to be excluded from comparison. Current generated columns `ogc_fid` and `data_library_version` are always ignored.

Will spit out a report. If you want to do bit by bit

```python3 -m dcpy.cli lifecycle scripts validate_ingest run dcp_specialpurpose``` (both library and ingest -> runs them, loads into db)

```python3 -m dcpy.cli lifecycle scripts validate_ingest load_single ingest dcp_specialpurpose``` (if you need to rerun just one of them)

```python3 -m dcpy.cli lifecycle scripts validate_ingest compare dcp_specialpurpose -k sdlbl```

If you don't normally operate in a dev container, you might need to run this in docker if your local version of gdal is 3.9.*

## Commits
1. minor dev container cleanup. simplify name of container so that commands can be run in it just specifying `de` rather than container id
2. make geoparquet files valid. On one hand, maybe should add a test, but this also feels trivial
3. time to start treating compare as a proper module -> split out models
4. rename a test file, to add in some functionality
5. add a pydantic class which can contain a pd.DataFrame but still support simple equality checks. See comment [here](https://github.com/NYCPlanning/data-engineering/pull/1236#discussion_r1840735779)
6. test df keyed report functionality - maybe a little futile. We don't actually use these right now, for the sql tables I compare them in sql rather than reading into pd dataframe and then comparing, mainly for performance reasons. However, if we later add something where we spin up a temp db for testing, it'd be easy to load these dfs and repeat these tests using the currently untested sql functions
7. cleanup from a previous change - decided type changes aren't needed if one table is missing the column entirely
8. move comparison models to a submodule. Feel more appropriate here without a silly wrapper class pretending to be a module
9. expand geom comparisons, finally! The whole reason for this PR. after digging around a bunch in postgres, this seems fine for now. Could easily expand this query later
10. rework all the ingest validation functions. As mentioned, they were clunky, and as I reran these commands it became clear that there should only be a few steps 1. run ingest or library or both and load, 2. compare them

### CODE REVIEW

File: dcpy/models/base.py [hunk 1]
```diff
@@ -131,3 +132,50 @@ def from_yaml(cls, yaml_str: str, *, template_vars=None) -> typing.Self:
     def from_path(cls, path: Path, *, template_vars=None) -> typing.Self:
         with open(path, "r", encoding="utf-8") as raw:
             return cls.from_yaml(raw.read(), template_vars=template_vars)
+
+
+class ModelWithDataFrame(BaseModel, arbitrary_types_allowed=True):
```
fvankrieken, line 137: @alexrichey - this is all for the purpose of testing and checking equality of objects like `comparison.Report`. I guess on one hand, with a pd DataFrame we're sort of getting outside of pydantic's wheelhouse (DFs are not super easily serializable - though I guess you could go to a json representation of the data), and I have no plans of dumping these to json or yml (or reading them). I still prefer these to dataclasses, mainly in that they serve an overlapping need and I'd rather stick with the object whose api I'm familiar with even if a dataclass might be more specific to my need here.

And I'd still have to write this equality function for a dataclass doing the same thing, since the same issue would exist.

Anyways just rambling and curious if you have any big thoughts

alexrichey: Yeah, I agree. I think Pydantic classes avoid some pitfalls of [regular|data]classes and provide better documentation, so happy to just default to inheriting from BaseModel everywhere

damonmcc:
@fvankrieken
> for the sql tables I compare them in sql rather than reading into pd dataframe and then comparing

nice. that does feel like an ideal place to compare stuff (when possible). do all of those queries live in `data.compare`?

fvankrieken:
> @fvankrieken
> 
> > for the sql tables I compare them in sql rather than reading into pd dataframe and then comparing
> 
> nice. that does feel like an ideal place to compare stuff (when possible). do all of those queries live in `data.compare`?

Yup

damonmcc (approved):
lgtm!

dcpy/test/models/test_base.py
```diff
@@ -0,0 +1,162 @@
+from __future__ import annotations
+import pandas as pd
+from typing import Literal
+
+from dcpy.models.base import SortedSerializedBase, ModelWithDataFrame
+
+
+class TestSerialization:
```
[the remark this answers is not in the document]
fvankrieken: They do not. And actually recently I've run into unittest (TestCase specifically) and pytest not playing nicely together. Breaks paramterization of test functions, for example

dcpy/models/base.py
```diff
@@ -131,3 +132,49 @@ def from_yaml(cls, yaml_str: str, *, template_vars=None) -> typing.Self:
...
+        therefore, df.equals() must be used
+
+        Only supported types so far are
+        - DataFrame
+        - dict[Any, DataFrame]
+
+        This could be extended to include other parameterized generic types
+        """
+
+        if type(other) is not type(self):
```
alexrichey, line 151: speaking of footguns that Pydantic avoids, this feels like the most common one I encounter. (since type(myRegularClass)==`type`)

alexrichey (approved):
Looks great!

PR #1241, opened 2024-11-08
Data Validation Framework: Source + Product data

### LINKED ISSUE #650: Ingest v2 - validate input data
[stub]

sf-dcp:
Partially resolves #650

## What
This PR implements a *raw* architecture of data checks with Pandera in dcpy.

In order to run data checks on a dataframe with Pandera, you need to construct a pandera object, `pandera.DataFrameSchema`. The Pandera object takes in`pandera.Column`s which consist of `pandera.Check`s. Most of the PR is about creating these pandera objects in `validate/pandera_utils.py` module. The rest of the PR is creating a format for data checks in our templates.

Example of running data checks from Pandera docs: 
```python
import pandas as pd
import pandera as pa

# data to validate
df = pd.DataFrame({
    "column1": [1, 4, 0, 10, 9],
    "column2": [-1.3, -1.4, -2.9, -10.1, -20.4],
    "column3": ["value_1", "value_2", "value_3", "value_2", "value_1"],
})

# define schema
schema = pa.DataFrameSchema({
    "column1": pa.Column(int, checks=pa.Check.le(10)),
    "column2": pa.Column(float, checks=pa.Check.lt(-1.2)),
    ]),
})

validated_df = schema(df)
print(validated_df)
``` 

Example of data check format in ingest templates/metadata files:
```yaml
columns:
  - id: bbl
    data_type: text
    is_required: True
    checks:
      - str_contains:
          args:
            pattern: "75"
          warn_only: False

  - id: custom_value
    data_type: integer
    is_required: False
    checks:
      - greater_than:
          description: This column must be greater than zero
          args:
            min_value: 0
      - unique_values_eq:
          description: This column must contain at least these values.
          warn_only: True
          args:
            values: [5, 10]
```

Check names in a template correspond to pandera check names or our custom checks (same approach as in preprocessing steps). In the last commit, I created a custom check as an example how we would do it; the check itself can be scraped later.

I recommend to review this PR commit by commit or starting with the `validate.pandera_utils.run_data_checks()` fn and work your way backwards.

## 🚨 Feedback needed
* Naming of models/functions
* `validate.pandera_utils.create_check()` fn: I have some validation logic in it where it checks correct input values (i.e. such check name or check parameters exist). I would like to move it under `dataset.CheckAttributes` model -- does it make sense to do that? 
* Better test structure?
* How to appease `mypy` in the meantime, before pandera checks are fully implemented. 

## Next steps/PRs
* Implement data types in pandera (map our custom defined data types in templates to pandera data types and custom defined data types like `bbl`)
    * Ex: `text` -> `pandera.dtypes.DataType.String`
* Implement custom data checks.
* Implement `validate.run()` function which reads in a local file into df, pulls in a template, and runs data checks. 
* Refactor ingest/distribution code to use new `checks` data type. Refactor distribution metadata files.

### References
- Discussion: #902
- Alex's draft PR: #859
- Finn's draft PR: #862


alexrichey:
FYI, just reading through. This all makes sense (in particular, how we're constructing the checks) but I want to read through their docs a little more to understand the ecosystem.

dcpy/lifecycle/validate/pandera_utils.py
```diff
@@ -0,0 +1,119 @@
...
+    elif isinstance(check, dict):
+        assert (
+            len(check) == 1
+        ), "`utils.create_pa_check` expects exactly 1 key-value pair in `check` param."
+        check_name, check_args = next(iter(check.items()))
+
+    if check_name not in allowed_check_names:
+        raise ValueError(f"Unregistered check name: '{check_name}'.")
+
+    create_check = getattr(pa.Check, check_name)
```
alexrichey, line 46: Should probably rename, to avoid shadowing the outer function_name. Maybe `pandera_registered_check` or similar?

dcpy/lifecycle/validate/pandera_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        raise ValueError(f"Unregistered check name: '{check_name}'.")
+
+    pandera_registered_check = getattr(pa.Check, check_name)
+
+    if check_args:
+        check_expected_params = signature(pandera_registered_check).parameters
+        invalid_check_keys = set(check_args.args.keys()) - set(
+            check_expected_params.keys()
+        )
+        if invalid_check_keys:
```
alexrichey, line 53: So looking at the Pandera [source](https://github.com/unionai-oss/pandera/blob/813420db5092a7ad4a4ce338b18cf56b1277fbba/pandera/api/checks.py#L294) for this, and I think the intent is to be able to pass **kwargs through the `cls.from_builtin_check_name` pa.Check constructor. I'm guessing that comes into play with custom-registered checks. Which maybe in the long-run means we shouldn't do this type-checking. For now, I think it's fine.

dcpy/models/dataset.py
```diff
@@ -29,4 +41,15 @@ class Column(SortedSerializedBase, extra="forbid"):
...
+        None  # TODO: delete Checks after refactoring metadata
+    )
+
+    @field_validator("checks", mode="after")
+    @classmethod
+    def check_checks(cls, checks: list[str | dict[str, CheckAttributes]]):
+        if isinstance(checks, list):
+            for check in checks:
+                if isinstance(check, dict) and len(check) != 1:
+                    raise ValueError(f"{check} must contain exactly one key-value pair")
```
alexrichey, line 54: thoughts about just adding a `name` attribute to `CheckAttributes` and storing it as a list, to avoid this type of validation?

dcpy/models/dataset.py
```diff
@@ -29,4 +41,15 @@ class Column(SortedSerializedBase, extra="forbid"):
...
+    checks: Checks | list[str | dict[str, CheckAttributes]] | None = (
+        None  # TODO: delete Checks after refactoring metadata
+    )
+
+    @field_validator("checks", mode="after")
+    @classmethod
+    def check_checks(cls, checks: list[str | dict[str, CheckAttributes]]):
+        if isinstance(checks, list):
+            for check in checks:
+                if isinstance(check, dict) and len(check) != 1:
```
alexrichey, line 53: thoughts about just adding a name attribute to CheckAttributes and storing it as a list, to avoid this type of validation?

dcpy/models/dataset.py
```diff
@@ -14,12 +16,22 @@
 ]
 
 
-# TODO: extend/modify Checks model
+# TODO: DELETE
 class Checks(SortedSerializedBase):
     is_primary_key: bool | None = None
     non_nullable: bool | None = None
```
damonmcc, line 22: does this mean `Checks` can be deleted in this PR?

dcpy/models/dataset.py
fvankrieken: Hmm. I do think on the python side having `list[str | CheckAttributes]` just feels better. But the current implementation gives us a very dbt-like structure for defining tests in yml, which while that shouldn't necessarily be taken as the "best" format for writing tests, being consistent feels useful. See code from cdbg

```yml
- name: int__lot_block_groups_details
  description: int__lot_block_groups joined to pluto for lot info
  columns:
  - name: bbl
  - name: block_group_geoid
    tests:
    - not_null
    - relationships:
        to: ref('int__block_groups_raw')
        field: geoid
```

I suppose maybe this could be done with some sort of pydantic class with very custom validation/serialization/deserialization? That also seems like a can of worms though. 

I'm not quite sure, need to think on it a little

dcpy/models/dataset.py
```diff
@@ -14,12 +16,22 @@
...
 
 
+class CheckAttributes(SortedSerializedBase, extra="forbid"):
+    """
+    Represents constraints and optional metadata for column checks.
+    """
+
+    args: dict[str, Any]
```
fvankrieken, line 30: Speaking of dbt consistency - it would be sort of nice to have args not an extra level down. I guess we'd need to have `extra="allow` and then some sort property to get all "extra" arguments to CheckAttributes. Seems like a similar thing to the issue in `Column` - decide whether we want to prioritize simplicity in yml or simplicity in python

dcpy/lifecycle/validate/pandera_utils.py
```diff
@@ -0,0 +1,78 @@
...
+            len(check) == 1
+        ), "`utils.create_pa_check` expects exactly 1 key-value pair in `check` param."
+        check_name, check_args = next(iter(check.items()))
+
+    if check_name not in allowed_check_names:
+        raise ValueError(f"Unregistered check name: '{check_name}'.")
+
+    pandera_registered_check = getattr(pa.Check, check_name)
+
+    if check_args:
```
fvankrieken, line 48: if you want to make use of it, you could call `dcpy.utils.introspect.validate_kwargs` here - it would validate types of arguments as well. You'd still need to raise the error based on the returned object

dcpy/test/lifecycle/validate/test_pandera_utils.py
```diff
@@ -0,0 +1,87 @@
...
+    ],
+)
+def test_create_check_success(input_data, expected_pa_check):
+    actual_check = pandera_utils.create_check(input_data)
+    assert actual_check == expected_pa_check
+
+
+@pytest.mark.parametrize(
+    "input_data, error_message_substring",
+    [
```
fvankrieken, line 68: nit - seems like a nice spot to `zip` two lists

dcpy/lifecycle/validate/run.py
fvankrieken, line 1: I'd be tempted to call this file `data.py` or something like that, since that's really the whole purpose here - we're validating data (and I can imagine other things we need to validate that live under `dcpy/lifecycle/validate`

dcpy/models/lifecycle/ingest.py
```diff
@@ -93,6 +93,7 @@ class Template(BaseModel, extra="forbid"):
     attributes: DatasetAttributes
     ingestion: Ingestion
     columns: list[Column] = []
+    checks: list[str | dict[str, Any]] | None = None
```
fvankrieken, line 96: Would we want this as a separate list? Or would we want `columns` above to be the new `Column` type created in this PR?

dcpy/lifecycle/validate/pandera_utils.py
```diff
@@ -0,0 +1,119 @@
...
+    elif isinstance(check, dict):
+        assert (
+            len(check) == 1
+        ), "`utils.create_pa_check` expects exactly 1 key-value pair in `check` param."
+        check_name, check_args = next(iter(check.items()))
+
+    if check_name not in allowed_check_names:
+        raise ValueError(f"Unregistered check name: '{check_name}'.")
+
+    pandera_registered_check = getattr(pa.Check, check_name)
```
sf-dcp, line 46: # TODO: rename this to "check constructor" and add a comment about extra kwargs being applied. lay out thought process

dcpy/lifecycle/validate/pandera_utils.py
alexrichey: Resolving. Chatted in person, and basically you can pass these kwargs either to the built-in fn constructor (which we're disallowing here) or to the Check constructor, which is 👍

dcpy/lifecycle/validate/run.py
alexrichey: you could also just move this into the __init__.py for the module, so you get a nice 
```
from dcpy.lifecyle import validate
validate.run(...)
```

dcpy/models/lifecycle/ingest.py
sf-dcp: These will be schema level checks that haven't been implemented. An example of such check is a unique combination of columns

dcpy/lifecycle/validate/run.py
sf-dcp: yeah that makes sense!

dcpy/models/dataset.py
sf-dcp: Yeah, he idea behind `<check name>: { <params> }` format was to follow `dbt` as close as possible.

dcpy/models/lifecycle/ingest.py
fvankrieken: Ah right - in addition to `columns` being the new columns type?

dcpy/models/dataset.py
sf-dcp: yeah, I agree it would be nice to follow `dbt` tests as close as possible. I decided to encapsulate check-specific params into `args` key here because I want it to have validation on both generic `Check` and check-specific params. I'm not how I can maintain validation with args being flattened. Ideas?

alexrichey: @fvankrieken pored over the Pandera source with @sf-dcp, and the rationale here is that there are basically two levels of configuration for these checks:
- function specific, e.g. what are the args that "less_than" requires, as [here](https://github.com/unionai-oss/pandera/blob/813420db5092a7ad4a4ce338b18cf56b1277fbba/pandera/api/checks.py#L300)
- more general args like `group_by` that are defined in the general constructor for a check, as [here](https://github.com/unionai-oss/pandera/blob/813420db5092a7ad4a4ce338b18cf56b1277fbba/pandera/api/checks.py#L27)

so `args` here are function specific. In pandera land, makes sense to have them separate... can't speak for dbt land though.

dcpy/models/dataset.py
fvankrieken: Gotcha. And it ties into [this comment](https://github.com/NYCPlanning/data-engineering/pull/1241#discussion_r1913460076). There a couple spots in where we need to decide between simpler yml vs simpler python. I'd be inclined to make the same decision in both spots (as in, leave this as is, but don't have the `args` param but instead have like an `args` __property__ which returns any extra inputs to the model)

Don't think it's needed as part of this PR as it's really an aesthetics/ergomonics choice, but just something to think about before we fully operationalize this stuff

dcpy/test/lifecycle/validate/resources/valid_columns_with_checks.yml
```diff
@@ -0,0 +1,22 @@
+- id: bbl
+  data_type: text
+  is_required: True
+  checks:
```
fvankrieken, line 4: would be nice to have a plain str check here for completeness!

fvankrieken (approved):
A couple small notes/thoughts but
1. this all seems great to me, big picture-wise
2. you and Alex seem to be in a good spot (and agreement) about some of the details, so I'll continue to peek and give any wanted feedback, but carry on!

dcpy/models/dataset.py
fvankrieken: Oh interesting. Would we have another field for defining those "generic constructor" args then potentially?

I was thinking something like this logic:

https://stackoverflow.com/questions/69617489/can-i-get-incoming-extra-fields-from-pydantic

But if you have two distinct categories, that would be tough. Maybe "extra" inputs would be handled as args, and we could have `pa_generic_constructor_args: dict[str, any]` to handle those args?

Anyways - this isn't meant to hold up this PR. As noted [here](https://github.com/NYCPlanning/data-engineering/pull/1241#discussion_r1913548773) this is a little more on the aesthetics/ergonomics side, and don't think it should hold up the work of iterating to a useful implementation

dcpy/test/lifecycle/validate/resources/valid_columns_with_checks.yml
fvankrieken: Oh okay. Sounds good

dcpy/models/lifecycle/ingest.py
sf-dcp: resolving as a result of in-person chat

dcpy/models/dataset.py
sf-dcp: @fvankrieken the generic args (only 2 for now) are explicitly defined inside of the `CheckAttributes` model: 

```python
class CheckAttributes(SortedSerializedBase, extra="forbid"):
    """
    Represents constraints and optional metadata for column checks.
    """
    # TODO: add more description. this class mirrors ...

    args: dict[str, Any]   # encapsulated check-specific args
    description: str | None = None   # generic `Check` arg
    warn_only: bool = False   # generic `Check` arg
```

Thanks for the link! I will try to implement it in this PR if it's not too messy 🙃

dcpy/lifecycle/validate/run.py
sf-dcp: I like both ideas. I think for now, I will rename to module to `data.py` in case we decide to extend things to validate. In that case, we can have multiple functions in the `init.py` like `run_data_checks` and `run_my_other_checks`

sf-dcp:
@fvankrieken @damonmcc @fvankrieken Should I merge with unhappy `mypy`?

fvankrieken:
> @fvankrieken @damonmcc @alexrichey Should I merge with unhappy `mypy`?

No, those should be resolved in some way, particularly the couple failures outside of `validate`.

You also have a pytest failure - looks like import issues due to renaming

PR #1243, opened 2024-11-12
Use ingest with fallback to library

### LINKED ISSUE #1234: Ingest - try ingest, default to library
Before we start rolling out ingest, we need
1. a dcpy cli target which will call `ingest` if the dataset is defined within `ingest/templates`, or `library` if it isnt
2. updated ghas to use this cli target

fvankrieken:
Closes #1234

Lingering - might make sense to rename all these recurring actions to all be called "ingest", however, that would make them not able to trigger via workflow dispatch on a feature branch. I can do that in a follow up PR

# Integration tests
## ingest single gha - use ingest
[run](https://github.com/NYCPlanning/data-engineering/actions/runs/11801531812/job/32875074160)

result:
image

## ingest single gha - fall back to library
[run](https://github.com/NYCPlanning/data-engineering/actions/runs/11801579126/job/32875247569)

result:
image

## bytes quarterly
(all fall back to library)

[run](https://github.com/NYCPlanning/data-engineering/actions/runs/11801638150)

Fails due to an issue with dcp_lion that is to be resolved, and congressional districts due to it being on bytes as "24c1" rather than "24c" like the others

## opendata recurring
(all fall back to library)

[run](https://github.com/NYCPlanning/data-engineering/actions/runs/11801633496)

## arcgis online recurring
(all fall back to library)

[run](https://github.com/NYCPlanning/data-engineering/actions/runs/11801545849)

Fails due to an issue with natural heritage communities unrelated to this PR. I'll investigate separately


alexrichey (approved):
Cool! I could also see putting `run_fall_back_to_library` in the lifecycle/scripts folder to avoid referencing library in ingest, but this seems good

fvankrieken:
> Great! Looks good

Going to add one more commit to rename GHAs

PR #1263, opened 2024-11-21
Extract XLSX logic + Add Data Dictionary + Add Artifacts

alexrichey:
TLDR: I'm working on a real writeup for this, and adding docstrings to make this more readable. If you get there beforehand, the commits are quite atomic and could be easily read that way. Also, there are two tests that we'd expect to fail atm, since they rely on changes in the metadata repo [here](https://github.com/NYCPlanning/product-metadata/pull/24/files)

There are a few things going on here:
1. A rewrite of the XLSX code to pull our business logic out, and make it a generalized generator of worksheets with tables in them. This bit I feel pretty good about, although I think we need to add back logic about table headers, rather than generating them abstractly.

2. (Here's where things get muddy) Facilities to convert Dataset attributes into a table format / abstract of a table. This includes design elements like Tables/Rows/Cells/Styles in `design.elements.py`, as well as `abstract_doc.py` which will turn a Pydantic model into a tabular_report'ish thing. **I wouldn't spend too much time on details here** but would love big thoughts about how to simplify this. [Here's](https://github.com/NYCPlanning/data-engineering/pull/1263#issuecomment-2517924549) a comment outlining my thoughts.

3. Adds a data dictionary for fields on our metadata. This concept feels solid, so nitpicks very welcome.

4. Adds declarative instructions for generating XLSX files, including what columns to include, and what objects to table'ify. Again, I think this is a solid concept that needs some help settling in. Please pick those nits.

5. Modifications to existing scripts to start from the full OrgMetadata instead of specific Datasets. In general, we should really never use the dataset metadata directly, as it will not inherit Org attributes, the data dictionary, or artifacts.

I general, I'd love to get input to clean up 1, 3, and 4 above, then merge. 

Then for 2, I think we can rewrite this to accommodate the HTML/PDF generation. So for 2) big picture thoughts welcome, but detailed review isn't needed, unless you notice something glaringly wrong.


dcpy/lifecycle/package/abstract_doc.py
fvankrieken, line 1: I do think with formatted cells and rows, we're really talking about a `Workbook` or `Sheet` or `Worksheet` here more so than a `doc`

fvankrieken: I know part of the idea is to abstract from that a bit, but I don't think we can quite escape it

fvankrieken:
Not to muddy the waters but have you looked at xlsxwriter? Seems to have a bit more support for formatting. Though it seems like you've managed to make things work pretty well with openpyxl

dcpy/lifecycle/package/abstract_doc.py
fvankrieken: Maybe its a "formatted table"? "tabular report"? The crux of it seems to be these things that are undoubtedly worksheet tables, with some decoration/metadata that then a specific implementation gets to decide exactly how to dump it out

alexrichey: I like tabular report. However, I do think you could pretty easily render this as HTML, so while it's currently just Excel workbooks, I think that might be overly specific.

fvankrieken: Even when we're talking about "rows" of "cells" in `models.design.elements`? That seems distinctly tabular in terms of language choice. But I think that's more me getting caught up on semantics - Is the thought that these still might be displayed in a more nested (json-like) format in an html/pdf? Where each "column" is say a header, each "row" a subheader, and each "cell" and element under that

alexrichey:
@fvankrieken I'm currently fixing a few things, and documenting. If you happen to start reviewing imminently, here's what I'm thinking kind of broadly:

I like your idea of a `tabular_report` - specifically I'm thinking of it as an object you can generate for any of our Pydantic models (though just starting with Metadata), that will translate a pydantic class into information displayable in a table, with the field metadata if it's available. Then we can pass the `tabular_report` to different renderers (ie XLSX, or HTML for @damonmcc). It would also carry around some suggested styling for cells, (e.g. if you render the column.values rows as a str, you should use a monospaced font) that renderers can use or ignore. 

So that'd mean probably removing Rows and Cells from design.elements, and making the XLSX writer a little more prescriptive about generating tables (as opposed to constructing them entirely abstractly)

But I think the DataDictionary and Artifact side of things are pretty solid concepts that need some sharpening. So would love feedback on that. 

Attached is a sample XLSX for Zoning Map Amendments. It's a good example for the table formatting in the `Column Information` tab.

[zoning_map_amendments_data_dictionary.xlsx](https://github.com/user-attachments/files/18011259/zoning_map_amendments_data_dictionary.xlsx)

dcpy/lifecycle/package/abstract_doc.py
```diff
@@ -0,0 +1,360 @@
...
+from dcpy.utils.logging import logger
+
+
+# TODO; Extract these into a generic style that we can pass to the XLSX renderer
+BLUE = "FF009DDC"
+TITLE_FONT_SIZE = 18.0
+MONOSPACED_FONT = "Consolas"
+
+
+def _make_title_subtitle_cell(title: str, subtitle: str):
```
alexrichey, line 17: @fvankrieken some additional context on a previous comment: this type of logic is what I'm thinking we'll move into the XLSX module, along with basically everything in _make_table_top. I think it's really only useful when made concrete in a renderer.

dcpy/lifecycle/package/abstract_doc.py
```diff
@@ -0,0 +1,360 @@
...
+            dictionary_section = org_metadata.data_dictionary.dataset["revisions"]
+        case "dataset.attributes":
+            dictionary_section = org_metadata.data_dictionary.dataset["attributes"]
+        case _:
+            raise Exception(f"Unknown data_source: {data_source}")
+    return dictionary_section
+
+
+# TODO: move to pydantic models
+def get_data_source(
```
alexrichey, line 236: @fvankrieken and this would be folded into the `tabular_report`, which could probably be a Pydantic class from which we'd inherit. (would probably move `get_field_metadata` into that class as well)

dcpy/models/base.py
```diff
@@ -26,6 +26,19 @@ class SortedSerializedBase(BaseModel):
     _exclude_falsey_values: bool = True
     _head_sort_order: list[str] = PrivateAttr(default=["id"])
     _tail_sort_order: list[str] = PrivateAttr(default=["custom"])
+    _repr_functions: dict[str, typing.Callable[[typing.Any], str]] = {}
+
+    def field_repr(self, field_name: str):
```
alexrichey, line 31: thinking we'll move this functionality into a `TabularReport` class

dcpy/test/lifecycle/package/test_abstract_docs.py
```diff
@@ -0,0 +1,176 @@
...
+            assert (
+                extra_note in extra_description_cell
+            ), "The extra description from the third party should be included in the description"
+
+    assert (
+        found_field_with_third_party_note
+    ), "Sanity check that we're actually testing the extra description"
+
+
+def test_generating_revisions(org_metadata: OrgMetadata):
```
fvankrieken, line 100: little nit - test name seems off

dcpy/models/product/data_dictionary.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class FieldSet(CustomizableBase):
+    fields: dict[str, FieldDefinition] = {}
+
+
+class DataDictionary(CustomizableBase, TemplatedYamlReader):
+    org: dict[str, dict[str, FieldDefinition]] = {}
+    product: dict[str, dict[str, FieldDefinition]] = {}
+    dataset: dict[str, dict[str, FieldDefinition]] = {}
```
fvankrieken, line 17: Slowly working through this. What are your thoughts on this vs each of these being a list of `FieldSet`s, and then FieldSet could also have a `label` field. If these are analagous to tabs in an excel sheet, at first thought
```yml
datasets:
  columns:
    column_name:
      summary: column_name
      extra_description: Name of ...
  attributes:
     ...
```
is less intuitive than
```yml

datasets:
- name: column
  fields:
    column_name:
      summary: column_name
      extra_description: Name of ...
```

You can have a similar argument for the inner dictionary as well. It's definitely more of a usual pattern for our codebase, so just wanted to know what drew you to this.

Also a little curious how you imagine `org` and `product` playing in here, but can talk about that in person

dcpy/models/product/metadata.py
```diff
@@ -123,6 +124,7 @@ class OrgMetadata(SortedSerializedBase, extra="forbid"):
     template_vars: dict = Field(default_factory=dict)
     metadata: OrgMetadataFile
     column_defaults: dict[tuple[str, COLUMN_TYPES], DatasetColumn]
+    data_dictionary: DataDictionary = DataDictionary()
```
fvankrieken, line 128: A bit pedantic, but I think I'd prefer something like `DataDictionarySpec` just to clarify that this is a definition of the format of the data dictionary rather than an instance of one

.github/workflows/template_build.yml
```diff
@@ -50,6 +50,15 @@ jobs:
...
       - uses: actions/checkout@v4
 
+      - uses: actions/checkout@v4
+        with:
+          repository: NYCPlanning/product-metadata
+          path: product_metadata
+
+      - name: set_product_metadata_path
+        run:  echo "PRODUCT_METADATA_REPO_PATH=$(pwd)/product_metadata" >> $GITHUB_ENV
+        working-directory: ./
```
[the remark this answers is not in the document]
alexrichey: Eventually, for sure something like that... we'd be using Github as a database at that point, and would probably have to rewire how metadata is deserialized (ie to grab over via http). It'd probably make more sense to just "compile" each dataset's metadata, ie compute the dataset with all the org/product overrides, and store that somewhere. Open to ideas though!

damonmcc:
@alexrichey from looking at the sample XLSX for Zoning Map Amendments you linked to ([zoning_map_amendments_data_dictionary.xlsx](https://github.com/user-attachments/files/18011259/zoning_map_amendments_data_dictionary.xlsx)), this is great!! I compared it to the XLSX on the (private for now) Open Data page for Zoning Map Amendments and they seem identical to me

when you get through the last fixes, happy to approve the PR. before and after that, I'll keep looking through the changes and following the convos

dcpy/lifecycle/package/abstract_doc.py
```diff
@@ -0,0 +1,360 @@
...
+    maybe_image_path: Path | None = None,
+    column_widths: list[float] | None = None,
+) -> de.Table:
+    """Make a table for an object.
+    Each row is an attribute of the object.
+    There are only two columns
+        - formatted key (summary + description)
+        - value
+    """
+    rows = _make_table_top(
```
fvankrieken, line 130: Not 100% on this - but at this level of abstraction it still feels slightly odd to me to have both the "top" and actual "table" all present as "row"s. Not that it's entirely improper - even for a non-excel output, to some extent it makes sense to turn each "table"/section(/tab) just into a series of rows/elements to be formatted. But it also seems similarly valid that these abstract "table"s could still have titles (or "header section", or whatever you want to call it - distinct entity from the tabular report itself), which could still already have some formatting but the individual formatters have their own opinion on how to put those together with the table.

That said, there's also some elegance in the specific formatters just getting a bunch of "row"s and knowing exactly what to do with each row. So really I'm just thinking out loud here without any helpful specific feedback lol

fvankrieken:
> @fvankrieken I'm currently fixing a few things, and documenting. If you happen to start reviewing imminently, here's what I'm thinking kind of broadly:
> 
> I like your idea of a `tabular_report` - specifically I'm thinking of it as an object you can generate for any of our Pydantic models (though just starting with Metadata), that will translate a pydantic class into information displayable in a table, with the field metadata if it's available. Then we can pass the `tabular_report` to different renderers (ie XLSX, or HTML for @damonmcc). It would also carry around some suggested styling for cells, (e.g. if you render the column.values rows as a str, you should use a monospaced font) that renderers can use or ignore.
> 
> So that'd mean probably removing Rows and Cells from design.elements, and making the XLSX writer a little more prescriptive about generating tables (as opposed to constructing them entirely abstractly)
> 
> But I think the DataDictionary and Artifact side of things are pretty solid concepts that need some sharpening. So would love feedback on that.
> 
> Attached is a sample XLSX for Zoning Map Amendments. It's a good example for the table formatting in the `Column Information` tab.
> 
> [zoning_map_amendments_data_dictionary.xlsx](https://github.com/user-attachments/files/18011259/zoning_map_amendments_data_dictionary.xlsx)

I like this - when you describe it this way, I feel like this is a (much) more complex version of the little things I was bumping up against with my `indented_print`. More complex formatting obviously. For generalizing to pydantic models, curious on your thoughts around levels of nested models - would you "truncate" at some point in building a more generic report? Or handle that at the level of renderer? This is your tabulate case - something that needs to be coerced to a cell itself has nested structure. So in other words, should the "abstract_doc" have some knowledge of how deep the data can go, or should a renderer have logic to, when it finds another table inside a cell, to call "tabulate" or something like that. Or do we leave all of that up to the specific artifact definition

dcpy/lifecycle/package/abstract_doc.py
fvankrieken: But I think a good path forward, as said in person, is to test this out on html/pdfs and see what feels useful there

fvankrieken (approved):

PR #1281, opened 2024-11-25
Source Data Checks: Refactor column class for ingest & distribution

sf-dcp:
~~Small~~ PR to refactor existing Column objects in ingest & distribution code into one. 

### Why? 
For our own sanity when implementing data checks on source & product data. 

Easiest to go commit by commit. 

### Major changes: 
1. Create base `models.dataset.Column` class. 
    * Note, I added `is_required` attribute which will be used in validation.
2. Revise old ingest and metadata `Column` objects to inherit from the new class
    * for metadata, remove a code snippet used in creating metadata raw templates
4. Move `metadata_v2.Checks` class to `models.dataset`
5. Appease mypy
    * fix inconsistencies in `data_type` allowed values

Data checks will be implemented in next PR.

### DIFF
--- a/dcpy/connectors/socrata/metadata.py
+++ b/dcpy/connectors/socrata/metadata.py
@@ -58,7 +58,9 @@ def make_dcp_col(c: pub.Socrata.Responses.Column) -> md.DatasetColumn:
 
     # model_construct() method doesn't perform validation on keys, need this sanity check here
     # instance keys == column model keys below:
-    dataset_column.__dict__.keys() == dataset_column.model_fields.keys()
+    assert (
+        dataset_column.__dict__.keys() == dataset_column.model_fields.keys()
+    ), "DatasetColumn instance keys don't match the DatasetColumn class keys"
 
     return dataset_column
 


dcpy/models/lifecycle/ingest.py
```diff
@@ -77,12 +78,10 @@ class Ingestion(SortedSerializedBase):
...
-class Column(SortedSerializedBase):
-    id: str
-    data_type: Literal[
-        "text", "integer", "decimal", "geometry", "bool", "date", "datetime"
-    ]
-    description: str | None = None
+class Column(BaseColumn):
+    _head_sort_order = ["id", "data_type", "description"]
+
+    data_type: COLUMN_TYPES  # override BaseColumn `data_type` to be required field
```
sf-dcp, line 84: In the `BaseColumn`, data_type can be of `COLUMN_TYPES` or `None`. Making it a required field for ingest.

sf-dcp:
There is one failing pytest that I'm not sure how to address -- would love an advice.

In `dcpy.connectors.socrata.metadata` module, we have a function (`make_dcp_col`) that generates metadata templates from socrata api where column `data_type: <FILL ME IN>`. So the test fails because <FILL ME IN> isn't a valid data_type value. 

One quick solution is to add <FILL ME IN> value to the list of allowed values for `data_type`.

 This is probably not something we will be using in the future; however, it could be useful for other agencies... But is it worth revising the validator to accommodate the edge case?

fvankrieken:
> There is one failing [pytest](https://github.com/NYCPlanning/data-engineering/actions/runs/12040579020/job/33570785388?pr=1281#step:6:1918) that I'm not sure how to address -- would love an advice.
> 
> In `dcpy.connectors.socrata.metadata` module, we have a function (`make_dcp_col`) that generates metadata templates from socrata api where column `data_type: <FILL ME IN>`. So the test fails because isn't a valid data_type value.
> 
> One quick solution is to add value to the list of allowed values for `data_type`.
> 
> This is probably not something we will be using in the future; however, it could be useful for other agencies... But is it worth revising the validator to accommodate the edge case?

Is there a way to avoid running validators at runtime in pydantic? Pass an extra arg to the constructor or something along those lines

dcpy/models/dataset.py
```diff
@@ -0,0 +1,32 @@
...
+class Column(SortedSerializedBase, extra="forbid"):
+    """
+    An extensible base class for defining column metadata in ingest and product templates.
+    """
+
+    id: str
+    data_type: COLUMN_TYPES | None = None
+    description: str | None = None
+    is_required: bool = True
+    checks: Checks | None = None
```
fvankrieken, line 32: Gut impression here - seems that `checks` should maybe be a list? Where then checks might take form
```
- name (or id): {name}
  parameter: value
```
Or something like that. Could just be a list of strings too. Will think more (and look at our old discussions) and revisit tomorrow

dcpy/models/lifecycle/ingest.py
fvankrieken: I had description as part of these columns?? I had no idea lol. This note of yours makes sense though - I don't think the code comment is needed

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -374,11 +350,13 @@ def validate_consistency(self):
...
 
     def apply_column_defaults(
-        self, column_defaults: dict[tuple[str, str], DatasetColumn]
+        self, column_defaults: dict[tuple[str, COLUMN_TYPES], DatasetColumn]
     ) -> list[DatasetColumn]:
         return [
-            c.override(column_defaults[c.id, c.data_type])
-            if c.data_type and (c.id, c.data_type) in column_defaults
-            else c
+            (
```
fvankrieken, line 356: tiny nit, sorry - :scissors: these parentheses?

dcpy/connectors/socrata/metadata.py
```diff
@@ -54,7 +54,7 @@ def make_dcp_col(c: pub.Socrata.Responses.Column) -> md.DatasetColumn:
         dcp_col["values"] = [
             {"value": s["item"], "description": FILL_ME_IN_PLACEHOLDER} for s in samples
         ]
-    md.DatasetColumn._validate_data_type = False
+    # md.DatasetColumn._validate_data_type = False # legacy attribute used during migration, no longer there
```
fvankrieken, line 57: Ah - this is why the error is popping up now?

fvankrieken: Given that this is code to create a placeholder, it seems fine to leave, no? And resolves our issue.

Or maybe it could even go in line 58? Not sure how this arg works. but maybe
```python
md.DatasetColumn(**dcp_col, _validate_data_type = False)
```
would work too?

fvankrieken: And per the commit message/comment - this I think is explicitly for the purpose of avoiding this error - I don't think it has to do with anything from migration from v1 to v2

fvankrieken:
Couple little notes but this seems very sensible

dcpy/models/dataset.py
sf-dcp: Not sure yet, I'm planning to focus on Checks in next PR. In this PR, I just moved the existing Checks obj Alex implemented to this file. If I don't do that, I run into an error where imports between `dataset.py` & `metadata_v2.py` are circular.

dcpy/connectors/socrata/metadata.py
sf-dcp: Yep, this is the reason for the error. If I revert the code back, it should solve the error. Counter q: would we won't this attribute `_validate_data_type` in the parent `Column` obj or `metadata.Column`? 

 Would love to get @alexrichey 's thoughts too on this

sf-dcp:
> > There is one failing [pytest](https://github.com/NYCPlanning/data-engineering/actions/runs/12040579020/job/33570785388?pr=1281#step:6:1918) that I'm not sure how to address -- would love an advice.
> > In `dcpy.connectors.socrata.metadata` module, we have a function (`make_dcp_col`) that generates metadata templates from socrata api where column `data_type: <FILL ME IN>`. So the test fails because isn't a valid data_type value.
> > One quick solution is to add value to the list of allowed values for `data_type`.
> > This is probably not something we will be using in the future; however, it could be useful for other agencies... But is it worth revising the validator to accommodate the edge case?
> 
> Is there a way to avoid running validators at runtime in pydantic? Pass an extra arg to the constructor or something along those lines

Need to look into this. Though `mypy` may still complain

dcpy/connectors/socrata/metadata.py
fvankrieken: I don't think we would. We only want to not validate in this case, when we're generating it with placeholders. Other than that, we should always validate at runtime on instantiation

dcpy/models/dataset.py
fvankrieken: Yeah I think that makes sense to keep for later, since we're not actually attempting to implement checks yet.

dcpy/connectors/socrata/metadata.py
alexrichey: Yeah, IIRC this was only for the purpose of spitting out a metadata yaml template with `column.type` of something like "< FILL ME IN! >". I'd just kill every reference to it.

fvankrieken: What do you mean specifically by kill every reference? Get rid of this util? Per @sf-dcp 's point about if we make this public, I think going from socrata metadata to dcpy metadata is a useful util to have

dcpy/models/product/dataset/metadata_v2.py
```diff
@@ -48,50 +48,26 @@ class CustomizableBase(SortedSerializedBase, extra="forbid"):
...
-    description: str | None = None
     notes: str | None = None
     example: str | None = None
-    checks: Checks | None = None
     deprecated: bool | None = None
     values: list[ColumnValue] | None = None
-
-    @field_validator("data_type")
-    def _validate_colum_types(cls, v):
-        if cls._validate_data_type:
```
fvankrieken, line 92: Ohh this was not a pydantic arg, this was a custom one defined by @alexrichey. Sorry @sf-dcp I thought that was some sort of built-in pydantic behavior

dcpy/connectors/socrata/metadata.py
fvankrieken: Found this in a github discussion

https://github.com/pydantic/pydantic/blob/6cda388c7a1af29d17973f72dfebd98cff626988/pydantic/main.py#L417

So `Column.construct(**args)` would return an unvalidated column object I think

alexrichey: oh, neat. @sf-dcp If 👆works, great. Overall, I'm in favor of whatever is fastest/easiest at the moment

sf-dcp: I like the idea of replacing all references of `< FILL ME IN! >` with `text`. This way, we keep the integrity while still offering an option to populate a raw template... 

Are we okay with that?

@alexrichey @fvankrieken

fvankrieken: I think I prefer having the placeholder - it really forces the user to actually enter something explicitly without making any assumptions. But again, just if this command works, otherwise let's just do away with it.

sf-dcp: let me try the placeholder... will report back

alexrichey (approved):
LGTM! though did you mean to delete `lpc_scenic_landmarks.yml` just now?

dcpy/connectors/socrata/metadata.py
sf-dcp: Tried the `construct` fn. It's now deprecated and the new [method](https://github.com/pydantic/pydantic/blob/6f31f8f68ef011f84357330186f603ff295312fd/pydantic/main.py#L297) is `model_construct()`. 

Sooo it seems like this method does no validation on input attributes & its values. For example, you can instantiate a model without any keys or random keys like below:

image
 
 Using this seems more problematic because we can potentially generate templates with invalid keys... @fvankrieken, I think we should go with the text field

sf-dcp:
> LGTM! though did you mean to delete `lpc_scenic_landmarks.yml` just now?

🤦‍♀️ didn't commit this before switching to the current branch. ty!

dcpy/connectors/socrata/metadata.py
fvankrieken: Hmm. At least having keys validated would be nice.

I'm not horribly concerned though - this is a util that's designed to create an invalid model, and this seems like a one-liner command to run it without error. Tests can make sure that structure is as expected. In terms of how a user would use it, this seems like less of a source of error to me than autofilling data_types that users might then forget to update. And this is slightly better than say just a dict in that it still shows developers what is expected.

sf-dcp: Makes sense. Added an assert statement for key validation!

PR #1282, opened 2024-11-25
Ingest - move connector objs, logic out of ingest and into classes

fvankrieken:
Don't think we should throw this wrench into `ingest` this second, but was thinking about "shotgun" surgery and how that applies in ingest. 

Currently, to add a new "source" (as @damonmcc will have to), you need to add logic in 3 or 4 places

With this sort of refactor, you'd only have to add it in one place. Also, it sort of sets us further down a path where users could register their own `Source`s

Would love people's thoughts - this is obviously half-baked as written, I haven't implemented all the other sources. And you do have to make sure you import the proper modules so that the code that registers the subclasses is actually run prior to pydantic parsing.

This needs the actual pydantic source classes to move out of models. Seems a little unavoidable if we want something along the lines of the simplicity of calling `source.download` without needing some sort of `ingest` handler that feeds a more declarative `source` model to a connector object that then could call the functions.


alexrichey:
I like this a lot. Could see a method in the ingest module to dynamically register a new connector to delegate to.

PR #1283, opened 2024-11-25
Ingest template generation - add attributes

fvankrieken:
Forgot this before - just missing the `attributes` section when you use `validate_ingest convert` to generate a template

### DIFF
--- a/dcpy/lifecycle/scripts/validate_ingest.py
+++ b/dcpy/lifecycle/scripts/validate_ingest.py
@@ -57,8 +57,7 @@ def convert_template(dataset: str):
         id=library_template["dataset"]["name"],
         acl=library_template["dataset"]["acl"],
         attributes=DatasetAttributes(
-            _exclude_falsey_values=False,
-            name=library_template["dataset"]["name"],
+            name="<CHANGE ME>",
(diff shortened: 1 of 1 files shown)


dcpy/lifecycle/scripts/validate_ingest.py
```diff
@@ -53,6 +56,12 @@ def convert_template(dataset: str):
     converter = Converter(
         id=library_template["dataset"]["name"],
         acl=library_template["dataset"]["acl"],
+        attributes=DatasetAttributes(
+            _exclude_falsey_values=False,
+            name=library_template["dataset"]["name"],
```
fvankrieken, line 61: Should I just make this an empty string? Something that a little more clearly needs to be changed

alexrichey: IIRC I used to just do something like "< CHANGE ME >" in metadata

alexrichey (approved):

dcpy/lifecycle/scripts/validate_ingest.py
fvankrieken: Yeah I'll just do that

Issue #561, opened 2024-01-29
DE<>GIS: Generate product READMEs from metadata


alexrichey:

damonmcc:
found an example of our code converting a markdown file to a pdf [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/ceqr/recipes/dep_cats_permits/runner.sh#L33)

alexrichey:
@damonmcc Is this the correct issue, or is there another?

damonmcc:
looking into using weasyprint (already the engine we're using in our use of pandoc) to do `html + css = pdf`

weasyprint samples [here](https://github.com/CourtBouillon/weasyprint-samples/tree/master)

alexrichey:
@damonmcc Here's how I'd go about the remainder of this task. Let's chat when you have a minute. In terms of timeline, the next check-in with Amanda+Matt+myself is Thursday next week, and I'd like to show progress on this item. POC/WIP is fine 🙂

## Data Dictionary (All fields come from metadata)
- [ ] POC of approach
     basic and ugly, but with some colors, and the DCP logo. Generated from manually created HTML with inlined CSS, fed into weasyprint. Goal is to prove out the approach, and get a sense of the difficulties.
- [ ] Create skeleton of Jinja HTML template. Probably start from scratch, but take inspo from Heng's work. Use real product metadata, but don't focus effort on transforming/testing the data.
    - [ ] get team feedback on template via draft PR
- [ ] Productionize:
    - Make the template look good
    - CSS should probably be in a separate file, and inlined*
    - tests, if applicable
- [ ] Implement auto-asset-autogeneration (similar to OTI data dict), and implement in TemplateDB


## README
Most fields from metadata, except changelog. Potentially other fields? TBD. e.g. rich text with images.
- [ ] POC of approach
     basic and ugly, but with some colors, and the DCP logo. Generated from manually created HTML with inlined CSS, fed into weasyprint. Goal is to prove out the approach, and get a sense of the difficulties.
- [ ] Create skeleton of Jinja HTML template with mix of real and mock data (mock the changelog). 
    - [ ] get team feedback on template via draft PR
- [ ] Productionize:
    - Make the template look good
    - CSS should probably be in a separate file, and inlined*
    - tests, if applicable
- [ ] Implement auto-asset-autogeneration (similar to OTI data dict), and implement in TemplateDB
[...]

PR #1296, opened 2024-12-02
compose and style data dictionary HTML/PDFs

damonmcc:
related to #561 

follow-up to #1269

These changes use the new approach for generating data dictionary HTML/PDFs from metadata yml:
- combine modular jinja templates
- apply css styling to HTML before generating the PDF
- combine org, product, and dataset metadata to determine all fields available for data dictionary generation

These changes do not apply all of the formatting seen in the Figma designs. I got stuck with nesting html elements and getting the text to wrap correctly so I focused on creating/using modular jinja files for a data dictionary.

## tests/outputs

Tested changes by running Template DB export stage locally and looking at the PDF and here's the PDF generated by the PR Template DB build [templatedb_data_dictionary.pdf](https://github.com/user-attachments/files/17983910/data_dictionary.pdf).

Also generated a PLUTO data dictionary: [pluto_data_dictionary.pdf](https://github.com/user-attachments/files/18098973/data_dictionary.pdf)

## ideas for follow-up work

- move code from Template DB to dcpy so this can easily be done for all products (same for XLSX generation)
- improve layout of content via css + html
- include the build's version in the data dictionary. might be best read from `recipe.lock.yml`


alexrichey (approved):
This is awesome. Very clean skeleton for making things look pretty. Have a few comments that to add about the github stuff, but this is great.

.github/workflows/template_build.yml
```diff
@@ -30,6 +30,7 @@ on:
 
 env:
   TOY_SECRET_GITHUB: ${{ secrets.TOY_SECRET }}
+  GHP_TOKEN: ${{ secrets.GHP_TOKEN }}
```
alexrichey, line 33: So we actually already check out the product-metadata repo in the steps below. I'm wondering if it makes sense to not invoke the Github repo clone python function. In general, this type of thing feels like environment setup, especially in the case that you're developing locally. E.g. if you want to build template DB with modified product-metadata on your machine. So I'd prefer we rely on cloning in the GHA, and not in the export step. Not a blocker though. (can discuss later)

products/template/build_scripts/export.py
```diff
@@ -24,23 +24,25 @@
...
+    org_metadata = OrgMetadata.from_path(org_metadata_path)
+
+    metadata = org_metadata.product("template_db").dataset("template_db")
+    metadata.write_to_yaml(dataset_metadata_path)
 
-def generate_metadata():
-    dataset_metadata_yml = product_metadata.download(
-        "template_db", PRODUCT_PATH / "data_dictionary.yaml", dataset="template_db"
-    )
     html_path = generate_metadata_assets.generate_html_from_yaml(
```
alexrichey, line 37: Seems like it'd be a lot easier to just pass the OrgMetadata to whatever method generates the assets, yeah? In general, I've been defaulting to only requiring a path for typer commands.

dcpy/connectors/github.py
```diff
@@ -36,6 +39,19 @@ def is_running(self) -> bool:
         return self.status in ["queued", "in_progress"]
 
 
+def download_repo(
```
alexrichey, line 42: Should we just call this `clone`, to mirror GH terminology? Also `f"https://github.com/NYCPlanning/` seems ripe for moving to dcpy.config.

.github/workflows/template_build.yml
damonmcc: yea this was tough to place. I went with python approach since having the org's metadata feels like a requirement of using these new parts of `dcpy`, so it felt weird to not have the approach to getting org metadata in `dcpy`

definitely feels like environment setup, just gotta figure out (later) what `dcpy` shouldn't be able to do

alexrichey: Definitely agree on the awkwardness of it. We can brainstorm some ways to make it more standard, e.g. retrieving it could be a part of dcpy setup/install. The reason I still favor just cloning it in the GHA (or in this case, using what's already cloned in this GHA job) retrieving the metadata repo feels more akin to injecting passwords into the environment, or installing dependencies, basically things that should be done at the beginning of the job, rather than tucked in a python script near the end. (cloning in GHA also mirrors our local dev setup.)

alexrichey:
> move code from Template DB to dcpy so this can easily be done for all products (same for XLSX generation)

For this TODO, we'll want to brainstorm, but I think we'll just want to invoke `dcpy.lifecycle.package.assemble` on everything that's specified in the OrgMetadata [artifacts](https://github.com/NYCPlanning/product-metadata/blob/main/packaging/artifacts.yml). Proto code to do that is [here](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/lifecycle/package/assemble.py#L216). Just need to add the hook in the HTML/PDF generator here.

dcpy/lifecycle/package/generate_metadata_assets.py
```diff
@@ -48,10 +51,9 @@ def _style_html_document(html: str, stylesheet_path: Path) -> str:
 def generate_pdf_from_html(
     html_path: Path,
     output_path: Path,
-    stylesheet_path: Path,
+    stylesheet_path: Path = DEFAULT_PDF_STYLESHEET_PATH,
```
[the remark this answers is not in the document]
damonmcc: (talked IRL) trying to modify an imported package's constants and expectingthat to affect the package does seem unreasonable

```python
from pandas import A_CONSTANT, ...

A_CONSTANT = 5
```

people shouldn't be modifying that constant. if users want an imported function to use a non-default value, they should use the function's args

Issue #1290, opened 2024-11-27
Ingest Migration - FacDB `script` sources


fvankrieken:
A satisfying group to migrate, because we get to delete the scripts, meaning higher test coverage of dcpy!! Reduce that denominator

- [x] dcas_ipis (holdover in this issue from original scope, which was just grab bag of script sources)
- [x] bpl_libraries
- [x] dcp_facilities_with_unmapped
- [x] dcp_colp (not really script source, but there is one that was temporarily used for formatting issues)
- [x] dcp_sfpsd
- [x] hra_centers (remove)
- [x] nypl_libraries
- [x] nysed_nonpublicenrollment
- [x] usdot_airports

fvankrieken:
dcas_ipis

Seems like we have one positive difference - these fields were always str, but now they're being properly cast to str without decimal place

```
________________________________________________________________________________
Tables
    Left: dcas_ipis_library
    Right: dcas_ipis_ingest
________________________________________________________________________________
Row count
    Left: 17229
    Right: 17229
________________________________________________________________________________
Column comparison
    Both
        agency
        alternate_address
        bbl
        block
        boro
        borough
        cd
        data_library_version
        house_number
        lot
        ogc_fid
        owned/leased
        owner
        parcel_name
        primary_use_code
        primary_use_text
        street_name
        u_a_use_code
        u_a_use_text
        u_f_use_code
        u_f_use_text
        unnamed:_22
        use_status
        x_coord
        y_coord
    Left only: None
    Right only: None
    Type differences
        Borough
            Left: character varying
            Right: text
        Y coord
            Left: character varying
            Right: text
        House number
            Left: character varying
            Right: text
        Parcel name
            Left: character varying
            Right: text
        X coord
            Left: character varying
            Right: text
        Street name
            Left: character varying
            Right: text
        Lot
            Left: character varying
            Right: text
        Unnamed: 22
            Left: character varying
            Right: double precision
        Owned/leased
            Left: character varying
            Right: text
        U f use code
            Left: character varying
            Right: text
        Primary use text
            Left: character varying
            Right: text
        Cd
            Left: character varying
            Right: text
        Block
            Left: character varying
            Right: text
        U a use code
            Left: character varying
            Right: text
        Boro
            Left: character varying
            Right: text
        Agency
            Left: character varying
            Right: text
        Owner
            Left: character varying
            Right: text
        U f use text
            Left: character varying
            Right: text
        Alternate address
            Left: character varying
[...]

PR #1299, opened 2024-12-03
maintenance for colp 2024-12

fvankrieken:
Takes one ingest dataset from #1290 - `dcas_ipis`. See report there.

Passing build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12144204955)


dcpy/data/compare.py
```diff
@@ -210,7 +210,7 @@ def compare_sql_rows(
     right_columns = client.get_table_columns(right)
 
     columns = set(left_columns) & set(right_columns) - set(ignore_columns or [])
-    query_columns = ",".join(list(columns))
+    query_columns = ",".join(f'"{c}"' for c in columns)
```
fvankrieken, line 213: This was giving me errors because one of the columns for this ends up (both by library and ingest) with a colon in the name

damonmcc (approved):

PR #1301, opened 2024-12-04
use recipe for kpdb

fvankrieken:
Nightly QA failed due to source datasets being moved over to library, but KPDB still using bash utils (that look for pgdumps only) to import recipe datasets. I've moved all recipe interaction in KPDB to a recipe file and updated ghas appropriately. There's still some dataloading logic that happens outside of recipe's realm, so not refactoring that for now.

[Passing run](https://github.com/NYCPlanning/data-engineering/actions/runs/12163458887)


damonmcc (approved):
yay!! gonna squash the commits?

PR #1314, opened 2024-12-11
Add GHA for running ingest utils

sf-dcp:
Extract data with ingest & library locally, put the datasets into database, and run a compare report util.

My local machine can't handle running migration utils for some big datasets like OTI building footprints. Note, I can't really test this GHA until we merge a first draft to main. If there are issues with the template, I will follow up with another PR.

### DIFF
--- a/.github/workflows/ingest_library_compare.yml
+++ b/.github/workflows/ingest_library_compare.yml
@@ -1,0 +1,44 @@
+name: Ingest 'Run and Compare' Tool
+run-name: "Ingest 'Run and Compare' Tool: ${{ inputs.dataset }}"
+
+on:
+  workflow_dispatch:
+    inputs:
+      dataset:
+        description: "Name of the dataset (required)"
+        required: true
+        default: dcp_mappluto
+      schema_name:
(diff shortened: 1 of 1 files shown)


alexrichey (approved):
Would be nice gift to future developers to add the other flags for for `run_and_compare` while you're here. Specifically:
    
    version: str | None = typer.Option(None, "--version", "-v"),
    key_columns: list[str] = typer.Option(None, "-k", "--key"),
    ignore_columns: list[str] = typer.Option(None, "-i", "--ignore"),
    columns_only_comparison: bool = typer.Option(False, "--columns-only", "-c"),

sf-dcp:
Oh great suggestion @alexrichey. I will add it to second PR (so that I can validate the template syntax)

alexrichey:
> Oh great suggestion @alexrichey. I will add it to second PR (so that I can validate the template syntax)

Ah, yeah, gotcha. FYI, I think this is one of the circumstances where we're all in agreement that you should feel free to just commit a very minimal starter action without review, just to get it registered on GHA and start getting it working.

PR #1315, opened 2024-12-11
Add GHA for `ingest` "run_and_compare" CLI

sf-dcp:
### What
This is a GHA ro run the `run_and_compare` ingest migration command which: pulls data from source using library & ingest, does some pre-processing steps for the "ingest" version, uploads both versions to database, and runs a comparison between the two. 

### Why to create this GHA
While migration of recipe templates is a one-time adventure and we could do all of this locally, I had an issue with a particularly big dataset. My local machine just can't handle the size and crashes. Thus the reason for creating this GHA. 

### Things to keep in mind
It requires presence of both recipe templates (in ingest & library). Most likely, this action will be run from a feature branch.

This is an example of a job:

image

#### Successful GHA run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12284126901/job/34279185084#step:6:723). This is the dataset I couldn't run locally by the way.

### DIFF
--- a/.github/workflows/ingest_library_compare.yml
+++ b/.github/workflows/ingest_library_compare.yml
@@ -11,6 +11,20 @@ on:
       schema_name:
         description: "Schema name where to store data tables (required)"
         required: true
+      version:
+        description: "The version of the dataset (i.e. 22v2, 21C) if needed (optional)"
+        required: false
+      key_columns:
+        description: "Key columns for comparison. Ex: 'column1,column2' (optional)"
+        required: false
+      ignore_columns:
+        description: "Columns to ignore during comparison. Ex: 'column1,column2' (optional)"
+        required: false
+      columns_only_comparison:
+        description: "Compare only column names (optional)"
+        type: boolean
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ingest_library_compare.yml
```diff
@@ -11,19 +11,33 @@ on:
       schema_name:
```
alexrichey, line 11: oh, sorry, forgot to ask about this, but one line up: `default: dcp_mappluto` do we want that?

alexrichey (approved):
One question, but looks good. Is there a link to a successful run? Curious what the output would look like for a diff like this.

sf-dcp:
> One question, but looks good. Is there a link to a successful run? Curious what the output would look like for a diff like this.

Yes! Please see the link in my last sentence of the issue description.

.github/workflows/ingest_library_compare.yml
sf-dcp: Eh, not really. Will remove

PR #1316, opened 2024-12-12
GHA template for `ingest` compare util: fix a bug with env vars

sf-dcp:
Didn't catch a bug with vars in my previous [PR]( #1315 ) (the GHA would still run successfully but with wrong vars). 

Successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12300148572/job/34327723403) (the job fails because 2 datasets don't match by data types which is okay). 

* Before fixing the bug: 
    image

* After fixing the bug (as expected)
    image

### DIFF
--- a/.github/workflows/ingest_library_compare.yml
+++ b/.github/workflows/ingest_library_compare.yml
@@ -34,7 +34,7 @@ jobs:
       run:
         shell: bash
     env:
-      BUILD_NAME: ${{ github.schema_name }}
+      BUILD_NAME: ${{ inputs.schema_name }}
       RECIPES_BUCKET: edm-recipes
       PUBLISHING_BUCKET: edm-publishing
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ingest_library_compare.yml
```diff
@@ -59,6 +59,6 @@ jobs:
           version: ${{ github.event.inputs.version && format('--version {0}', github.event.inputs.version) || '' }}
           key_columns: ${{ github.event.inputs.key_columns && format('--key {0}', github.event.inputs.key_columns) || '' }}
           ignore_columns: ${{ github.event.inputs.ignore_columns && format('--ignore {0}', github.event.inputs.ignore_columns) || '' }}
-          columns_only_comparison: ${{ github.event.inputs.columns_only_comparison && '--columns-only' || '' }}
+          columns_only_comparison: ${{ github.event.inputs.columns_only_comparison == 'true' && '--columns-only' || '' }}
```
fvankrieken, line 62: classic GHA fun right here

fvankrieken (approved):

PR #1317, opened 2024-12-12
Add Processing Steps Logging

alexrichey:
Adds Summaries for Processing Steps in Ingest Configs. Sample for COLP:
image

### Testing

I ran (locally) about five ingest datasets, including COLP, and created an ingest template for PLUTO to test performance for a dataset with more records. I specifically added the `strip_columns` step, as it's probably the most performance intensive type to summarize (we just do a before-after column comparison) 

Before Changes (~8 seconds)
image

After Changes (~10 seconds)
image

So on a dataset with about a million rows, the summary functionality ___may__ have added two second, but this could just be noise.

### Known Issues
- when we coerce to strings, the resulting type is just `object`. I left a note about potentially using `.astype("string")` which coerces to dTypeString. When I tested it, it worked fine, however it breaks a few test cases around handling of `None` values. 

### Callouts / Questions
- Instead of having a section for `processing_steps_summaries`, do we instead want to insert the summaries into the their entry under `processing_steps`? Just to have everything in one place.

### DIFF
--- a/dcpy/lifecycle/ingest/transform.py
+++ b/dcpy/lifecycle/ingest/transform.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Callable, Literal
 
 from dcpy.models import file
+from dcpy.models.base import SortedSerializedBase
 from dcpy.models.lifecycle.ingest import ProcessingStep, Column
 from dcpy.utils import data, introspect
 from dcpy.utils.geospatial import transform, parquet as geoparquet
@@ -12,6 +13,36 @@ from dcpy.utils.logging import logger
 from dcpy.connectors.edm import recipes
 
 OUTPUT_GEOM_COLUMN = "geom"
+
+
+class ProcessingSummary(SortedSerializedBase):
+    """Summary of the changes from a data processing function."""
+
+    description: str | None = None
+    row_modifications: dict
+    column_modifications: dict
+
+
+class ProcessingResult(SortedSerializedBase, arbitrary_types_allowed=True):
+    df: pd.DataFrame
+    summary: ProcessingSummary
+
+
+def make_generic_change_stats(
+    before: pd.DataFrame, after: pd.DataFrame, *, description: str | None
+) -> ProcessingSummary:
+    """Generate a ProcessingSummary by comparing two dataframes before and after processing."""
+    initial_columns = set(before.columns)
+    final_columns = set(after.columns)
+
+    return ProcessingSummary(
+        description=description,
+        row_modifications={"count_before": len(before), "count_after": len(after)},
+        column_modifications={
+            "added": list(final_columns - initial_columns),
+            "removed": list(initial_columns - final_columns),
+        },
+    )
 
 
 def to_parquet(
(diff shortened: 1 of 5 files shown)


alexrichey:
@fvankrieken Officially throwing in the towel here. My WIP PR is [here](https://github.com/NYCPlanning/data-engineering/pull/1317)

dcpy/lifecycle/ingest/transform.py
```diff
@@ -68,49 +95,111 @@ class ProcessingFunctions:
...
 
     def rename_columns(
-        self, df: pd.DataFrame, map: dict[str, str], drop_others=False
-    ) -> pd.DataFrame:
+        self, df: pd.DataFrame, drop_others=False, **kwargs
+    ) -> ProcessingResult:
+        assert "map" in kwargs, "map must be supplied to rename_columns"
+        col_map: dict[str, str] = kwargs[
+            "map"
+        ]  # doing this to avoid shadowing the builtin `map` fn
```
fvankrieken, line 182: oops

fvankrieken: If you just want to rename this arg (and update any templates that use it) I am in full favor

dcpy/test/lifecycle/ingest/shared.py
```diff
@@ -93,5 +93,5 @@ class Sources:
     (Sources.api, f"{TEST_DATASET_NAME}.json"),
     (Sources.socrata, f"{TEST_DATASET_NAME}.csv"),
     (Sources.s3, "test.txt"),
-    (Sources.esri, f"{TEST_DATASET_NAME}.json"),
+    # (Sources.esri, f"{TEST_DATASET_NAME}.json"),
```
fvankrieken, line 96: ?

fvankrieken (approved):
Great first pass at this, this is very fun.

I do think longer term this is another point that maybe these should move to classes -> would be super simple then to just have a default implementation for computing the diff, and most of the implemented steps don't have to think about it then. But there's obviously some subtleties there too.

On a similar note I agree with the point of maybe getting this into one object -> maybe having `Template`s have `ProcessingStepDefinitions` to make sure that a user can't supply a result and then `Config`s having `ProcessingStepDefinition | ProcessingStep`s so that we can just drop them in once it's done. Again though, I think this isn't needed right now - this is a great first implementation so that we have an MVP of this, both great just to get some df comps but also the more "required" stuff like the custom field w/ version of append_prev.

fvankrieken:
Oh what's up with the commented arcgis feature server source in tests though?

alexrichey:
> Oh what's up with the commented arcgis feature server source in tests though?

@fvankrieken  🤦 Yeah, sorry for not calling that out. I had added an explainer in the commit [here](https://github.com/NYCPlanning/data-engineering/pull/1317/commits/3921e67d0c3b0894b9485f233b7b93a4b8d6dc80) but TLDR that test case is taking a while, because it's actually making a callout

fvankrieken:
> > Oh what's up with the commented arcgis feature server source in tests though?
> 
> @fvankrieken 🤦 Yeah, sorry for not calling that out. I had added an explainer in the commit [here](https://github.com/NYCPlanning/data-engineering/pull/1317/commits/3921e67d0c3b0894b9485f233b7b93a4b8d6dc80) but TLDR that test case is taking a while, because it's actually making a callout.
> 
> I'll add this to that PR we've got open around test case annoyances (e.g. files being left over after running a test)

Oh it's not using the mocked arcgis response? There should be one mocked for `library` tests. Happy to let that be punted

fvankrieken:
> > > Oh what's up with the commented arcgis feature server source in tests though?
> > 
> > 
> > @fvankrieken 🤦 Yeah, sorry for not calling that out. I had added an explainer in the commit [here](https://github.com/NYCPlanning/data-engineering/pull/1317/commits/3921e67d0c3b0894b9485f233b7b93a4b8d6dc80) but TLDR that test case is taking a while, because it's actually making a callout.
> > I'll add this to that PR we've got open around test case annoyances (e.g. files being left over after running a test)
> 
> Oh it's not using the mocked arcgis response? There should be one mocked for `library` tests. Happy to let that be punted

And that'll be moot with the connector PR to some extent. Well not quite moot -> these tests will probably remain, but connector-specific logic should really move out of ingest tests too

alexrichey:
@fvankrieken 
> Oh it's not using the mocked arcgis response? There should be one mocked for `library` tests. Happy to let that be punted

Yeah, it's actually using `requests.post` which isn't mocked. I started mocking that out, but realized it's probably the wrong direction. Better to separate the concerns (ie have tests for arcgis, probably of the integration variety) and just register a mocked connector, with our new approach.

PR #1319, opened 2024-12-12
rename a `lifecycle.package` module

damonmcc:
to align with `xlsx_writer`

failed tests are related to a docker image/python package issue @fvankrieken is working on


products/template/product-metadata
alexrichey, line 1: was this intentional?

damonmcc: oops!

damonmcc: dropped

PR #1330, opened 2024-12-16
migrate ArcGIS Online datasets to ingest

### LINKED ISSUE #1249: Ingest Migration - Arcgis Online Datasets
All datasets in the [arcgis feature server weekly job](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/ingest_arcgis_feature_server.yml) should be migrated.

Glancing at code.. I haven't actually defined this as a source type yet. So we need to
- [ ] define a source for arcgis online. It will need 
      - [ ] model in `dcpy/models/ingest.py`, as option for both `Template` and `Config`
      - [ ] case in helper in `configure.py` to get filename from `Source`
      - [ ] case in helper in `configure.py` to call arcgis feature server connector api to get version
      - [ ] case in helper in `extract.py` to call  arcgis feature server connector api to get data
      - [ ] tests for above 3

Then, we need to migrate datasets

- [ ] dcp_cscl_commonplace
- [ ] dcp_cscl_complex
- [ ] dcp_waterfront_access_map_wpaa
- [ ] dcp_waterfront_access_map_pow
- [ ] dcp_wrp_recognized_ecological_complexes
- [ ] dcp_wrp_special_natural_waterfront_areas
- [ ] nysdec_freshwater_wetlands_checkzones
- [ ] nysdec_freshwater_wetlands
- [ ] nysdec_lands
- [ ] nysdec_natural_heritage_communities
- [ ] nysdec_priority_lakes
- [ ] nysdec_priority_shorelines
- [ ] nysdec_priority_estuaries
- [ ] nysdec_priority_streams
- [ ] nysdec_tidal_wetlands
- [ ] nysparks_historicplaces
- [ ] nysparks_parks
- [ ] nysshpo_historic_buildings_points
- [ ] nysshpo_historic_buildings_polygons
- [ ] nysshpo_archaeological_buffer_areas
- [ ] usnps_parks

### LINKED ISSUE #1464: Add arcgis online as ingest source
Split off from #1249 which originally had both this feature plus migration of actual datasets

in progress pr from months ago #1330

This needs
- [x] model in `dcpy/models/ingest.py` for this `Source`, as option for both `Template` and `Config`
- [x] case in helper in `configure.py` to get filename from `Source`
- [x] case in helper in `configure.py` to call arcgis feature server connector api to get version
- [x] case in helper in `extract.py` to call  arcgis feature server connector api to get data
- [x] tests for above 3. There are some `library` tests that can be pillaged

damonmcc:
resolves #1464

## What
Expand `ingest` to pull data from arcgis servers. To do that, I added a new ingest model, downstream functions to version and pull data from arcgis, and tests.

## Notes:
* I went with `.json` extension, not `geojson`. This is to match our approach with parquet files in ingest: we store geoparquet files as `*.parquet`. Also, library used `*.json`


dcpy/test/lifecycle/ingest/resources/templates/nysparks_historicplaces.yml
sf-dcp, line 1: This test template is actually not used in tests. I added it for future possible cases. It's a valid template

dcpy/models/lifecycle/ingest.py
```diff
@@ -47,6 +49,15 @@ class ESRIFeatureServer(BaseModel, extra="forbid"):
     def feature_server(self) -> esri.FeatureServer:
         return esri.FeatureServer(server=self.server, name=self.dataset)
 
+    @property
+    def feature_server_layer(self) -> esri.FeatureServerLayer:
```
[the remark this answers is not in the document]
fvankrieken: Might be worth making this a `cached_property` since resolve_layer can actually make a call to arcgis to determine layer metadata

dcpy/models/lifecycle/ingest.py
```diff
@@ -35,6 +37,28 @@ class DEPublished(BaseModel, extra="forbid"):
     filename: str
 
 
+class ESRIFeatureServer(BaseModel, extra="forbid"):
+    type: Literal["esri"]
+    server: esri.Server
+    dataset: str
+    layer_name: str | None = None
+    layer_id: int | None = None
+    crs: str = "EPSG:4326"  # The default value here is geojson specification
```
fvankrieken, line 46: Should we allow crs to be specified at all? I'd sort of lean no as a gut response based on our earlier convo, interested in your thoughts

sf-dcp: Looks like arcgis takes in crs as one of the params. I think it's worth keeping it in case we decide to extend this connector to choose a data format at the source.

fvankrieken: What do you mean specifically by "arcgis" in this case? Like the api for a feature server?

(sorry I cannot remember the code I wrote or why I wrote it that way lol)

dcpy/models/lifecycle/ingest.py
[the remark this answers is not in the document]
sf-dcp: Is what you saying that `feature_server_layer` may be called at instance initiating and then cached property doesn't?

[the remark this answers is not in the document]
fvankrieken: No - what I'm saying is that the call `source.feature_server_layer` makes to `arcgis_feature_service.resolve_layer` will potentially query the server to determine some of the info, so if we want the returned value to be available at multiple points (when getting version and when pulling data), it would be nice if it didn't have to repeat the steps to resolve the layer but just returned the cached value from the first time it is called. This will happen automatically if you use `@cached_property` instead of `@property`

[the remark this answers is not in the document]
sf-dcp: Uh, really cool! Thanks for explaining

dcpy/models/lifecycle/ingest.py
sf-dcp: @fvankrieken , yep, the api. The api takes in a bunch of different params like crs, desired data format, and many more. 

[Here](https://github.com/NYCPlanning/data-engineering/blob/bf66ddee01b2a14b9fe0ff8ed8299db7673c3c24/dcpy/connectors/esri/arcgis_feature_service.py#L126) is wrapper we use for querying argis api.

fvankrieken:
> > This looks great - have you run it (end to end)?
> 
> Yep, I ran ingest locally for that test template:
> 
> image
> I deleted this test data from `edm-recipes` because I didn't migrate the template

Use your dev bucket next time!

dcpy/models/lifecycle/ingest.py
fvankrieken: Gotcha. As a reference here's the feature server api

https://developers.arcgis.com/rest/services-reference/enterprise/query-feature-service-layer/

fvankrieken (approved):

dcpy/models/lifecycle/ingest.py
sf-dcp: yep, I believe that's the same link we have in the doctstring of the wrapper

PR #1337, opened 2024-12-18
Make bytes->Socrata distribution a little more friendly

alexrichey:
These are the changes we were using last week to distribute datasets. This improves the CLI to distribute from Bytes, fixes a bug around data dictionary generation, and makes a few quick additions to the README. (Documentation in general needs to be re-evaluated, but this was mostly for the purposes of facilitating the mobbing session last week)

Commits are atomic, and the details are included in the commit description.

### DIFF
--- a/dcpy/lifecycle/scripts/package_and_distribute.py
+++ b/dcpy/lifecycle/scripts/package_and_distribute.py
@@ -3,26 +3,38 @@ import typer
 from typing import Unpack
 
 from dcpy.configuration import PRODUCT_METADATA_REPO_PATH
+
 from dcpy.models.product import metadata as product_metadata
 from dcpy.lifecycle.distribute import socrata as soc_dist
 from dcpy.lifecycle.package import assemble
 from dcpy.utils.logging import logger
 
 
-def from_bytes_to_tagged_socrata(
+def package_and_dist_from_bytes(
     org_metadata_path: Path,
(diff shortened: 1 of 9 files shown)


dcpy/lifecycle/scripts/package_and_distribute.py
```diff
@@ -93,11 +125,18 @@ def from_bytes_to_tagged_socrata_cli(
         help="Only push metadata (including attachments).",
     ),
 ):
-    results = from_bytes_to_tagged_socrata(
+    datasets_set: set[str] = set()
+    if datasets:  # TODO: can typer do this natively?
```
fvankrieken, line 129: So this can be done natively - see https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/lifecycle/scripts/validate_ingest.py#L235. Then can be called with `python3 -m ... -k boro -k block -k lot` which would be read by typer as `["boro", "block", "lot"]`

fvankrieken: I don't know that you need to do that here though

alexrichey: right... so, I think if that's what typer supports, it's probably best to just go with that. Will change

fvankrieken (approved):

PR #1340, opened 2024-12-20
improve data dictionary pdf

damonmcc:
related to #561 

only the last commit is related to pdf formatting. all the other commits are refactoring for quality-of-life and consistency

## Template DB pdf with new tags

Screenshot 2024-12-27 at 1 50 52 PM

### DIFF
--- a/dcpy/lifecycle/package/resources/document_templates/attributes.jinja
+++ b/dcpy/lifecycle/package/resources/document_templates/attributes.jinja
@@ -20,7 +20,9 @@
 
     <div class="attribute-section">
       <span class="attribute-name">Tags: </span>
-      <span class="attribute-value">{{ metadata.attributes.tags|join(', ') }}</span>
+      {% for tag in metadata.attributes.tags %}
+      <span class="attribute-value-tag">{{ tag }}</span>
+      {% endfor %}
     </div>
 
   </div>
--- a/dcpy/lifecycle/package/resources/document_templates/document.css
+++ b/dcpy/lifecycle/package/resources/document_templates/document.css
@@ -22,6 +22,12 @@ html {
 .attribute-name {
   font-weight: bold;
 }
+.attribute-value-tag {
+  background: #eeeeee;
+  border-radius: 8px;
+  padding: 5px;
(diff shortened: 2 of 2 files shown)


products/template/build_scripts/export.py
```diff
@@ -28,20 +28,15 @@
...
-    metadata = org_metadata.product("template_db").dataset("template_db")
-    metadata.write_to_yaml(dataset_metadata_path)
-
-    html_path = PRODUCT_PATH / "data_dictionary.html"
-    pdf_writer.generate_html_from_yaml(
-        dataset_metadata_path,
-        html_path,
-        pdf_writer.DEFAULT_DATA_DICTIONARY_TEMPLATE_PATH,
-        pdf_writer.DEFAULT_DATA_DICTIONARY_STYLESHEET_PATH,
+    yaml_writer.write_yaml(
```
fvankrieken, line 31: Nice, this is much cleaner. It would be nice to get this sort of logic into `lifecycle/package` (as in, outputs are configured within metadata, so we just need something like `metadata.generate_attachments`) but that certainly seems like a separate issue

damonmcc: definitely wanna move this stuff into `lifecycle` next

fvankrieken (approved):

products/template/build_scripts/export.py
alexrichey: yeah, def plus +1 to moving this to [packaging artifacts](https://github.com/NYCPlanning/product-metadata/blob/main/packaging/artifacts.yml). I'm foreseeing dynamic dispatch based on the `artifact.type`, which would make it extensible for the outside world.

alexrichey: Was initially hesitant about a separate module for just writing the dataset yml, but I think it really fits with that approach 👆. Though I do think it's probably less of a general `yaml_writer` and more of a `dataset_yaml_writer` or similar. We should probably align on a standard interface for these methods that are writing artifacts... maybe 
`.generate(dataset_md, out_path, **kwargs)`

PR #1344, opened 2024-12-23
Use socratapy api for column updates, informative logging for column mismatches

fvankrieken:
Alright - this ended up going in circles quite a bit, and in the end is barely making any code changes.

Mainly
1. just because I did this when I was exploring changing transformations and determining these things a tad more programatically, moved to using socratapy's api for managing column metadata (simple refactor)
2. a more helpful error message when mapping doesn't line up

Since the thought is that if we have this rare case where mapping doesn't line up (which we don't expect operationally but has happened during the current effort of getting automation going), we get a simple error message, resolve it via UI, and then the issue is fixed moving forwards.

Went through the exercise of making sure that solving this in the UI once fixes the issue moving forward. Here's a "simple" [revision](https://data.cityofnewyork.us/dataset/dcp_test/5n4u-q8gi/revisions/4) of a dataset where everything lines up as expected

image

Then, we have a revision where there are column changes (this was applied after this screenshot)
image

Once this is applied, we send that same file in a [new revision](https://data.cityofnewyork.us/City-Government/dcp_test/5n4u-q8gi/revisions/28), and mappings are applied correctly
image

And it does this without modifying the actual column metadata. 
image
In this case, it's a little weird. I've made the dataset column names not line up with the dataset metadata. So column mappings are sort of their own distinct little weird entity defined for the socrata dataset. But that can be fixed by pushing a new revision where we'll have to specify the column mappings manually again ([here](https://data.cityofnewyork.us/City-Government/dcp_test/5n4u-q8gi/revisions/27))

### DIFF
--- a/dcpy/connectors/socrata/publish.py
+++ b/dcpy/connectors/socrata/publish.py
@@ -236,27 +236,21 @@ class RevisionDataSource:
         "The field names in the uploaded data do not match our metadata"
     )
 
-    uploaded_columns: list[Any]
-    column_update_endpoint: str
     _raw_socrata_source: dict[str, Any] | None = None
 
     @classmethod
     def from_socrata_source(cls, socrata_source):
-        return RevisionDataSource(
-            _raw_socrata_source=socrata_source,
-            # AR Note: It's not clear to me why/when you would have multiple schemas/output
-            # schemas for any one datasource. In any case, I haven't yet encountered it,
-            # and for our use case (single dataset upload per product) I don't expect we will
-            uploaded_columns=socrata_source.attributes["schemas"][0]["output_schemas"][
-                0
-            ]["output_columns"],
-            column_update_endpoint=f"https://{SOCRATA_DOMAIN}"
-            + socrata_source.input_schemas[0].links["transform"],
-        )
+        return RevisionDataSource(_raw_socrata_source=socrata_source)
+
+    def get_latest_output_schema(self):
+        return self._raw_socrata_source.get_latest_input_schema().get_latest_output_schema()
 
     @property
     def column_names(self) -> list[str]:
-        return [c["field_name"] for c in self.uploaded_columns]
+        return [
+            c["field_name"]
+            for c in self.get_latest_output_schema().attributes["output_columns"]
+        ]
 
     def calculate_pushed_col_metadata(self, our_columns: list[md.DatasetColumn]):
         # TODO: using c.id or c.name?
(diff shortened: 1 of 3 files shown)


dcpy/connectors/socrata/publish.py
```diff
@@ -281,39 +275,35 @@ def calculate_pushed_col_metadata(self, our_columns: list[md.DatasetColumn]):
...
-            new_col["display_name"] = our_col.name
-            new_col["description"] = our_col.description
-            new_uploaded_cols.append(new_col)
+            output_schema.change_column_metadata(col, "display_name").to(our_col.name)
+            output_schema.change_column_metadata(col, "description").to(
+                our_col.description
+            )
+            if our_col.custom and our_col.custom.get("api_name"):
+                logger.info(f"Mapping column {our_col.id} to {col}")
+                output_schema.change_column_transform(col).to(f"`{our_col.id}`")
```
fvankrieken, line 300: this probably should be `our_col.name.lower()` rather than `our_col.id` - as far as I can tell, that's how socrata reads in the datasets

dcpy/connectors/socrata/publish.py
```diff
@@ -594,6 +593,7 @@ def push_dataset(
         for attachment_id in socrata_dest.attachment_ids
     ]
 
+    dataset.discard_open_revisions()
```
fvankrieken, line 596: I should remember to take this out 🙂

dcpy/connectors/socrata/publish.py
```diff
@@ -281,39 +275,43 @@ def calculate_pushed_col_metadata(self, our_columns: list[md.DatasetColumn]):
...
+            field_name = col["field_name"]
+            our_col = our_cols_by_field_name[field_name]
+            our_col_index = list(our_api_names).index(field_name)
 
-            new_col["position"] = our_col_index + 1
-            new_col["initial_output_column_id"] = new_col["id"]
+            output_schema.change_column_metadata(field_name, "position").to(
+                our_col_index + 1
+            )
+            #    .change_column_metadata(col, "initial_output_column_id").to(col.id)
```
fvankrieken, line 290: left this as a comment but seems like this is handled under the hood

fvankrieken:
Also @alexrichey - I did not specify any transformation, seems that OpenData does this by default for geom columns

image

fvankrieken:
Error right now looks like this

```
ERROR:dcpy:Error Updating Column Metadata! However, the Dataset File was uploadedand the revision can still be applied manually, here: https://data.cityofnewyork.us/d/5n4u-q8gi/revisions/26
ERROR:dcpy:Error:
ERROR:dcpy:    Socrata 'transformations' failed. See revision -> 'Review & Configure Data' -> 'Column Mapping'
ERROR:dcpy:    Failures by columns:
ERROR:dcpy:        {
ERROR:dcpy:            "boro_code": {
ERROR:dcpy:                "type": "transform_compilation_failure",
ERROR:dcpy:                "message": "No such column `borocode'"
ERROR:dcpy:            }
ERROR:dcpy:        }
```

Open to feedback on use of textwrap and how this looks obvs

dcpy/connectors/socrata/publish.py
```diff
@@ -653,9 +655,9 @@ def push_dataset(
             except Exception as e:
                 # Upating column Metadata is tricky, and there's still some work to be done
                 logger.error(
-                    f"""Error Updating Column Metadata! However,
-                    the Dataset File was uploaded and the revision can still be applied manually, here: {rev.page_url}
-                    Error: {e}"""
+                    "Error Updating Column Metadata! However, the Dataset File was uploaded"
+                    f"and the revision can still be applied manually, here: {rev.page_url}\n"
```
alexrichey, line 659: uber-nit: "... was uploadedand the revision..."

alexrichey (approved):
Had one question, and one nit. But otherwise, lgtm! (after the failing test case is fixed)

dcpy/connectors/socrata/publish.py
fvankrieken: Not uber!

dcpy/lifecycle/distribute/_cli.py
```diff
@@ -31,11 +31,10 @@ def _dist_from_local(
...
     md = m.Metadata.from_path(metadata_path or (package_path / "metadata.yml"))
-    result = distribute.to_dataset_destination(
+    distribute.to_dataset_destination(
         metadata=md,
         dataset_destination_id=dataset_destination_id,
         publish=publish,
         dataset_package_path=package_path,
         metadata_only=metadata_only,
     )
-    print(result)
```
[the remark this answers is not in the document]
fvankrieken: That happens here - it's pretty thorough in its logging. There's maybe a question as to whether it's handy to have the distribute cli log some sort of result in addition to trusting connectors to log useful info, but this was not helpful, and there's no similar logging when calling `bytes_to_socrata`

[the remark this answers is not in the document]
fvankrieken: Oops - https://github.com/NYCPlanning/data-engineering/pull/1344/files#diff-3df1e94315a27c8749c1ddf037b04bb9c376c1e286d00833d7fde0da00809996R665

fvankrieken:
And forgot testing obvs. Will fix that in addition to the uber nit

Issue #1348, opened 2024-12-27
CDBG 2025


damonmcc:
primary outputs:
- csvs with details about the CDBG eligibility of 2020 census block groups, tracts, and boroughs
- an excel file with three sheets to align with [May 2019 output](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/DCP%20DATA%20SHARING/OMB/Census%20Tract%20Database%20(May%202019).xlsx?d=w64f504b40e194689b7c622fe6e901982&csf=1&web=1&e=yo4eh5): Tract, Borough Block Group

source data:
- 2020 census tracts
- PLUTO 24v4
- 2020 HUD Low and Moderate Income Summary Data (LMISD)

source data we seem to be missing, but would only need for documentation:
- ACS data for total population in each census tract
- latest "Low and Moderate 4-Person Family Income Limit: $xx,xxx"

Intending to release this data on Byte in sync with Geosupport 25A on 2/24.

### details from GDE Manual

NYC receives Community Development Block Grants from the U.S. Department of Housing and Urban Development (HUD). These grants may be used for projects in residential areas where over half the population is living in low- and moderate-income households.

A census tract is considered residential if at least 50.0% of the total built floor area is residential. This is determined by summing ResArea and BldgArea for MapPLUTO lots in the census tract. If at least 90% of a lot's area is in the tract, the entire lot is assigned to that tract. Otherwise, the building area is assigned proportionately based on the percent of the lot's area in each census tract. 

The clipped version of MapPLUTO was used because otherwise waterfront properties on the East River in Queens and Brooklyn had a portion of their building area assigned to Manhattan. An additional check was added to verify that a lot's area was fully captured. This was a problem for waterfront lots that extended outside the tract boundaries. 

### new/active stuff

[new source data in sharepoint](https://nyco365-my.sharepoint.com/shared?id=%2Fsites%2FNYCPLANNING%2Fitd%2Fedm%2FShared%20Documents%2FPROJECTS%2FDATA%20ENGINEERING%2FDCP%20DATA%20SHARING%2FOMB&listurl=https%3A%2F%2Fnyco365%2Esharepoint%2Ecom%2Fsites%2FNYCPLANNING%2Fitd%2Fedm%2FShared%20Documents&sortField=FileLeafRef&isAscending=true&login_hint=dmccullough%40planning%2Enyc%2Egov)

[...]

PR #1355, opened 2024-12-31
add cdbg data product

damonmcc:
related to #1348


fvankrieken:
Just noting that some block groups have 0 population - for these, should percentages be 0 (as reported in the source data) or null?

damonmcc:
> Just noting that some block groups have 0 population - for these, should percentages be 0 (as reported in the source data) or null?

when I hit "divide by 0" issues, I went with 0 as the result for now since I didn't wanna have to handle nulls yet

this made the final eligibility queries simpler and allows a `not_null` test in product models (started with those just as sanity checks)

but I could see why we might wanna use nulls to be less misleading

PR #1359, opened 2025-01-02
use total population in CDBG

damonmcc:
related to #1348 

according to HUD's data dictionary, the column `LOWMOD_PCT` isn't what we expect because the denominator `LOWMODUNIV` is "Persons with the potential for being deemed Low-, Moderate- and Middle-income" rather than the total population

so we have to get total population from a different dataset

## changes
1. add `dcp_censusdata_blocks` to get the total population of each census block
2. change where census blocks are grouped to census block groups.
    - This helped make the new data usable because it doesn't have a column with GEOID values to parse and determine the census block group. So joining all census block data before grouping seem best.
3. use new total population data


fvankrieken (approved):
I think this all seems fine for the incremental change needed. I do think we should sit down with some quick white boarding to simplify the repo

fvankrieken:
Dbt tests currently failing though

damonmcc:
> Dbt tests currently failing though

a column in `dcp_censusdata_blocks` is named `pop1.1` and sqluff seems to think it "refers to table/view not found in the FROM clause or found in ancestor statement"

gonna try different single/double quote combos (locally first)

fvankrieken:
> > Dbt tests currently failing though
> 
> a column in `dcp_censusdata_blocks` is named `pop1.1` and sqluff seems to think it "refers to table/view not found in the FROM clause or found in ancestor statement"
> 
> gonna try different single/double quote combos (locally first)
> 
> edit: might have to ignore it for now, couldn't fix it with quotes

Yeah that seems like a sqlfluff bug

PR #1360, opened 2025-01-02
fix inconsistent s3 metadata field names

### LINKED ISSUE #1350: Scheduled Action Failure - Ingest - 📁 Open Data Routine Updates
[Failed action](https://github.com/NYCPlanning/data-engineering/actions/runs/12530935501)

fvankrieken:
Resolves #1350 

When attempting to update the config of an existing version of a recipe dataset, we get a weird "signature does not match" error. This occurs on calling `put_object` when the object already exists. Googling has been entirely unhelpful - this error seems to occur when people have bad credentials. ~This makes me think that this might be something to do with Digital Ocean - I can't imagine there are that many folks using boto3 with DO, which would be a possible reason for lack of other people having this issue~ Issue was with the metadata I was attempting to push. Due to a bug in our code, when pulling metadata and then pushing again (and "merging" with standard generated metadata), there ended up being a both a "date-created" field and "date_created" field, with differing values. Resolving this resolved the issue

This logic maybe needs a little reworking in general - not great that I had to change multiple literals.

## Actions (all using dev bucket)
These were run in order
- [Run on this branch](https://github.com/NYCPlanning/data-engineering/actions/runs/12586088257) (could have been main) to initially archive all opendata datasets
- [Failing update](https://github.com/NYCPlanning/data-engineering/actions/runs/12586286671/job/35079800704) (run from main)
- [Now succeeding update](https://github.com/NYCPlanning/data-engineering/actions/runs/12587016106)

### DIFF
--- a/dcpy/connectors/edm/publishing.py
+++ b/dcpy/connectors/edm/publishing.py
@@ -174,11 +174,11 @@ def get_source_data_versions(product_key: ProductKey) -> pd.DataFrame:
 def generate_metadata() -> dict[str, str]:
     """Generates "standard" s3 metadata for our files"""
     metadata = {
-        "date_created": datetime.now(pytz.timezone("America/New_York")).isoformat()
+        "date-created": datetime.now(pytz.timezone("America/New_York")).isoformat()
     }
     metadata["commit"] = git.commit_hash()
     if CI:
-        metadata["run_url"] = git.action_url()
+        metadata["run-url"] = git.action_url()
     return metadata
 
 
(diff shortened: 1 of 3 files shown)


fvankrieken:
The fact that this is failing is bizarre. Those jobs have no problem deleting the existing config, but then two lines later when pushing the new config fail. I wonder if something is weird with trying to keep the existing metadata - like there's something internal that I'm trying to set that DO/S3 doesn't like

alexrichey (approved):
good catch!

PR #1365, opened 2025-01-03
fix order of args when building 'child' image

fvankrieken:
See https://github.com/NYCPlanning/data-engineering/blob/main/admin/ops/docker_build_and_publish.sh#L14

This isn't super testable, but I think the code is quite evident that these args were just swapped

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -190,7 +190,7 @@ jobs:
           DOCKER_PASSWORD: "op://Data Engineering/Dockerhub/password"
       - name: publish images
         if: needs.determine_image_logic.outputs.child_action == 'Build'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -190,7 +190,7 @@ jobs:
           DOCKER_PASSWORD: "op://Data Engineering/Dockerhub/password"
       - name: publish images
         if: needs.determine_image_logic.outputs.child_action == 'Build'
-        run: ./admin/ops/docker_build_and_publish.sh "${{ matrix.image }}" "${{ needs.determine_image_logic.outputs.base_tag }}" "${{ needs.determine_image_logic.outputs.child_tag }}"
```
fvankrieken, line 193: Script is here

https://github.com/NYCPlanning/data-engineering/blob/main/admin/ops/docker_build_and_publish.sh#L14

the tag of the image being built should be before the tag of the base image

sf-dcp (approved):

PR #1373, opened 2025-01-06
fix CDBG population data types

damonmcc:
related to #1348 

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-cdbg-zeros)

## before

`int__tracts.low_mod_income_population_percentage` was mostly 0s because `low_mod_income_population and `potential_lowmod_population` are integers and division using integers doesn't return (the expected) decimal values

value counts from a recent build are either 0 or 100:
Screenshot 2025-01-06 at 12 08 25 PM

`int__block_groups.low_mod_income_population_percentage` doesn't have this issue because it isn't computed, it comes straight from `stg__low_mod_by_block_group`

## after

value counts from build on this branch make more sense:

Screenshot 2025-01-06 at 12 22 40 PM

### DIFF
--- a/products/cdbg/models/product/_product_models.yml
+++ b/products/cdbg/models/product/_product_models.yml
@@ -86,7 +86,7 @@ models:
         tests: [not_null]
 
       - name: low_mod_income_population
-        data_type: bigint
+        data_type: integer
         tests: [not_null]
 
       - name: low_mod_income_population_percentage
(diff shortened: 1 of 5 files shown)


products/cdbg/models/product/_product_models.yml
damonmcc, line 1: even though they have to be `decimal` during the build, still seemed good to export these as `integer`

fvankrieken: agreed

fvankrieken (approved):

PR #1375, opened 2025-01-06
Add Connector Registration and Dynamic Dispatch

alexrichey:
Implements dynamic (ie non-static / hardcoded) dispatcher for our connectors, at the lifecycle level. The idea is that if you define an destination type as, say, `ginger_ftp` as below
image
then the distribution code should delegate the push to a connector registered as `ginger_ftp`.

#### Impetus
This approach would enable others to register custom connectors. E.g. if another agency needed a `snowflake` connector, they could just register one. (example in the test case). We can implement something similar for ingest and package.

#### Changes
1. refactor `lifecycle.distribute` to be a generic distributor (ie there should be nothing specific to Socrata or BYTES). The distributor should delegate push/pull requests to whatever connector is registered by the dispatcher. 
2. Move anything socrata specific, or anything that references multiple stages, into `lifecycle.scripts`.

#### Other Notes
1. I tried to start pulling code into the __init__ files for lifecycle modules to implement a nice API. In general, I think it would be nice to expose an API per lifecycle module that looks something like [lifecycle stage].[method_name]. For example, `distribute.to_dataset_destination` feels really nice. However, there's some work to be done here, as some of the methods I added have a stutter, e.g. `package.validate_package(...)`. It would be nice to just call `package.validate(...)` BUT there's also a `package.validate` module so we'd have a clash. There are a few ways around this, e.g. renaming the file to `package/_validate.py`. Open to thoughts! 
4. I went a little heavy on `Unpack` for kwargs, but I'm not convinced it's actually all that helpful here. I might be nicer to just use a class. 
5. `dcpy/test/models/test_connectors.py` is mostly for my sake to understand what variance/covariance mean for typing in the context of protocols vs regular functions. I'm happy to just delete it before merging.

#### TODOS
1. We probably need to add extra generics to the `ConnectorDispatcher` for result types. 
2. `lifecycle.scripts.package_and_distribute` has a little more logic about distribution than it ought to. I started moving this into `lifecycle.distribute`, but it got a little messy. 

#### Minor Changes
- Sometimes we called it `distribute` and other times `distribution`. Standardize on `distribute`.
- Change the working file directory from .[stage] -> .lifecycle/[stage] for package.
- I'm trying to adhere to our general emerging pattern of only passing around metadata.yml paths in typer functions, and just passing around metadata objects in the rest of our code.

Successful Job showing what a failure looks like:
https://github.com/NYCPlanning/data-engineering/actions/runs/13018821848


sf-dcp:
I like this too. Is it possible to define required/optional args at a connector level rather than unpacking kwargs and receiving an error when pushing/pulling? Or would it conflict with the underlying ABC class?

alexrichey:
@fvankrieken @sf-dcp I've changed things to find a nice potential middle-ground where we can type-safety, by just binding/configuring connectors at the lifecycle level. The idea being that for `ingest`/`package`/`distribute` we have a pretty fixed set of args that we'd pass to these dispatchers, so it makes sense to establish interfaces/protocols at the lifecycle level. That leaves us free to keep using sensible arguments in the connectors themselves.

dcpy/lifecycle/distribute/connectors.py
fvankrieken, line 1: What's the thought in having this connectors within lifecycle/distribute?

alexrichey:
@damonmcc FYI, my big refactor of the `package_and_distribute` file is functional - it needs cleanup and some fixes to testing etc. But it's worked so far with manual testing. Here's an example output from the logs:
image

I also need to hook this up to GHA. But... as of now, there's not a single reference to `bytes` or `socrata` in `lifecycle.package`, and the only references to Socrata in `lifecycle.distribute` are in the README (which we have to rewrite) and in the registration of the socrata connector. So we're pretty thoroughly generalized 🎉

dcpy/models/product/metadata.py
```diff
@@ -87,10 +87,10 @@ def get_datasets_by_id(self) -> dict[str, DatasetMetadata]:
     def query_destinations(
         self,
         *,
-        datasets: set[str] | None = None,
+        datasets: frozenset[str] | None = None,
```
alexrichey, line 90: immutability helps the ol' typechecker.

alexrichey:
@damonmcc I think I'm pretty much done here. Feel free to go ahead and implement that FTP connector. Or we can just review this and merge.

damonmcc:
> @damonmcc I think I'm pretty much done here. Feel free to go ahead and implement that FTP connector. Or we can just review this and merge. Another real connector might give us a better sense of what we actually need though.

a lotta changes here so let's get it merged and I'll add FTP in a follow-up. I'll review and let you know if a walkthrough would be helpful

shall I rebase and skip the failing FTP test?

alexrichey:
@damonmcc cool, marked as ready for review. Happy to walk through in person, if anything is murky. It's definitely a shotgun surgery PR.

.github/workflows/distribute_socrata_from_bytes.yml
```diff
@@ -43,9 +52,17 @@ jobs:
       image: nycplanning/build-base:latest
     env:
       PUBLISHING_BUCKET: edm-publishing
+      RECIPES_BUCKET: edm-recipes
+      PRODUCT_METADATA_REPO_PATH: product-metadata
+      _TYPER_STANDARD_TRACEBACK: 1
```
damonmcc, line 57: [looks like](https://typer.tiangolo.com/tutorial/exceptions/#disable-pretty-exceptions) this disables the use of `rich` for error messages. are "pretty" tracebacks in github action logs hard to read when distributing?

damonmcc (approved):
lgtm! separate from this, would love to set our test coverage to ignore CLI code and then revisit coverage of stuff like `package_and_distribute()`

PR #1383, opened 2025-01-08
rework boroughs, tracts, block group output tables

fvankrieken:
build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12677940091)


fvankrieken:
Still need to rework tests. Do we like this? Would we rather have one "product" table that's a little more aligned with our naming of things, and then another table that ONLY renames to desired export column names?

damonmcc:
I like the idea of a separate (subfolder?) product model for exporting column names that are out of our control. so that our primary product model can have logic in it and be tested, while any downstream model is only renaming columns for a specific use

but this definitely works for now if you wanna use this

fvankrieken:
But yeah maybe something like the top level files in "product" are very "DE-friendly" so-to-speak - bool columns that are just bools, snake case column names, etc. Then we have `product/export` or something that has the actual exports with any required odd formatting

products/cdbg/models/product/cdbg_zap_eligibility.sql
```diff
@@ -1,6 +1,3 @@
--- Ignore "Line is too long" sqlfluff rule
--- noqa: disable=LT05
```
fvankrieken, line 2: Sorry, keeping my fix on the individual lines lol

fvankrieken:
All expected outputs

image

damonmcc (approved):

PR #1390, opened 2025-01-13
FacDB - move from csv to parquet

fvankrieken:
A few notes here.

Commits 1 and 2 were made with the purpose of rerunning 24v2 but using parquet when possible -> not library to ingest, just pg_dump/csv to parquet within library. See run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12772102619). It fails, but data loading (and as part of that importantly, preprocessing) succeeds.

Commit 5 then reverts these commits (to get the code change needed that commit 2 reverted), so we can just build with the latest data. See build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/12772503176)

All ready for you to start troubleshooting :)


products/facilities/recipe.yml
```diff
@@ -3,98 +3,271 @@ version_strategy: bump_latest_release
...
+      function: dispatch
+  - name: nysparks_historicplaces
+    preprocessor:
+      module: facdb.pipelines
+      function: dispatch
+  - name: nysparks_parks
+    preprocessor:
+      module: facdb.pipelines
+      function: dispatch
+  - name: nysoasas_programs
```
damonmcc, line 273: does this one also need the preprocessor specified? it was last in the list before so I think the default was being applied to it

fvankrieken: Oops that would be an oversight, thanks

fvankrieken:
Also whether you or I merge, please squash to get rid of the reverting and un-reverting

Issue #1400, opened 2025-01-22
CDBG output formats


damonmcc:
there are two sets of outputs (destinations?) for CDBG:
- Geosupport requires a csv
  - one column (BOROCT) and every row is an eligible tract
  - to be shared with GR team via email
- [Bytes](https://www.nyc.gov/site/planning/data-maps/open-data/dwn-cdbg.page) requires csv, FGDB
  - (schemas TBD)
  - to be shared with GIS team/Webteam via email
- [DCP website](https://www.nyc.gov/site/planning/data-maps/community-development-block-grant.page) requires excel
  - to be shared with GIS team/Webteam via email

currently, the build exports csvs for the DCP website excel to be manually put together. generating an excel file during the build would be nice but isn't necessary for this release

PR #1402, opened 2025-01-23
add more outputs to CDBG

damonmcc:
related to #1400 

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-cdbg-gru)


products/cdbg/models/product/bytes/cdbg_tracts_bytes_csv.sql
```diff
@@ -0,0 +1,15 @@
+SELECT
+    ROW_NUMBER() OVER () AS "OBJECTID",
```
fvankrieken, line 2: Let's add this row number to `cdbg_tracts` just so we only generate it once and guarantee its consistency across outputs

products/cdbg/models/product/_product_models.yml
```diff
@@ -199,8 +199,45 @@ models:
...
+    tests: [ unique, not_null ]
+  - name: '"BoroCT2020"'
+  - name: '"NTACode"'
+  - name: '"NTAName"'
+  - name: '"BldgArea"'
+  - name: '"ResArea"'
+  - name: '"Res_pct"'
+  - name: '"TotalPop"'
+  - name: '"LowMod_Population"'
+  - name: '"LoMod_pct"'
```
fvankrieken, line 235: Is this consistent with bytes? Just with "LowMod_Population" and "LoMod_pct". Even if it is consistent with bytes, I think we change it so that these two are consistent with each other

products/cdbg/models/product/bytes/cdbg_tracts_bytes_fgdb.sql
```diff
@@ -0,0 +1,14 @@
+SELECT
+    ROW_NUMBER() OVER () AS "OBJECTID",
+    bct2020 AS "BoroCT2020",
+    NULL AS "NTACode",
+    NULL AS "NTAName",
```
damonmcc, line 5: @fvankrieken these columns are present in the fGDB on bytes

I forgot i added these but didn't check if any source data has NTA data. can you take a look and see if it's an easy change?

fvankrieken: `dcp_ct2020_wi` has these fields

fvankrieken: Just rejoining at the end on geoid seems easier than passing them through the transformations

products/cdbg/models/staging/stg__census_tracts.sql
```diff
@@ -2,7 +2,8 @@ SELECT
     geoid,
     borocode::integer AS borough_code,
     boroname AS borough_name,
-    boroct2020 AS bct2020,
+    boroct2020 AS bct2020_source,
+    borocode::integer || ct2020 as bct2020,
```
fvankrieken, line 6: why the cast to int?

damonmcc: hmmm no good reason. dropped it

products/cdbg/models/product/_product_models.yml
damonmcc: you mean change them to align with `cdbg_tracts` or `cdbg_boroughs_excel `? both bytes models use "LowMod_Population"/"LoMod_pct"

fvankrieken: I mean change them all so that we have "LowMod" everywhere instead of both "LowMod" and "LoMod"

damonmcc:
gonna add `dc-cdbg` to the database cleanup action

products/cdbg/models/product/_product_models.yml
```diff
@@ -199,8 +203,45 @@ models:
...
+  - name: '"LowMod_pct"'
+  - name: '"Eligibility"'
+  - name: '"CT_text"'
+
+- name: cdbg_tracts_bytes_fgdb
+  description: Census tracts and their Community Development Block Grant (CDBG) eligibility details for use in the DCP Bytes ESRI Geodatabase
+  columns:
+  - name: '"OBJECTID"'
+    tests: [ unique, not_null ]
+  - name: '"BoroCT2020"'
```
fvankrieken, line 231: If we're testing anything for uniqueness, should probably be this? I'd still like to see (not in this PR) if we can just drop `OBJECTID` entirely

fvankrieken (approved):

products/cdbg/models/product/_product_models.yml
damonmcc: for sure. also wouldn't mind enforcing contracts on all product models

PR #1405, opened 2025-01-27
Fix CPDB nightly build

### LINKED ISSUE #1399: CPDB: clean bins
This issue is the result of investigation of CPDB nightly build [fail](https://github.com/NYCPlanning/data-engineering/actions/runs/12823037097/job/35756930089#step:9:71). 

**TLDR**: There are 2 problems that need to be fixed: change pre-processing of `doitt_buildingfootprints` and adding logic to clean  `cpdb/data/dcp_cpdb_agencyverified.csv` source dataset.

### Why the build is failing?
In [this](https://github.com/NYCPlanning/data-engineering/blob/62d1198a1066cc96b84d8dfc4c5d91425606861e/products/cpdb/sql/attributes_agencyverified_geoms.sql#L28) place of the build, the `bin` column of the `dcp_cpdb_agencyverified` table has invalid values that can't be casted to an integer. Example of invalid values from that column: 

Image

### Why have the build started failing now if there were no code changes?

CPDB started failing after we updated `doitt_buildingfootprints` recipe to use ingest. This dataset has remained mostly the same except the `bin` string representation: library version stored `bin` as decimal string (ex: `123.0`) and ingest - integer string (ex. `123`). 

Then, this dataset is cleaned with our pre-processing python script where we remove records with invalid bins here:

Image

This specific line of code doesn't act as expected. It removes ALL records if bin values are not integer strings which was the case with previous `doitt_buildingfootprints` dataset. The table was empty and no join was performed and the casting issue wasn't shown (the failing spot).

I submitted a GH issue for `pandas` here.

sf-dcp:
Closes #1399.

## What
Nightly qa build for CPDB was [failing](https://github.com/NYCPlanning/data-engineering/actions/runs/12823037097/job/35756930089#step:9:71), and this PR is meant to fix it. I recommend reviewing the GH issue linked above for more details. 

**Note**: the build started passing again but it's the same behavior as in builds prior fails where `doitt_buildingfootprints` is empty prior the join. This PR is still very much needed.

## Changes
1. Clean up input csv `dcp_cpdb_agencyverified`: remove `?` from values in the `bin` column
2. Move pre-processing of `doitt_buildingfootprints` dataset from `pandas` to SQL and keep original raw version (previously it was dropped after pre-processing making it hard to debug)  
    * During the pre-processing step, the dataset is filtered where the`bin` column only has valid numeric values. Very basic.
    * Why not to keep pre-processing in `pandas`? Because `pandas` changes the raw `bin` column to decimal (there are NULL values present) and then uploads the dataset back to database with the column as decimal. We do correct data types in `ingest` and this `pandas` behavior adds more complexity. 
3. Update the join query between dcp_cpdb_agencyverified` & `doitt_buildingfootprints` to filter out invalid `bin` values.

Successful build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/13016090532/job/36305543896).


products/cpdb/bash/01_preprocessing.sh
```diff
@@ -7,7 +7,11 @@ echo "fixing dot_bridges"
 run_sql_command "ALTER TABLE dot_projects_bridges RENAME COLUMN fmsid TO fms_id;"
 python3 python/dot_bridges.py
 
-echo "fixing doitt_buildingfootprints"
-run_sql_command "ALTER TABLE doitt_buildingfootprints RENAME TO doitt_buildingfootprints_source;"
-python3 python/doitt_buildingfootprints.py
-run_sql_command "DROP TABLE IF EXISTS doitt_buildingfootprints_source;"
+echo "fixing doitt_buildingfootprints_source and saving result into doitt_buildingfootprints"
+run_sql_command "
+    DROP TABLE IF EXISTS doitt_buildingfootprints;
```
fvankrieken, line 12: I think I might lean dropping this entirely, and just handling null bins as needed. But I think that's getting into more of a general refactor - if we ever dbtify cpdb something like that can happen then

products/cpdb/data/dcp_cpdb_agencyverified.csv
```diff
@@ -1121,16 +1121,16 @@ Canceled - remove from map,DCAS,,,,, Fixed Asset, BPL - MILL BASIN BR LED RETROF
 ,DOHMH,,,,, Fixed Asset, APICHA COMMUNITY HEALTH CENTER: OUTFITTING OF CLINICAL EXAM  ,Pass-through,850HLQNAPICH,unmapped,,,
 ,DOHMH,,,,, , CATHOLIC MANAGED LONG TERM CARE (D/B/A ARCHCARE SENIOR LIFE  ,Pass-through,850HL82ARCHC,unmapped,,,
 ,DOHMH,,,,, Fixed Asset, DOH CLINIC CODE CORRECTIONS                                  ,Pass-through,81901199601,unmapped,,,
-385 Throop Ave.,DOHMH,3017960001.0,3050268???,Brooklyn,, Fixed Asset, DOHMH - LED LIGHTING UPGRADE (Bedford)                           ,Yes - Multi-site,816ACEDOH502,unmapped,,,11221.0
```
fvankrieken, line 1124: Have you investigated these bbls/bins at all? Do we know if this is a valid bin?

fvankrieken: They do seem to all be 7 digits after removing question marks which at least means they're all theoretically valid bins now. I'm so fascinated by the question marks

sf-dcp: @fvankrieken, great question. Yes, I did investigate them in `doitt_buildingfootprints` and found them to be valid bins but without a question mark in the end.

fvankrieken: I think we might just want to check with AD if she has any idea why these were like this in the first place. 

We have other invalid rows like 
image

Which do have at least a valid BIN. Not that we should do anything about those right now, but maybe the question marks were included because they were unsure?? Just would be worth trying to make sure there's no reason/meaning. Don't see anything in the [cpdb wiki](https://github.com/NYCPlanning/db-cpdb/wiki)

fvankrieken:
At some point we need to have an effort to go through our whole codebase and make sure bbls and bins are strings everywhere. But today is not that day.

products/cpdb/bash/01_preprocessing.sh
sf-dcp: Yeah, we can drop this step. I don't think we need to handle the null beans

sf-dcp:
> At some point we need to have an effort to go through our whole codebase and make sure bbls and bins are strings everywhere. But today is not that day.

Yeah, this should probably happen after we fully implement data checks for source data to validate data type (rn column data type in a ingest recipe is ignored)

products/cpdb/data/dcp_cpdb_agencyverified.csv
sf-dcp: Would you like me to check with AD prior merging the PR?

products/cpdb/bash/01_preprocessing.sh
damonmcc: agree we can drop this. looks to me like `attributes_agencyverified_geoms.sql` handles null bins during the `UPDATE dcp_cpdb_agencyverified` part

and if this is dropped, than maybe there's no need to do `import_as: doitt_buildingfootprints_source` in the recipe?

sf-dcp: @damonmcc  yep, I will make the updates. Just waiting to meet with AD tmw to finalize this PR

sf-dcp:
Met with AD: decision is to filter out all non-numerical bin values like `123???` or `123 (additional bin 321)`. 

CPDB run with new code updates [here](https://github.com/NYCPlanning/data-engineering/actions/runs/13081252004/job/36505084778).

fvankrieken:
So there are two points I can find that refer to `doitt_buildingfootprints` without either joining on `bin` column or specifying `where bin is not null`
- https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bbl.sql#L43
- https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bin.sql#L48

Can you assess what impact that might have on these queries and any downstream logic?

fvankrieken:
I'm running repeat builds of 24adopt on main and this branch and would like to look at the final tables and differences between them before merging. I can't imagine anything that would prevent merging, but just seems that we should assess impact before getting this fix on main

sf-dcp:
> So there are two points I can find that refer to `doitt_buildingfootprints` without either joining on `bin` column or specifying `where bin is not null`
> 
> * https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bbl.sql#L43
> * https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bin.sql#L48
> 
> Can you assess what impact that might have on these queries and any downstream logic?

We haven't used `doitt_buildingfootprints` during CPDB at all since the table had been empty after its pre-processing step.

fvankrieken:
> > So there are two points I can find that refer to `doitt_buildingfootprints` without either joining on `bin` column or specifying `where bin is not null`
> > 
> > * https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bbl.sql#L43
> > * https://github.com/NYCPlanning/data-engineering/blob/main/products/cpdb/sql/attributes_maprojid_bin.sql#L48
> > 
> > Can you assess what impact that might have on these queries and any downstream logic?
> 
> We haven't used `doitt_buildingfootprints` during CPDB at all since the table had been empty after its pre-processing step. Can you clarify how to assess the impact? @fvankrieken

Sorry - we made a decision to not filter `where bin is not null`, I am referring to the impact of that decision

sf-dcp:
@fvankrieken, the most recent `doit_buildingfootprints` version has a total of 2 records with null bins (geometry is not null). So the impact is not dramatic. But it makes sense to add a clause 'where bin is not null' just in case, before we add actual data checks.

sf-dcp:
@fvankrieken, filtering for non-null geoms in `doitt_buildingfootprints` didn't solve the problem: the projects `846P-4OLMST3` & `846P-400ZOO8` still got a NULL value. No `datasource` is present. 

When you were comparing the tables between my older branch and main, was the datasource `doitt_buildingfootprints` present?

fvankrieken (approved):
The vanishing 2 dpr geoms is a funny one, but seems like an unrelated bug. Let's chat tomorrow about root of the issue, and I'll write up something about next steps after that.

PR #1409, opened 2025-01-29
GRU - `Historical Footprint BINs vs PAD` report enhancement

### LINKED ISSUE #1310: GRU - Historic Footprints BIN vs PAD Enhancement
For the upcoming 25A cycle due around mid-February, GRU is requesting a change to said check. Requires changes in this repo (for QA app) but primarily in the db-gru repo. 

The request:

> In the output reports, we discovered that there are BINs in the Historic Building Layer that are also present in the current Building Footprint Layer. We are asking to add a field to the output report denoting cases where BIN is being reused so we can filter for such cases. In other words, after running all Historic BINs through Geosuppport also check if the BIN is present in the latest Footprint file. If the BIN is in latest Footprint File indicate so it in the separate field.

Seems relatively straightforward. Offhand, I don't remember the format of the output - we might need a separate csv if the current output is not inclusive of all bins, which seems likely. However, this is a relatively new test and I don't remember the format. Will take just a little research and then some slight back and forth with GRU to confirm

sf-dcp:
Last PR related to #1310, the other PRs have been merged in the GRU repo. Closes #1310.

Adding new source to the report as shown below:

image

## TODO after merge

- [ ] re-deploy streamlit
- [ ] communicate update to GRU


apps/qa/src/pages/gru/constants.py
```diff
@@ -24,7 +24,7 @@
             "Historical Footprint BINs vs PAD",
             "historical-footprints-vs-pad",
             ["all_results_historicalfootprintbin_padbin"],
-            ["doitt_buildingfootprints_historical"],
+            ["doitt_buildingfootprints_historical", "doitt_buildingfootprints"],
```
fvankrieken, line 27: hmm. Not sure why past Finn hard-coded these if we could get them from the csvs? But this whole page this could use a decent sized refactor at some point

fvankrieken (approved):

PR #1411, opened 2025-01-31
drop checkbook from nightly qa builds

damonmcc:
related to #1338 

reasons to drop it from nightly QA:
- it isn't a finished data product that we'd like to ensure is always buildable
- it often fails due to an intermittent/flaky DB connection during it's `build.py` script
- it's currently failing because it expects csvs from edm-recipes but Ingest only produces parquet files now


sf-dcp (approved):
😮‍💨

PR #1421, opened 2025-02-04
Ingest Migration - FacDB OpenData datasets

### LINKED ISSUE #1420: Ingest Migration - FacDB socrata datasets
- [x] dca_operatingbusinesses
- [x] dcla_culturalinstitutions
- [x] dfta_contracts
- [x] doe_busroutegarages
- [x] dohmh_daycare
- [x] doi_evictions
- [x] dot_pedplazas
- [x] sca_enrollment_capacity
- [x] dsny_donatenycdirectory
- [x] dsny_electronicsdrop
- [x] dsny_fooddrop
- [x] dsny_leafdrop
- [x] dsny_specialwastedrop
- [x] dsny_textiledrop
- [x] hhc_hospitals
- [x] hra_jobcenters
- [x] hra_medicaid
- [x] hra_snapcenters
- [x] nycha_communitycenters
- [x] nycha_policeservice
- [x] nysdec_solidwaste
- [x] nysdoh_healthfacilities
- [x] nysdoh_nursinghomes
- [x] nysomh_mentalhealth
- [x] nysopwdd_providers
- [x] qpl_libraries
- [x] sbs_workforce1

fvankrieken:
closes #1420

All socrata datasets that go into FacDB!

When possible, I went geojson. Just makes things a lot simpler. I've also gone with the norm of coercing certain data types in the `dtype` field for csv source rather than a processing step - for these cases, that seems like where in the future we'd maybe try to apply logic when reading in fields that are "special" data types (i.e. "bbl" type instead of "text", which enforces structure). Doing this at read time of csv is nice because it preserves leading zeroes without need to lpad after the fact.

See issue for reports


sf-dcp (approved):
🎉

dcpy/utils/geospatial/transform.py
```diff
@@ -90,6 +90,7 @@ def df_to_gdf(df: pd.DataFrame, geometry: file.Geometry) -> gpd.GeoDataFrame:
             geometry=gpd.points_from_xy(df[x_column], df[y_column]),
             crs=geometry.crs,
         )
+        gdf.geometry = gdf.geometry.apply(lambda geom: None if geom.is_empty else geom)
```
fvankrieken, line 93: This was resulting in a lot of `POINT(NAN NAN)` before due to making a point out of two null values. Which apparently is a valid point! Figured it made more sense for `NULL, NULL` to get coerced to NULL rather than an empty point/polygon

PR #1422, opened 2025-02-04
fix CPDB projects outside of NYC

### LINKED ISSUE #1410: CPDB has projects outside of NYC
## problem

> Hello @NYCPlanning/data-engineering. We found some data issues while packaging this version of CPDB up for ArcGIS Online publication. In the points file, there are a cluster of 8 projects that appear in the mountains of Kyrgyzstan!
> 
> Flagging the maprojid values here for review:
> 
> 846P-1HUNTPT
> 846P-4FMCPML
> 846P-303ASRC
> 846P-2PEPWIN
> 846P-306ASEQ
> 846P-3REC54E
> 846P-3SRAWIN
> 846P-3TDCREC
> 
> cc @croswell81 @caseysmithpgh  

 _Originally posted by @jackrosacker in [#1135](https://github.com/NYCPlanning/data-engineering/issues/1135#issuecomment-2627650590)_

## cause

For relevant projects, DPR source data has longitudes values that are positive but should be negative.

All of these projects have a MULTIPOINT geometry type and have exactly one point with a positive longitude. For the single of these projects that only has one point, it has a positive longitude.

damonmcc:
resolves #1410

all builds (repeat builds of 24adopt) on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/repeat_build.yml?query=branch%3Adm-cpdb-coords). latest outputs are in `edm-publishing/db-cpdb/build/dm-cpdb-coords`

I checked the new points shapefile in QGIS and all points are now in NYC.

I imagine the SQL approach here could be simpler but I wanted to fit into the "update columns" pattern and fix this ASAP so GIS can redistribute this release.

### DIFF
--- a/products/cpdb/sql/attributes_badgeoms.sql
+++ b/products/cpdb/sql/attributes_badgeoms.sql
@@ -30,3 +30,54 @@ WHERE
         )
     )
     AND geom IS NOT NULL;
+
+-- Fix projects outside of NYC
+-- For now, only fixes point projects with inverted longtiude values
+-- All points in NYC are around latitude -73.9
+DROP TABLE IF EXISTS cpdb_dcpattributes_not_in_nyc_fixed;
+
+WITH point_projects_not_entirely_in_nyc AS (
+    SELECT
+        maprojid,
+        geom
+    FROM cpdb_dcpattributes
+    WHERE
+        ST_GEOMETRYTYPE(geom) = 'ST_MultiPoint'
+        AND ST_XMAX(geom) > 0
+),
+
+project_points AS (
+    SELECT
+        maprojid,
+        (ST_DUMP(geom)).geom AS single_point
+    FROM point_projects_not_entirely_in_nyc
+),
+
+fixed_points AS (
+    SELECT
+        maprojid,
+        CASE
+            WHEN ST_X(single_point) > 0
+                THEN ST_SETSRID(ST_MAKEPOINT(-ST_X(single_point), ST_Y(single_point)), 4326)
+            ELSE single_point
+        END AS single_point_fixed
+    FROM project_points
+),
+
+projects AS (
+    SELECT
+        maprojid,
+        ST_MULTI(ST_UNION(single_point_fixed)) AS geom
+    FROM fixed_points
+    GROUP BY maprojid
+)
+
+SELECT *
+INTO cpdb_dcpattributes_not_in_nyc_fixed
+FROM projects;
+
+UPDATE cpdb_dcpattributes
+SET geom = cpdb_dcpattributes_not_in_nyc_fixed.geom
+FROM cpdb_dcpattributes_not_in_nyc_fixed
+WHERE
+    cpdb_dcpattributes.maprojid = cpdb_dcpattributes_not_in_nyc_fixed.maprojid;
--- a/products/cpdb/bash/02_build.sh
+++ b/products/cpdb/bash/02_build.sh
@@ -156,4 +156,5 @@ run_sql_command "\COPY cpdb_badgeoms FROM './data/cpdb_geomsremove.csv' DELIMITE
 run_sql_file sql/attributes_badgeoms.sql	
 
 # create final table
+echo 'Creating projects tables'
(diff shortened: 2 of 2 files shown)


products/cpdb/sql/attributes_badgeoms.sql
```diff
@@ -30,3 +30,54 @@ WHERE
...
+projects AS (
+    SELECT
+        maprojid,
+        ST_MULTI(ST_UNION(single_point_fixed)) AS geom
+    FROM fixed_points
+    GROUP BY maprojid
+)
+
+SELECT *
+INTO dm_cpdb_coords.cpdb_dcpattributes_not_in_nyc_fixed
```
fvankrieken, line 76: hard-coded schema here and on line 37

fvankrieken:
I don't love this big blanket "fix" applied to all points when we're implementing it because of a specific limited data issue with a single source dataset, especially when hopefully this. Obviously there could be another factor or two in these runtimes, but just glancing, looks like this more than doubles build time

image

Maybe we apply this logic to the source dataset before the points are grouped? And still hope to drop that code if the source data issue is resolved I think.

fvankrieken:
> I don't love this big blanket "fix" applied to all points when we're implementing it because of a specific limited data issue with a single source dataset, especially when hopefully this. Obviously there could be another factor or two in these runtimes, but just glancing, looks like this more than doubles build time
> 
> image
> 
> Maybe we apply this logic to the source dataset before the points are grouped? And still hope to drop that code if the source data issue is resolved I think.

I do feel like that runtime has to be a weird unrelated issue though - I don't quite see how this update could take so long. An update on a good sized table with no index is expensive... but not THAT expensive. Going to trigger another build on your branch

products/cpdb/sql/attributes_badgeoms.sql
damonmcc: oops! and of course the build worked with this

damonmcc:
investigating the change in build time

damonmcc:
this is from the raw logs of a 29 minute build
```
2025-02-05T16:20:21.7845883Z 
2025-02-05T16:20:21.7846316Z ###########################
2025-02-05T16:20:21.7846808Z # final geometry clean-up #
2025-02-05T16:20:21.7847191Z ###########################
2025-02-05T16:20:21.7847430Z 
2025-02-05T16:20:21.7847609Z Cleaning geometries: lines to polygons and geoms to multi
2025-02-05T16:20:32.5034780Z Removing bad geometries
2025-02-05T16:20:34.5208084Z psql:sql/attributes_badgeoms.sql:37: NOTICE:  table "cpdb_dcpattributes_not_in_nyc_fixed" does not exist, skipping
2025-02-05T16:20:35.3899707Z psql:sql/projects.sql:1: NOTICE:  table "cpdb_projects_geom" does not exist, skipping
2025-02-05T16:20:39.5539243Z psql:sql/projects.sql:136: NOTICE:  table "cpdb_projects_without_budget_data" does not exist, skipping
2025-02-05T16:20:39.7821304Z "run_sql_file sql/projects.sql -v ccp_v=$ccp_v" command filed with exit code 0.
2025-02-05T16:20:39.7982721Z 
2025-02-05T16:20:39.7984149Z #########################################
2025-02-05T16:20:39.7984929Z # create admin bounds relational tables #
2025-02-05T16:20:39.7985434Z #########################################
2025-02-05T16:20:39.7985730Z 
```

the new logic would have run in between:

```
2025-02-05T16:20:32.5034780Z Removing bad geometries
2025-02-05T16:20:34.5208084Z psql:sql/attributes_badgeoms.sql:37: NOTICE:  table "cpdb_dcpattributes_not_in_nyc_fixed" does not exist, skipping
2025-02-05T16:20:35.3899707Z psql:sql/projects.sql:1: NOTICE:  table "cpdb_projects_geom" does not exist, skipping
```

so it seems like the changes here are not causing the longer build time. gonna keep looking for a cause in the logs and during a rebuild

fvankrieken (approved):
Can you also try to follow up with DPR about these points?

PR #1423, opened 2025-02-05
`dcpy`: Add `geoalchemy2` python package

sf-dcp:
The package `geoalchemy2` is needed for `to_postgis()` fn in `dcpy` when importing a geospatial dataset into database (it currently exists only [here](https://github.com/NYCPlanning/data-engineering/blob/1769be219e7302bceb6a7752dc5883ee59e1d374/admin/run_environment/requirements.in#L14)). The missing package came up when using `dcpy` in gru repo.

I compiled python reqs to check that CI tests pass... Some weirdness with upgrading `sqlfluff` where tests are failing for not having unique aliases. Other than that the tests are ok, but I rolled compiled reqs back.

### DIFF
--- a/admin/run_environment/constraints.txt
+++ b/admin/run_environment/constraints.txt
@@ -25,26 +25,26 @@ attrs==25.1.0
     #   jsonschema
     #   rasterio
     #   referencing
-babel==2.17.0
+babel==2.16.0
     # via agate
-beautifulsoup4==4.13.3
+beautifulsoup4==4.12.3
     # via
     #   -r /__w/data-engineering/data-engineering/admin/ops/../run_environment/requirements.in
     #   gdown
 blinker==1.9.0
     # via streamlit
-boto3==1.36.14
+boto3==1.36.11
     # via
     #   -r /__w/data-engineering/data-engineering/admin/ops/../run_environment/requirements.in
     #   moto
-boto3-stubs==1.36.14
+boto3-stubs==1.36.11
     # via -r /__w/data-engineering/data-engineering/admin/ops/../run_environment/requirements.in
-botocore==1.36.14
+botocore==1.36.11
     # via
     #   boto3
     #   moto
     #   s3transfer
-botocore-stubs==1.36.14
+botocore-stubs==1.36.11
     # via boto3-stubs
 bqplot==0.12.44
     # via leafmap
@@ -148,7 +148,7 @@ diff-cover==9.2.2
     # via sqlfluff
 distlib==0.3.9
     # via virtualenv
-duckdb==1.2.0
+duckdb==1.1.3
     # via leafmap
 et-xmlfile==2.0.0
     # via openpyxl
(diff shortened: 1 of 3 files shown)


fvankrieken:
So this PR declares it geoalchemy2 as a dependency for dcpy. However, this doesn't actually install it anywhere - this is more for a case where you'd install dcpy without having the dependencies already installed. However, our images do already have the dependencies installed - see the various requirements files

sf-dcp:
> So this PR declares it geoalchemy2 as a dependency for dcpy. However, this doesn't actually install it anywhere - this is more for a case where you'd install dcpy without having the dependencies already installed. However, our images do already have the dependencies installed - see the various requirements files

The current gru image, `docker-geosupport`, doesn't have this package. I installed `dcpy` successfully in that image but ran into missing package problem when using dcpy import function.

fvankrieken:
> > So this PR declares it geoalchemy2 as a dependency for dcpy. However, this doesn't actually install it anywhere - this is more for a case where you'd install dcpy without having the dependencies already installed. However, our images do already have the dependencies installed - see the various requirements files
> 
> The current gru image, `docker-geosupport`, doesn't have this package. I installed `dcpy` successfully in that image but ran into missing package problem when using dcpy import function.
> 
> @fvankrieken

Gotcha. This would solve that issue, but then you have to install all dependencies at runtime - you can use `build-geosupport` docker image which already has all the requirements of dcpy installed

sf-dcp:
> > > So this PR declares it geoalchemy2 as a dependency for dcpy. However, this doesn't actually install it anywhere - this is more for a case where you'd install dcpy without having the dependencies already installed. However, our images do already have the dependencies installed - see the various requirements files
> > 
> > 
> > The current gru image, `docker-geosupport`, doesn't have this package. I installed `dcpy` successfully in that image but ran into missing package problem when using dcpy import function.
> > @fvankrieken
> 
> Gotcha. This would solve that issue, but then you have to install all dependencies at runtime - you can use `build-geosupport` docker image which already has all the requirements of dcpy installed

Yes, I was planning to use that image but decided not to go with it because it installs to many unneeded things like `gdal`

fvankrieken (approved):

fvankrieken:
I still think that's preferable to installing all python dependencies on the fly in terms of speed. Also, using our images with dependencies already installed guarantees that dcpy is running in an environment it's been tested in, as opposed to installing the latest dependencies at runtime

fvankrieken:
regardless, this change should happen!

sf-dcp:
> I still think that's preferable to installing all python dependencies on the fly in terms of speed. Also, using our images with dependencies already installed guarantees that dcpy is running in an environment it's been tested in, as opposed to installing the latest dependencies at runtime

Good point. How about installing `dcpy` during build of `docker-geosupport` image?

fvankrieken:
> > I still think that's preferable to installing all python dependencies on the fly in terms of speed. Also, using our images with dependencies already installed guarantees that dcpy is running in an environment it's been tested in, as opposed to installing the latest dependencies at runtime
> 
> Good point. How about installing `dcpy` during build of `docker-geosupport` image?

docker-geosupport is a legacy image that we don't really update or maintain actively (in original intent, meant for public use as well - just an environment that can run python-geosupport) and it has a fairly different build process than the rest of our images

The `build` images are our standard images for running builds or really any actions in, and I think it's more useful to get the GRU scripts more in line with the rest of our processes than maintaining adding a bespoke image for them. 

Also, if we install dcpy on any image, some form of gdal is going to be installed on that image since it underlies geopandas, geoalchemy, etc. Might be worth doing some investigating at some point to see how/if this aligns with the separate installed binaries... but there's no getting around gdal getting installed somehow, however lightweight an image you may desire

sf-dcp:
Okay. I will prioritize getting the import GRU pr in asap, and then will poke around docker images.

PR #1425, opened 2025-02-06
Ingest GHA cleanup

fvankrieken:
Lot of repeated logic across these jobs, so wanted to just make one job that you feed a list of datasets to

- [bytes action](https://github.com/NYCPlanning/data-engineering/actions/runs/13183409001)
- [opendata](https://github.com/NYCPlanning/data-engineering/actions/runs/13183417116)
- [arcgisonline](https://github.com/NYCPlanning/data-engineering/actions/runs/13183414608)

Right now it expects whitespace-delimited list. Should I maybe tweak so that it could be commas (or commas and whitespace) as well? Might be more useful for triggering via workflow dispatch


alexrichey (approved):
LGTM!

PR #1435, opened 2025-02-10
Enable Assembling/Distributing from a Multilayer shapefile (e.g. dcm.shp)

alexrichey:
Contains a few fixes/additions to make automated distribution possible for products like DCM and Zoning, where we are given a shapefile on Bytes with multiple layers in it, but we'd like to push a single layer to Socrata. The main feature here (commit 3) enables dis/re-assembling a shapefile from an external source like Bytes.

Related PR in [Product Metadata Repo](https://github.com/NYCPlanning/product-metadata/pull/34) to show what metadata changes look like to enable this.

There isn't a successful job run, since I'd need to merge the related PR in the metadata repo. But I've tested locally using `zoning.commercial_overlay_district`, and here's what the output looks like:

image

Long-term, we should add declarative packaging steps to the product-metadata repo to make assembly instructions explicit. Basically, thinking of a file with a list of package steps like so:
- pull multi_layer_shapefile [x]
- reassemble x into n other shapefiles
- generate datadictionary.pdf
- generate datadictionary.xlsx
- zip files x, y, and z
... etc.


dcpy/lifecycle/package/assemble.py
```diff
@@ -44,6 +44,69 @@ def _file_id_to_zipped_paths(
     return mapping
 
 
+SHAPEFILE_SUFFIXES = {"shx", "shp.xml", "shp", "sbx", "sbn", "prj", "dbf", "cpg", "shx"}
+
+
+def unpack_multilayer_shapefile(
```
fvankrieken, line 50: makes sense in this context, but curious if you thought at all about a more naive util that doesn't use metadata but just splits a shapefile based on the contents found inside the zip? And if so what pushed you this way

dcpy/lifecycle/package/assemble.py
```diff
@@ -44,6 +44,69 @@ def _file_id_to_zipped_paths(
...
+            out_path = package_path / file_ids_to_paths[shapefile_id]["package_path"]
+            assert out_path.parent.exists()
+
+            logger.info(f"Packaging shapefile at {unpackaged_path} -> {out_path}")
+            shutil.make_archive(
+                str(out_path).rsplit(".", 1)[0],
+                root_dir=unpackaged_path,
+                format="zip",
+            )
+        print("all done!")
```
fvankrieken, line 106: either use `logger` or ✂️ ?

fvankrieken:
Think this largely looks great but would love to chat tomorrow! A couple thoughts/questions on longer-term vision (mainly just thinking about assemblies potentially living at the product level and how that would interact with what we have now)

dcpy/lifecycle/package/assemble.py
alexrichey: d'oh. yes, oversight.

alexrichey:
> Think this largely looks great but would love to chat tomorrow! A couple thoughts/questions on longer-term vision (mainly just thinking about assemblies potentially living at the product level and how that would interact with what we have now)

For sure about this living at the product-level. Case in point: in the case of DCM or Zoning, we would end up downloading the multi layer shapefile for every dataset in the product. Would love explicit instructions somewhere above the dataset level.

dcpy/lifecycle/package/assemble.py
alexrichey: yeah, there's a TODO that hints at something like that, but it would be nice to extract this into a generic zipfile-splitter util. We do use the metadata in this case to say, effectively, "this zipfile contains these other file-ids, so find the filename we expect for those file-ids, and split the zip accordingly, then put stuff in the correct folder" (usually /dataset_files). Happy to do that here, if you want.

fvankrieken: Nah I don't think that's worth holding it up. 

It'd still be shapefiile-specific I think - not sure in what other scenarios we'd expect a bunch of files in a zip to share a stem.

fvankrieken (approved):

PR #1441, opened 2025-02-10
Enable Excel cross-file keyed updates

alexrichey:
TLDR: this enables automating changes from one excel table to another in a different file. 

Motivated by PFF. Basically, we're handed a gigantic Excel file with Census data, and a smaller file with corrections. I wanted to make the most minimal changes to the existing file, keying off specific values in the corrections file. This was the fastest reproducible way to do it. 

#### Test Case Explanation
To illustrate (and to save you from having to open the xlsx file itself)
The test case has an Excel file with two tabs:

The base tab, to be updated.
image

The tab with "corrections" / "modifications"
image

As a result of running the update, all the non-yellow cells in the last three columns should be updated with vals from the second tab, keying off the first three columns. This is, in miniature, what's happening with PFF and the corrections files.


dcpy/utils/excel.py
```diff
@@ -0,0 +1,145 @@
+import typer
+import openpyxl  # type: ignore
```
fvankrieken, line 2: of course

dcpy/utils/excel.py
```diff
@@ -0,0 +1,145 @@
...
+from pathlib import Path
+
+from dcpy.utils.logging import logger
+
+MISSING_KEYS_ERROR = "Supplied keys must be found in the target table"
+MISSING_DESTINATION_CORR_COL_ERROR = "Destination is missing column for correction"
+DUPLICATE_KEYS_ERROR = "Found duplicate keys"
+
+
+def _unpivot_table(
```
fvankrieken, line 12: semantics... but `unpivot` seems like a misleading term here - it's sort of indexing, or more than anything just creating a dict of key to key-pair values. So a sort of jsonification of the table. But really preserving the same number of rows (# of keys in returned dict) and columns (length of key + number of keys in inner dict). So something like `_table_by_keys` (VERY open to other specific names) would make more sense to me

dcpy/utils/excel.py
```diff
@@ -0,0 +1,145 @@
...
+        key = tuple([str(row_dict[k]) for k in keys])
+        if key in unpivoted:
+            duplicate_keys.add(key)
+        unpivoted[key] = {k: v for k, v in row_dict.items() if k not in keys}
+    if duplicate_keys:
+        raise Exception(f"{DUPLICATE_KEYS_ERROR}. Keys: {duplicate_keys}")
+    return unpivoted
+
+
+def _update_cells(
```
fvankrieken, line 36: And on a similar note, maybe like `_update_cells_by_keys` or something just to make the context of this function quite clear (if we end up adding many more excel utils in the future or whatever - sincerely hope that wouldn't be the case but you never know)

fvankrieken (approved):
A couple minor comments but looks great.

dcpy/utils/excel.py
alexrichey: `_table_by_keys` sounds much clearer. Maybe even `_table_values_by_keys`. Will change!

PR #1445, opened 2025-02-11
update numbldgs to new four-four

### LINKED ISSUE #1403: Scheduled Action Failure - PLUTO Inputs - 📁 Number of Buildings Processing
[Failed action](https://github.com/NYCPlanning/data-engineering/actions/runs/12989728131)

fvankrieken:
closes #1403.

In #1379 we updated the four-four for `doitt_buildingfootprints` because the old page was deprecated and OTI migrated to a new page. That old dataset, somewhat obscured, is also the actual source of `pluto_input_numbldgs`, which gets geocoded and minorly processed before being archived.

Longer term - we should probably have it pull from our archived `doitt_buildingfootprints` - something that we could use for other recipe datasets too where we
- archive `doitt_buildingfootprints` -> gets version from socrata
- above triggers archive of `doitt_buildingfootprints_geocoded`, which gets latest of above and pins version to the same version

We have other workflows where we do similar things - DevDB inputs off the top of my head. But for now, just want to fix this issue.

### DIFF
--- a/products/pluto/pluto_build/python/numbldgs.py
+++ b/products/pluto/pluto_build/python/numbldgs.py
@@ -23,7 +23,7 @@ def get_bbl(inputs):
 
 
 def get_bins(uid):
-    # https://data.cityofnewyork.us/Housing-Development/Building-Footprints/nqwf-w8eh
+    # https://data.cityofnewyork.us/Housing-Development/Building-Footprints/5zhs-2jue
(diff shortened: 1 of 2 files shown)


sf-dcp:
oh, I missed this. LGTM!

sf-dcp (approved):

fvankrieken:
> oh, I missed this. LGTM!

We checked in on it when it first failed! And just decided to not prioritize with all the other things going on.

PR #1453, opened 2025-02-12
Package and Dist - Test Cases Refactor

alexrichey:
This is much less daunting than it seems. I'd recommend that you go commit-by-commit and read the commit message bodies for context.

This cleans up existing test cases to:
1. use a single metadata repo, and within that repo, use shared constants, setup, etc.
2. refactor test cases into the `lifecycle.packaging` structure that @damonmcc and I had talked about. Namely that within the `packaging` phase, there's a sub-phase for `assembly` and `qa`

EDIT: Squashed Commits to rebase/fix conflicts


fvankrieken (approved):
One clarification which I feel like I've lost track of - what specifically does `assemble` mean distinct from `package` in our lingo? Or really what is in `package` that's outside of `assemble` I guess

alexrichey:
> One clarification which I feel like I've lost track of - what specifically does `assemble` mean distinct from `package` in our lingo? Or really what is in `package` that's outside of `assemble` I guess

I think it's not so much you losing track of it, as me not communicating! Chatted with @damonmcc and we were thinking of `packaging` as the lifecycle stage, then within it there's the sub-phase where we `assemble` the package, then the next sub-phase where `qa` it for consistency. (using `qa`, but we could just as easily call it `validation`)

Open to thoughts, of course!

Issue #1334, opened 2024-12-18
FacDB 25v1


damonmcc:
# FacDB Source Data Updates

Like most of our data products, source data must be updated in data library before FacDB is run. As there are are many source datasets with varied update processes, this issue template should be opened to track progress towards updating all source data

**All source data listed is to be uploaded as .csv files**

## Scraped by data library

- [x] bpl_libraries
      Source: Scraped from BPL website
      Source url: https://www.bklynlibrary.org/locations/json
- [x] nypl_libraries
      Source: Scrape from NYPL website
      Source url: https://www.nypl.org/locations/list
- [x] uscourts_courts
      Source: Court locator for NY state
      Source url: http://www.uscourts.gov/court-locator/city/New%20York/state/NY
- [x] dcp_colp
      Source: Bytes
      Source url: https://www1.nyc.gov/site/planning/data-maps/open-data.page#city_facilities
      Need to specify version when archiving the data (check on the website for date last updated).
- [x] nysoasas_programs
      Source: Scraped from OASAS website
      Source url: https://webapps.oasas.ny.gov/providerDirectory/index.cfm?search_type=2
      Need to set version to today’s date as DD-Mon-YY. Example: 13-Nov-20

## Source data from OpenData portals

To see if a dataset needs to be uploaded, check date last updated in open data/bytes against version in data library.

- [x] dca_operatingbusinesses https://data.cityofnewyork.us/Business/Legally-Operating-Businesses/w7w3-xahh
- [x] dcp_facilities_with_unmapped https://www1.nyc.gov/site/planning/data-maps/open-data/dwn-selfac.page
- [x] dcla_culturalinstitutions https://data.cityofnewyork.us/Recreation/DCLA-Cultural-Organizations/u35m-9t32
- [x] dfta_contracts https://data.cityofnewyork.us/Social-Services/Department-for-the-Aging-NYC-Aging-All-Contracted-/cqc8-am9x
      NOTE: `providertype` column has two types of meal categories. One of them, **CITY MEALS ADMINISTRATIVE SERVICES CONTRACTS**, should only have one record (city meals). If there are more records in this category, need to revise code in `dfta_contracts.sql`.
- [x] doe_busroutesgarages https://data.cityofnewyork.us/Transportation/Routes/8yac-vygm

- [x] dot_pedplazas https://data.cityofnewyork.us/Transportation/Routes/k5k6-6jex

[...]

PR #1454, opened 2025-02-13
Ingest QOL tweaks

fvankrieken:
All coming up as part of #1334 and the associated datasets moving over to ingest.

1. originally moved too much logic to __init__.py, now mainly just moves cli target and `ingest` function (the function that runs the whole ingest process) to __init__ so that it's easily accessible
2. make use of the `.lifecycle` dir for "staging" the process
3. make use of the `.lifecycle` dir to create a local file system reminiscent of `edm-recipes/datasets` where the outputs are moved to. Just seemed to be nice for troubleshooting (slash running without a connection to s3)
4. ability to provide the path to a source dataset file when running via cli. Just overrides whatever source is found in the template
5. just make one "archive_dataset". @alexrichey based on our pairing yesterday though, maybe I'll scrap this - you could argue we want a different connector class for "raw" datasets and our parquet datasets, since we interact with them in fairly different ways

### DIFF
--- a/dcpy/lifecycle/ingest/run.py
+++ b/dcpy/lifecycle/ingest/run.py
@@ -1,15 +1,15 @@
 import json
 from pathlib import Path
-import typer
 
 from dcpy.models.lifecycle.ingest import Config
 from dcpy.connectors.edm import recipes
 from . import configure, extract, transform, validate
 
+
 TMP_DIR = Path("tmp")
 
 
-def run(
+def ingest(
     dataset_id: str,
     version: str | None = None,
     *,
@@ -84,34 +84,3 @@ def run(
             case _:
                 pass
     return config
-
-
-app = typer.Typer(add_completion=False)
-
-
-@app.command("run")
-def _cli_wrapper_run(
-    dataset_id: str = typer.Argument(),
-    version: str = typer.Option(
-        None,
-        "-v",
-        "--version",
-        help="Version of dataset being archived",
-    ),
-    mode: str = typer.Option(None, "-m", "--mode", help="Preprocessing mode"),
-    latest: bool = typer.Option(
-        False, "-l", "--latest", help="Push to latest folder in s3"
-    ),
-    skip_archival: bool = typer.Option(False, "--skip-archival", "-s"),
-    csv: bool = typer.Option(
-        False, "-c", "--csv", help="Output csv locally as well as parquet"
-    ),
-):
-    run(
-        dataset_id,
-        version,
-        mode=mode,
-        latest=latest,
-        skip_archival=skip_archival,
-        output_csv=csv,
-    )
--- a/dcpy/lifecycle/ingest/run.py
+++ b/dcpy/lifecycle/ingest/run.py
@@ -3,10 +3,11 @@ from pathlib import Path
 
 from dcpy.models.lifecycle.ingest import Config
 from dcpy.connectors.edm import recipes
+from dcpy.lifecycle import BASE_PATH
 from . import configure, extract, transform, validate
 
-
-TMP_DIR = Path("tmp")
+INGEST_DIR = BASE_PATH / "ingest"
+STAGING_DIR = INGEST_DIR / "staging"
 
 
 def ingest(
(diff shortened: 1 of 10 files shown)


dcpy/lifecycle/ingest/run.py
```diff
@@ -1,15 +1,17 @@
...
 from pathlib import Path
-import typer
-
+import shutil
 from dcpy.models.lifecycle.ingest import Config
 from dcpy.connectors.edm import recipes
 from . import configure, extract, transform, validate
 
-TMP_DIR = Path("tmp")
+INGEST_DIR = Path(".lifecycle") / "ingest"
```
alexrichey, line 8: Had started trying to define some shared path'ing stuff up in [lifecycle](https://github.com/NYCPlanning/data-engineering/blob/2f0a75aa2f9bb60ae6d2f93d64be45000d63f354/dcpy/lifecycle/__init__.py#L6) FYI

dcpy/test/conftest.py
```diff
@@ -97,6 +98,12 @@ def _test_product(
     yield constants
 
 
+@pytest.fixture(scope="function")
+def tmp_dir():
```
alexrichey, line 102: can we just use [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html)?

dcpy/lifecycle/ingest/run.py
```diff
@@ -3,10 +3,11 @@
 
 from dcpy.models.lifecycle.ingest import Config
 from dcpy.connectors.edm import recipes
+from dcpy.lifecycle import BASE_PATH
 from . import configure, extract, transform, validate
 
-
-TMP_DIR = Path("tmp")
+INGEST_DIR = BASE_PATH / "ingest"
+STAGING_DIR = INGEST_DIR / "staging"
```
sf-dcp, line 10: a bit of a nit: naming this var as `INGEST_STAGING_DIR` feels more intuitive to me

dcpy/lifecycle/scripts/validate_ingest.py
```diff
@@ -108,7 +108,7 @@ def library_archive(
 def ingest(
     dataset: str,
     version: str | None = None,
-    ingest_parent_dir: Path = TMP_DIR,
+    ingest_parent_dir: Path = STAGING_DIR,
 ) -> None:
     ingest_dir = ingest_parent_dir / dataset / "staging"
```
sf-dcp, line 113: Need a fix here: we don't need the line 113 sinc you are using the STAGING_DIR path. Also, I would rename `ingest_parent_dir` to `ingest_staging_dir` in the fn signuture

dcpy/lifecycle/ingest/run.py
```diff
@@ -19,20 +22,31 @@ def run(
...
     if not staging_dir:
         staging_dir = (
-            TMP_DIR / dataset_id / config.archival.archival_timestamp.isoformat()
+            STAGING_DIR / dataset_id / config.archival.archival_timestamp.isoformat()
         )
         staging_dir.mkdir(parents=True)
     else:
         staging_dir.mkdir(parents=True, exist_ok=True)
 
+    with open(staging_dir / "config.json", "w") as f:
```
sf-dcp, line 44: replace "config.json" with `CONFIG_FILENAME`?

dcpy/lifecycle/ingest/run.py
sf-dcp: same comment for `OUTPUT_DIR`

sf-dcp:
Curious, what is the need for 4th commit? In which case you wouldn't want to use the template for defining local source?

fvankrieken:
> Curious, what is the need for 4th commit? In which case you wouldn't want to use the template for defining local source?

Great question. Rationale is that for something that's a local csv (or emailed csv) that might have a different filename each time. I don't think that it makes sense to have to commit changes (or just modify the template) to archive it, it's something that should be valid to override at runtime

dcpy/lifecycle/scripts/validate_ingest.py
fvankrieken: I should clean this up, but am still going to provide a folder for this to run in that's separated from the "usual" ingest folders, just to keep it separate from anyone running an "official" job on their own comp

fvankrieken: See the latest commit for renaming - mainly, worked to distinguish any path that's set for `ingest` (i.e. a variable or override that's set for these sort of "parent" folders that contain many datasets) vs any folder that's specific to the dataset being run.

It's still a little more complicated than it should be but given how this operates now I think it makes sense

fvankrieken: This is how the 'validate ingest" script outputs stuff now

image

alexrichey (approved):

sf-dcp (approved):

PR #1457, opened 2025-02-13
Ar pff 2024

alexrichey:


dcpy/lifecycle/data_loader.py
```diff
@@ -61,32 +62,55 @@ def load_dataset_into_pg(
...
         # make column names more sql-friendly
         columns = {
             column: column.strip().replace("-", "_").replace("'", "_").replace(" ", "_")
             for column in df.columns
         }
+        df.rename(columns=columns, inplace=True)
+        pg_client.insert_dataframe(df, ds_table_name)
+
         if include_ogc_fid_col:
-            df.rename(columns=columns, inplace=True)
```
alexrichey, line 88: 🐛! Not one we'd encountered bc we default `include_ogc_fid_col` to True

dcpy/lifecycle/data_loader.py
```diff
@@ -181,3 +201,17 @@ def _import_dataset(
     )
     client = postgres.PostgresClient(schema=database_schema, database=database)
     import_dataset(ds=ds, stage=lifecycle_stage, pg_client=client)
+
+
+@app.command("get_versions")
+def _cli_wrapper_get_versions(
```
alexrichey, line 207: wanted this just to test that db-factfinder versions were being queried correctly. It does indicate that `data_loader` might not be the correct module name.

fvankrieken: Eh it's still roughly there - a util to check that it loaded correctly doesn't seem too far out of scope

fvankrieken: Unless you more mean that data_loader maybe shouldn't know what a "version" is, in which case, time to ponder

fvankrieken: Sorry reading comprehension is not great today - this is a bit separate from what I was saying. Is there a reason you wouldn't want this cli target just say in the connector registry?

products/factfinder/README.md
fvankrieken, line 1: 🙏

dcpy/models/connectors/edm/recipes.py
```diff
@@ -24,6 +24,7 @@ class DatasetType(StrEnum):
     csv = "csv"
     parquet = "parquet"
     xlsx = "xlsx"  # needed for a few "legacy" products. Aim to phase out
+    json = "json"
```
fvankrieken, line 27: sadness

fvankrieken: did you actually put anything in recipes with json format?

fvankrieken: Or is this enum used in a few places now

dcpy/test_integration/connectors/edm/test_plan_load.py
fvankrieken, line 1: You've run this locally?

dcpy/lifecycle/data_loader.py
alexrichey: > Sorry reading comprehension is not great today - this is a bit separate from what I was saying. Is there a reason you wouldn't want this cli target just say in the connector registry?

Oh... yeah, that actually makes a lot of sense. Will move it.

dcpy/test_integration/connectors/edm/test_plan_load.py
alexrichey: affirmative

dcpy/models/connectors/edm/recipes.py
alexrichey: yeah, I'm pulling in a PFF data dictionary JSON file into the QA recipe. Just reflecting a bit: does it make sense to just use the file extension instead of this enum? The only case where we might need to specify are for ambiguously zipped things. (e.g. my_file.zip might be a shapefile)

fvankrieken: sorry - I meant "put in `edm-recipes`". We've gotta axe that name

fvankrieken: Just because this is in the models for edm recipes connector

alexrichey: oh... I didn't even catch onto that. Where would you think these should go?

fvankrieken: we have a `models/files.py` which seems appropriate if this is being used by multiple places? But it's also sort of specific to the connectors. I'm fine with punting

PR #1458, opened 2025-02-14
bump python to 3.13, compile requirements

fvankrieken:

[Python 3.13](https://docs.python.org/3/whatsnew/3.13.html)

Latest sqlfluff is a bit more opinionated - doesn't like subqueries with same aliases as the parent query. I think that makes sense. Made some minimal seemingly sensible tweaks to aliases in offending queries


alexrichey (approved):
LGTM - thanks for digging into all the sqlfluff fixes. Big changes for a patch release!

fvankrieken:
> LGTM - thanks for digging into all the sqlfluff fixes. Big changes for a patch release!

I know - and not the first time we've had suddenly failing code from a so-called "patch"

PR #1463, opened 2025-02-19
Enable Recipes to Specify non-Recipe Sources (e.g. EDM Publishing)

alexrichey:
### The Motivation
I'd like use the existing recipe approach for importing datasets for `lifecycle.builds.qa`, for PFF in particular. This would require using non EDM Recipe sources, e.g. builds and published datasets from Publishing. 

### TLDR: What's In this PR
1. A class-based definition of a connectors, and a refactor/wrapping of edm recipes, publishing, and drafts. This enables the use of other sources (like `edm.publishing.drafts`) in our recipes. 
2. An initial swing at integration tests.

### Longer: What's In this PR
A `Connector` used to encompass many things, like s3 functionality, postgres specifics, database logging, etc. This PR introduces a concrete set of behaviors for our standard connectors (pull, push, and version utils) and exposes them in a connector registry that can be dynamically invoked by lifecycle stages.

It also introduces the start of integration tests for our connectors.

### The long term architecture goal
1. any module under dcpy/connectors primarily interacts only with external sources (e.g. Socrata, edm-publishing, etc.) and not with lifecycle stages, postgres, and such. Connectors should generally not be aware of (ie they should not import) anything under `dcpy.lifecycle`. (Note: that goal is not fully achieved here. Looking at you `edm.recipes`). One notable quirk to call out is that connectors **do** know how their data should be organized when pulled to your machine. `edm.publishing.drafts` needs to organize things differently than `edm.publishing.published` because one has a concept of drafts, and the other doesn't
2. `dcpy.lifecycle` stages should generally not know anything specific about connectors. They should retrieve connectors by name from the registry, and call the methods on the VersionedConnector implementations.
3. For lifecycle specific configuration (ie. "where do I put data for builds`) the lifecycle.config should be referenced.

In diagram form, the dependency graph should look like this below. This is largely achieved for recipes, publishing.drafts and publishing.published.

image

Note re the text in orange: Calling this out as current behavior that maybe we don't want. Similar to loading data, we could expand `dcpy.data_loader` to handle version querying, etc. 

### Future Work
- [ ] Implementing the integration tests locally and in GHA
- [ ] lifecycle.builds.load (and maybe plan) still has a few references to `connectors.edm.recipes`. I didn't want to fight with it at the moment, but it wouldn't be too difficult. 
- [ ] duplicated `InputDataset.id` in a recipe will cause problems in LoadResults. It's not a problem if you don't care about the returned LoadResults (as I don't for PFF) but it does break test cases. 
- [ ] Refactor the distribution dispatcher logic just use the Connector Registry. The Protocols with the generics were an experiment motivated by type-safety, but they suuuuck to use and reason about.
- [ ] the `pull_conf` that we put in recipes is not well-defined. I think that's fine, but something to keep an eye on.
- [ ] parquet files are read into memory, and can cause Out of Memory errors

### Other Callouts
We'd talked about just registering a subset of connectors for certain stages, to limit bad behavior. For example, it would be bad to have Socrata as a recipe source for a build (since Socrata effectively has no versions). For now I thought it was simpler to just register one set of connectors. We'll have to just _not do stupid things like using Socrata in recipes_.


dcpy/models/connectors/__init__.py
```diff
@@ -14,10 +14,22 @@ def push(self, arg: _O, /) -> Any:
     def pull(self, arg: _I, /) -> Any:
         """pull"""
 
+    def list_versions(self, key: str, sort_desc: bool) -> list[str]:
```
alexrichey, line 17: add version querying methods to the connector protocol. We'll need these in order resolve recipes dataset.versions in a generic way.

dcpy/connectors/edm/publishing.py
```diff
@@ -896,5 +897,28 @@ def _cli_wrapper_download_file(
...
+
+    # In lifecycle.builds, we would always expect to receive an
+    # InputDataset (from the recipe) so we adapt the existing
+    # edm.publishing connector to work with that type.
+    def pull(self, input_ds: InputDataset):
+        key = parse_version(input_ds.id, input_ds.version) # <- convert to key
+        out_path = publishing.download_file(key)
+        # dump into a database, file, etc.
+
+    def list_versions(self, key: str, sort_desc: bool = True) -> list[str]:
```
alexrichey, line 914: stubbing out these methods for now.

dcpy/models/connectors/__init__.py
```diff
@@ -14,10 +14,22 @@ def push(self, arg: _O, /) -> Any:
...
+        return ""
+
+    def version_exists(self, key: str, version: str) -> bool:
+        return False
+
 
 class ConnectorDispatcher(Generic[_O, _I]):
     _connectors: dict[str, _ConnectorProtocol[_O, _I]]
 
+    def __getitem__(self, item):
```
alexrichey, line 30: in the current implementation, we wrap the ConnectorDispatcher's methods. Not sure that really makes sense, tbh... might make more sense to just make the connectors retrievable, like so

dcpy/connectors/edm/publishing.py
```diff
@@ -896,5 +897,28 @@ def _cli_wrapper_download_file(
     download_file(key, filepath, output_dir)
 
 
+@dataclass
+class RecipeConnectorAdapter:
```
fvankrieken, line 901: One just gut reaction - recipes and publishing are somewhat special connectors in two main ways, one maybe frivolous (the first) and the other maybe a little more pertinent. 
1. They're mapped to very specific entities (or implementations) rather than something generalized. i.e. we pull from various socrata endpoints using the socrata connector, so the connector is an generalized connector api for theoretically any socrata endpoint. Recipes and publishing you could argue could be generalized (or are, just depending on env vars), but really they're tightly coupled to a specific entity (of ours) in the cloud at the moment. I can't decide how important this distinction is or if it really matters though
2. They're "trusted". That is to say, they serve very specific roles in our pipelines - builds can only pull from the recipe connector (at the moment - argument that they should pull from publishing as well). qa will (presumably) only pull from recipe and publishing connectors. So there might be logic specific to these that we wouldn't want applying to other connector

I don't think that this means they _need_ to be something other than a `connector`, but I am curious as to how you've thought about those sorts of distinctions and how to handle them as we try to move towards a generalized connector interface

fvankrieken: Thinking about that more for the second point, I feel like validation might be good, but if anything this is a very _good_ reason to have these simply be other connectors actually. Like if someone wanted to use `lifecycle` code to pull in data from socrata, rather than a long term data store, why not let them? But also let them configure their long term data store and pull that data into pipelines easily

dcpy/models/connectors/__init__.py
fvankrieken: Thoughts on what this would look like for, say, socrata? Like would we prefer an error because socrata cannot have multiple versions? Or just return `["2025-02-19"]`

fvankrieken:
Love it. Very few notes, I think this makes a ton of sense

fvankrieken:
> Products.datasets aren't as standardized as recipes. How do we specify particular files from a product? e.g. acs.csv in the factfinder.acs_current dataset?

Feels like this is a bit orthogonal, but maybe this we start with a barebones implementation in a `recipe`?

It would make sense for a recipe to contain information on how to generate outputs
```yml
inputs:
  dataset_defaults:
    destination: postgres
  datasets:
    - id: dcp_pop_acs2010_demographic
outputs:
  files:
    - id: acs
      filepath: acs.csv
      source: acs # as in, say the table in the database that this gets exported from
```

Longer term, we could hopefully just at the end of a build run like `dcpy.cli lifecycle builds export recipe.lock.yml` and dcpy would have the info needed to generate them all, but for now it could just be more a log of what is expected

products/factfinder/recipe_qa.yml
```diff
@@ -0,0 +1,30 @@
...
+    # Business as usual. Pull from recipes
+    - id: dcp_pop_acs2010_demographic
+      version: latest
+      source: edm.recipes # <- pull this one from recipes
+
+
+    # Datasets below will default to pulling from edm.publishing
+
+    # grab my build. In reality, we'd generalize this by using version env vars.
+    - id: factfinder.acs_current # <- proposal: use the form of [product].[dataset] for edm.publishing. would need to make a string resolver
```
fvankrieken, line 21: This gives me the slightest ick but I think it's the cleanest way to do it. Otherwise we'd need custom pydantic types per connector which is exactly what we're trying to move away from with this sort of dispatch logic. And also in terms of a literal "identifier", it's hard to imagine what else would be a great unique identifier of a file/dataset within one of our products than a string like this (certainly don't want the `id` field to be a pydantic record instead of a string)

dcpy/models/connectors/__init__.py
alexrichey: hadn't thought about it! The list-with-one-item approach looks nice for this odd destination type

fvankrieken: Don't know if it'd be completely odd - AGOL I think is similar, and in general there might be many connectors where there's only "current" data and "version" is a little less distinct as an entity

dcpy/connectors/edm/publishing.py
fvankrieken: All makes sense.

Though the Azure note has me thinking slightly - you could imagine wanting literally the exact same api for "recipes" for Azure vs DO spaces, just with a different "backend" if that makes sense. Similarly, someone might want to use ingest but have the same folder system/versioning/etc set up on a local file system. There are a couple spots to separate out that logic
- one connector that applies to all of them, but the backend is configurable. Essentially, we're doing something like setting `dcpy.connectors.recipes.backend=s3` or something in a pyproject.toml for our repo (not for dcpy package, but like a runtime configuration)
- multiple connectors, which you're suggesting and writing this out I think I'm more agreeing with. I guess my hang up is this way, we'll implement the same api a couple times. That's completely fine - that's the idea behind connectors, is each have their implementation. I guess where I'm having a slight hangup is that each of them is not simply a different connector but each is a different implementation of a more specific connector - our long term data store where we archive raw and processed data and store them in a similar format. So not only would each one say, implement QueryVersion, but each is going to have a `get_config` function, and query_version that uses `get_config` in the same way, and each is making the same assumptions about the underlying structure of the data store.

Sorry obviously thinking out loud here. But I think that's a lot of writing to say that 1. this all makes sense, and to my concerns/thoughts here 2. maybe we have something along the lines of a folder called `edm/recipes` which contains both a more generic representation of our `recipes` api and then submodules like `s3` or `azure_blob_storage` etc that implement the specifics.

dcpy/connectors/edm/recipes.py
```diff
@@ -241,31 +254,25 @@ def read_df(
...
+):
+    logger.info(f"preproc: {preprocessor}")
     ds_table_name = import_as or ds.id
-    logger.info(
-        f"Importing {ds.id} {ds.file_type} into {pg_client.database}.{pg_client.schema}.{ds_table_name}"
-    )
-
-    local_dataset_path = fetch_dataset(ds, local_library_dir)
-
     if ds.file_type == DatasetType.pg_dump:
```
fvankrieken, line 276: Keeping this logic here for now? Or trying to get into `data_loader` as part of this pr? Haven't gotten to `data_loader` yet

dcpy/connectors/registry.py
```diff
@@ -0,0 +1,122 @@
...
+from dcpy.utils.logging import logger
+
+
+### Dispatchers
+
+_O = TypeVar("_O", contravariant=True)
+_I = TypeVar("_I", contravariant=True)
+
+
+class PushPullProtocol(Protocol, Generic[_O, _I]):
```
fvankrieken, line 14: Do we want to just scrap these?

dcpy/connectors/edm/recipes.py
alexrichey: yeah, IIRC I started refactoring this into data_loader, but it started to unravel into a lot of changes. It's a definite TODO though... This leaves thing in a somewhat kludgey state, but there are already a lot of changes here. Thoughts? Happy to work on this together to clean this up.

dcpy/connectors/registry.py
alexrichey: Yeah, we do, but they're used in Distribution. Shouldn't be terribly difficult to refactor that.

dcpy/test/lifecycle/builds/test_load.py
```diff
@@ -132,66 +71,24 @@ def test_import_parquet_file(self):
...
-            load.import_dataset(ds, None)
-
-
-@patch("dcpy.connectors.edm.recipes.fetch_dataset", return_value=RESOURCES_DIR)
-@patch("dcpy.connectors.edm.recipes.read_df", return_value=_test_df)
-@patch("dcpy.lifecycle.builds.metadata.write_build_metadata")
-@patch("dcpy.lifecycle.builds.plan.write_source_data_versions")
-@patch("dcpy.utils.postgres.PostgresClient")
-@patch("dcpy.connectors.edm.recipes.purge_recipe_cache")
-def test_load_source_data(
```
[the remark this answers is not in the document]
alexrichey: 💔

dcpy/connectors/edm/recipes.py
fvankrieken: Lemme take a quick glance, but happy to defer to your judgement

fvankrieken: Looks like recipes.import_dataset is now only used by my ingest validation scripts? So if we just get that using data loader, I think this can just be scrapped entirely

alexrichey: Cool - I'll take a swing at killing it

dcpy/connectors/registry.py
alexrichey: Though (to be clear) would like to do that in a followup PR

alexrichey:
@fvankrieken fyi [nightly qa](https://github.com/NYCPlanning/data-engineering/actions/runs/13774196663) job passing on this branch

fvankrieken:
> @fvankrieken fyi [nightly qa](https://github.com/NYCPlanning/data-engineering/actions/runs/13774196663) job passing on this branch

Want to just rebase latest main? Just with a big one I'm more comfortable with making sure nothing is out of sync

alexrichey:
> > @fvankrieken fyi [nightly qa](https://github.com/NYCPlanning/data-engineering/actions/runs/13774196663) job passing on this branch
> 
> Want to just rebase latest main? Just with a big one I'm more comfortable with making sure nothing is out of sync
> 
> Just for PR tests, not re-running nightly qa

Sure thing, rebased. In this case, it just adds your GHA change, and templates folder move. But... rebasing will give us fully green PR checks, which is nice.

fvankrieken:
> Sure thing, rebased. In this case, it just adds your GHA change, and templates folder move. But... rebasing will give us fully green PR checks, which is nice.

Hmm maybe not - seems like that setting I removed defaults to true anyways?

fvankrieken:
> > Sure thing, rebased. In this case, it just adds your GHA change, and templates folder move. But... rebasing will give us fully green PR checks, which is nice.
> 
> Hmm maybe not - seems like that setting I removed defaults to true anyways?

[Docs](https://github.com/codecov/codecov-action) says defaults to false, so it should have passed

PR #1471, opened 2025-02-21
get FacDB 25v1 building

fvankrieken:
Many ingest switchover-related tweaks

[Passing build](https://github.com/NYCPlanning/data-engineering/actions/runs/13488685627)

There are a bunch of little things, think each commit is largely self-explanatory though.

Part of #1334 but we're still blocked by COLP on actual build

### DIFF
--- a/.github/workflows/ingest_multiple.yml
+++ b/.github/workflows/ingest_multiple.yml
@@ -75,6 +75,7 @@ jobs:
       DEV_FLAG: ${{ inputs.dev_bucket && 'true' || 'false' }}
     strategy:
       fail-fast: false
+      max-parallel: 10
       matrix:
         dataset: ${{ fromJSON(needs.set_matrix.outputs.matrix) }}
     steps:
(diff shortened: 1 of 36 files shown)


dcpy/lifecycle/ingest/templates/nysdec_solidwaste.yml
```diff
@@ -6,6 +6,7 @@ attributes:
   url: https://data.ny.gov/Energy-Environment/Solid-Waste-Management-Facilities/2fni-raj8
 
 ingestion:
+  target_crs: EPSG:4326
```
fvankrieken, line 9: All other datasets coming into facdb are assumed to have wgs84, so just made consistent. This wasn't handled in ingest migration because coordinates were never converted into geom column before

products/facilities/facdb/sql/_create_facdb_boro.sql
```diff
@@ -1,16 +1,5 @@
...
-    borocode,
-    (CASE
-        WHEN boro = 'MANHATTAN' THEN 'NEW YORK'
-        WHEN boro IN ('BROOKLYN', 'BRONX', 'STATEN ISLAND') THEN boro
-        ELSE city
-    END) AS city,
-    NULLIF(NULLIF(REGEXP_REPLACE(LEFT(zipcode, 5), '[^0-9]+', '', 'g'), '0'), '') AS zipcode
-INTO facdb_boro
-FROM (
+WITH coalesced_boro AS (
```
fvankrieken, line 2: Just felt cleaner to have a CTE than a subquery (and sqlfluff was unhappy)

.github/workflows/ingest_multiple.yml
```diff
@@ -75,6 +75,7 @@ jobs:
       DEV_FLAG: ${{ inputs.dev_bucket && 'true' || 'false' }}
     strategy:
       fail-fast: false
+      max-parallel: 10
```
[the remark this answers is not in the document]
fvankrieken: I just figured with our socrata issues this made sense - I used it to archive 30 at once so wanted to increase odds that if we were throttled that it wasn't an issue

PR #1475, opened 2025-02-24
fix use of inputs in action run name

damonmcc:
[before](https://github.com/NYCPlanning/data-engineering/actions/runs/13508535872) and [after](https://github.com/NYCPlanning/data-engineering/actions/runs/13508623247)


alexrichey (approved):
Cool Cool

PR #1477, opened 2025-02-26
add tabulate to dcpy requirements

fvankrieken:


damonmcc:
nice. for printing dataframes?

fvankrieken:
It's actually used in dcpy already, just wasn't an explicit requirement. So if you installed dcpy in a clean environment before this, it wouldn't install tabulate

PR #1478, opened 2025-02-27
Get facdb 25v1 passing... again

fvankrieken:
Because bytes quarterly update datasets had been hard-coded to filetype pgdump, there were 404s when trying to pull them after the 25a update yesterday (first time they were archived by ingest)

[successful run](https://github.com/NYCPlanning/data-engineering/actions/runs/13568330146)

I looked at the QA app and results seem okay... EXCEPT we have way fewer records than last time coming from dca_operatingbusinesses. I assume I bungled something in a different PR, will address separately

### DIFF
--- a/products/facilities/recipe.yml
+++ b/products/facilities/recipe.yml
@@ -11,26 +11,17 @@ inputs:
   - name: dcp_mappluto_wi
     file_type: pg_dump
   - name: dcp_boroboundaries_wi
-    file_type: pg_dump
   - name: dcp_ct2010
-    file_type: pg_dump
   - name: dcp_ct2020
-    file_type: pg_dump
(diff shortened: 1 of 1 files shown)


sf-dcp (approved):
LGTM. Can you check what other products use hardcoded pg_dump format for these quarterly datasets?

damonmcc (approved):
the `dca_operatingbusinesses` might be related to the weird filtering we do in `pipelines.py` (expecially the `where df.expiration_date >= today`)

fvankrieken:
> LGTM. Can you check what other products use hardcoded pg_dump format for these quarterly datasets?

None others pin - FacDB needed this because of the default to parquet (previously csv)

PR #1484, opened 2025-02-28
sftp

damonmcc:


dcpy/connectors/sftp.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+    return response
+
+
+class FTPConnector:
+    def push(self, dest_path: str, ftp_profile: str):
+        raise Exception("Push not implemented for FTP")
+
+    def pull(self, **kwargs):
+        raise Exception("Pull not implemented for FTP")
```
damonmcc, line 74: will implement when the dust settles around connects in @alexrichey's open PR

dcpy/connectors/sftp.py
```diff
@@ -0,0 +1,74 @@
+import os
+import tempfile
+import pysftp  # type: ignore
+import paramiko  # type: ignore
```
alexrichey, line 4: Can you add the type-stubs for these?

dcpy/connectors/sftp.py
```diff
@@ -0,0 +1,74 @@
...
+
+from dcpy.utils.logging import logger
+from dcpy.models.connectors.sftp import SFTPServer, SFTPUser
+
+
+def _temp_private_key_from_env() -> str:
```
alexrichey, line 10: Is this actually a temporary file? Seems like it's not going to be deleted

dcpy/connectors/sftp.py
```diff
@@ -0,0 +1,74 @@
...
+        f"Wrote sftp private key from environemnt variable SFTP_PRIVATE_KEY to {tmp_path} ..."
+    )
+    return tmp_path
+
+
+def _connection(server: SFTPServer, user: SFTPUser):
+    key_path = (
+        user.private_key_path if user.private_key_path else _temp_private_key_from_env()
+    )
+    # disable host key checking
```
alexrichey, line 24: Host checking is definitely something we should enable, yeah? Might be out of scope here, but seems important.

dcpy/test_integration/test_sftp.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    assert entries == ["authorized_keys", "keys"]
+
+
+def test_get_file(default_sftp_kwargs: dict):
+    sftp.get_file(**default_sftp_kwargs)
+    assert default_sftp_kwargs["local_file_path"].exists()
+
+
+def test_put_file(default_sftp_kwargs: dict):
```
alexrichey, line 50: Should we start this with an assertion that the file doesn't exist, via list_directory? I could see stubbing my toe here developing locally.

alexrichey:
@damonmcc This looks really great. I'd like to refresh myself on best practices around ssh key management here, and will revisit tomorrow. In the meantime, had a few questions/suggestions

dcpy/test_integration/test_sftp.py
damonmcc: makes sense. if it can't be pushed as a temp file, may have to delete it after pushing so that running the test twice doesn't fail

dcpy/connectors/sftp.py
damonmcc: it does seem important enough to make it an option. but since the Ginger server reject connections that check for host keys, I hardcoded it here

maybe connection options like these would best be set in a connector's implementation (thinking new connector apporach) e.g. `class GingerSFTPConnector(SFTPConnector): ...`

dcpy/connectors/sftp.py
damonmcc: i'll run these again and check. temp files baffle me sometimes

alexrichey:
> @damonmcc This looks really great. I'd like to refresh myself on best practices around ssh key management here, and will revisit tomorrow. In the meantime, had a few questions/suggestions

Sorry, just getting back to this. Re: keys, I'm not sure I quite follow the integration test setup. Specifically, with the two private keys you checked in, it looks like they're the same? I assume that's the private key that the client is using to connect... in which case we wouldn't want the server to have a copy - the server should just have the public key.

dcpy/connectors/sftp.py
damonmcc: two ways:
1. connections fail with `SSHException: No hostkey for host ginger.nyc.gov found.`
2. we set `StrictHostKeyChecking=no` in our current bash-based connections [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/pluto/pluto_build/bash/ssh.sh)

dcpy/connectors/sftp.py
damonmcc: tried fixing to actually use a temp file and the added complexity of the necessary context management isn't worth it

especially since
1. the current bash-bashed approach [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/pluto/pluto_build/bash/ssh.sh) adds files to `~./ssh`
2. the key is already a very not-temp envar, so using a temp file doesn't seem to add value/security

changed this to write the key to a file `/.ssh/`, a standard folder for keys that we gitignore

damonmcc:
@alexrichey 
> Sorry, just getting back to this. Re: keys, I'm not sure I quite follow the integration test setup. Specifically, with the two private keys you checked in, it looks like they're the same? I assume that's the private key that the client is using to connect... in which case we wouldn't want the server to have a copy - the server should just have the public key.

seems like a private key is user-specific and both the server and the client need to know it. so one of the checked in copies is for the server (mounted volume in dev container server) and the other is in test resources for tests to use. also added `SFTP_PRIVATE_KEY=` to example.env for tests that expect the envar

I'll rebuild the dev container without the mounted copy and see how it goes

alexrichey:
> @alexrichey
> 
> > Sorry, just getting back to this. Re: keys, I'm not sure I quite follow the integration test setup. Specifically, with the two private keys you checked in, it looks like they're the same? I assume that's the private key that the client is using to connect... in which case we wouldn't want the server to have a copy - the server should just have the public key.
> 
> seems like a private key is user-specific and both the server and the client need to know it. so one of the checked in copies is for the server (mounted volume in dev container server) and the other is in test resources for tests to use. also added `SFTP_PRIVATE_KEY=` to example.env for tests that expect the envar
> 
> I'll rebuild the dev container without the mounted copy and see how it goes

@damonmcc No, the server in general should not have a copy of your private key. It needs your public key, which it can then use to verify that you have the associated private key.

dcpy/connectors/sftp.py
alexrichey: > connections fail with SSHException: No hostkey for host ginger.nyc.gov found.

This probably means that you don't have the hostkey locally, not that Ginger doesn't allow it.

> we set StrictHostKeyChecking=no in our current bash-based connections

I'm guessing they just never figured out the checks?

dcpy/connectors/sftp.py
alexrichey: Is it actually that complicated to do the context management? Happy to brainstorm, or pair. I think the current approach is very bad-practice because If you run this locally, you're going to have private keys laying around afterwards.

damonmcc:
@alexrichey dropped the private key from the server container's volume and tests pass! rebasing to reflect that

dcpy/connectors/sftp.py
damonmcc: gotcha. I'll add host checking when I implement connectors in a follow-up

dcpy/test_integration/test_sftp.py
damonmcc: thought about it and adding that assertion doesn't seem like a good idea:
- we talked about the value of tests like these only testing one thing, e.g. `test_put_file` shouldn't fail if `list_directory` breaks
- if this tests expects the file to not be there and then puts it there, it'd also have to delete it
- overwriting a file is expected behavior. we can add a "don't overwrite flag" if we want later
- if `sftp.put_file` doesn't succeed, it'll raise an error. so I'm not sure what you could stub your toe on locally

dcpy/connectors/sftp.py
damonmcc: let's chat IRL about this

from the docs and examples of usage with this stuff, using a non-temp file seems like normal practice. and local runs would leave the private key file laying around in your repo directory at `data-engineering/.ssh/`, a directory that gitignored and right next the `.env` file which already has the private key in it and isn't a temp file

alexrichey:
@damonmcc and @fvankrieken, I think we should set some time aside to chat about the 1. handling of SSH keys and 2. host checking. With security related work, it's important to really think things through and be on the same page.

dcpy/test_integration/test_sftp.py
alexrichey: For the first item, an analogue would be database CRUD tests. The industry standard integration tests will insert some data and re-query it. (While you could rely on a success response from the database client, that's not generally considered sufficient)

sf-dcp:
> @damonmcc and @fvankrieken, I think we should set some time aside to chat about the 1. handling of SSH keys and 2. host checking. With security related work, it's important to really think things through and be on the same page.

@sf-dcp plz invite me too

damonmcc:
> > @damonmcc and @fvankrieken, I think we should set some time aside to chat about the 1. handling of SSH keys and 2. host checking. With security related work, it's important to really think things through and be on the same page.
> 
> @alexrichey plz invite me too

per chat during standup

1. change sftp utils to expect a private key file to be in the running machine's standard directory for ssh keys: `/home/.ssh/` and not depend on envars
2. don't disable host key checking
3. do all necessary local env setup to get 1 and 2 working
4. punt on issues related to how this will work in github actions (like getting and saving creds from 1password)

example.env
```diff
@@ -4,4 +4,5 @@ AWS_SECRET_ACCESS_KEY=
 AWS_ACCESS_KEY_ID=
 RECIPES_BUCKET=
 RECIPE_ENGINE=
-EDM_DATA=
\ No newline at end of file
+EDM_DATA=
+SFTP_PRIVATE_KEY=""
```
fvankrieken, line 8: ✂️ ?

dcpy/connectors/sftp.py
damonmcc: @fvankrieken yup I'm workin on that. other than dropping the lines that disable it, also have to add the host keys to the dev container so tests pass

just changed the PR to draft

sf-dcp:
### The following changes have been done:
1. Create sample host and client authentication keys for dev container
2. Enable host checking when connecting to sftp server
3. Modify `SFTPUser` model to have custom `known_hosts` and authentication key paths. If not provided, a default value will be used. For authentication keys, the sftp connection object will look for available keys.

fvankrieken:
>     4. Replace `pysftp` library interacting with sftp server with `paramiko`. First of all, `pysftp` is a thin layer on top of `paramiko`. So why to replace? Because `pysftp` doesn't support the signature algorithm `ssh-rsa` which is used with the DOF ginger server authentication keys.

also looks like pysftp hasn't been updated in 9 years and paramiko is actively maintained, so seems like a good move in general too!

alexrichey (approved):
This is so good! Wouldn't change a single thing.

Would be super convenient if git stored file permissions. Maybe we can add a container startup script to chmod 600 the ssh keys, or do that in the test fixture. I suppose we can get to that when we automate the integration tests in GHA.

fvankrieken:
> > This is so good! Wouldn't change a single thing.
> > Would be super convenient if git stored file permissions. Maybe we can add a container startup script to chmod 600 the ssh keys, or do that in the test fixture. I suppose we can get to that when we automate the integration tests in GHA.
> 
> git does store file permissions!

Oh nevermind - just if it's executable or not

PR #1485, opened 2025-03-01
Compile python requirements

fvankrieken:

duckdb dropped from our requirements due to it no longer being a req - we currently use it in dcpy (in an unused util file) and we certainly like having it in our dev environments so added it in.

Also added more reqs for dcpy to pyproject.toml (used pipreqs to generate list)


.github/workflows/test.yml
```diff
@@ -179,6 +179,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         if: needs.determine_image_logic.outputs.child_action == 'Build'
+      - name: Set up Python 3.13
```
fvankrieken, line 182: Annoyed that this was needed but there's a [step in setting up the build images](https://github.com/NYCPlanning/data-engineering/blob/main/admin/ops/docker_build_and_publish.sh#L26) that makes sure all requirements of dcpy are installed on the images, even if they're not in the image-specific `requirements.txt`. This step was failing as the runners default to python 3.10 and for whatever reason the required packages could not be resolved with our constraints with python 3.10

PR #1490, opened 2025-03-03
`ingest`: add feature to configure template directory

sf-dcp:
Inspired by our upcoming presentation during Open Data week. 

### What: 
Currently, template directory in ingest is hardcoded. We want anybody to be able to use ingest with their own templates, not ours. The following are added in this PR:
- Ability to specify env var with path to a local directory with templates
- Ability to provide an optional CLI arg with template dir. If both env var and arg present, the CLI arg takes precedence over the env var.
- Move templates folder from `./dcpy/lifecycle/ingest/templates` to `./templates`. This is done so that our templates folder isn't included when someone installs `dcpy` via pip install.

#### GHA
Ingest is used in 3 GHA templates. These are test runs to make sure all work as expected.
* ingest single dataset [run](https://github.com/NYCPlanning/data-engineering/actions/runs/13706290988).
* ingest multiple datasets [run](https://github.com/NYCPlanning/data-engineering/actions/runs/13706331654).
* zoning dataloading [run](https://github.com/NYCPlanning/data-engineering/actions/runs/13706900300/job/38334283109#step:6:91). The link shows the template being read in successfully. The job fails presumably because I run it in my dev bucket which doesn't have the gis datasets (source).

#### Running CLI: 
* with flag:
    image

* without flag (use env var):
    image

### Considerations:
* The `TEMPLATE_DIR` env var is checked inside of `ingest()` fn. First, I had this var evaluated at a top level, outside of the function; it would error out if it didn't find the env var. However, this approach would cause non-ingest tests (validating product metadata) fail if the var wasn't present. So I decided to go with a lazy evaluation approach.


dcpy/test/lifecycle/ingest/shared.py
```diff
@@ -26,6 +26,7 @@
 TEST_OUTPUT = RESOURCES / TEST_DATA_DIR / f"{TEST_DATASET_NAME}.parquet"
 FAKE_VERSION = "20240101"
 TEST_DATASET = Dataset(id=TEST_DATASET_NAME, version=FAKE_VERSION)
+PROD_TEMPLATE_DIR = Path(__file__).parent.parent.parent / "templates"
```
sf-dcp, line 29: We need this variable to test validate actual templates. I know this looks ugly but we need an actual static path. Open to suggestions how to make this prettier

dcpy/test/lifecycle/ingest/shared.py
```diff
@@ -26,6 +26,7 @@
 TEST_OUTPUT = RESOURCES / TEST_DATA_DIR / f"{TEST_DATASET_NAME}.parquet"
 FAKE_VERSION = "20240101"
 TEST_DATASET = Dataset(id=TEST_DATASET_NAME, version=FAKE_VERSION)
+PROD_TEMPLATE_DIR = Path(__file__).parent.parent.parent.parent.parent / "templates"
```
sf-dcp, line 29: We need this variable to test validate actual templates. I know this looks ugly, but we need an actual static path to validate our "production" templates in tests. Open to suggestions how to make this prettier!

templates/bpl_libraries.yml
fvankrieken, line 1: Should we maybe go with `ingest_templates` as folder name?

dcpy/configuration.py
```diff
@@ -26,3 +27,7 @@
 IGNORED_LOGGING_BUILDS = ["nightly_qa", "compile_python_reqs"]
 
 PRODUCT_METADATA_REPO_PATH = env.get("PRODUCT_METADATA_REPO_PATH")
+
+TEMPLATE_DIR: Path | None = (
```
fvankrieken, line 31: Any thoughts on having default value here if not supplied, instead of the literal being an optional value? Haven't fully thought through which I'd prefer, just curious

dcpy/configuration.py
```diff
@@ -26,3 +27,7 @@
...
+    Path(env.get("TEMPLATE_DIR", "")) if env.get("TEMPLATE_DIR") else None
```
fvankrieken, line 32: nittiest nit - given that we check `if env.get("TEMPLATE_DIR"`, the first part can just be `Path(env["TEMPLATE_DIR"])`

sf-dcp: Unfortunately `mypy` doesn't like it, so I had to change it to this 🙃

fvankrieken: This is why we need easier functional syntax in python! `env.get("TEMPLATE_DIR") |> Option.map(Path)`

templates/bpl_libraries.yml
sf-dcp: Yeah, sounds good

dcpy/configuration.py
sf-dcp: Thinking for an external use case, I would want the folder to be defined explicitly, rather than having a default path. And as soon as a user tries to run ingest, it will loudly fail with a clear message. 

Additionally, I think it makes the logic a bit more complex because there will be 3 possible values: env var, CLI arg, and a default value. Meaning we would need to decide on a precedence (I think I'm using the right word but not sure) which value is considered a priority.

dcpy/test/lifecycle/ingest/shared.py
fvankrieken: Hmm... if we're pulling this out of dcpy, this test shouldn't be run as part of dcpy's pytests.

Alex's tests of product metadata offer an example of we might test this outside of dcpy, see [here](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/test_helper.yml#L216). And with that, you'd need to add a `filter` to make sure we run the test whenever templates are changed (but dcpy isn't - that test that Alex has there only runs when there are changes to dcpy. For the filters, that logic lives in test.yml instead of test_helper.yml)

dcpy/configuration.py
sf-dcp: Actually, I didn't look closely to your comment, sorry. I meant that `mypy` didn't like having Path(env.get("TEMPLATE_DIR") where Path may receive `None` or `str`; `mypy` doesn't see the if condition. 

What you are proposing essentially returns a string or an error, correct? I can try this approach and see

fvankrieken: I don't think that's right. Running mypy on 
```python
from os import environ as env
from pathlib import Path

a: Path | None = Path(env["var"]) if env.get("var") else None
```

Yields `Success: no issues found in 1 source file`

fvankrieken:
Also sorry - you have a merge conflict from my dcp_colp fix :( should be easy to resolve though

dcpy/configuration.py
sf-dcp: Aha, nice. By error I meant querying `env["var"]` when "var" key doesn't exist returns error, not `None`

dcpy/test/lifecycle/ingest/shared.py
fvankrieken: I think this is worth implementing in this PR - if you have any questions around the path filters happy to chat, beyond that I'd hope it's relatively straightforward

sf-dcp: Should I implement the test migration here or in a separate PR? Not sure what the scope size looks like.

fvankrieken: Eh it's maybe a little more work than I'm giving it credit for. Sorry let me stop thinking while I type and actually think about what it'd take

fvankrieken: It would need
- removal of code in pytest
- new cli target for ingest - validate a template (and it's processing steps). This isn't complicated per se but there are definitely some subleties, what's the scope of this cli, does it need new helpers, etc. Plus how would a user like to use this (validating their own templates - makes sense as a use case)
- new path filter in `test.yml`
- new job in `test_helper.yml`

The second is the only one that seems like it could complicate things, and I think makes me feel that it would be worth keeping this PR as is, since it preserves how we test now, and then worry about refining that behaviour next sprint

sf-dcp: I see. Thanks for looking into it. I can create a follow up GH issue with the steps you outlined

dcpy/configuration.py
sf-dcp: Yeah, I agree about the hustle with an env var. Default path seems easier than an env var. I'm leaning towards limiting to 2 options, whatever they are

fvankrieken (approved):
Squash please!

PR #1493, opened 2025-03-04
fix failing gft dbt source test

fvankrieken:
gft build is our only failing build on nightly qa, and I don't know when if ever NYSDEC is going to fix (or respond) about the one little data issue (two lakes near binghamton NY having the same pwl_id) which doesn't actually impact the build itself. So I wanted to resolve it.

A couple thoughts on options fixing this
1. just set to warn, and add a "todo"
2. above, but also test the staging table. Long term, there's an argument for testing staging tables rather than source tables as general practice - given the scope of staging tables, it sort of makes sense that we'd want to impose these tests when the tables have been staged and are "transformation-ready". In this case, the data issue is with lakes outside of NYC - after clipping to NYC boundary, pwl_id would actually be unique. So the action of staging gives us some amount of opportunity to handle any known/expected potential issues with source data. Also, any true requirements of the source data are maybe better handled at time of ingestion rather than build time
3. implemented here, don't do unique column but unique combo of columns. this keeps in line with our current setup of testing source and not staging tables, and this test aligns with how we actually set the unique key for the variable in the staging table (concatenate pwl_id and name). But still don't love this here - the fact that it's testing the unique key of the staging table sort of assumes knowledge of what happens downstream and that's a little odd. But seemed like the simplest way to resolve for now

Very curious on folks' thoughts

### DIFF
--- a/products/green_fast_track/models/_sources.yml
+++ b/products/green_fast_track/models/_sources.yml
@@ -471,7 +471,6 @@ sources:
         columns:
           - name: pwl_id
             tests:
-              - unique
               - not_null
           - name: name
             tests:
@@ -479,6 +478,9 @@ sources:
           - name: wkb_geometry
             tests:
               - not_null
+        tests:
+          - dbt_utils.unique_combination_of_columns:
+              combination_of_columns: [ pwl_id, name ]
 
       - name: nysdec_priority_estuaries
         columns:


fvankrieken:
> The second option makes most sense to me.

Why?

fvankrieken:
> > The second option makes most sense to me.
> 
> Why?

And do you mean in this case specifically, or in general with testing staging tables instead of source tables for build-required checks?

damonmcc:
for now, doing #3 like you did here seems good to align with how we actually use the data

longterm, I like #2

**(assuming there's a staging table for every source that is doing a little as possible)**

I'd love if we only raise errors on staging tables and not on sources. I like treating sources as things we just need to put in the database and can't control, so we might test them but only warn if we do. we'll probably end up repeating ourselves a lot if we test sources because we'd be afraid to not test staging

seems better to enforce expectations on the things we can actually control: the queries in a staging model. if source data changes in a way that matters, either the staging model fails to build or the tests on it fail

damonmcc (approved):

sf-dcp:
> > > The second option makes most sense to me.
> > 
> > 
> > Why?
> 
> And do you mean in this case specifically, or in general with testing staging tables instead of source tables for build-required checks?

In my understanding, modifying a test on source data (use a compound unique key instead of primary key) could potentially result in duplicate data. Sample data below: 

| id | bbl | bin  |
|----|-----|------|
| 1  | 123 | 1111 |
| 2  | 234 | 2222 |
| 2  | 345 | 2222 |

Again, a hypothetical example, where we care for bin column to be unique. 

That's the primary reason I was leaning towards the 2nd solution. And also agree with your and Damon's perspective to test data in staging where we can actually control things.

fvankrieken:
> In my understanding, modifying a test on source data (use a compound unique key instead of primary key) could potentially result in duplicate data. Not saying that it's applicable to this case but in general. Sample data below:
> 
> id	bbl	bin
> 1	123	1111
> 2	234	2222
> 2	345	2222
> Again, a hypothetical example, where we care for bin column to be unique while creating a composite key on id and bbl.
> 
> That's the primary reason I was leaning towards the 2nd solution. And also agree with your and Damon's perspective to test data in staging where we can actually control things.

So as mentioned, the implemented unique compound key test is the actual unique requirement in our pipeline (the current failing test is more just our understanding of how the source data is intended - more focused on the actual unique entities in the source data than the requirement of the pipeline). I don't think we'd ever just introduce a compound key for the sake of arbitrary uniqueness

sf-dcp (approved):

PR #1498, opened 2025-03-05
Use uv to compile, add option to not upgrade all packages

fvankrieken:
Go by commit - commit 1 has a bunch of the lines (just simplified the way it lists any package that was explicitly in requirements.in)

Motivation was that it's maybe a little annoying to add a single package right now.

So now you can run `.admin/ops/python_compile_requirements --no-upgrade`.  When you do pip compile without upgrading, it defaults to what it finds in the output requirements file. We default to upgrading since this is largely for our recurring job to keep upgrading packages on a regular basis, but it's obviously annoying if you're adding one package and then need to change your whole environment.

Also, uv makes this run locally in like 3 seconds. So I think this would be a good way to get away from running that action to compile requirements -> just do it locally and everything will happen just fine

[Action running properly](https://github.com/NYCPlanning/data-engineering/actions/runs/13730081024) (and fast)

#1503 showing that the output of the action looks like we'd expect


admin/run_environment/requirements.txt
```diff
@@ -31,16 +29,16 @@ bcrypt==4.3.0
...
+    #   -r requirements.in
     #   gdown
 blinker==1.9.0
     # via streamlit
 boto3==1.37.6
     # via
-    #   -r /__w/data-engineering/data-engineering/admin/ops/../run_environment/requirements.in
+    #   -r requirements.in
     #   moto
-boto3-stubs[s3]==1.37.6
```
fvankrieken, line 42: Ooh this is actually a problem. Not sure why uv pip compile would drop the optional installation

PR #1508, opened 2025-03-10
Ar fix failing nightly

alexrichey:
the processing steps to ingest COLP were changed to remove a `wkb_geometry` field, and leave the duplicated `geom` field. This cleans out references in two products that were failing nightly QA

PR #1509, opened 2025-03-10
github action maintenance

fvankrieken:
By commit is easier - some autoformatting in commit 1

Main motivation was pluto build failures seen in #1508 - there are times when we don't want these "take-action" failures to actually fail, since we regularly run builds that aren't "production" and don't need to meet the same stringent requirements.

Also
- remove unused postgis service for colp build (seems like just detritis)
- change ci failure for codecov

## Testing test_severity logic

- [Run w/ 'warn'](https://github.com/NYCPlanning/data-engineering/actions/runs/13773396224/job/38517029360) (should pass)
- [Run w/ 'error'](https://github.com/NYCPlanning/data-engineering/actions/runs/13773852992/job/38518537088) (should fail)

### DIFF
--- a/.github/workflows/test_helper.yml
+++ b/.github/workflows/test_helper.yml
@@ -210,7 +210,6 @@ jobs:
       env:
         CODECOV_TOKEN: ${{ env.CODECOV_TOKEN }}
       with:
-        fail_ci_if_error: true
         verbose: true
 
   validate_product_metadata:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,11 @@ on:
         description: "Version of dataset being built (overrides version defined in recipe)"
         type: string
         required: false
+      test_severity:
+        description: "Test severity (should post-build tests cause action to fail?)"
+        type: choice
+        required: true
+        options: [ error, warn ]
       dev_image:
         description: "Use dev image specific to this branch (if exists)"
         type: boolean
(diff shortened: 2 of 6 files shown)


.github/workflows/test_helper.yml
```diff
@@ -210,7 +210,6 @@ jobs:
       env:
         CODECOV_TOKEN: ${{ env.CODECOV_TOKEN }}
       with:
-        fail_ci_if_error: true
```
fvankrieken, line 213: @damonmcc @alexrichey don't remember where our last group discussion fully ended up, it was a while back and I thought we had somewhat settled on not failing pr tests. I am definitely on the side of use codecov as just another thing in a developer/reviewer's toolbox to assess PRs rather than have a binary failure when a certain threshold is not met

alexrichey: I believe the outcome was to not fail builds

damonmcc: yea no fail seems right

sf-dcp:
Q about colp: I'm assuming colp isn't using a database for builds, correct?

alexrichey (approved):
thanks!

fvankrieken:
> Q about colp: I'm assuming colp isn't using a database for builds, correct?

It uses edm-data

sf-dcp:
> > Q about colp: I'm assuming colp isn't using a database for builds, correct?
> 
> It uses edm-data. See https://github.com/NYCPlanning/data-engineering/pull/1509/files#diff-c2e55e8985d0a5957887af56af44b8f4b5f626079709deae6fa2a5ced8ebc0ebR49

Blanking out. Thanks for clarifying

PR #1514, opened 2025-03-11
cbbr refresh for updated source data

### LINKED ISSUE #1512: cbbr refactor for fy26
## The problem

cbbr submissions look a bit different than before - CAPS/AE coordinated on a new form for community boards to submit requests with that should make geocoding much more straightforward. However, pipelines need to be tweaked.

cbbr also has a slightly funny data flow. CAPS (I believe) sends the "raw" request data to OMB, who collect agency responses, and send back the dataset. So essentially we have a "raw" output from the app/caps as well as a slightly mutated dataset with agency responses from OMB, but they're roughly the same data otherwise. Historically, we've just used the OMB data as input to the pipeline. However, it doesn't maintain the easily-geocodable columns in the data we send them.

## Acceptance Criteria
- [ ] our pipeline runs with updated source data
- [ ] our pipeline geocodes as well as CAPS's manually geocoded file, which was also shared with us

## Not In Scope

N/A - pretty self-explanatory

## Solution

Going to go from one source dataset "cbbr_submissions" to two
- "dcp_cbbr_requests" -> the raw csv we get from CAPS (who presumably get it from DH or the app maintained by AE. Longer-term, we should maybe automate this
- "omb_cbbr_agency_responses" -> essentially the same file, but with agency responses (and explanations for them) to the requests

They're linked by a unique id column. The former has the properly formatted location fields which come from geosearch, meaning if a request is location-specific it should be 100% geocode-able. The omb dataset theoretically could be our only dataset (and roughly corresponds to the old "cbbr_submissions", but they seem to squish the location fields together, so easier to use "dcp_cbbr_requests" for geocoding and then join in agency responses

fvankrieken:
Closes #1512.

A little messy but by-commit is the way to go.

Main change/motivation for changes is that the structure of the source data changed. AE/CAPS worked to try to make sure that if people were entering location information, it was entered in a way that could be geocoded. So instead of an address field that we worked our best to parse, it's now either
- geosearch-verified address
- intersection (two streets specified)
- street segment (on street, cross street 1, cross street 2)
- a loose "site name" field
So there are a bunch of sql changes as well as python geocoding changes.

Things fall into a couple buckets
- reworking sql/python for change in source data. We've gone from one source for "submissions"/requests to the internal one generated by AE as well as the one given to us by OMB with agency responses
- moving to a single source for manual mappings
- reworking geocoding a little to try a few new things with the data. There seems to be a case where people did not properly enter intersections, but used the "on/between cross streets" to denote an intersection instead.
- some dcpy bugs


dcpy/connectors/filesystem.py
```diff
@@ -37,7 +37,10 @@ def pull(
         source_path = self._path(key, path_prefix)
         if source_path != destination_path:
             shutil.copy(source_path, destination_path)
-        return {"path": destination_path}
+        if destination_path.is_file():
+            return {"path": destination_path}
+        else:
+            return {"path": destination_path / source_path.name}
```
alexrichey, line 43: ha.... I had just fixed this one in my PR. You beat me to it.

alexrichey (approved):
All looks good!

PR #1518, opened 2025-03-12
Add Lifecycle Connectors Excalidraw source

alexrichey:
So we can edit this later, if we'd like


alexrichey:
> I saw the title and really thought it meant adding a `Connector` for Excalidraw and my disappointment is immeasurable.
> 
> Was very interested to hear your proposal for using excalidraw as a long term data store

Haha, sorry to disappoint. Re long-term vision... don't have one. Just wanted somewhere to put this in case anyone wanted to iterate.

PR #1519, opened 2025-03-12
ingest - implement and make use of connector classes

### LINKED ISSUE #1515: Ingest: implement connector classes for extracting and archiving
## The problem

ingest code is currently very coupled to connector code - `configure.py`, `extract.py`, etc import specific connectors and have logic in ingest specific to retrieving data/metadata from the various sources. With #1463, we have a framework to move this logic into connector classes

## Acceptance Criteria
- [ ] `dcpy.ingest` does not import any specific connectors, only a registry
- [ ] a dataset of each source type passes (using gha, archiving in dev bucket)

## Not In Scope

- issue in #1516. I had initially talked about these being somewhat coupled, but that's a slightly different issue that really needs its own solution, whether it's as simple as not erroring until the env var is actually needed or broader configuration for dcpy

## Solution
- implement something like a "non-versioned pull" ABC connector to handle things like arcgis online, socrata. We still have a way of reflecting the current "version" in these cases, but there's no handling of multiple versions, there's just a single "pull"
- implement connector classes for each of the source types in ingest, as well as passing of kwargs to them.
- implement "push" for the existing edm.recipes connector as well as a new "raw_dataset" edm.recipes connector

fvankrieken:
Closes #1515 

Multiple things big and small going on here. I tried to be quite detailed in commit descriptions, so look there as you go by commit. **Definitely go by commit**, I think it's relatively sane

#### all the connector-related ingest changes are in one commit (9th) - if you're unsure of how the functionality in a specific commit is relevant, I recommend checking that one as a reference.

Main takeaways
- rework connectors a bit, generally
  - different apis that can be available: versioned vs nonversioned, etc
  - move towards kwargs instead of dict object passed in
  - registry "subregistries" so that you can declare what api is needed at various points in codebase
- implement connectors for storage in ingest. For now, this is separate from edm.recipes, which is a little more coupled to being backwards compatible with library
- use connectors in ingest

My biggest open questions
- Resolve `edm.recipes` vs `ingest_datastore` -> should this new connector just go in recipes? Should it become one with the current recipes connector? 
  - arg against -> nice to have clean break. An external user stays further away from any code that has a concept of `library`. Also don't have to worry about refactoring name (ingest_datastore is generic, better to expose slightly to users, whereas `edm.recipes` is very much OUR datastore)
  - arg for -> might be easier to take advantages of the new connector if we just make one and have it backwards compatible. Less room for confusion moving forward with two parallel connectors, one pushing one pulling.
- handling raw filename in ingest (see comment)


dcpy/connectors/registry.py
```diff
@@ -185,9 +192,13 @@ def get_subregistry(self, cls: Type[_C2]) -> ConnectorRegistry[_C2]:
         return ConnectorRegistry(connectors=connectors)
 
     @property
-    def nonversioned(self) -> ConnectorRegistry[NonVersionedConnector]:
-        return self.get_subregistry(NonVersionedConnector)
+    def versioned(self) -> ConnectorRegistry[VersionedConnector]:
+        return self.get_subregistry(VersionedConnector)  # type: ignore[type-abstract]
```
fvankrieken, line 196: these type: ignores are a little bit of a smell. Ongoing discussion around this here - https://github.com/python/mypy/issues/4717. I'd be slightly tempted to ignore this globally. Basically, risk is that mypy is unhappy given that the type constraint is an abstract class but when we use it, we're doing a concrete class... I think.

alexrichey:
@fvankrieken Some really cool stuff here, though this does remind me a lot of the rabbit-hole I went down with the distribution protocols. I eventually started to feel like all the `mypy` warnings were probably a good indicator of potential issues down the road. 

I do see the value, for certain connectors, in forcing implementers to explicitly call out versions. But I think there's probably a cleaner approach that doesn't require any `# type: ignore`s. For example just adding a push_version method (sketched out below)

```python
# A simplified version of what you've got here:
class Connector:
    def push(self, key: str, **kwargs: dict):
        print(kwargs)

class VersionedConnector(Connector):
    def push(self, key: str, version: str, **kwargs):
        print(kwargs)

# The problem is that this breaks inheritance, because 
# VersionedConnector.push must always be usable when Connector.push can be. (ie without `version`) 
# It also subtly changes the meaning of `key` by implying that the version isn't a part of the key.


# Alternate Idea: add `push_version` method
# It would force implementors to implement two different
# push methods, but that's not so bad, as seen in the example
# Publish Connector below. This approach also makes it pretty clear that
# this connector still, at the end of the day, has a key (with version as a part of it)
class VersionedConnector2(Connector):
    def push(self, key: str, **kwargs):
        pass

    def push_version(self, version: str, **kwargs):
        pass


class SocrataConnector(VersionedConnector2):
    def push(self, key: str, **kwargs):
        # this method is really more just a binding from the connector
        # registry kwargs to methods with actual descriptive type signatures
        self._my_actual_push_method(key=key, **kwargs)

    def push_version(self, version: str, **kwargs):
        key = self._version_to_key(version, **kwargs)
        self.push(key, **kwargs)

    def _version_to_key(self, version: str, **kwargs):
        return f"my/key/from/{version}/and/kwargs"

    def _my_actual_push_method(self, key, version, foo: str, bar: dict, baz: int):
        pass
```

Would this work for your use case? Thinking that you could still pull the versioned connectors out of the registry and call .push_version instead of .push.

fvankrieken:
> @fvankrieken Some really cool stuff here, though this does remind me a lot of the rabbit-hole I went down with the distribution protocols. I eventually started to feel like all the `mypy` warnings were probably a good indicator of potential issues down the road.
> 
> ...

Oops never responded online. This is all much more sensible, and gives just the slightest bit more type safety. In progress implementing

dcpy/connectors/registry.py
```diff
@@ -113,32 +113,73 @@ def push_versioned(self, key: str, version: str, **kwargs) -> dict:
...
     def clear(self):
         self._connectors = {}
 
     def list_registered(self) -> list[str]:
         return list(self._connectors.keys())
 
-    def __getitem__(self, item):
-        if item not in self._connectors:
+    @overload
+    def __getitem__(self, item: tuple[str, Type[_C2]]) -> _C2: ...
```
fvankrieken, line 164: This is so that if you have a `ConnectorRegistry[GenericConnector]`, you could do
```python
a: GenericConnector = registry["a"]
a: VersionedConnector = registry["a", VersionedConnector]
```
Types of a here being what mypy would interpret the variables as

dcpy/connectors/registry.py
fvankrieken: These properties exist because they just seem like things that we'd want commonly - get all connectors that implement `Pull`, etc

dcpy/lifecycle/connector_registry.py
```diff
@@ -1,15 +1,26 @@
...
 def _set_default_connectors():
     connectors.clear()
     connectors.register(connector=recipes.Connector())
     connectors.register(connector=publishing.DraftsConnector())
     connectors.register(connector=publishing.PublishedConnector())
+    connectors.register(connector=publishing.GisDatasetsConnector())
+    connectors.register(connector=SocrataConnector())
+    connectors.register(connector=ArcGISFeatureServiceConnector())
+    connectors.register(connector=web.WebConnector())
+    connectors.register(connector=web.WebConnector(), conn_type="api")
```
fvankrieken, line 24: ingest handles "api" and "file download" slightly differently (and they are just slightly different philosophically, even if they're handled by the same code)

fvankrieken: And api might require more tweaks down the road - it's own connector that can add query params etc.

dcpy/connectors/registry.py
```diff
@@ -113,6 +113,19 @@ def push_versioned(self, key: str, version: str, **kwargs) -> dict:
...
+# TODO - this doesn't really belong. If we moved from key/version to more of a path, this could be done away with
+class StorageConnector(Connector, ABC):
+    """A connector that does not version datasets but only stores the "current" or "latest" versions"""
+
+    @abstractmethod
+    def exists(self, key: str) -> bool:
+        return False
+
+    @abstractmethod
+    def get_subfolders(self, prefix: str) -> list[str]:
```
fvankrieken, line 132: I had originally gone with "get_suffixes" -> essentially, give a key and get all paths that follow. What I really needed for ingest was subfolders, which is a little less generalized (as in, this as written doesn't really support something like ingesting straight to a database). But that seems like a subtlety that can be resolved after we continue to iterate on connectors

dcpy/lifecycle/ingest/configure.py
```diff
@@ -175,11 +112,10 @@ def get_config(
 
     archival = ArchivalMetadata(
         archival_timestamp=run_details.timestamp,
-        raw_filename=filename,
+        raw_filename="",  # TODO - this is now set after pulling file
```
fvankrieken, line 115: I have run out of juice here and could use input - I think it makes sense to not necessarily know the filename until pulling, and I don't think it makes a ton of sense to add a function that makes a connector call to determine filename based on key. For now, this is blank here, then gets set after the dataset is pulled (and before it's archived) but this obviously isn't optimal

dcpy/connectors/registry.py
```diff
@@ -44,79 +46,160 @@ def pull(self, source_type: str, arg: _I) -> str:
...
+
+        c = self._connectors[conn_type]
+        if type_validator:
+            assert isinstance(c, type_validator)
+        return c
+
+    def __contains__(self, item):
+        return item in self._connectors
+
+    def get_subregistry(self, cls: Type[_C2]) -> ConnectorRegistry[_C2]:
```
alexrichey, line 189: I think the type-checkers in the implementations are getting legitimately mad here, because you're passing abstract classes where `get_subregistry` wants a concrete one. Playing around with it, if you just do this, I think you're fine to remove the type: ignore's in the implementation 
```python
get_subregistry(self, cls: Type[_C2] | _C2)
```
(might have to add one or two more here, though.

dcpy/connectors/drive.py
alexrichey, line 1: maybe `local_filesystem`? wordier, but a little more concrete.

dcpy/connectors/registry.py
```diff
@@ -44,79 +46,160 @@ def pull(self, source_type: str, arg: _I) -> str:
...
 
-    def __getitem__(self, item):
-        if item not in self._connectors:
+    @overload
+    def __getitem__(self, item: tuple[str, Type[_C2]]) -> _C2: ...
+
+    @overload
+    def __getitem__(self, item: str) -> _C: ...
+
+    def __getitem__(self, item: str | tuple[str, Type[_C2]]) -> _C2 | _C:
```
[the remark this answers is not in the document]
fvankrieken: ty

dcpy/connectors/drive.py
fvankrieken: Yeah I'm with you on `drive` not being ideal - but the rationale was that a shared/network drive path could work for this too (I think? I guess I should make sure shutil can handle that). So a "drive" seemed like the best generic. That being said I still don't like it. Thoughts?

dcpy/connectors/registry.py
fvankrieken: Oh really? I'll poke around a little more

dcpy/connectors/registry.py
```diff
@@ -44,79 +46,160 @@ def pull(self, source_type: str, arg: _I) -> str:
...
+    @abstractmethod
+    def get_latest_version(self, key: str, **kwargs) -> str:
+        pass
 
-    def query_latest_version(self, key: str) -> str:
-        return ""
 
-    def version_exists(self, key: str, version: str) -> bool:
+# TODO - this doesn't really belong. If we moved from key/version to more of a path, this could be done away with
+class StorageConnector(Connector, ABC):
```
alexrichey, line 124: I actually do think something like this might belong here. Though, the other Abstract connectors are more about defining a behavior that must be implemented, and it seems like what this implements is path'ing. So maybe `PathedConnector`?

Just imagining that something like the recipes connector might have Pathing, Keys, Versions, etc... though now we're veering a little more towards protocols.

dcpy/connectors/registry.py
```diff
@@ -44,79 +46,160 @@ def pull(self, source_type: str, arg: _I) -> str:
...
-class VersionedConnectorRegistry:
+class ConnectorRegistry(Generic[_C]):
     """A Registry for VersionedConnectors.
 
     Connectors can be dynamically registered and invoked."""
 
     MISSING_CONN_ERROR_PREFIX = "No registered connector named:"
 
-    _connectors: dict[str, VersionedConnector]
+    _connectors: dict[str, _C]
```
alexrichey, line 147: I do wonder if these should be `type[_C]`. (Totally fine not addressing anything like this now though!)

dcpy/connectors/ingest_datastore.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class Connector(VersionedConnector):
+    conn_type: str
+    _storage: StorageConnector
+
+    def __init__(
+        self,
+        storage_type: str,
+        registry: ConnectorRegistry,
```
alexrichey, line 23: Can we just pass in the StorageConnector instance here?

dcpy/test/connectors/test_drive.py
```diff
@@ -0,0 +1,92 @@
...
+        output_file = tmp_path / "subfolder" / "test2.csv"
+        assert not output_file.exists()
+        self.connector.push(key=str(output_file), filepath=tmp_file)
+        assert output_file.exists()
+
+    # next two are a little duplicated with _path test
+    # but this is the more important functionality to test
+    def test_push_drive_kwarg(self, tmp_file: Path, tmp_path: Path):
+        key = "test2.csv"
+        print(tmp_path)
```
alexrichey, line 80: nit: use logger

dcpy/lifecycle/ingest/configure.py
alexrichey: ha... facing a similar issue with recipe stage configuration, where you won't know the key ahead of time. I think this is fine.

dcpy/connectors/drive.py
alexrichey: gotcha... just `filesystem` then?

alexrichey (approved):
a few nits and questions, but nothing blocking. This is great.

dcpy/connectors/registry.py
damonmcc: don't mind ignoring here or globally, especially if it means more readable code. seems like making mypy happy sometimes has a cost of added complexity

dcpy/connectors/ingest_datastore.py
fvankrieken: Yeah that probably makes more sense.

dcpy/test/connectors/test_drive.py
fvankrieken: Oops - really "nit: ✂️"

dcpy/lifecycle/ingest/configure.py
fvankrieken: I might leave it and then add an assert statement before archival, just to make sure things happened properly

fvankrieken:
> a few nits and questions, but nothing blocking. This is great.

Any thoughts on ingest_datastore vs edm.recipes connector?

dcpy/lifecycle/data_loader.py
```diff
@@ -21,9 +21,11 @@ def pull_dataset(ds: InputDataset, stage: str) -> Path:
         raise Exception(f"Cannot import a dataset without a resolved version: {ds}")
     if ds.file_type is None:
         raise Exception(f"Cannot import a dataset without a resolved file type: {ds}")
+    if ds.source is None:
+        raise Exception(f"Cannot import a dataset with no source: {ds}")
 
     ds.custom["file_type"] = ds.file_type
-    connector = connectors[ds.source]
+    connector = connectors.pull[ds.source]
```
damonmcc, line 28: why the change from `connectors` to `connectors.pull`? just having trouble following, this would be a `ConnectorRegistry.get_subregistry(Pull)[ds.source]`?

fvankrieken: Because otherwise as these types and the registry are currently implemented, mypy would complain. That's why I mention that maybe too stringent -> might make sense to assume that we can `pull` and `push` and then more use subregistries to either limit behavior, or allow random more specific usages. 

But basically yeah, without getting that subregistry (which you are correct in terms of what this is short for), mypy wouldn't know that the returned connector implements the `Pull` api. So first, we get the connectors which do implement pull, and then look for the one we need. It's definitely at least a tad awkward.

fvankrieken: It might be a little clearer if we went the route of `connectors[ds.source, Pull]`... but then you have to import `Pull` everywhere which is annoying.

alexrichey:
> > a few nits and questions, but nothing blocking. This is great.
> 
> Any thoughts on ingest_datastore vs edm.recipes connector?

Oh, sorry, I did have thoughts, yes. I assume that the end-state is that in the pyproject, there will probably a tool entry for `ingest.datastore_connector` or something like that, that defaults to a generic connector like what you put together in `ingest_datastore`. Definitely nice that it's not edm recipes. So I like this approach. And maybe we eventually just kill the edm.recipes one?

dcpy/connectors/registry.py
fvankrieken: I don'tt hink that's right - I'm getting errors - the `type[_C2]` doesn't really have much to do with abstract vs concrete. The difference in usage is more this

```
class A(BaseModel):
    attribute: str

a = A(attribute="a")

T = TypeVar("T", bound=BaseModel)

def my_func(a: T) -> T:
    ...

obj = my_func(a)

def my_func_2(a: type[T]) -> T:
    ...

obj = my_func(A)
```

So here I don't want `_C2` to be an option - you shouldn't be able to pass an instance of a storageconnector. But you can provide `StorageConnector`, the (uninstantiated) class type to the function

dcpy/connectors/registry.py
fvankrieken: Per above comment, it should not

dcpy/connectors/registry.py
fvankrieken: Gonna punt this for the moment until we come back to any thoughts about paths implemented more generally

dcpy/connectors/registry.py
alexrichey: FYI, [this](https://github.com/NYCPlanning/data-engineering/pull/1583/commits/f2736f5b7c09a95c0b773a7ea06293008dce5dfb) is what I mean. (apologies, it's on my branch, rebased on this branch, but that shouldn't effect anything)

alexrichey: Also, you could probably get rid of the # type: ignores that I added, but I didn't futz long enough. But TLDR: I don't think `type`[...] or `Type`[...] means you're passing an instance, but rather a concrete class. At least that's my reading of the Type docs

```
    builtins.type or typing.Type can be used to annotate class objects.
    For example, suppose we have the following classes::

        class User: ...  # Abstract base for User classes
        class BasicUser(User): ...
        class ProUser(User): ...
        class TeamUser(User): ...

    And a function that takes a class argument that's a subclass of
    User and returns an instance of the corresponding class::

        def new_user[U](user_class: Type[U]) -> U:
            user = user_class()
            # (Here we could write the user object to a database)
            return user

        joe = new_user(BasicUser)
```

fvankrieken: So this here https://github.com/NYCPlanning/data-engineering/pull/1583/commits/f2736f5b7c09a95c0b773a7ea06293008dce5dfb#diff-63057eb582411f6c1cbc08dcb5ce3b3b580cd4c7a0342baab5204d9c4699c0edR193 would very much error if an arg of type `_C2` rather than `type[_C2]` was provided. Let's just chat in person

dcpy/lifecycle/ingest/validate.py
```diff
@@ -7,15 +6,7 @@
...
-class ArchiveAction(StrEnum):
-    push = "push"
-    update_freshness = "update_freshness"
-    do_nothing = "do_nothing"
-
-
-def validate_against_existing_versions(
-    ds: recipes.Dataset, filepath: Path
-) -> ArchiveAction:
+def validate_against_existing_versions(ds: recipes.Dataset, filepath: Path) -> bool:
```
sf-dcp, line 9: can we rename this function to something like `is_new_version` and also update the doctsring where it says that the config file will be updated in line 16? From my understanding, the config file will no longer record the date checked.

fvankrieken: Oh wow the docstring is way inaccurate. Yeah will update. 

Maybe `compare_to_existing_version`? `is_equivalent_to_existing_version`? I think the former maybe

fvankrieken: You know what while I'm add it let me go through the whole PR and add docstrings where appropriate

sf-dcp (approved):
Reviewed at high level -- LGTM!

Issue #1438, opened 2025-02-10
Open Data Week 2025: ingestion


damonmcc:
Demo repo here: https://github.com/NYCPlanning/open_data_ingest_demo

Slides here (request access): https://docs.google.com/presentation/d/1-sS4y0AuAw6-7jbvCHwtL28ybe0K9G84GkTfVFFG5ak/edit?usp=sharing

## Shared Todos

See #1437 

## Ingest Todos
- [ ] ~#1515~ making out of scope for this
- [x] #1516
- [x] acl -> optional (required only when interacting w/ s3
- [x] ergonomics around cli (default to not interacting w/ s3)
- [ ] ~~nice-to-have version of above: configure storage as dir vs s3~~

## Demo Repo Todos
- [ ] https://github.com/NYCPlanning/open_data_ingest_demo/issues/4
- [x] https://github.com/NYCPlanning/open_data_ingest_demo/issues/3

## Presentation Todos (some overlap with Shared Todos)
- [x] Slide formatting
- [x] shared dcpy.lifecycle slides
- [x] shared "get in touch" slides
- [x] "the problem" slides
- [ ] "the answer: ingest" slides
- [ ] "how it's helped us" slides
- [x] "roadmap" slides

fvankrieken:
I created a discussion a while back for requirements to get `dcpy.ingest` somewhat ready for users

https://github.com/NYCPlanning/data-engineering/discussions/955

fvankrieken:
Basic idea of pres structure from me and @sf-dcp today

1. brief intro to dcpy.lifecycle. Thought that this could be some nice SYNERGY between the talks (like even have the same slides). Briefly discuss the python work we've been doing the last year and a half and why, maybe some diagram of the "product lifecycle", then zoom in to our specific parts
2. The problem that ingest solves
  - include notes on existing tooling - fivetran etc
3. Intro to the tool!
  - what it does at a high level (i.e. "python cli that solves the above problem"
  - who it's for
  - some picture of long term vision?
 4. Demo!!
   - from zero to data archived -> write simple yaml template live
   - 4-5 templates? From different socrata websites, arcgis online feature servers
 5. The benefits we see from it
 6. Roadmap? Next steps, stuff to pull people in
   - data checks
   - geocoding as preprocessing

PR #1523, opened 2025-03-14
make s3 interaction opt-in for ingest

fvankrieken:
originally meant to address #1516, but after starting to implement, think that will be easier post connector stuff after allsince that will simplify interactions between ingest and recipes just a little bit

Instead, just addresses smaller points in todos in #1438
 - default to not interact with s3 and only validate dataframe against s3 if flag is present
 - make acl values optional for this reason (only need acl if we're pushing to a server, otherwise it's just clutter for users)

### DIFF
--- a/dcpy/lifecycle/ingest/__init__.py
+++ b/dcpy/lifecycle/ingest/__init__.py
@@ -20,7 +20,7 @@ def _cli_wrapper_run(
     latest: bool = typer.Option(
         False, "-l", "--latest", help="Push to latest folder in s3"
     ),
-    skip_archival: bool = typer.Option(False, "--skip-archival", "-s"),
+    push_to_s3: bool = typer.Option(False, "--s3", "-s"),
     csv: bool = typer.Option(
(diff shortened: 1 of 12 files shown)


dcpy/lifecycle/ingest/__init__.py
```diff
@@ -20,7 +20,7 @@ def _cli_wrapper_run(
     latest: bool = typer.Option(
         False, "-l", "--latest", help="Push to latest folder in s3"
     ),
-    skip_archival: bool = typer.Option(False, "--skip-archival", "-s"),
+    push_to_s3: bool = typer.Option(False, "--s3", "-s"),
```
damonmcc, line 23: would be great to rename `--s3` to `--push_to_s3` and use the full name in the github actions instead of `-s`. the extra characters are worth the clarity

sf-dcp (approved):
Looks great! Agree with @damonmcc  to rename and use full arg in CLI

PR #1526, opened 2025-03-18
don't apply codecov status to ci check

fvankrieken:
Need to remove dummy commit, but needed to demonstrate.

I think this looks how I'd actually like it - we still get codecov commenting its report, but CI isn't failed

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,0 @@
-coverage:
-  status:
-    project: off
-    patch: off
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,4 @@
+coverage:
+  status:
+    project: off
+    patch: off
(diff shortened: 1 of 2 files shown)


codecov.yml
fvankrieken, line 1: don't love needing to add a file for this, but doesn't seem like we can do pyproject.toml

damonmcc: no worries. and now that we need this file we might prefer to move all codecov stuff to it (later)

damonmcc (approved):

codecov.yml
fvankrieken: What other codedov stuff do you mean?

alexrichey (approved):
great, thanks

fvankrieken:
Alright given that all we're doing is modifying default behavior, did so in codecov in our account - for this type of change, I think that made sense. If we want to get more specific, we can have a codecov.yml

CI passed with the yml reverted, so we're good

Issue #1557, opened 2025-04-02
CEQR Data Hub datasets for internal review


damonmcc:
There are 13 analytical chapters ([Data Sources excel](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/cp/_layouts/15/Doc.aspx?sourcedoc=%7Be9735a2d-c211-4f87-b0b5-095fe8bbc7ff%7D&action=edit&activeCell=%27DE%20datasets%27!I4&wdinitialsession=6730c478-00c0-9373-1614-7a06a69ca79a&wdrldsc=3&wdrldc=1&wdrldr=AccessTokenExpiredWarningGatekeeperCookieMismatch%2C) in sharepoint). We plan to provide links to download/navigate to all datasets for first 5 chapters for the initial review ("Friends and Family").

The sheet called DE Datasets in that excel file is the source of truth for DE.

Adding datasets to excel sheet 

## chapters completed
- [ ] Analysis Framework
- [ ] Land Use
- [ ] Socioeconomic Conditions
- [ ] Community Facilities
- [ ] Open Space

PR #1558, opened 2025-04-02
add and use more data in CEQR Data Hub

damonmcc:
related to #1557


dcpy/library/templates/dcp_wrp_coastal_zone_boundary.yml
fvankrieken, line 1: Is there a reason to not do ingest? feature server is implemented, and doesn't mutate the data like library does (just column name cleaning that happens by default in library, so not the biggest issue).

products/ceqr/build_scripts/build_markdown.py
fvankrieken, line 1: Is the file generated by this pointed at directly by the github pages site in s3? Or does it need to be copied and checked in?

damonmcc: needs to be copied and checked in (see PR description)

fvankrieken: Gotcha - yeah I just did some googling and the former doesn't exactly seem possible, at least hard to find docs asserting so.

I don't love the fact that the pipeline is thinking about design decisions - in general I think it's easier to maintain/enhance these types of things when the pipeline doesn't have to worry about frontend representation and then when making tweaks to the UI, a developer doesn't have to go look at pipeline code.

That being said, those design decisions being made here are pretty minimal. This is nifty and it works.

Question though, do these outputs need much from the build itself? As in, with these joins happening here is there actual build info joined? Or just manipulations of the seed files? If they're more static, it might make some sense in the future for them to just live in the github pages repo, similar to reasoning above. Just curious - no changes requested.

fvankrieken: Nit on this file - many mentions of "markdwon"

sf-dcp: Socrate moment

sf-dcp (approved):
@damonmcc walked me through the functionality and it looks good to me. Only nit/question why not using `ingest` for the 4 sources.

products/ceqr/build_scripts/build_markdown.py
fvankrieken: 🤌

dcpy/library/templates/dcp_wrp_coastal_zone_boundary.yml
damonmcc: just that the two `dcp_wrp_*` templates we already have are library ones so I wen't with consistency + move them all at once later instead of a mix of library/ingest for the same feature server (`server: dcp, name: WRP`)

the two existing ones are `dcp_wrp_recognized_ecological_complexes` and `dcp_wrp_special_natural_waterfront_areas`

fvankrieken: Oh yeah I forgot you mentioned this, sorry

fvankrieken (approved):

Issue #1560, opened 2025-04-07
COLP 2025-03 QA Round 1


fvankrieken:

fvankrieken:
See https://github.com/NYCPlanning/edm-overview/issues/1125#issuecomment-2779710677

PR #1563, opened 2025-04-08
COLP 2025-03

fvankrieken:
follow up to #1560 

- update ipis_modified_names to add flags for records that have been reviewed. This file is essentially an output of the build -> all rows that have had parcelname cleaned programmaticaly. GIS then looks through and verifies that they seem alright, and we add a flag for those rows that they have approved.
  - I also added logic to sort the output - hopefully in the future we don't have quite as many "changed rows" in the git diff because of this
- add a manual correction for overzealous programmatic cleaning. It seems like there's precedence for this (see row above new row)
- sqlfluff colp, because we never did!


damonmcc:
nice. when merging, let's rebase instead of squashing since these are so distinct

damonmcc (approved):

PR #1573, opened 2025-04-08
CSCL ETLC POC - add atomic polygon fields

### LINKED ISSUE #1568: CSCL ETL POC - Atomic Polygon-based fields
The standard workflow for these cscl chapters is pretty consistent, just with different sections in the documentation and different fields needed in the final product table
- [x] Source data (ingest templates, recipe/dbt setup) added as needed if specified in 2.2.3
- [x] intermediate tables added to perform transformations outlines in 2.2.3
- [x] consider/implement constraints/tests of table (i.e. 1-1 rows with centerline, etc)
- [x] final product table modified with int table joined in to align with Table 3 for relevant fields:
  - L19-20, L25-29, L34-36, L72-73

fvankrieken:
addresses #1568, though some questions remain there. See issue for description of problem and implementation questions.

There are also a few weird data things - see comments in `int__centerline_offsetpoints`

build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14361781898)

Before any more of these issues are tackled, I'm going to flush out the project in GH especially around validation - that's probably worth focusing on before getting too far into any specific transformation. But it seemed like this and the node-based joins were the big "prickly" transformations, so just wanted to make sure that they were going to be feasible and not too hard (to write the code, or in terms of runtime)

### DIFF
--- a/products/lion/models/intermediate/2.2.3/int__centerline_offsets.sql
+++ b/products/lion/models/intermediate/2.2.3/int__centerline_offsets.sql
@@ -1,0 +1,44 @@
+-- int__centerline_offsets
+-- takes centerline segment, finds adjoining atomic polygons
+-- method in docs is 
+--     "To determine the given segment’s left and right APs, the ETL tool will generate 
+--     temporary left and right “offset points” positioned on both sides of the segment,
+--     offset two feet perpendicularly from the segment’s midpoint"
+-- the current method is an approximation. See comments in #1568
+-- TODO: decide on current method or appropriate replacement
+
+{{ config(
+    materialized = 'table'
+) }}
+
+WITH centerline AS (
+    SELECT
+        segmentid,
+        st_linemerge(geom) AS geom
+    FROM {{ ref("stg__centerline") }}
+    -- TODO remove this -> likely from odd source data.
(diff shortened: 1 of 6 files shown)


products/lion/models/intermediate/2.2.3/int__centerline_offsets.sql
```diff
@@ -0,0 +1,44 @@
...
+WITH centerline AS (
+    SELECT
+        segmentid,
+        st_linemerge(geom) AS geom
+    FROM {{ ref("stg__centerline") }}
+    -- TODO remove this -> likely from odd source data.
+    WHERE segmentid NOT IN (
+        354785, -- Large circular segment extending into other states
+        9008702 -- Strange small simple line that is discontinuous
+    )
```
damonmcc, line 23: nit: could do this in the staging model?

damonmcc (approved):

products/lion/models/intermediate/2.2.3/int__centerline_offsets.sql
fvankrieken: I did that originally - but figured these segments might not produce errors in other processing steps, so just filter them out where they were producing errors

PR #1578, opened 2025-04-11
Refactor Build Publishing and Implement for PFF

alexrichey:
## TLDR
This Implements:
1) build directory and upload changes for PFF
2) recipe changes to enable stage confs, with env vars. 

I'd start by glancing at the recipe files for both the acs.yml recipe and the test recipes. That should give a sense of the changes here. 

Also, apologies, commits are not atomic. After having to rebase a few times, I lost the patience to keep things tidy.

### PFF
defaults to dumping data in the configured lifecycle directory
image

This changes to call ACS data files `current` and `previous` so that dataset names don't change over time. (at least not until 2030, when there's a new census). We should probably also add JSON file with info about which years are included in these files. 

### Recipe Changes
This is a first step towards declaring build variables upfront, so that we can accomplish [this](https://github.com/NYCPlanning/data-engineering/discussions/1575). There's potentially a little more planning to do, but for this side of things, I'd start reviewing the test recipes.

### DIFF
--- a/dcpy/lifecycle/builds/plan.py
+++ b/dcpy/lifecycle/builds/plan.py
@@ -164,7 +164,7 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
             )
 
     # Resolve any unresolved conf values (e.g. from the environment)
-    for conf in recipe.get_unresolved_conf_values():
+    for conf in recipe.get_unresolved_stage_config_values():
         if "env" in conf.value_from:
             env_var = conf.value_from["env"]
             if env_var not in os.environ:
--- a/dcpy/lifecycle/builds/plan.py
+++ b/dcpy/lifecycle/builds/plan.py
@@ -88,20 +88,21 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
         recipe.vars["VERSION_TYPE"] = recipe.version_type.value
 
     # Determine previous version
-    try:
-        previous_recipe = publishing.get_previous_version(
-            product=recipe.product, version=recipe.version
-        )
-        logger.info(
-            f"Previous version of {recipe.product}: {previous_recipe.label} ({previous_recipe})"
-        )
-        recipe.vars["VERSION_PREV"] = previous_recipe.label
-    except (
-        LookupError,
-        ValueError,
-        TypeError,
-    ) as e:  # versions not found, or don't parse correctly
-        logger.error(f"Error: {e}")
+    if "VERSION_PREV" not in recipe.vars:
+        try:
+            previous_recipe = publishing.get_previous_version(
+                product=recipe.product, version=recipe.version
+            )
+            logger.info(
+                f"Previous version of {recipe.product}: {previous_recipe.label} ({previous_recipe})"
+            )
+            recipe.vars["VERSION_PREV"] = previous_recipe.label
+        except (
+            LookupError,
+            ValueError,
+            TypeError,
+        ) as e:  # versions not found, or don't parse correctly
+            logger.error(f"Error: {e}")
 
     # Add vars to environ so both can be accessed in environ
     logger.info(f"Export envars: {recipe.vars}")
@@ -161,6 +162,14 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
             ds.file_type = ds.file_type or recipes.get_preferred_file_type(
                 ds.dataset, RECIPE_FILE_TYPE_PREFERENCE
             )
+
+    # Resolve any unresolved conf values (e.g. from the environment)
+    for conf in recipe.get_unresolved_conf_values():
+        if "env" in conf.value_from:
+            env_var = conf.value_from["env"]
+            if env_var not in os.environ:
+                raise Exception(f"build.plan requires missing env var: {env_var}")
+            conf.value = os.environ[env_var]
 
     return recipe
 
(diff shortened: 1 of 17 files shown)


dcpy/lifecycle/builds/plan.py
```diff
@@ -88,20 +88,21 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
...
-            f"Previous version of {recipe.product}: {previous_recipe.label} ({previous_recipe})"
-        )
-        recipe.vars["VERSION_PREV"] = previous_recipe.label
-    except (
-        LookupError,
-        ValueError,
-        TypeError,
-    ) as e:  # versions not found, or don't parse correctly
-        logger.error(f"Error: {e}")
+    if "VERSION_PREV" not in recipe.vars:
```
alexrichey, line 91: Just enabling manually setting the previous version (for convenience, in my case.)

dcpy/models/lifecycle/builds.py
```diff
@@ -64,6 +65,31 @@ class RecipeInputs(BaseModel):
...
 
 
+class StageConfValue(BaseModel, extra="forbid"):
+    UNRESOLVABLE_ERROR: ClassVar[str] = (
+        "Stage Conf Value requires either `value` or `value_from`"
+    )
+
+    name: str
+    value: str | None = None
+    value_from: Optional[dict[str, str]] = None
```
fvankrieken, line 75: nittiest of nits sorry - here and below you have Optional instead of `| None`, let's stick with the latter for consistency?

dcpy/models/lifecycle/builds.py
```diff
@@ -64,6 +65,31 @@ class RecipeInputs(BaseModel):
...
+
+    @model_validator(mode="after")
+    def check_resolvable(self) -> Self:
+        if not self.value and not self.value_from:
+            raise ValueError(self.UNRESOLVABLE_ERROR)
+        return self
+
+
+class StageConf(BaseModel, extra="forbid", arbitrary_types_allowed=True):
```
fvankrieken, line 84: I'm all nits today - can this and other (var) references also be "config" instead of "conf" for consistency?

fvankrieken: Seems like that's just these two pydantic models actually

alexrichey:
@fvankrieken nits resolved [here](https://github.com/NYCPlanning/data-engineering/pull/1578/commits/d5efc653e36b721571d0981b0810e3a8205775a5). Appreciate the thoroughness!

alexrichey:
> In terms of next steps/where to go from here

@fvankrieken Yeah, this needs some iterating to get closer to the grand vision. This was all I could really justify, under the auspices of working on PFF. The next phase is probably having stringing together plan/load/build/upload using each phase's output files, but this a definite step closer. Would be good to brainstorm a little bit... I think there's a way to do this that actually reduces the complexity of the entire process, rather than increasing it.

damonmcc:
got any PPF builds to link to showing it works with these changes?

dcpy/models/lifecycle/builds.py
```diff
@@ -73,13 +99,26 @@ class Recipe(BaseModel, extra="forbid", arbitrary_types_allowed=True):
...
+        return (
+            self.version is not None
+            and (
+                len(self.inputs.datasets) == 0
+                or len([x for x in self.inputs.datasets if not x.is_resolved()]) == 0
+            )
+            and not self.get_unresolved_conf_values()
         )
 
+    def get_unresolved_conf_values(self):
```
damonmcc, line 114: this seems more like `get_unresolved_stage_conf_values`, especially since the results of this are only a subset of what determines if a Recipe `is_resolved`.  would a rename make sense?

damonmcc (approved):
agree with @fvankrieken's thoughts. changing `conf` -> `config` is worth it in all cases, thanks for doing that for the new changes! would be great to change the existing uses of `conf` in the future

dcpy/lifecycle/builds/plan.py
```diff
@@ -162,6 +163,14 @@ def plan_recipe(recipe_path: Path, version: str | None = None) -> Recipe:
                 ds.dataset, RECIPE_FILE_TYPE_PREFERENCE
             )
 
+    # Resolve any unresolved conf values (e.g. from the environment)
+    for conf in recipe.get_unresolved_conf_values():
+        if "env" in conf.value_from:
```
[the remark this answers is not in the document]
alexrichey: yeah, great question...  this was probably mostly me just reflexively copying the K8s configmaps [pattern](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) but I could see using value_from `dcpy.lifecycle.config`, or references to fields on the recipe itself, e.g. suppose you wanted to pass the version to the connector_args:
```yaml
- name: version
  value_from:
    recipe: "version"
```
or 
:
```yaml
- name: previous_version
  value_from:
    recipe: "vars.VERSION_PREV"
```

dcpy/models/lifecycle/builds.py
alexrichey: very much so. Changed [here](https://github.com/NYCPlanning/data-engineering/pull/1578/commits/3c45b7542ad0805e91a7efe915366823a7654653)

alexrichey:
> got any PPF builds to link to showing it works with these changes?

@damonmcc not yet, I'd just uploaded test files for PFF (the dataset is fairly large) and there's not yet anything integrated into existing builds. Come to think of it, we should probably add it to template-db

PR #1588, opened 2025-04-17
don't use pydantic when checking dataset version

fvankrieken:
https://github.com/NYCPlanning/data-engineering/actions/runs/14524485082/job/40752905845

Issue is that I changed some of the string literals that pydantic uses to determine what "type" a "source" is... so now when recipes attempts to read an older config for one of these datasets, it's unable to determine the type.

Two solutions for this sort of thing
- have allowed "aliases" in pydantic for these fields -> let's us change the field names but allow backwards compatibility
- don't read config as pydantic model when doing simple things like getting version

The second is what I've done here. The first does seem preferable in some ways. But with registerable connectors, we're maybe going to be moving away from having strict constraints on these fields anyways. And if all we ever really need is a version that we know is going to be there even if the models change, it's nice to keep that function as simple as possible.

Going to merge for now to unblock Damon but this is certainly something we could discuss and I'd be happy to make tweaks

### DIFF
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -113,13 +113,18 @@ def set_latest(key: DatasetKey, acl):
     )
 
 
-def get_config(name: str, version="latest") -> library.Config | ingest.Config:
+def get_config_obj(name: str, version="latest") -> dict:
     """Retrieve a recipe config from s3."""
     bucket = _bucket()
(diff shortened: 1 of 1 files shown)


fvankrieken:
Another solution would be to have something like a `SparseConfig` pydantic model that just has the barebone fields you'd need for pulling data, and we'd have to enforce that this model doesn't change. Maybe overly complicated but you know I love my type safety

Issue #1585, opened 2025-04-17
CSCL ETL POC - Final product table and .dat export


fvankrieken:
The relevant portion of the [etl docs](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BB3C7B8CF-45B2-4C01-BA36-4ED2FD0BA25E%7D&file=ETL_V7.doc&action=default&mobileredirect=true) is section 2.1, which contains table 2. This contains the formatting of fields.

In sql/dbt, we should
- ensure schema
- ensure formatting of fields
  - length of fields (lpadding with zeroes or spaces as defined in docs)
  - handling of nulls
  - possible values of fields

Then, we generate the .dat file. This is a headerless text file, no delimiter. Columns/fields are determined based on positioning in the line (characters 8-9 are special district flag, that sort of thing). Not quite sure the best way to go about this 
- if postgres's `COPY` can be used, that might be simplest
- we can always use python, but then we might need to validate/format fields both in dbt/sql and in python, which seems like it will be frustrating and wasted effort

Then, we need to validate outputs. So as I write this, maybe this ticket should include the more "general" qa aspects of #1577, leaving that issue more for validating against [existing outputs](https://nyco365.sharepoint.com/sites/NYCPLANNING/itd/edm/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FNYCPLANNING%2Fitd%2Fedm%2FShared%20Documents%2FDOCUMENTATION%2FGRU%2FCSCL%2FETL%2F25A%20GDB%20and%20ETLs&CT=1744725340076&OR=OWA%2DNT%2DMail&CID=62a2af4c%2D86c8%2Dd075%2Ddb77%2Df0e96f6b6c7b&csf=1&web=1&e=yYv8al&clickParams=eyJYLUFwcE5hbWUiOiJNaWNyb3NvZnQgT3V0bG9vayBXZWIgQXBwIiwiWC1BcHBWZXJzaW9uIjoiMjAyNTA0MDQwMDUuMTEiLCJPUyI6Ik1hYyBPUyBYIHVuZGVmaW5lZCJ9&cidOR=Client&FolderCTID=0x0120001114839845564543AB5C3DCB09500AC5)

For reference in those existing outputs, any of the `{BOROUGH}.dat` are the ones we're talking about.

fvankrieken:
[...]

PR #1589, opened 2025-04-17
start lion export

damonmcc:
related to https://github.com/NYCPlanning/data-engineering/issues/1585

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-lion-exports)

Screenshot 2025-04-17 at 10 22 58 PM

### DIFF
--- a/products/lion/bash/export.sh
+++ b/products/lion/bash/export.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+source ../../bash/utils.sh
+set_error_traps
+
+# Export tables to DAT files
+# fixed-width text files with no header or delimeter are specific to LION
+function dat_export {
+    local table=${1}
+    local output_file=${2:-${table}}
+    run_sql_command \
+        "\COPY (\
+            SELECT * FROM ${table}\
+        ) TO STDOUT;">${output_file}.dat
+}
+
+echo "Export LION outputs"
+rm -rf output
+mkdir -p output 
+(
+    cd output
+
+    echo "Export LION dat files"
+    dat_export manhattan_lion_dat ManhattanLION &
+    wait
+    
+    echo "Export source data versions and build metadata"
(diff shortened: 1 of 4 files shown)


products/lion/bash/export.sh
```diff
@@ -0,0 +1,36 @@
...
+    cd output
+
+    echo "Export LION dat files"
+    dat_export manhattan_lion_dat ManhattanLION &
+    wait
+    
+    echo "Export source data versions and build metadata"
+    csv_export source_data_versions
+    cp ../build_metadata.json ./
+    echo "${VERSION}" > version.txt
```
fvankrieken, line 29: ✂️ , let's not keep this norm. Also, let's just `cp` in source_data_versions.csv rather than exporting it?

products/lion/models/product/export/manhattan_lion_dat.sql
```diff
@@ -0,0 +1,10 @@
+WITH manhattan AS (
+    SELECT *
+    FROM {{ ref('lion_dat') }}
+    WHERE
+        "Borough" = '5'
```
fvankrieken, line 5: that ain't manhattan!

products/lion/bash/export.sh
damonmcc: the norm you're talkin about is version.txt right?

products/lion/models/product/export/manhattan_lion_dat.sql
damonmcc: gotta love boro codes. also this should become a macro

damonmcc:
@fvankrieken changes made and latest build passed

products/lion/bash/export.sh
fvankrieken: Si

fvankrieken (approved):

PR #1600, opened 2025-04-20
deprecate dcp_mappluto recipes dataset

damonmcc:
these changes deprecate a redundant source dataset

## motivation

after [nightly qa failed](https://github.com/NYCPlanning/data-engineering/actions/runs/14545957020/job/40811443416) because dcp_pluto 25v1.1 wasn't archived, I ran ingest/library for the four pluto datasets: `dcp_pluto`, `dcp_mappluto`, `dcp_mappluto_wi`, `dcp_mappluto_clipped`

but two of those are identical: [`dcp_mappluto`](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/templates/dcp_mappluto.yml) and [`dcp_mappluto_wi`](https://github.com/NYCPlanning/data-engineering/blob/main/dcpy/library/templates/dcp_mappluto_wi.yml). they both pull `nyc_mappluto_[VERSION]_unclipped_shp.zip` from Bytes

seems better to keep the more explicit `wi` and `clipped` . this would align with how the data is presented on the [DCP website](https://www.nyc.gov/content/planning/pages/resources/datasets/mappluto-pluto-change)

## details on the changes

DevDB is the only data product that has `dcp_mappluto` in it's recipe file. successful build on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-resolve-pluto)


fvankrieken:
One other reference that would be nice to scrub: `ingest_single` workflow - default dataset_id is "dcp_mappluto"

fvankrieken (approved):

Issue #1611, opened 2025-04-21
Update ingest/library bytes links


fvankrieken:
With the new website, many links have changed. We need to go through templates and update all links (see tweaks in #1544 for example)

PR #1439, opened 2025-02-10
facdb ingest sources, cont

fvankrieken:
closes #1432. Done as part of #1334 

Forgot about this not being merged, I think I was going to do more but lost steam. Not going to migrate any more as part of this build

PR #1612, opened 2025-04-21
update colp source link

fvankrieken:
part of #1611 

https://github.com/NYCPlanning/data-engineering/actions/runs/14576778171

[successful facdb run](https://github.com/NYCPlanning/data-engineering/actions/runs/14576920433) with newly archived data

### DIFF
--- a/dcpy/utils/data.py
+++ b/dcpy/utils/data.py
@@ -92,6 +92,11 @@ def read_data_to_df(
         case file.Csv():
             df = pd.read_csv(
                 local_data_path,
+                index_col=False,
+                encoding=data_format.encoding,
+                delimiter=data_format.delimiter,
+                names=data_format.column_names,
+                dtype=_get_dtype(data_format.dtype),
             )
             gdf = (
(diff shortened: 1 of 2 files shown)


damonmcc (approved):

fvankrieken:
Added another commit - there was a bug introduced by me in #1439 

Need to think/dig for a moment to see if this affected any other archived datasets

fvankrieken:
Okay - re-archived COLP with the bug fix, and it seems good.

I dug through every ingest dataset that specifies `csv` filetype, and checked
- if they specify any of the kwargs that were not being used per the bug
- if they were archived since 3/10 (date that bug was merged into main)

And there were none.

Funnily, https://github.com/NYCPlanning/data-engineering/issues/1524 seems like it was caused by this - edes csv was being pulled but didn't match what was already archived with the same version. Makes sense if ingest suddenly stopped reading in columns with the right dtype

alexrichey (approved):

PR #1615, opened 2025-04-21
Fvk lion export

fvankrieken:


products/lion/macros/format_text.sql
fvankrieken, line 1: I think I will delete this. Just wanted to leave as an option

products/lion/macros/create_format_text_proc.sql
fvankrieken, line 1: Rationale for this being a pg function and not a dbt macro is that it's really specific in scope.

dbt macros are great for things that are sort of one-level-up from a pg function (like coalesce, any of our geom functions, etc). If we want to reference another table, or iterate over many fields, etc. But this is a case where I really just needed to format text. It's a lot nicer to have the calls to this (see `lion_dat_fields.sql`) not use jinja syntax, and instead directly reference columns rather than pass along string references in jinja to be re-substituted.

Or in simpler words - if all we're doing is processing text without needing fancy jinja logic or joining other tables, a pg function seemed like the most ergonomic way to handle it

fvankrieken: By having it as a macro and then adding the line to dbt_project, this gets created when the project first runs

PR #1616, opened 2025-04-21
add USFWS critical habitat tempalte

damonmcc:
resolves https://github.com/NYCPlanning/data-engineering/issues/1609


dcpy/utils/data.py
```diff
@@ -88,6 +89,7 @@ def read_data_to_df(
                 layer=data_format.layer,
             )
         case file.GeoJson():
+            pyogrio.set_gdal_config_options({"OGR_GEOJSON_MAX_OBJ_SIZE": 0})
```
sf-dcp, line 92: what does this do?

sf-dcp:
I'm seeing 17 records under the natural resources sheet of the excel where 1 isn't available, so 16 total. There are 14 records added in this PR. Just confirming that it's expected.

dcpy/utils/data.py
sf-dcp: After we delete library from `dcpy`, I don't think we have `gdal` used anywhere in `dcpy`. @fvankrieken, is that right?

damonmcc: this prevented an arcserver error related to pulling large geometries via API. but I'll double check that this is still needed because the API error may have been because I was polling it too frequently

if this is still needed I'll add a comment to this line

damonmcc:
> I'm seeing 17 records under the natural resources sheet of the excel where 1 isn't available, so 16 total. There are 14 records added in this PR. Just confirming that it's expected.

yup that's expected. there are 14 rows in DE Chapter Datasets sheet which these changes should align with.

There's a discussion via comments in the original natural resources sheet and I've been erring on the side of not modifying their original sheets and using our sheets as the "final drafts"

dcpy/utils/data.py
damonmcc: so [here's](https://github.com/NYCPlanning/data-engineering/actions/runs/14581439480/job/40898836202#step:6:302) ingest failing due to frequent API calls to a feature layer with big geometries

and [here's](https://github.com/NYCPlanning/data-engineering/actions/runs/14581439480/job/40899947412) a re-run of that action (no code changes) 30 minutes later failing with a `pyogrio` error: `Failed to read GeoJSON data`. I tried [this solution](https://stackoverflow.com/questions/79162109/geopandas-fails-reading-geojson-solved) and it worked on this run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14581903790)!

damonmcc: > After we delete library from `dcpy`, I don't think we have `gdal` used anywhere in `dcpy`. @fvankrieken, is that right?

from looking at our requirements files:
```
pyogrio==0.10.0
  # via geopandas
```

so I think this will still work. I think the `library` use of GDAL that we don't like is `from osgeo import gdal`

sf-dcp: oh interesting! Thanks for clarifying. You may wanna add `pyogrio` on line 117 ("dependencies") of the `pyproject.toml` file.

sf-dcp (approved):
LGTM! The only potential tweak is adding `pyogrio` to the `pyproject.toml` file (dcpy dependencies section).

dcpy/utils/data.py
fvankrieken: So clarifying - this error is happening when geopandas is reading a local geojson file, after it's already been retrieved via api

fvankrieken: I also don't think it needs to be added as a dependency - geopandas is already a dependency, and if geopandas removed pyogrio as a dependency, then this line would cease to do anything so some sort of error seems like it'd be appropriate.

In terms of where this happens - I think it might make more sense to either
1. declare this globally (like in dcpy's top level config file)
2. only tweak the parameter when catching this specific error when reading geojson, and then reset it

I would lean 1? I don't think there's a downside, I don't think we anticipate any issues from too-large geojson features being read?

damonmcc: > So clarifying - this error is happening when geopandas is reading a local geojson file, after it's already been retrieved via api

yup

> 1. declare this globally (like in dcpy's top level config file)

move it to `__init__.py` since we already do this sort of "package setting" there. it seems better than `configuration.py` where we're only setting envars

Issue #1624, opened 2025-04-22
ingest with source type `s3` failing


damonmcc:
on main [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14605923100/job/40974801629) and with a new template on a dev branch [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14604774383)

`get_version()` in `ingest/configure.py` does `connector.get_latest_version(source.get_key(), **source.model_dump())`

but if if `source` has a top-level item named `key`, `connector.get_latest_version` fails with "got multiple values for argument 'key'"

the class `S3Source` has a top-level item named `key` here

PR #1625, opened 2025-04-22
fix s3 ingest

damonmcc:
closes https://github.com/NYCPlanning/data-engineering/issues/1624

successful ingest run on main [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14606072502) and with a new template [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14606126367)

this seemed like the simplest fix for now. changing `S3Source` to use `path` instead of `key` would be nice but I didn't feel confident in that change being simple (would old config.json files parse correctly?)

### DIFF
--- a/dcpy/lifecycle/ingest/configure.py
+++ b/dcpy/lifecycle/ingest/configure.py
@@ -48,7 +48,9 @@ def read_template(
 
 def get_version(source: Source, timestamp: datetime):
     connector = source_connectors[source.type]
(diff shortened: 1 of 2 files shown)


fvankrieken (approved):
Yeah, and the issue that "key" really is the correct term for a unique identifier. Anything else would be sort of inaccurate. But agree that this solution (same solution that I used elsewhere) isn't super optimal either.

This is perfect for now, and thanks for adding the test case (though please rename it!)

PR #1628, opened 2025-04-23
Fvk lion v8 export

fvankrieken:


products/lion/macros/format_text.sql
fvankrieken, line 1: I think I will delete this. Just wanted to leave as an alternative

products/lion/macros/create_format_text_proc.sql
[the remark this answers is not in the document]
fvankrieken: I'll add a docstring - this is so that we can format things as specified in section 1.4 and table 2

products/lion/macros/create_format_text_proc.sql
```diff
@@ -0,0 +1,28 @@
+{% macro create_format_text_proc() %}
+
+CREATE OR REPLACE FUNCTION format_lion_text(
+    value varchar,
+    n int,
+    fill varchar DEFAULT '0',
+    blank_if_none boolean DEFAULT FALSE,
+    left_ boolean DEFAULT FALSE
```
damonmcc, line 8: I'm not seeing any uses of `format_lion_text` with  `left_` set to true. is this for future columns that need it?

products/lion/models/product/dat/_dat.yml
damonmcc, line 1: (for later) I've seen some project examples where there's a yml for every model instead of one for the whole subfolder. maybe a convention we should consider

products/lion/macros/create_format_text_proc.sql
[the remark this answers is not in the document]
damonmcc: appreciate the plpgsql vs dbt macro thoughts. makes sense!

the idea of a seed table in the future seems great too

damonmcc (approved):

products/lion/macros/create_format_text_proc.sql
fvankrieken: I don't think anything actually will - I wrote this based on the description of the fields before putting it into use, but every single field ended up being right-justified 🤷

Issue #1577, opened 2025-04-10
CSCL ETL POC - Validating POC outputs vs existing ETLs


fvankrieken:
These fall into two categories (or sort of a matrix, but whatever)

## POC/Migration Validation
i.e. ensuring that we're recreating the original pipeline.

Per #1559, the scope of the output we are aiming to produce is
- LION file (.dat)
- Only rows coming from centerline layer
- all columns

So we should be perfectly recreating a subset of rows from LION. In this [sharepoint folder](https://nyco365.sharepoint.com/sites/NYCPLANNING/itd/edm/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FNYCPLANNING%2Fitd%2Fedm%2FShared%20Documents%2FDOCUMENTATION%2FGRU%2FCSCL%2FETL%2F25A%20GDB%20and%20ETLs&CT=1744725340076&OR=OWA%2DNT%2DMail&CID=62a2af4c%2D86c8%2Dd075%2Ddb77%2Df0e96f6b6c7b&csf=1&web=1&e=yYv8al&clickParams=eyJYLUFwcE5hbWUiOiJNaWNyb3NvZnQgT3V0bG9vayBXZWIgQXBwIiwiWC1BcHBWZXJzaW9uIjoiMjAyNTA0MDQwMDUuMTEiLCJPUyI6Ik1hYyBPUyBYIHVuZGVmaW5lZCJ9&cidOR=Client&FolderCTID=0x0120001114839845564543AB5C3DCB09500AC5), we have both an input for archival and use as well as outputs to test against. There is not actually one lion dat file but one per borough. So whatever check we come up with should measure against these.

There are a few ways to approach this and we maybe want a couple
### File validation
Maybe the simplest binary output -> Is each row we produce in a headerless .dat file found in a respective reference .dat file? We need this certainty to have confidence in our outputs. However, when we have failures, outside of returning the raw rows it could be a little tougher to troubleshoot. There's definitely ways of handling it though. It maybe makes sense to
- use grep-based approach for fast yes/no
- use python string manipulation to get more details about rows that have incorrect data

Or, we could think more about the actual data hidden in the strings

### Data validation
Thinking about more of a tabular comparison - does the output match up just from a perspective of `SEGMENTID`, etc? This would be good to have for troubleshooting. The main issue here is that we have to go from formatted text .dat files to a tabular format, and there's room for error there

With that though, there are some maybe more standard "operational" checks

[...]

PR #1634, opened 2025-04-25
Lion - validating against current etl outputs

fvankrieken:
Start of #1577 

## Motivation - how to use the loaded dat tables to spot check

I've loaded parsed .dat outputs of the current etl tool into the db under the schema "production_outputs". We should have more complete tools around validation, but for now I've just been running some spot checks like

```sql
select 'ref' as source, * from production_outputs.manhattan_lion where "Segment ID" = '0018067'
union all
select 'poc' as source, * from fvk_lion_validation.lion_dat_fields ldf where "Segment ID" = '0018067'
```

For easy comparison. Again, it's not a thorough validation of the whole result set but it's a good sanity check while developing.
image

## Review
a couple main things here
- a few little cleanup things
- dat loader script -> python script to go from .dat files (from sharepoint - outputs of existing etl) to tables in postgres
- seed file that's used by the script. Has all fields and their respective indices in the .dat file
- some fixes I discovered when doing my first spot checks

build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14674723934) with all tests passing! Exciting for the dat fields and the one-column dat table

### DIFF
--- a/products/lion/seeds/lion_dat_formatting.csv
+++ b/products/lion/seeds/lion_dat_formatting.csv
@@ -1,0 +1,109 @@
+field_number,field_label,field_length,start_index,end_index,formatting
+L1,Borough,1,1,1,Digit > 0
+L2,Face Code,4,2,5,RJZF
+L3,Sequence Number,5,6,10,RJZF
+L4,Segment ID,7,11,17,RJZF
+L5,5-Digit Street Code (5SC),5,18,22,RJZF
+L6,LGC1,2,23,24,RJZF
+L7,LGC2,2,25,26,RJZF - blank if none
+L8,LGC3,2,27,28,RJZF - blank if none
+L9,LGC4,2,29,30,RJZF - blank if none
+L10,Board of Elections LGC Pointer,1,31,31,Blank if none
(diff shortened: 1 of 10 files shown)


products/lion/seeds/lion_dat_formatting.csv
fvankrieken, line 1: For now, this is just being used by the `dat_loader` script to parse dat files generated by existing etl app into pd dataframe/postgres tables. They could be used down the road to generate a lot of the code in `lion_dat_fields.sql` if this seems easier to maintain here than in the sql file. So just made it a seed because I figured I might as well.

sf-dcp: Ok, I'm a bit lost. From the script, we load the dat file into Postgres, resulting with a single column for all data. I'm confused how the dat files relates to this csv file. I initially thought that the csv is the formatted dat file but it doesn't sound like it

sf-dcp: Nm! It all makes sense

sf-dcp (approved):
Easy solution for now, looks great!

Issue #1597, opened 2025-04-18
CSCL ETL POC - lingering data questions


fvankrieken:
Follow ups to #1559 

Just wanted a place to consolidate these. These should be very specific data questions

## General
- [ ] when to use `st_linemerge`? It seems there might be a few weird segments that are truly multilinestrings, but maybe don't need to be.
- [ ] strange giant circular segment -> hopefully resolved by archiving new data

## 2.2.1
- [ ] The segment `0182706` in the source segment table has 2 records with a duplicate local group code (`lgc`). This results in additional `lgc2` = `03` and `lgc3` = `03` columns in the prod data. In our code, we have `lgc2` = `3` and `lgc3` = `NULL` as the result of using unique records. Are duplicates allowed in the lgc columns?

## 2.2.2
- [ ] is a non-spatial join actually fine for joining endpoints to nodes?
  - 70ish segments missing to or from node. Is this consistent with current process? If so, this is moot question 
- [ ] some endpoints are on boundary of sectionalmap, leading to duplicate rows. How should these be handled?
  - this is actually a topology issue. I looked into one of the points, and it turns out there's some overlap between two sectional map polygons, with actual intersections with both. See sectionalmap 5 and 6 (polygons have opacity slightly down
    ![Image](https://github.com/user-attachments/assets/a3876fbe-dbaa-429c-a173-7e8e2494505d)
    ![Image](https://github.com/user-attachments/assets/cd9e3cce-c605-47b9-9b2b-6c327d811422)
    overlap seems to be ~6 inches

## 2.2.3
- [ ] we've implemented a slightly different algorithm. Evaluate whether these fields line up with what comes from existing process, but also determine validity of new algorithm 
  - #1568 has summary
- #1642 was opened originally to address a specific issue, but broader decision really needs to be made. Might close it out

## 2.2.6
- [ ] There is a big mismatch, around 7,000 records, between dev and prod tables with respect to the `2.2.6` field. Verified, the dev code is correct. From looking at the prod code, it incorrectly uses 2010 census tracts for the field calculations; the docs specify to use census 2020.

## 2.2.8
- [ ] More of a note: around ~100 records in dev table don't match prod. The calculation uses left & right atomic polygons. All ~100 records have mismatching atomic polygons between dev and prod.

## 2.2.10
[...]

PR #1636, opened 2025-04-28
Lion ETL: Centerline & Face Code Fields

### LINKED ISSUE #1566: CSCL ETL POC - Centerline Street/Face Code Fields
The standard workflow for these cscl chapters is pretty consistent, just with different sections in the documentation and different fields needed in the final product table
- [x] Source data (ingest templates, recipe/dbt setup) added as needed if mentioned in 2.2.1
- [x] intermediate tables added to perform transformations outlines in 2.2.1
- [x] consider/implement constraints/tests of table (i.e. 1-1 rows with centerline, etc)
- [x] final product table modified with int table joined in to align with Table 3 for relevant fields:
  - L2, L5-10, L74-78

sf-dcp:
Closes #1566.

2.2.1 in the [docs](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BD3194825-3026-4295-8A31-6FB030749AA1%7D&file=ETL_V8_02012024%20-%20Copy.docx&action=default&mobileredirect=true)

Latest build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14767666484/job/41462156546)

### What
Adds 4 types of fields in this PR:
- Local group codes in fields `lgc1` - `lgc9`. You could think of them as various ids for a given centerline segment. `lgc1` is the "preferred" local group code. A centerline segment can have between 1 and 9 local group codes.
- Board of Elections local group code. It's derived based on a flag and the lgc number. If a flag is in a street segment corresponding to `lgc4`, then the value for this field is 4.
- Face code
- `5sc` field. This field is a street code but without a borough in the beginning.

The logic in dbt collects all these fields into one table, `int__centerline_streetcode_and_facecode` which has 1-1 mapping with the final lion table. The visual:

image

### Testing data
Please refer to the issue. In summary, there are mismatches that I'm not sure how to address. The dev values in the new fields make sense logically, matching to the docs description.


products/lion/models/intermediate/2.2.1/_int_221.yml
```diff
@@ -0,0 +1,88 @@
...
+  - name: five_digit_street_code
+    tests:
+    - not_null
+  - name: face_code
+    tests:
+    - not_null
+  # the test below should be here but dbt errors while building staging tables
+  # tests:
+  # - dbt_expectations.expect_table_row_count_to_equal_other_table:
+  #    compare_model: ref("stg__centerline")
```
sf-dcp, line 88: Would be nice to have this test but `dbt` throws an error when building stage tables saying `int__centerline_streetcode_and_facecode` table cannot be found.

products/lion/models/intermediate/2.2.1/int__nonpreferred_local_group_code.sql
```diff
@@ -0,0 +1,28 @@
...
+    WHERE
+        preferred_lgc_flag <> 'Y'
+),
+nonpreferred_ranked_by_lgc AS (
+    SELECT
+        segmentid,
+        lgc,
+        RANK() OVER (
+            PARTITION BY segmentid
+            ORDER BY lgc ASC
```
fvankrieken, line 19: I think we maybe should order by `ogr_uid` here - I had another one where it had similar wording like "take the first" or "in order" and the way I got my code to match up with production was by going by row number in the original table

fvankrieken: How many rows actually get assigned an lgc2?

fvankrieken: Looks like a lot, it would be surprising if your ranking were not accurate

sf-dcp: I will try with `ogr_uid`. 

Answering your question: there are 49,664  *centerline* records with an assigned `lgc2`. The total number of records from this table with `lgc2` is 72,525 which includes all feature classes.

fvankrieken: I take it back!

products/lion/models/intermediate/2.2.1/_int_221.yml
fvankrieken: Huh that really shouldn't be... I'm gonna poke around

fvankrieken:
Looks great. My only notes
- run tests for boe pointer
- decide whether we want to "fix" [this](https://github.com/NYCPlanning/data-engineering/issues/1566#issuecomment-2845203494) or note it in #1597 
- I'm gonna quickly look into the weird failing dbt test

sf-dcp:
* I ran the tests for the boe pointer - please see the issue. Unfortunately got a big chunk of mismatched records back.

* I added a note to the issue

products/lion/models/intermediate/2.2.1/_int_221.yml
```diff
@@ -0,0 +1,88 @@
...
+  columns:
+  - name: segmentid
+    tests:
+    - not_null
+  - name: lgc
+  - name: lgc_rank
+    tests:
+      - dbt_utils.accepted_range:
+          min_value: 2
+          max_value: 9  # a segment shouldn't have more than 9 lgcs
```
fvankrieken, line 35: Longer term, we wouldn't want to error here but output to a report (while still generating LION), but that's obviously out of scope of the poc

products/lion/models/intermediate/2.2.1/_int_221.yml
fvankrieken: Seems like an odd bug due to the fact that our build step has [multiple commands for the different stages](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/lion_build.yml#L79), and since this test has a reference to a staging table dbt somehow is messing up when it should be run (and running it only after building staging tables rather than building the table the test actually corresponds to)

Fixed action [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14779853699) by [just running `dbt build`](https://github.com/NYCPlanning/data-engineering/blob/fvk-sf-lion/.github/workflows/lion_build.yml#L72), which I think is fine, we don't get that much out of separate bash commands and print statements

fvankrieken (approved):
Approving with a couple github requests
- detail issues around that single LGC3 mismatch in #1597 (i.e. are duplicate LGCs allowed?)
- " for BOE flag. It really seems like this is a bug in the existing ETL
- up to you if you want to get that test working. As I linked in the comment, it's a pretty simple fix (to a very strange bug IMO) to get it working. Happy to punt it towards general pipeline cleanup

fvankrieken:
Also oops merge conflict from curve-related fields that you have to resolve, should be super simple

PR #1641, opened 2025-04-29
Lion - Census Tract Formatting

### LINKED ISSUE #1633: CSCL ETL POC - format census tracts and house numbers
Little bit of a stub, but these two fields have very specific formatting as specified in 1.4. In the `lion_dat_fields` product table (or maybe in just the `lion` product table), these need to be handled. 
- [ ] census tracts. It most likely makes sense to break these into basic and suffixes, then merge (or don't merge - they'll get concatenated in the final table anyways)
- [ ] house numbers. Haven't fully read the docs so not sure on best approach

fvankrieken:
Closes #1633 . Two formatting issues in scope here
- census tract formatting
- house number formatting

build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14741284887/job/41379671459)

- census tracts needed little tweaks. They come in as 008700 (tract 87, no suffix) or 051001 (tract 51, suffix 01). The output expects those as "  87  " and " 051001" respectively -> base is RJSF, suffix is spaces if null, RJZF otherwise
  - A handful in `production_outputs` don't match up, but only in cases where one is null and the other is populated. This is an unresolved issue with 2.2.3 logic that I need to look into. No mismatches due to formatting
- 1.4 also has detailing on house numbers. However, in my querying of the tables, it seems that they come from CSCL centerline in the exact format that they are needed in the export


products/lion/models/intermediate/2.2.3/_int_223.yml
```diff
@@ -6,15 +6,45 @@ models:
     tests: [ unique, not_null ]
   - name: left_atomicid
   - name: left_2000_census_tract
+    tests:
+    - dbt_expectations.expect_column_value_lengths_to_equal:
+        value: 6
```
fvankrieken, line 11: I added these tests here because this is where the tract values are split into two. I guess I could test stg__atomic_polygons instead, but this feels nice logically (if it were to go wrong, it feels useful to see the error at this table where census tract is used). It is a bit of a self-dependency though. I think that's fine.

products/lion/models/intermediate/2.2.3/int__centerline_atomicpolygons.sql
```diff
@@ -12,8 +12,14 @@ SELECT
     co.segmentid,
     left_poly.atomicid AS left_atomicid,
     left_poly.censustract_2000 AS left_2000_census_tract,
```
fvankrieken, line 14: These are still selected here, but not selected in `lion` right now

products/lion/models/product/dat/lion_dat_fields.sql
```diff
@@ -25,19 +25,21 @@ SELECT
     format_lion_text(to_nodeid::TEXT, 7, '0') AS "To-Node ID",
     format_lion_text(round(to_x)::INT::TEXT, 7, '0') AS "To-X Coordinate",
     format_lion_text(round(to_y)::INT::TEXT, 7, '0') AS "To-Y Coordinate",
-    format_lion_text(left_2000_census_tract, 6, ' ', TRUE) AS "Left 2000 Census Tract", -- TODO section 1.4
+    format_lion_text(left_2000_census_tract_basic::TEXT, 4, ' ', TRUE) AS "Left 2000 Census Tract Basic",
+    format_lion_text(left_2000_census_tract_suffix::TEXT, 2, '0', TRUE) AS "Left 2000 Census Tract Suffix",
```
fvankrieken, line 29: We could merge these back to one field? But it felt better to be able to test them individually in this model - they all get squished together in the end anyways

sf-dcp:
> A handful in production_outputs don't match up, but only in cases where one is null and the other is populated.

Could you give an example what it looks like? I don't think I understand this part

products/lion/models/intermediate/2.2.3/_int_223.yml
sf-dcp: I like it here for documentation purposes too

products/lion/models/product/dat/lion_dat_fields.sql
fvankrieken: I think I'm gonna push back given that both the basic and suffix have different formatting expectations, it's nice to be testing them each separately in this table

fvankrieken: And in the etl docs table, these census tract fields explicitly have subfields

sf-dcp (approved):

Issue #1571, opened 2025-04-08
CSCL ETL POC - Segment locational status


fvankrieken:
Blocked by #1568 

- [ ] Source data (ingest templates, recipe/dbt setup) added as needed if specified in 2.2.6
- [ ] intermediate tables added to perform transformations outlines in 2.2.6
- [ ] final product table modified to align with Table 3 for relevant fields: L41

sf-dcp:
### `Borough Boundary Indicator`

```sql
select 
d."Borough",
d."Sequence Number",
d."Segment ID",
d."Borough Boundary Indicator" as bbi_dev,
p."Borough Boundary Indicator" as bbi_prod
from lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Sequence Number", "Segment ID")
where d."Borough Boundary Indicator" <> p."Borough Boundary Indicator";
```

This query yielded 95 unique records. Sample: 

|Borough|Sequence Number|Segment ID|bbi_dev|bbi_prod|
|-------|---------------|----------|-------|--------|
|1|00016|0344644|R| |
|1|00100|0023540|L| |
|1|00100|0307317|L| |
|1|00195|0346982|L| |

All records in the table have `NULL` in the prod column though they shouldn't. The dev values are populated as expected by the docs:

> If there is no AP on one side of the given segment, the segment lies along the City
limit. The tool will populate the Borough Boundary Indicator with the value (‘L’ or
‘R’) that identifies the side of the segment on which there is no AP. 

I did some research and it looks like the discrepancy is due to the underlying differences in atomic polygons for a given polygon. I ran the following query and it yielded zero records:

```sql
select
d."Borough",
d."Sequence Number",
d."Segment ID",
d."Borough Boundary Indicator" as bbi_dev,
p."Borough Boundary Indicator" as bbi_prod
from lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Sequence Number", "Segment ID")
where d."Borough Boundary Indicator" <> p."Borough Boundary Indicator" and 
	(d."LEFT CENSUS BLOCK 2020 BASIC" = p."LEFT CENSUS BLOCK 2020 BASIC" and d."RIGHT CENSUS BLOCK 2020 BASIC" = p."RIGHT CENSUS BLOCK 2020 BASIC")
```

sf-dcp:
### `Segment Locational Status`

```sql
select 
d."Borough",
d."Sequence Number",
[...]

PR #1655, opened 2025-05-02
Lion: `2.2.6` & `2.2.8`

sf-dcp:
Related to #1571

Successful [build](https://github.com/NYCPlanning/data-engineering/actions/runs/14851255477/job/41695181946)

This PR adds 2 fields to the `lion` table. 

For the most part, the logic between the 2 fields is very similar: it's based on a segment's atomic polygons. For `2.2.6`, commit with part || covers the case when a segment is within same atomic polygon and the output value depends on whether the segment is a dead-end (There are no other segments attached to it). 

I added the tables in red:

image

### Testing data

Please refer to the issue. There is a mismatch for the `Borough Boundary Indicator` field due to underlying atomic polygons.

### DIFF
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -1,0 +1,9 @@
+models:
+- name: int__centerline_segment_locational_status
+  description: |
+    Centerline segments with Locational Status & Borough Boundary Indicator columns.
+  columns:
+  - name: segmentid
+    tests: [ unique, not_null ]
+  - name: borough_boundary_indicator
+  - name: segment_locational_status
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -19,6 +19,9 @@ models:
     tests:
       - accepted_values:
           values: ['9', 'H', 'I', '1', '2', '3', '4', '5', 'X']
+  tests:
+  - dbt_expectations.expect_table_row_count_to_equal_other_table:
+     compare_model: ref("int__centerline_atomicpolygons")
 
 - name: int__centerline_segment_neighbors
   description: |
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -13,6 +13,8 @@ models:
     tests:
       - accepted_values:
           values: [ False ]
+          config:
+            severity: warn
   - name: segment_locational_status
     tests:
       - accepted_values:
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -6,4 +6,11 @@ models:
   - name: segmentid
     tests: [ unique, not_null ]
   - name: borough_boundary_indicator
+    tests:
+      - accepted_values:
+          values: ['R', 'L']
+  - name: is_ap_boro_boundary_error
+    tests:
+      - accepted_values:
+          values: [ False ]
   - name: segment_locational_status
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -18,3 +18,12 @@ models:
       - accepted_values:
           values: ['9', 'H', 'I', '1', '2', '3', '4', '5', 'X']
 
+- name: int__centerline_segment_neighbors
+  description: |
+    Centerline segments with offset nodes mapped to the neighboring segments.
+    Only contains segments having both offset nodes.
+  columns:
+  - name: segmentid
+  - name: node
+  - name: to_segment
+  - name: node_missing_neighbor
--- a/products/lion/models/intermediate/2.2.6/_int_226.yml
+++ b/products/lion/models/intermediate/2.2.6/_int_226.yml
@@ -14,3 +14,7 @@ models:
       - accepted_values:
           values: [ False ]
   - name: segment_locational_status
+    tests:
+      - accepted_values:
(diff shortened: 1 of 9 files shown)


products/lion/models/intermediate/2.2.6/int__centerline_segment_locational_status.sql
```diff
@@ -0,0 +1,116 @@
...
+    FROM atomicpolygons AS ap
+    LEFT JOIN segments_with_orphan_node AS o
+        ON ap.segmentid = o.segmentid
+    WHERE ap.left_atomicid = ap.right_atomicid
+),
+different_aps_different_boros AS (
+    SELECT
+        *,
+        CASE
+            WHEN centerline_borocode <> left_borocode AND centerline_borocode = right_borocode THEN 'L'
```
fvankrieken, line 69: For these `WHEN`s, we're already specifying that left <> right in the filter. So we can just check one of these two conditions, the other is guaranteed to be true.

This goes for all four `WHEN` clauses in this CTE. I get maybe being explicit about the conditions, but I think given how complicated this query is, keeping the CASE WHENs simpler based on the filter conditions makes it easier to read and digest

products/lion/models/intermediate/2.2.6/int__centerline_segment_locational_status.sql
```diff
@@ -0,0 +1,116 @@
...
+        *,
+        NULL AS borough_boundary_indicator,
+        TRUE AS is_ap_boro_boundary_error,
+        NULL AS segment_locational_status
+    FROM atomicpolygons
+    WHERE left_atomicid IS NULL AND right_atomicid IS NULL
+)
+
+SELECT * FROM left_city_boundary_aps
+UNION
```
fvankrieken, line 107: Again, let's use UNION ALL

products/lion/models/intermediate/2.2.6/int__centerline_segment_locational_status.sql
```diff
@@ -0,0 +1,116 @@
...
+        *,
+        CASE
+            WHEN centerline_borocode <> left_borocode AND centerline_borocode = right_borocode THEN 'L'
+            WHEN centerline_borocode = left_borocode AND centerline_borocode <> right_borocode THEN 'R'
+        END AS borough_boundary_indicator,
+        centerline_borocode <> left_borocode AND centerline_borocode <> right_borocode AS is_ap_boro_boundary_error,
+        CASE
+            WHEN
+                centerline_borocode <> left_borocode AND centerline_borocode = right_borocode
+                THEN left_borocode::char(1)
```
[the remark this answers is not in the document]
sf-dcp: I believe this value gets casted to string here by default anyways because the `segment_locational_status` column has non-numeric values as well. So, I either casted to be explicit or because the query wasn't working

products/lion/models/intermediate/2.2.6/_int_226.yml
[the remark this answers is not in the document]
sf-dcp: Yeah, I tried to do one big table but it was hard to parse the logic. I will add the test to cross-check record count with the atomic polygons table (the underlying table that gets chunked up)

sf-dcp:
@fvankrieken I added all the suggestions and updated the GHA to run `dbt build` for all tables. Successful run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/14865272150/job/41740403330). Interestingly, the build ran only 5 minutes! Maybe building entire thing without separating dbt commands expedites the run...

products/lion/models/intermediate/2.2.6/int__centerline_segment_locational_status.sql
```diff
@@ -67,16 +67,14 @@ different_aps_different_boros AS (
         *,
         CASE
             WHEN centerline_borocode <> left_borocode AND centerline_borocode = right_borocode THEN 'L'
-            WHEN centerline_borocode = left_borocode AND centerline_borocode <> right_borocode THEN 'R'
+            ELSE 'R'
```
fvankrieken, line 70: Sorry, this is not entirely what I meant - I meant within the same WHEN clause

Given that we have a clause filtering down WHERE left <> right (boro), all we need for one of these clauses is

```
CASE
    WHEN centerline_borocode <> left_borocode THEN 'L'
    WHEN centerline_borocode <> right_borocode THEN 'R'
END AS borough_boundary_indicator
```

and then something similar for seg loc status

To your code, the second `WHEN` clause is somewhat unnecessary as well, but I think it helps with understanding the logic to leave it. But my point was more around readability and flow of logic of these individual `WHEN` clauses
- we're in the CTE where left boro <> right boro
- in those cases, we are assigning the boundary indicator to the side whose borough differs from centerline

fvankrieken: I guess I'm missing the case where neither left or right boro equals the centerline boro though

fvankrieken: So then maybe

```
CASE
    WHEN centerline_borocode = right_borocode THEN 'L'
    WHEN centerline_borocode = left_borocode THEN 'R'
END AS borough_boundary_indicator
```

fvankrieken: Let's just drop this post-review commit for now. I do think this could be cleaned up slightly but to your point we maybe should have more upstream tests in that sort of case.

Sorry about this

sf-dcp: Per teams chat, I revised the logic.

sf-dcp:
> Assuming the pr test doesn't suddenly freak out for no reason!

Yes, re-ran the build and everything passed! Yeah, it's def faster with using the dbt build command... Took only 6 min for the whole run where the build is 3 minutes.

Issue #1391, opened 2025-01-16
GRU - change names of SAF Addresses vs PAD QA reports


damonmcc:
GRU has requested updates to the names of the output reports for the Generic SAF Addresses vs PAD Roadbed SAF Addresses vs PAD QAQC job

- SAF_gen_1R_pad  update to        GN SAFs vs GRID1R
- SAF_gen_1_pad    update to        GN SAFs vs GRID1G
- SAF_rb_1R              update to       RB SAFs vs GRID1R
- SAF_rb_1_pad       update to        RB SAFs vs GRID1G

 
If possible, for the post 25A cycle which will start around 2/25/2025

damonmcc:
filenames seem to be declared by `db-gru-qaqc` repo in `saf-vs-pad.py` [here](https://github.com/NYCPlanning/db-gru-qaqc/blob/main/python/saf-vs-pad.py#L228)

```python
df.to_sql(
    f"saf_{INPUT_TYPE}_{FUNCTION+ROADBED}_pad".lower(),
    con=engine,
    if_exists="replace",
    index=False,
    method=psql_insert_copy,
)
```

PR #1668, opened 2025-05-12
gru page refresh

fvankrieken:
A few things
- wanted to list filenames dynamically since gru has requested renaming some of them in #1391 
- reworked the looking up of archival information to be just a little less finicky. Annoying to lose type safety but these issues weren't being caught anyways because the function that was being used is so unmaintained
- noticed that the page is broken because the bucket no longer defaults to completely publicly available (thanks @alexrichey ). Instead of making these outputs public read, figured it'd make sense to just use presigned urls. Right now am doing them with an hour before they expire, seems perfectly reasonable that someone would need to refresh the page after an hour

It works now locally
image

### DIFF
--- a/apps/qa/src/pages/gru/components.py
+++ b/apps/qa/src/pages/gru/components.py
@@ -5,8 +5,9 @@ import streamlit as st
 import time
 
 from dcpy.connectors.github import WorkflowRun
+from dcpy.utils import s3
 from src.shared.components.github import dispatch_workflow_button
-from .constants import qa_checks
+from .constants import qa_checks, bucket
 from .helpers import get_source_versions, get_geosupport_versions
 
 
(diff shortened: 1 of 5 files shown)


sf-dcp:
- A nit to add a little comment on the streamlit about the link expiring in an hour. 
- Q: this pr doesn't close the linked ghi, correct?

fvankrieken:
> * A nit to add a little comment on the streamlit about the link expiring in an hour.
> * Q: this pr doesn't close the linked ghi, correct?

I don't really want to add a comment that might just confuse users, I think this is a case where it's simpler for them if they don't have to think about it.

To that end, added logic to autorefresh the page every 10 min (and shortened the expiration time - figured it's better to not just be generating tons of these lingering around)

To your question, nope. Just removes hardcoding of filenames so that changing filenames won't break anything. The work to close that is here https://github.com/NYCPlanning/db-gru-qaqc/pull/274, but this can merge first

sf-dcp (approved):

Issue #1653, opened 2025-05-01
CEQR Data Hub: Broken Shapefiles


sf-dcp:
[Copying from a teams chat]

> https://github.com/NYCPlanning/caps-process/issues/48#issuecomment-2835280430 Julia from parks documented a few cases of shapefiles with a missing `.proj` file in the POPs file, NYS Parks, Federal Parks, and Parks Capital Project Tracker which we verified as true- no `.proj` component was included. Sally is going to help look but we would appreciate if your team could help in the QA of data stored and downloadable through digital ocean. 

The `.proj` file stores the projections info, the crs. This issue could be due to some weirdness in `library` or CEQR build.

fvankrieken:
Did they compile a more specific list or is this what we have in terms of info? No worries if this is as specific as it is, just if there was already a list compiled I would use that obviously rather than dig through the excel myself

sf-dcp:
@fvankrieken, yes! Here is the list:

* MapPLUTO Shoreline Clipped
* housing database
* facilities database
* POPS
* NYS Parks
* federal parks
* parks capital projects tracker
* historic building (nyc)
* scenic landmarks
* historic districts (national)

[Link](https://nycplanning.github.io/test-ceqr-data-github-pages-theme/pages/datasets.html) to the CEQR Hub where these websites are available for download.

fvankrieken:
I __think__this is a library/gdal issue. library uses gdal to generate (geo)parquet files. On the version of gdal that library is pinned to, the geoparquet files generated are missing crs, so when they're then read into postgis, they no longer "know" what projection they are. Gonna run library with my dev bucket and latest gdal to see if that fixes it. If that doesn't, then maybe it's just a simple missing option in the library templates.

Assuming that it is a version of the parquet library driver, a few ways to fix this
- update gdal and re-archive these datasets with library
- migrate these datasets from library to ingest
- use the raw datasets for these instead of the "processed" one, though this is blocked by option 2 so this is irrelevant

1 might not be too hard, and would maybe be best for our infrastructure in that we'd fix this bug that's certainly affecting other datasets. I have a WIP branch from a while ago that upgrades gdal and changes the logic that broke with new gdal (#1669). But it also wouldn't be too hard to just try to quickly crank through migrating these to ingest.

fvankrieken:
[...]

PR #1670, opened 2025-05-13
CEQR - pin various library datasets to pg_dump

fvankrieken:
Band-aid solution to #1653 

library-generated parquet files are missing projection.

I tried archiving w/ library with later version of gdal, and projection still seems to be missing from parquet metadata, though ogrinfo happily returns the proj info of the file. It seems like maybe there's some misalignment with gdal parquet driver and geoparquet spec?? Not sure. Not going to look much further into it since it seems there's not an immediate fix.

So this will have all datasets properly getting proj, and longer term getting these datasets over to `ingest` will fix the issue longer term. I'll do some work on that this week. Below is one of the outputs of the "build"
image

### DIFF
--- a/products/ceqr/recipe.yml
+++ b/products/ceqr/recipe.yml
@@ -7,14 +7,19 @@ inputs:
 
   dataset_defaults:
     file_type: parquet
+    # several datasets are pinned to pg_dump
+    # this is because they are still archived with library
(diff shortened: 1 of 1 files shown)


sf-dcp (approved):
Awesome!

PR #1679, opened 2025-05-19
point geosupport version scraping at new website

fvankrieken:
Geosupport is now available on the new website, and bytes is no longer updated. Since we determine latest geosupport version by scraping that webpage (something that should ideally change long term...), the logic needed a little update

https://www.nyc.gov/content/planning/pages/resources/geocoding/geosupport-desktop-edition

output of edited python script:
image

run to publish [here](https://github.com/NYCPlanning/data-engineering/actions/runs/15122360608)

### DIFF
--- a/admin/run_environment/docker/geosupport_versions.py
+++ b/admin/run_environment/docker/geosupport_versions.py
@@ -13,10 +13,6 @@ logging.basicConfig(
     datefmt="%Y-%m-%d %H:%M:%S",
 )
 
-IGNORE_UPAD_RELEASE = (
-    True  # TODO this is temporary while an image with 22c4 UPAD needs to be built
-)
-
 CALLER_ENVIRONMENT_VARIABLE_NAME = "VERSIONSTRING"
 GEOSUPPORT_RELEASE_URL = "https://apps.nyc.gov/content-api/v1/content/planning/resources/geocoding/geosupport-desktop-edition"
 # to convert a release letter to a number for the docker image tag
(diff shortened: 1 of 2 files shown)


admin/run_environment/docker/geosupport_versions.py
```diff
@@ -27,74 +27,54 @@
...
-            )
+    if primary_release == upad_primary_release:
+        logging.info("Matching Primary and UPAD's Primary releases")
+        # UPAD should be incorporated
+        release = upad_release
+    else:
+        logging.info("WARNING! Mismatch between posted Primary and UPAD releases")
+        # posted UPAD is not meant for current release
+        # TODO this is temporary while an image with 22c4 UPAD needs to be built
+        # note 2025-05 - not sure if this is needed
```
fvankrieken, line 61: It makes sense that we'd ignore it? In this case, UPAD is 25A4 - we obviously want to build 25B

sf-dcp:
I'm a bit confused why or for what we have scraped the geosupport version... In library/ingest template, don't we supply the version to feed into the url to extract data from?

fvankrieken:
> I'm a bit confused why or for what we have scraped the geosupport version... In library/ingest template, don't we supply the version to feed into the url to extract data from?

So good question.

For one, there's no library/ingest involved here. This is a part of building our geosupport-related docker images. We publish these and they're tagged by geosupport version. See https://hub.docker.com/r/nycplanning/docker-geosupport/tags

But in terms of why it couldn't just be put in manually... originally, the docker build/publish ran daily, so it just kept it in sync. Right now it doesn't run daily, only when we make changes to docker code. It maybe doesn't need to be determined programmatically. But there is something nice about maybe running it on a schedule so we don't need to remember when a UPAD update goes out mid-cycle and go publish a new docker image. 

Longer term, that "flow" definitely should happen without scraping our public website (like ideally, GRU runs some process which triggers some job of ours that builds and publishes this thing. But that's not super scoped right now in any way)

fvankrieken:
So if you run "docker - build and publish" and build one of the geosupport images, it starts by seeing what the latest versions available are

admin/run_environment/docker/geosupport_versions.py
sf-dcp: I don't understand this logic all together, so can't advise. Perhaps silly question... do we need both of these datasets for geosupport? If so, is there a certain rule around versions for geosupport?

sf-dcp: I see. This makes sense. Then I feel like we don't really need the if/else below: it could be simplified to the `IGNORE_UPAD_RELEASE` logic. 

Because why would we want to build an image with Geosupport = `25b` and tpad = `24a4`?

sf-dcp (approved):

Issue #1680, opened 2025-05-19
PLUTO: investigate area for bbl `3063430031`


sf-dcp:
From an email received on 5/19:

> I am working on a project with BBL 3063430031 and have come across a 10k sf lot area discrepancy between the PLUTO data and the official tax map. Since this seems to be a significant discrepancy, I wanted to bring it to your attention! Please see the details below:
 
> BBL: 3063430031
> PLUTO Lot Area: 24,683 sf
> Tax Map Area: 35,087 sf

The area comes from either DTM or PTS depending on the lot shape (regular vs irregular). We need to research the lot and correct if needed. 

TODO: 

- [ ] Inform Bianca of the outcome.

fvankrieken:
Seems like a pretty normal lot, with a very under-reported area

<img width="910" alt="Image" src="https://github.com/user-attachments/assets/0bef1eae-54fa-4e07-b11e-6473225a1ec8" />

<img width="910" alt="Image" src="https://github.com/user-attachments/assets/1021fc9d-0fb3-41a6-b4de-ad51c643bf44" />

I'm going to add a manual correction and maybe we should let DOF know? Don't recall at this moment if this is a type of issue that we notify them of

fvankrieken:
Closing, moving the todo to the pluto 25v2 issue

Issue #1661, opened 2025-05-06
PLUTO: Condo Lots with NULL values in APPBBL field


croswell81:
The definition of **Apportionment BBL (APPBBL)** from the [PLUTO Data Dictionary](https://s-media.nyc.gov/agencies/dcp/assets/files/pdf/data-tools/bytes/pluto_datadictionary.pdf) is the originating BBL (borough, block and lot) from the apportionment prior to the merge, split **or property’s conversion to a condominium**. Only available for mergers, splits, and conversions since 1984.

Based on this we would expect every condo lot "75xx" in PLUTO to have a value in the APPBBL field (unless it occurred before 1984). 

When querying PLUTO with Lot >= 7500 And Lot <= 7599 And APPBBL IS NULL **there are 2,237 null lots out of 10,502 condo lots** (i.e. 75xx).

Please look into how the APPBBL is populated and determine if there is a reason why some condo lots do not have a value.

fvankrieken:
I would describe this largely as a PTS issue. However, we seemingly can fix a lot of these.

Not sure if it's just because I'm looking at the latest (proto-25v2) build, but I see **2326 records** that are condos but have NULL appbbl. If you recall, PTS has rows both for the primebbl and for individual units. Our logic in pluto populates appbbl solely from the corresponding primebbl row in PTS. So I did a bit of digging into these records to see what their source PTS data looked like
- 2326 records
- all have no appbbl data associated with primebbl record in PTS
- 2184 have appbbls associated with their **unit** PTS record
  - 2078 have only one distinct appbbl associated with their various unit PTS records. **these seem like great candidates to have appbbl set during the build**
  - 106 have multiple appbbls associated with their various unit PTS records. As in, this prime bbl has multiple units, each of which are associated with a different appbbl. These maybe we could say assign the lowest or something, but worth discussing how we might want to handle
- 142 have no appbbls associated with them
  - 123 don't have unit PTS records at all
  - 19 have unit PTS records but these are also missing appbbl data

Happy to talk this over and/or get some data dumps for you

fvankrieken:
Per 6/3 convo - ideally all primebbl pts records have appbbl defined, so maybe we make a report out of this to send to DOF

- data quality note - should not allow unit appbbls - there are some weird cases where a unit row has it's own bbl as its appbbl
- rows with one distinct appbbl for the units -> take this appbbl
- ones with multiple - add the filtering of unit appbbls but then take lowest
- the rest - leave null

[...]

Issue #1660, opened 2025-05-06
PLUTO: Wrong CD value for Central Park


croswell81:
Wai at NYCEM reported that the **Community District (CD) value** for central park is listed as _'108'_ but it should be **'164'** which is the joint interest area (JIA) for Central Park. This applies to **BBL 1011110001**.

He also noted they first noticed this in version 23v1 but have not conducted a thorough review.

fvankrieken:
Funny case happening here.

For PTS we start with bbl. We
- geosupport BL to get address info. This generally has more than one address
- grab the "first" address that has both a 5 digit sc and 
- feed this address to 1A and 1E to get additional info (BL function does not return things like CD)

For Central Park, one of the addresses that gets returned is 831 5 ave. This is a funny address - not sure what is really associated it. Google maps puts it in the middle of 5th Avenue, and GOAT says it's on the right/east side -> across the street from the park. So if that part is correct, then the cd of 108 is correctly being returned for this address.

So it seems maybe either
- BL(1011110001) should not include address 831 5 Ave
- 1A/1E(831 5 Ave) should return it on the west side of 5th Ave (and associated political districts)

Issue #1638, opened 2025-04-29
PLUTO Enhancement: encorporate new FAR table


fvankrieken:
[This table](https://github.com/NYCPlanning/data-engineering/blob/main/products/pluto/pluto_build/data/dcp_zoning_maxfar.csv) has some outdated values. SL will share an updated table by May 11, which will include data for manufacturing or industrial FAR and affordable FAR; we will not be adding any new fields into PLUTO 25v2, but will be doing the following
- [x] incorporate the updated table so that PLUTO build uses updated residential, commercial, and facility FAR values will be incorporated into PLUTO 25v2.
- [x] update the Residential FAR description in PLUTO's data dictionary to state that the Residential FAR does not include the Universal Affordability Preference (UAP) FAR allotment.  We will share the updated description with zoning for review before publication.
- [ ] create a new report to flag when a zoning district exists without an entry in the linked table (maybe fail the build?), as well as potentially outputting the table from build for easy sharing with zoning. (or the file from the repo can just be shared - that part tbd)

fvankrieken:
Expecting table by EOW or next Monday

fvankrieken:
Data shared [here](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/RANDOM%20DATA/dcp_zoning_maxfar_202505.xlsx?d=wcc003126310d4950abfddfd67bde1fd4&csf=1&web=1&e=xVy4Hu)https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/RANDOM%20DATA/[dcp_zoning_maxfar_202505](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/RANDOM%20DATA/dcp_zoning_maxfar_202505.xlsx?d=wcc003126310d4950abfddfd67bde1fd4&csf=1&web=1&e=xVy4Hu).xlsx?d=wcc003126310d4950abfddfd67bde1fd4&csf=1&web=1&e=xVy4Hu

damonmcc:
@fvankrieken good to close?

PR #1687, opened 2025-05-27
pluto 25v2 tweaks

fvankrieken:
#1564 

Closes
- #1562 
- #1680
- #1661
- #1660 
- #1693 - though I revert the change, as it looks like this week the issue was resolved in source data

Partially addresses #1638 - adds the new table. I'm not yet automating a check that makes sure that each zoning type in pluto is actually represented, for a couple reasons. I'll write up a new issue for that

Go by commit - descriptions in the commit messages when needed. There really 2 main tweaks - change to corrections logic because I was tired of writing boilerplate code, and the appbbl stuff. Well, and the actual corrections. Given the changes to the process, I did make sure that all these new corrections are still being applied.

[Latest run](https://github.com/NYCPlanning/data-engineering/actions/runs/15783708531/job/44495318672)


sf-dcp:
@fvankrieken lmk when it's ready for review

products/pluto/pluto_build/sql/corr_ownername_city.sql
```diff
@@ -1,216 +0,0 @@
...
-    b.type,
-    b.reason,
-    b.version
-FROM pluto AS a, pluto_input_research AS b
-WHERE
-    a.ownername = b.old_value
-    AND a.bbl NOT IN (
-        SELECT bbl FROM pluto_changes_applied
-        WHERE field = 'ownername'
-    );
```
sf-dcp, line 18: this query is bit weird when compared to other corrected fields... It matches based on owner name which could technically be the same across multiple bbls. But then we correct owner name in pluto based on bbl & matched ownername. 

All to say, your new implementation for `pluto_changes_applied` table (for `ownername` field) is correct but the records between new and old implementation may differ. 

On the side note, so happy to see these correction files gone!

products/pluto/pluto_build/data/pluto_input_research.csv
```diff
@@ -27321,8 +27321,6 @@ bbl,field,old_value,new_value,Type,reason,version
 2033310053,unitstotal,117,0,2,PTS error - vacant lots with units,24v2
 "3010407502",unitsres,100,10,2,PTS error - units,"24v2"
 "3010407502",unitstotal,100,10,2,PTS error - units,"24v2"
-"4017687502",unitsres,100,10,2,PTS error - units,"24v2"
-"4017687502",unitstotal,102,12,2,PTS error - units,"24v2"
```
sf-dcp, line 27325: I don't think we should remove these and the other 2 records below. If we need to re-build PLUTO for an older version (and older data), we will get different results

products/pluto/seeds/ignored_bbls_for_unit_count_test.csv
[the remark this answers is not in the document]
sf-dcp: let me think about this more and get back with my thoughts

products/pluto/pluto_build/data/pluto_input_research.csv
fvankrieken: That's not how the corrections are applied - the version is not used at all, meaning at any point in time if we attempt to rebuild older versions with a newer codebase, we'll get different results. The only way to rebuild with the same corrections as before currently is by checking out an appropriate commit/tag.

products/pluto/pluto_build/sql/corr_ownername_city.sql
fvankrieken: Yeah it is a little odd and I thought about if maybe we should have two separate cases - the "normal" corrections and the null bbls (applied to many records) ones. But the new logic does correctly add to pluto_changes_applied for either option

products/pluto/pluto_build/data/pluto_input_research.csv
fvankrieken: I mean I'm right too in the specifics of what I was saying ... but totally misunderstood your point

fvankrieken: Yeah it makes sense that if a correction is wrong (as in we know that the new value is wrong, we should update the correction - delete the old one), but in cases like this they are both "valid" corrections. That being said, source control does handle this too, but we don't really operationally make use of that in rebuilding. So I'm with you on this. I'll add them back

fvankrieken: Updated in https://github.com/NYCPlanning/data-engineering/pull/1687/commits/e06f767ef508666ec3aa129669ad2191b12dc2c8

products/pluto/pluto_build/sql/corr_template.sql
```diff
@@ -16,13 +18,39 @@ WHERE
...
     :FIELD = b.new_value,
     dcpedited = 't'
 FROM pluto_input_research AS b
 WHERE
-    a.bbl = b.bbl
+    b.bbl = a.bbl
+    AND b.field = :'FIELD'
+    AND a.:FIELD IS NOT DISTINCT FROM b.old_value;
+
+----- null bbl in corrections - all that match -----
```
sf-dcp, line 31: do we only correct owner name with bbl not being specified in the corrections file?

Issue #1690, opened 2025-05-28
CPDB: Agency Lookup, fix UK and EDC


TangoYankee:
# Description
OMB recently shared [a spreadsheet of agency codes](https://cloud.digitalocean.com/spaces/ae-data-backups?path=zoning-api%2F). Comparing the [cpdb agency lookup](https://github.com/NYCPlanning/data-engineering/blob/a6e4ede09526a460003c42a27477536ddbe7ba72/products/cpdb/data/agencylookup.csv) to this spreadsheet, we can improve a few of the codes in the agency lookup. Notably, not every code maps to a true agency.

## Acceptance criteria for agency lookup 
- [ ] Change `cape` values for code `111` from `UK/Unknown` to `City Wide Equipment Purchases/TBD`
- [ ] Change `cape_agency` value of `96` from `UK/Unknown` to `Human Resources Administration/HRA`
- [ ] Change `cape_agency` value of `998` from `EDC/Economonic Development Corporation` to `Transit Authority/NYCTA`

TangoYankee:
While we're looking at CPDB agencies, the build includes an [override for the agency](https://github.com/NYCPlanning/data-engineering/blob/b987bac6bab3dc5cbc0a9907f9c2b9529892c9e9/products/cpdb/sql/ccp_projects.sql#L48) when the description claims it is for the Trust for Governor's Island (TGI) or Brooklyn Navy Yard (BNY). However, it does not override other agencies, such as when `111` (city-wide services) applies to a specific agency or when `071` is for "Volunteers of America".

Unless there is a specific reason for overriding the agency for TGI or BNY, I think we should be consistent in overriding agencies. We should either override every agency or override none of them. Practically, I think we should override none of them. It is impractical to manually check every description, identify when it includes an organization, and override the agency value. It also complicates normalization of the data; agency codes appear to be reused for different agencies when they're only supposed to be used for a single agency.

fvankrieken:
> While we're looking at CPDB agencies, the build includes an [override for the agency](https://github.com/NYCPlanning/data-engineering/blob/b987bac6bab3dc5cbc0a9907f9c2b9529892c9e9/products/cpdb/sql/ccp_projects.sql#L48) when the description claims it is for the Trust for Governor's Island (TGI) or Brooklyn Navy Yard (BNY). However, it does not override other agencies, such as when `111` (city-wide services) applies to a specific agency or when `071` is for "Volunteers of America".
> 
[...]

PR #1691, opened 2025-05-29
CPDB: Update agencylookup (#1690)

TangoYankee:
* Ground-truth OMB source data for agency codes

* Add class column for description of agency code

* Add agency codes from OMB source data


sf-dcp:
Addressed in #1854

Issue #1601, opened 2025-04-20
DevDB A2 jobs


damonmcc:
ITD Project Intake tracking #[PR2025-01-14-012](https://nyco365.sharepoint.com/sites/NYCPLANNING/electronic-submission/Lists/Project%20Intake/DispForm.aspx?ID=149&e=x1e3QE)

## problem

There is a class of DOB permits (A2s) that are currently not included in the Housing Database, but which do affect unit counts in the city. Housing Division has reviewed many of the past A2 permits and would like to incorporate them into the HDB going forward.

We need to build a process with DE to identify future permits, make corrections to them, and add them to the HDB to give a fuller picture of housing changes.

The next DevDB version will be 25Q2 and is scheduled for July 2025.

## definition of done

- [x] identify A2 jobs from DOB NOW and add them to the current A2 export
- [x] Add corrections mechanism for A2 jobs

#### DOB BIS records (`dob_jobapplications`)

These are records from DOB's old system and have job type values `A1, A2, DM, NB`. We get them from Open Data page [`ic3t-wcy2`](https://data.cityofnewyork.us/Housing-Development/DOB-Job-Application-Filings/ic3t-wcy2/about_data) and filter records to only include those job types while archiving to `edm-recipes`.

#### DOB NOW applications (`dob_now_applications`)

There are records from DOB's new system and have job type values seen below. We get them from DOB via FTP and archive to `edm-private`.

<img width="459" alt="Image" src="https://github.com/user-attachments/assets/8a6ce6f2-85b9-4870-a019-19d45f8330b4" />

#### DOB NOW permits (`dob_now_permits`, `dob_permitissuance`)

`dob_now_permits` has useful columns like `permit_type`. `dob_permitissuance` also has a `jobtype` column which has values like A2.

Records from `dob_permitissuance` with recent issuance dates:
<img width="844" alt="Image" src="https://github.com/user-attachments/assets/56142c5d-9795-4a99-be13-f86c7cb35cd6" />

We currently only use NOW permits to assign/recode date fields: `date_permittd`, `permit_year`, `permit_qrtr`. Permits are joined to applications using `job_number` (modified version of the source column `job_filing_number`).

### filtering BIS and NOW source data

[...]

PR #1703, opened 2025-06-04
Dm devdb a2

damonmcc:
related to #1601


products/developments/sql/_now.sql
```diff
@@ -73,10 +75,12 @@ DROP TABLE IF EXISTS _init_now_devdb;
 
 WITH source_data AS (SELECT * FROM dob_now_applications),
 
+work_types_lookup AS (SELECT * FROM lookup_now_types),
+
 jobs_of_interest AS (
     SELECT * FROM source_data
     WHERE
-        (jobtype IN ('New Building', 'Full Demolition') OR jobtype ~* 'CO')
+        lower(jobtype) != 'no work'
```
damonmcc, line 83: this where clause is now more permissive so that A2 jobs from NOW are no longer dropped so early

personally, this is such a "staging" table I'd rather not drop any records at all but I'll save that for for dbt refactor

products/developments/sql/_units.sql
```diff
@@ -94,12 +96,19 @@ SELECT
             OR (classa_init IS NOT NULL AND classa_init != '0')
             OR (classa_prop IS NOT NULL AND classa_prop != '0')
             THEN 'Residential'
-    END AS resid_flag
+    END AS resid_flag,
+    job_type IN ('Alteration (A2)', 'Alteration (Non-CO)') AS presumed_a2_alteration,
```
damonmcc, line 100: this boolean column is how all `export_a2_devdb` records are selected

products/developments/data/lookup_now_types.csv
```diff
@@ -0,0 +1,24 @@
+code,description
```
fvankrieken, line 1: Where did these come from?

sf-dcp: I believe from [here](https://www.nyc.gov/site/buildings/industry/permit-type-and-job-status-codes.page).

fvankrieken: Cool ty.

Hmm...  nice gfi/intern issue would be adding an ingest source/file type - html table!

damonmcc: I forgot I had an idea to add a `source` column to this, especially since the DOB website doesn't cover all codes so we might need to add some `source = DCP` mappings

I'll do that later after getting input/feedback from Housing team on the mappings

sf-dcp (approved):
LGTM!

products/developments/sql/_units.sql
```diff
@@ -94,12 +96,19 @@ SELECT
...
+    CASE
+        WHEN datasource = 'bis' THEN TRUE
+        WHEN datasource = 'now' THEN (
+            job_desc IS NOT NULL
+            AND lower(job_desc) LIKE '%combin%'
+            AND lower(job_desc) NOT LIKE '%sprinkler%'
+        )
+    END AS meets_bis_a2_inclusion_rules
```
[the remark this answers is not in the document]
damonmcc: happy to change the `CASE... WHEN` to the conditional approach. but I think I'll need to keep the `job_desc IS NOT NULL` part because otherwise, the column could be either `True/False/NULL`. since this about a binary exclusion in BIS data, I'd rather have no NULLs

screenshot of comparing the two ways to make `meets_bis_a2_inclusion_rules`:
Screenshot 2025-06-23 at 1 16 55 PM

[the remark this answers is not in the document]
fvankrieken: Oops - yeah makes complete sense

PR #1712, opened 2025-06-06
LION - perpendicular to midpoint offsets

### LINKED ISSUE #1711: LION - 2.2.3/Atomic Polygons/Offset Points
I implemented a slightly different algorithm than the one in the ETL. The original should be implemented. Further questions around changes/enhancements to the algorithm discussed [insert issue here].

There are 4 issues
- [x] different algorithm -> #1712. Only 4 rows actually change
- [ ] with implementation of original algorithm, 1 row that matched before no longer matches prod
- [x] borough boundary issue -> #1712
- [ ] "missing" aps. 58 for right polygons, 43 for left. See #1752 
  - [ ] this should resolve all remaining "Boro Boundary Indicator" differences as well, currently 93

fvankrieken:
Covers about 2/3 of #1711.

## Summary
per meetings with GSS, issues with the algorithm for joining atomic polygons was about the top level of concern of any remaining issues. I had tried to quickly do this before meeting with them, but was unable to wrap up.

I originally implemented a different algorithm (see #1568). For this, I've implemented the original as a plpgsql function, which
- finds midpoint of line
- finds segment of line which contains midpoint
- determines slope of line perpendicular to this segment
- generates offset points of a given distance from the midpoint along this perpendicular

It's much faster than my original implementation (18s vs 80), and I think if we just lower the offset distance we should resolve most weird edge cases. And anyways, for now, we just want to go that route.

I had hoped this would resolve all 200ish issues of mismatched AP fields from the POC and from prod, but it only fixed 4 rows. In digging, I discovered that I didn't read the ETL docs carefully enough. When a lot is on a borough boundary (but is adjacent to APs in two different boroughs), all fields relating to the AP that is not in the same borough as the centerline segment should be null

## Reviewing
Just go by commit, should be straight forward


products/lion/models/product/lion.sql
```diff
@@ -172,21 +172,30 @@ SELECT
...
 FROM centerline
-LEFT JOIN atomic_polygons AS ap ON centerline.segmentid = ap.segmentid
 LEFT JOIN nodes ON centerline.segmentid = nodes.segmentid
 LEFT JOIN saf ON centerline.segmentid = saf.segmentid AND centerline.boroughcode = saf.boroughcode
 LEFT JOIN curve ON centerline.segmentid = curve.segmentid
 LEFT JOIN streets ON centerline.segmentid = streets.segmentid
 LEFT JOIN sedat ON centerline.segmentid = sedat.segmentid AND centerline.boroughcode = sedat.boroughcode
 LEFT JOIN nypd_service_areas ON centerline.segmentid = nypd_service_areas.segmentid
 LEFT JOIN segment_locational_status ON centerline.segmentid = segment_locational_status.segmentid
+LEFT JOIN
```
fvankrieken, line 191: I could have edited the original AP int table instead of doing this multi join (as in, only join APs that are in the same borough). 

However, the int ap model actually still needs these values that are nullified when on the border, as the ap model is used in 2.2.6 - segment locational status. This double join doesn't add much time, it's a simple index lookup, and it seemed preferable to having a TON of `NULLIF`s in the select clause

sf-dcp:
From what I'm reading, it sounds like the change in boro boundary impacts 4 records only, correct? And nothing else fundamentally changed?

It's interesting that the speed with this implementation went up. Do you know why that could be? I'm just curious

PR #1725, opened 2025-06-12
allow overwrite for ingest

### LINKED ISSUE #1724: ZTL Failing: FileExistsError: Archived dataset XYZ already exists and has different data.
ZTL dataloading is failing. We had to rebuild with new zoning data, and it seems to be throwing an error based on an archival step.

Let me know how I can assist - this time sensitive.

```
FileExistsError: Archived dataset id='dcp_commercialoverlay' version='20250531' 
already exists and has different data.
Error: Process completed with exit code 1.
```

[Relevant run](https://github.com/NYCPlanning/data-engineering/actions/runs/15595906997)

### GIS Action Items
- [ ] Provide feedback on GH action override language

fvankrieken:
Closes #1724

[successful run](https://github.com/NYCPlanning/data-engineering/actions/runs/15612083539)

So our concept of immutability for ingest datasets doesn't really fit GIS's workflow. For ZTL, they
- process various zoning-related datasets
- load them to edm-publishing/datasets
- run "ztl loading" action, which
  - calls ingest on all these datasets
  - builds ztl

Finally, they QA ztl - this is treated as the QA of the source datasets.

So... if an issue is found, they correct data, overwrite the files in edm-publishing/datasets, and re-run ingest/build ztl. So they very much need to overwrite the files in edm-recipes.

Long term solution - this data maybe does not make it into edm-recipes until we "publish" ztl. I think this makes sense - ztl build could pull from `edm.publishing.gis` connector which already exists, and then we essentially have a dependency on the "publish" action that signals that we're good to archive data into edm-recipes for use in other data products

Short term - let GIS overwrite them with this checkbox

image

### DIFF
--- a/dcpy/lifecycle/ingest/run.py
+++ b/dcpy/lifecycle/ingest/run.py
@@ -101,12 +101,11 @@ def ingest(
     shutil.copy(dataset_staging_dir / config.filename, dataset_output_dir)
 
     version_exists = processed_datastore.version_exists(dataset_id, config.version)
-    if version_exists:
+    if version_exists and not overwrite_okay:
         validate.validate_against_existing_version(
             dataset_id,
             config.version,
             dataset_staging_dir / config.filename,
-            overwrite_okay=overwrite_okay,
         )
 
     if push and (overwrite_okay or not version_exists):
--- a/dcpy/lifecycle/ingest/run.py
+++ b/dcpy/lifecycle/ingest/run.py
@@ -30,6 +30,7 @@ def ingest(
     output_csv: bool = False,
     template_dir: Path | None = TEMPLATE_DIR,
     local_file_path: Path | None = None,
+    overwrite_okay: bool = False,
 ) -> Config:
     if template_dir is None:
         raise KeyError("Missing required env variable: 'TEMPLATE_DIR'")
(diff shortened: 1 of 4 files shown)


damonmcc (approved):

dcpy/lifecycle/ingest/run.py
```diff
@@ -102,17 +103,20 @@ def ingest(
     version_exists = processed_datastore.version_exists(dataset_id, config.version)
     if version_exists:
         validate.validate_against_existing_version(
-            dataset_id, config.version, dataset_staging_dir / config.filename
+            dataset_id,
+            config.version,
+            dataset_staging_dir / config.filename,
+            overwrite_okay=overwrite_okay,
```
alexrichey, line 109: @fvankrieken  Would it be cleaner to keep `validate_against_existing_version` as it was, and just alter the condition on 104 to be `if version_exists and not overwrite_ok:` ?

fvankrieken: Maybe! I sort of like still getting the warning from the validation, though I guess if you've run with overwrite you sort of explicitly don't care... so sure.

sf-dcp (approved):
LGTM

alexrichey (approved):

PR #1726, opened 2025-06-12
Cleave integration tests from dev container

fvankrieken:
Slightly moot point given that he eventually got things working outside of it, but wanted to separate integration tests from dev container so that Harris could more easily run the dev container (and also, it's good to be able to run them without launching a dev container)

With this, there are two ways to run these tests now. If you're in the container, all you need to do is to navigate to `dcpy/test_integration/docker` and run `docker compose up -d` and you're good to go. If you don't run the dev container, you'll need a little extra (the first time): `docker network create de && docker compose up -d`. After that, just the compose. Basically, needed to specify a docker "network" that gets created when you build the dev container, and that the integration tests look for when spinning up, so that if you're running both separately they can still talk to each other easily. There was probably another possible way, this seemed simplest.

I bumped the port for the postgis - I just personally kept getting port coliisions, I figured 2 above the default should be fine for most.

This doesn't resolve some slightly broader issues - it might make sense for these to get spun up and down (and destroyed) at test time - really try to make sure that the environment is fully controlled. Honestly, wouldn't be bad to use a roadmapping session or some other brainstorm to think about longer term what this should look like. But this seemed useful for now to try to keep our laptops from exploding.


fvankrieken:
> @fvankrieken Definitely in favor of anything that makes setup easier. Let me poke around with this locally for a bit, though. I need to re-think my dev environment setup anyway



> @fvankrieken Definitely in favor of anything that makes setup easier. Let me poke around with this locally for a bit, though. I need to re-think my dev environment setup anyway

See the readme in `dcpy/test_integration` for details on local setup

alexrichey (approved):
Per in-person chat, totally cool if you want to merge!

PR #1730, opened 2025-06-16
fix failing DevDB github action

damonmcc:
resolves https://github.com/NYCPlanning/data-engineering/issues/1729

action fails because I deleted bash utils from `products/developments` in a recent PR

two options were:
1. restore all bash utils from `devdb.sh`
2. refactor all use of it to use equivalent python utils

option 2 would be nice but isn't it easy enough to be fast. I did option 1 and only restored the bash utils that the github action needs

this action does other weird things like rely on library templates in `products/developments`, do we we can save a more thorough refactor later

[successful action run](https://github.com/NYCPlanning/data-engineering/actions/workflows/developments_datasync.yml?query=branch%3Adm-action-failures)

### DIFF
--- a/.github/workflows/developments_datasync.yml
+++ b/.github/workflows/developments_datasync.yml
@@ -89,8 +89,8 @@ jobs:
 
       - name: Load to Database
         run: |
-          ./devdb.sh import hpd_hny_units_by_building ${{ github.event.inputs.version }}
-          ./devdb.sh import hpd_historical_units_by_building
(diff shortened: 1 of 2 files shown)


fvankrieken (approved):
One note - would be nice to add type to template

damonmcc:
> One note - would be nice to add type to template

for sure, sadly I don't think it's an option yet like labels 😞

Issue #1717, opened 2025-06-09
distribute data updates (sprint L)


damonmcc:
see GIS's [Open_Data excel file](https://nyco365.sharepoint.com/:x:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7B6735537b-8d8e-437a-b5e5-3a6b54194b24%7D&action=edit&wdenableroaming=1&wdlcid=en-US&wdorigin=Teams-HL.Sharing.DirectLink.Copy.LOF&wdhostclicktime=1734629518613&wdredirectionreason=Force_SingleStepBoot&wdinitialsession=2a2b33b6-e1b0-d655-31ed-6ebe4e2c927f&wdrldsc=2&wdrldc=1&wdrldr=ServiceCreationTimedOut) in sharepoint for data updates in need of distribution

[github action](https://github.com/NYCPlanning/data-engineering/actions/workflows/distribute_socrata_from_bytes.yml) to distribute data from bytes 

[product metadata repo](https://github.com/NYCPlanning/product-metadata)

- [x] make an issue template to easily replicate this every sprint
- [x] Zoning
- [x] FacDB
- [ ] Edes (sheet says bytes has `20250609`, but bytes has `20250403`)
- [ ] CDBG (Finn's workin on it)
- [ ] MIH (Finn's workin on it)

damonmcc:
recent issues DE used to track distributing datasets:
- https://github.com/NYCPlanning/data-engineering/issues/1682
- ...

PR #1731, opened 2025-06-16
add issue template for data updates distribution

damonmcc:
An issue template makes it easy to create an issue like https://github.com/NYCPlanning/data-engineering/issues/1717 every sprint


fvankrieken (approved):

damonmcc:
after merging the option to use this wasn't showing up. an error message is visible when [viewing the file in github like this](https://github.com/NYCPlanning/data-engineering/blob/fcfad005a21f30f2a827382a396a6bb2b6fea035/.github/ISSUE_TEMPLATE/distribute_updates.yml): `Body must contain at least one non-markdown field`

will add a "sprint" field. unfortunately the value can't then be automatically placed into the issue title

PR #1733, opened 2025-06-16
add at least one field to the issue form

damonmcc:
follow-up to https://github.com/NYCPlanning/data-engineering/pull/1731


damonmcc:
nevermind, a form isn't worth it. changing it to a simple template

PR #1738, opened 2025-06-17
LION - join segments to nodes with tolerance

### LINKED ISSUE #1710: LION - 2.2.2/Nodes
127 records end up with a NULL (or all zeroes in the formatted table) "to" or "from" node id. This is likely a tolerance issue when nodes in the source data are not "snapped" to segment endpoints (or vice-versa).

```sql
select 
d."Borough",
d."Face Code",
d."Sequence Number",
d."Segment ID",
d."To-Node ID",
d."From-Node ID",
p."To-Node ID",
p."From-Node ID"
from  {schema}.lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Face Code", "Sequence Number", "Segment ID")
where d."To-Node ID" <> p."To-Node ID" or d."From-Node ID" <> p."From-Node ID";
```

We should
- first off, explore: is this really a tolerance issue? For these segments, can we do a spatial operation to join to nodes
- check with RZ - for these nodes, is this a data issue that should be fixed in the source data? If so, provide him with a report of segments/nodes (and their xy coordinates) that should be aligned
- If he feels that the source data is fine, implement an approach. I think doing this sequentially makes sense - first attempt joins with no tolerance, then attempt a join with a tolerance (to not do a more costly operation for all nodes)

Per chat with RZ, some small discrepancies are allowed. Definitely a potential future feature would be to evaluate how big a discrepancy is tolerable, but punting for now

fvankrieken:
closes #1710 - see issue for a description of the issue. Essentially, it was a tolerance issue with joining ends of segments to nodes.

Originally (before this PR), I joined essentially non-spatially - just checking point equality. This worked for 99+ % of rows, whereas the rest are not perfectly "snapped" in CSCL so need to be joined with some spatial tolerance.

For this PR, I had originally thought a "two-tiered" approach made sense

```mermaid
flowchart LR
    A[centerline] --> B[segment_2_node_only_exact]
    C[nodes] --join exact--> B
    B --without nodes--> D[segments_missing_nodes]
    D --> E[segment_2_node_with_tolerance]
    C --join with tolerance--> E
```

With the theoretical advantage being that the first join should be much more performant, and the second more costly spatial join could only happen on rows that didn't get joined to begin with. However, we could just do

```mermaid
flowchart LR
    A[centerline] --> B[segment_2_node]
    C[nodes] --join with tolerance--> B
```

Option 1 took 6 seconds to run, option 2 took 12... So not exactly a world of difference. Given that, I think it makes most sense to just to option 2

## Results
ALL centerline segment to and from nodes now match production!

### DIFF
--- a/products/lion/models/intermediate/2.2.2/int__centerline_segments_with_nodes.sql
+++ b/products/lion/models/intermediate/2.2.2/int__centerline_segments_with_nodes.sql
@@ -24,11 +24,10 @@ SELECT
     from_sm.sectional_map AS from_sectionalmap,
     to_sm.sectional_map AS to_sectionalmap
 FROM centerline
--- ~75 segments are missing nodes
--- TODO research whether this is consistent with current method and if we might need to join spatially with some minimal buffer
-LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_from ON centerline.from_geom = n_from.geom
-LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_to ON centerline.to_geom = n_to.geom
--- 6 segments with endpoint on border. This currently results in duplicate rows
+LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_from
+    ON st_dwithin(centerline.from_geom, n_from.geom, 0.001)
+LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_to
+    ON st_dwithin(centerline.to_geom, n_to.geom, 0.001)
 LEFT JOIN
     {{ source("recipe_sources", "dcp_cscl_sectionalmap") }} AS from_sm
     ON st_contains(from_sm.geom, centerline.from_geom)
--- a/ingest_templates/dcp_cscl_addresspoints.yml
+++ b/ingest_templates/dcp_cscl_addresspoints.yml
@@ -8,7 +8,7 @@ ingestion:
   source:
     type: s3
     bucket: edm-private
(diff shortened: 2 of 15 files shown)


products/lion/models/intermediate/2.2.2/int__centerline_segments_with_nodes.sql
```diff
@@ -24,11 +24,10 @@ SELECT
     from_sm.sectional_map AS from_sectionalmap,
     to_sm.sectional_map AS to_sectionalmap
 FROM centerline
--- ~75 segments are missing nodes
--- TODO research whether this is consistent with current method and if we might need to join spatially with some minimal buffer
-LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_from ON centerline.from_geom = n_from.geom
-LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_to ON centerline.to_geom = n_to.geom
--- 6 segments with endpoint on border. This currently results in duplicate rows
+LEFT JOIN {{ source("recipe_sources", "dcp_cscl_nodes") }} AS n_from
+    ON st_dwithin(centerline.from_geom, n_from.geom, 0.001)
```
fvankrieken, line 28: .001 (feet) still works. Trade-offs are too high and we could have duplicate rows (one segment joined to multiple nodes), too low and we could end up being too stringent with the tolerance. This currently seems like a nice trade-off. Though logically, .01 feet (1/8 inch) also seems perfectly reasonable. I say we start lower and if it ever starts failing, we can adjust

fvankrieken:
Run of ingest with updated templates: https://github.com/NYCPlanning/data-engineering/actions/runs/15858129694

PR #1756, opened 2025-06-25
add DevDB A2 corrections

### LINKED ISSUE #1745: correct DevDB A2 jobs
for DevDB builds, Housing team needs to be able to include and correct A2 job records. currently these records are filtered out from both BIS and NOW source data from DOB

two features:
- ability to whitelist records so they aren't excluded by A2 job filters
- ability to correct residential units of A2 job records (`units init` and `units prop`)

damonmcc:
resolves #1745


products/developments/sql/_units.sql
```diff
@@ -143,8 +154,8 @@ SELECT
     otherb_prop,
     resid_flag
 INTO _units_devdb
-FROM _units_devdb_resid_flag
-WHERE NOT presumed_a2_alteration;
+FROM _units_devdb_a2_details
+WHERE NOT (presumed_a2_alteration AND NOT include_a2_record);
```
damonmcc, line 158: this is where the inclusion of specified A2 jobs takes place

products/developments/data/archive/housing_input_hny.csv
fvankrieken, line 1: Let's maybe just delete this? It's in the git history forever if we really need to look back

products/developments/sql/_units.sql
```diff
@@ -121,11 +115,28 @@ CORRECTIONS
 CALL apply_correction(:'build_schema', '_UNITS_devdb_resid_flag', '_manual_corrections', 'resid_flag');
 
 /*
-Separate A2 job types from other types of records with units
+Determine all A2 job details
 */
+DROP TABLE IF EXISTS _units_devdb_a2_details;
+SELECT
+    _units_devdb_resid_flag.*,
+    job_type IN ('Alteration (A2)', 'Alteration (Non-CO)') AS presumed_a2_alteration,
```
fvankrieken, line 123: Should we maybe just call this "alteration" or something like that? Or maybe upstream just map these so that they all have jobtype "Alteration"? 

On that note, what do the job types look like in the final output? Are they like this? Or just "Alteration". My recollection is that they just get passed through the pipeline (after some manual adjusting in _bis and _now sql fiels), but that we'd want the latter, but I'm not quite sure

products/developments/sql/_units.sql
fvankrieken: Bit of a nit again, but just had to think for a second reading this to figure out what's happening. Seems we want one of two things to be true
- record is not an alteration (keep)
- record is an alteration but it's whitelisted

So maybe just `WHERE (NOT presumed_a2_alteration) OR include_a2_record`? To me helps the readability

fvankrieken:
Are you still planning on handling blacklisting functionality as well? As in, A2s/other alterations they've seen in the export that they've marked as not relevant to HDB and should not be included A2 export in the future

products/developments/sql/_units.sql
damonmcc: I like that, this felt hard to read. and happy to revisit column names for consistency/readability

damonmcc:
> Are you still planning on handling blacklisting functionality as well? As in, A2s/other alterations they've seen in the export that they've marked as not relevant to HDB and should not be included A2 export in the future

blacklisting a specific job: no that hasn't come up in my chats with Housing on what they need

blacklisting by categorically filtering out records like `_bis.sql` already does: yes but I'm gonna address that separately and they confirmed it isn't as important as being able to include+correct A2s

fvankrieken:
Just because otherwise, they'll have to wade through all the records that they've gone through before and identified as not wanting to include in HDB each time they do A2 QA

products/developments/sql/_units.sql
damonmcc: see upstream mappings in [_bis.sql](https://github.com/NYCPlanning/data-engineering/blob/main/products/developments/sql/_bis.sql#L72-L77) and [_now.sql](https://github.com/NYCPlanning/data-engineering/blob/main/products/developments/sql/_now.sql#L91-L95). A1's have historically been mapped to "alteration" so I didn't change that

changing these mappings is out of scope for this PR and the issue it address so I made #1760 to revisit this

damonmcc:
@fvankrieken 
> Just because otherwise, they'll have to wade through all the records that they've gone through before and identified as not wanting to include in HDB each time they do A2 QA
>
> Hmm okay. Can you check with Sam? When Sasha and I met with them to kick this off, that was something that we had said we'd do. Noted in [#1601 (comment)](https://github.com/NYCPlanning/data-engineering/issues/1601#issuecomment-2935345791)

either way (whether they want it or not and I'll confirm in our meeting with them on Monday), I think it's a nice-to-have, isn't essential for the upcoming first inclusion of A2s in DevDB 25Q2, and can be added iteratively on top of the changes in this PR. I made #1761 to revisit the blacklist feature

products/developments/data/corrections/hny_corrections.csv
sf-dcp, line 1: does renaming of this file impact the QAQC devdb page in any way? Do we use this file there?

sf-dcp (approved):
LGTM (contingent on response to my question about a corrections file renaming)

fvankrieken:
> either way (whether they want it or not and I'll confirm in our meeting with them on Monday), I think it's a nice-to-have, isn't essential for the upcoming first inclusion of A2s in DevDB 25Q2, and can be added iteratively on top of the changes in this PR. I made #1761 to revisit the blacklist feature

Sounds good

Issue #1768, opened 2025-06-30
DevDB A2 job filters


damonmcc:
In DevDB we label records as A2s and inconsistently filter them out. The csv of A2 jobs that we export for Housing team review is large and outdated.

The Housing team
- has suggestions to improve this filtering to help them focus on reviewing the records that are most likely to be worth including in DevDB
- would like us to stop removing BIS records before they get to the A2 export

## approach

In general, it's preferable to not exclude records from review and to use boolean columns to indicate filter matching

Suggested filter ideas:

1. comfortable focusing on only those work types that include General Construction (General Construction; General Construction/Structural; General Construction/Mechanical Systems; etc)
2. indicate and don't drop records with "combin"
3. look for instances of "separate apartment" and "separate unit"
4. look for terms like "restore", "merge"

damonmcc:
don't think I should change the current exclusion of A2 BIS jobs. as seen in the table below, 800K of the 1.5M A2 records are currently excluded because they have `description NOT LIKE '%sprinkler%' + description NOT LIKE '%combin%'`

|jobtype|desc_not_sprinkler|desc_combin|count|
|-------|------------------|-----------|-----|
|A2|false|false|112,459|
|A2|false|true|1,349|
|A2|true|false|857,404|
|A2|true|true|18,109|

PR #1771, opened 2025-06-30
refactor DevDB BIS script

damonmcc:
related to https://github.com/NYCPlanning/data-engineering/issues/1768

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-devdb-a2-filters)

This is just to make the script easier to read and modify. I did a similar refactor for the companion NOW script [here](https://github.com/NYCPlanning/data-engineering/blob/main/products/developments/sql/_now.sql).

### `_init_bis_devdb` table not changing after refactor

Screenshot 2025-06-30 at 2 08 26 PM

### DIFF
--- a/products/developments/sql/_bis.sql
+++ b/products/developments/sql/_bis.sql
@@ -65,185 +65,276 @@ OUTPUTS:
 
 
 DROP TABLE IF EXISTS _init_bis_devdb;
-SELECT
-    bis.jobnumber::text AS job_number,
 
-    -- Job Type recoding
-    CASE
-        WHEN jobtype = 'A1' THEN 'Alteration'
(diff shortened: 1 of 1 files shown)


fvankrieken:
I have to admit... I actually find the original easier to read. There's a single select statement with the logic for how to handle fields, a single join (so only two entities that data could be coming from, as opposed to all the CTEs), and a relatively simple `where` clause at the end. I get that the CTEs help logically chunk things a little, but sometimes I think they can also make things more obscure - you have more entities (tables/named clauses in the CTEs) to manage, more ambiguity about what logic is happening where, etc.

PR #1773, opened 2025-07-01
Compile python requirements

fvankrieken:


dcpy/lifecycle/package/validate.py
```diff
@@ -220,12 +220,11 @@ def validate_df(
             accepted_values = {str(v.value) for v in col.values} | {""}
             invalids = df_no_col_nulls[~df_no_col_nulls[col.name].isin(accepted_values)]
             if not invalids.empty:
-                invalid_counts = dict(invalids.groupby(by=col.name).size())
-                invalid_counts_int = {k: int(v) for k, v in invalid_counts.items()}
```
fvankrieken, line 224: @alexrichey I think this cast was just to handle mypy originally right? I think this solution is better - just gives what you want (dict of index: value)

alexrichey: seems plausible! This looks better though

alexrichey (approved):

Issue #1715, opened 2025-06-06
LION - 2.2.10/SAF


fvankrieken:
## Field Description
SAF - Special Address Flag. Look for in the docs for more

## Resolution
We'll need to check in with RZ and GSS on two things 
- [x] Main issue is that there are 3 rows with 'F' instead of 'D'. However, 'D' flags only come from altsegmentdata table, and no entries with value 'D' are present for the given segmentids (or their lionkeys). Not sure what's going on here, 
- [x] For the final category of saf, the language describes "populate L48 with the first SAFTYPE value encountered". Not entirely sure what "first" means in this case. This is muddied by the fact that there are 3 tables to join -> in one table, we could take row number as priority. I've gone with altsegmentdata > addresspoints > commonplace for these cases, which seems to recreate the outputs. But would like to confirm that this is desired.

fvankrieken:
```sql
select 
d."Borough",
d."Sequence Number",
d."Segment ID",
d."Special Address Flag",
p."Special Address Flag"
from  fvk_lion_offsets.lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Face Code", "Sequence Number", "Segment ID")
where d."Special Address Flag" <> p."Special Address Flag"
```

damonmcc:
emailed GR/GSS about both issues

unless there's a data source we're missing to resolve the D/F difference, I don't expect we'll have any changes to make. awaiting confirmation from GR/GSS

damonmcc:
reply from GR on the D/F issue:

> The SAFTYPE "F" in CSCL is a "D" in LION/Geosupport. Duplicate Address Pseudo-Streetnames (DAPs) include two subcategories of Ds that are distinguished in CSCL as a "real" DAP and a "pseudo" DAP. I am assuming the "F" is for fake or faux, "P" is already taken up. For LION/Geosupport purposes, there is just a "D". It is likely the code treats the "F" as a "D" and just writes in a "D".

excerpt from a SAF PDF attached in that reply:

> Type D records involve situations in which there exist duplicate house numbers on the same
street.

that PDF has details for every SAF value. there is no entry for F

damonmcc:
@fvankrieken think we're good to close this? didn't get an explicit acknowledgment from GR/GSS on part 2 but they helped resolve part 1 so now all the outputs align

PR #1775, opened 2025-07-02
LION SAF value mismatch

damonmcc:
related to #1715

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-lion-saf)

GS confirmed that `Special Address Flag` values of `F` from CSCL should become `D` for LION/Geosupport

## `poc_dat_comparison` table showing zero mismatch records for this field
Screenshot 2025-07-02 at 4 10 18 PM

### DIFF
--- a/products/lion/bash/export.sh
+++ b/products/lion/bash/export.sh
@@ -3,7 +3,7 @@ source ../../bash/utils.sh
 set_error_traps
 
 # Export tables to DAT files
-# fixed-width text files with no header or delimeter are specific to LION
+# fixed-width text files with no header or delimiter are specific to LION
 function dat_export {
     local table=${1}
     local output_file=${2:-${table}}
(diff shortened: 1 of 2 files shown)


products/lion/models/product/lion.sql
```diff
@@ -109,7 +109,12 @@ SELECT
     CASE
         WHEN twisted_parity_flag = 'Y' THEN 'T'
     END AS twisted_parity_flag,
-    saf.special_address_flag,
+    CASE
+        -- The SAFTYPE "F" in CSCL is a "D" in LION/Geosupport
+        -- Future exports may expect this mapping so this may move upstream later
+        WHEN saf.special_address_flag = 'F' THEN 'D'
+        ELSE saf.special_address_flag
+    END AS special_address_flag,
```
fvankrieken, line 117: I think we should probably have this in int/2.2.10 - still with the comment, but then if we need to make changes in the future it's in the SAF-specific model

damonmcc: I couldn't remember if we'd landed on moving it. will do

damonmcc: changed the comment since a future change would be less about moving it upstream

fvankrieken (approved):

PR #1783, opened 2025-07-07
CPDB Versioning: Part 2

sf-dcp:
Follow-up to #1723

Baby PR to fix CPDB versioning across years. Previously, bumping of a CPDB version would only work for same year (e.g. `24exec` -> `24adopt`)

The new code lets us bump like `25adopt` -> `26prelim`.

~~Ignore the failing test: it's related to Socrata.~~

### DIFF
--- a/dcpy/test/utils/test_versions.py
+++ b/dcpy/test/utils/test_versions.py
@@ -171,6 +171,7 @@ class TestVersions(TestCase):
             [None, 2, "23Q4.1", "24Q2"],
             [None, 1, "24exec", "24adopt"],
             [None, 1, "24adopt", "25prelim"],
+            [None, 2, "24adopt", "25exec"],
             [None, 6, "23prelim", "25prelim"],
             [None, 2, "24prelim", "24adopt"],
             ["patch", 1, "24prelim", "24prelim.1"],
--- a/dcpy/test/utils/test_versions.py
+++ b/dcpy/test/utils/test_versions.py
@@ -170,6 +170,8 @@ class TestVersions(TestCase):
             ["minor", None, "23v2.2.1", "23v2.3"],
             [None, 2, "23Q4.1", "24Q2"],
             [None, 1, "24exec", "24adopt"],
+            [None, 1, "24adopt", "25prelim"],
+            [None, 6, "23prelim", "25prelim"],
             [None, 2, "24prelim", "24adopt"],
             ["patch", 1, "24prelim", "24prelim.1"],
             ["patch", 2, "24prelim.2", "24prelim.4"],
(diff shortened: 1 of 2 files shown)


dcpy/utils/versions.py
```diff
@@ -20,6 +20,20 @@ class CapitalBudgetRelease(IntEnum):
     exec = 2
     adopt = 3
 
+    @classmethod
+    def resolve_bumped_release(cls, value: int) -> CapitalBudgetRelease:
```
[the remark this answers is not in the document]
sf-dcp: Are you suggesting to move this logic under the `bump()` function?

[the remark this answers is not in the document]
fvankrieken: Oh sorry okay. But yes, and more my point is that the `if... else` is really just calling `CapitalBudgetRelease(value)`, which means this function is really just calling modulo and so likely isn't needed

dcpy/utils/versions.py
```diff
@@ -442,9 +442,20 @@ def bump(
                 f"Version subtype {bump_type} not applicable for Date versions"
             )
         case CapitalBudget(), None:
+            total_bumped_value = previous_version.release_num + bump_by
+            bumped_year = previous_version.year + ((total_bumped_value - 1) // 3)
+
+            # Map total bump value to a valid release value: adopt (0), prelim (1), exec (2)
+            if total_bumped_value % 3 == 0:
```
fvankrieken, line 449: I still think you should do away with this if statement - It's still basically just doing modular math, so with a little tweak this could probably be a one-liner akin to line 446

sf-dcp: Like this? 

```python
bumped_release = CapitalBudgetRelease(3) if total_bumped_value % 3 == 0 else CapitalBudgetRelease(total_bumped_value % 3)
```

FYI: following your comment somewhere in the thread, I'm also revising `release_num` to `CapitalBudgetRelease`. Will re-request review from you once it's ready

fvankrieken: Well the case of `adopt` isn't really unique, it's just shifted. So more

```python
bumped = (total - 1) % 3 + 1
```

(or the enum could be zero index - then we wouldn't need the +1 or -1, but that's not a big deal)

sf-dcp: The last one unfortunately won't work for cases where the bump value is greater than 3. The one prior to that is great, thanks for helping figure out the math :D

fvankrieken (approved):

dcpy/test/utils/test_versions.py
```diff
@@ -124,7 +149,9 @@ def test_sort_invalid_versions(self):
             ].sort()
         with self.assertRaises(ValueError):
             [
-                versions.CapitalBudget(year=25, release_num=1),
+                versions.CapitalBudget(
+                    year=25, release=versions.CapitalBudgetRelease(1)
```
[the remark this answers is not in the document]
sf-dcp: 💯

PR #1786, opened 2025-07-08
Add metadata shapefile utilities

### LINKED ISSUE #1785: Write initial shapefile metadata interaction utilities

jackrosacker:
Closes #1785 

### Context
Basic utility functions to interact with shapefile metadata, though stops short of interacting with the actual content of the metadata beyond writing to file.

#### Utilities
Adds three completed utilities:
1. `write_metadata()` - to xml file, can be set to overwrite or not when metadata exists
2. `read_metadata()` - from xml file
3. `metadata_exists()` - boolean test

And a stubbed out utility:
1. `remove_metadata()` - seems useful, can implement if this feels in scope

All utilities require a strict input shapefile path convention, and can handle both bare and zipped shapefiles. 

I opted to use the same input path convention that [geopandas uses](https://geopandas.org/en/stable/docs/user_guide/io.html#:~:text=zipfile%20%3D%20%22zip%3A///Users/name/Downloads/gadm36_AFG_shp.zip!data/gadm36_AFG_1.shp%22) for the `read_file` command for zip files, and added the requirement that the path passed to my functions must resolve all the way to a ".shp" file extension. No inferring of shapefiles from folder names, or whatever. My rules:

##### Supported input path patterns

The zip file, specifying an un-nested shapefile:
- `zip:///Users/name/Downloads/gadm36_AFG_shp.zip!gadm36_AFG_1.shp`

The zip file, containing at least one directory level, specifying a shapefile:

- `zip:///Users/name/Downloads/gadm36_AFG_shp.zip!data/gadm36_AFG_1.shp`

The directory, specifying a shapefile:

- `/Users/name/Downloads/gadm36_AFG_shp/data/gadm36_AFG_1.shp`

A bare shapefile within the current working directory:

- `gadm36_AFG_1.shp`

#### Tests
Added tests that cover all functions noted above, as well as some of the helper functions. Tests do not cover all possible permutations of utility usage, i.e. they currently test for:
- metadata already exists, and "overwrite" arg is False -> raise exception
- metadata does not exist, and "overwrite" arg is False -> write metadata

Does not yet test for:
- metadata already exists, and "overwrite" is True -> write metadata
- metadata does not exist, and "overwrite" is True -> write metadata

I read about using parametrized tests for this type of situation, but that felt excessive without checking in first. See the test for [metadata_exists()](https://github.com/NYCPlanning/data-engineering/blob/a2ff3cc9a09be8caeee56e94a80158b181df5d25/dcpy/test/utils/geospatial/test_shapefile.py#L159) for some repetitive nonsense where I opted for explicitly testing multiple conditions, but stopped short of parametrized tests.

### Questions
- Any glaring issues?
- Introduce parametrized tests for better coverage?
- All existing functions operate on shapefile in place - do we want to add an option to edit a copy of the shapefile?
- Do we like the rigid input path convention, or should I scale back and just parse for ".zip" to determine if shp is zipped or not?
- Add `remove_metadata()` function?

### Possible next steps

Tagging you @caseysmithpgh in case you want to check this out too!


alexrichey:
@jackrosacker Thanks for your patience! I'll add some comments today, and we can chat about next steps

fvankrieken:
Both of you - anything in particular you'd like me to pay attention to, or just general review?

Thanks for the detailed write-up @jackrosacker

fvankrieken:
Also [this](https://geopandas.org/en/stable/docs/user_guide/io.html#:~:text=zipfile%20%3D%20%22zip%3A///Users/name/Downloads/gadm36_AFG_shp.zip!data/gadm36_AFG_1.shp%22) is great - I actually didn't know this

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,322 @@
...
+            for item in old_zip.infolist():
+                if item.filename != file_to_remove:
+                    data = old_zip.read(item.filename)
+                    new_zip.writestr(item, data)
+    os.replace(temp_zip, zip_file)
+
+
+def _validate_shp_input(shp_string: str | Path) -> None:
+    """Ensure the input string conforms to the required format:
+    Format if zip file: 'zip://path/to/file.zip!shapefile.shp'
```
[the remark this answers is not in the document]
alexrichey: nm! reading the description, and this is for (geo)pandas

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,322 @@
...
+
+    # process zip files ------------------------
+    if shp_info["is_zip"]:
+        with ZipFile(shp_info["path_to_zip"], "r") as archive:
+            return archive.namelist()
+
+    # process non zipped files -----------------
+    elif not shp_info["is_zip"]:
+        if not shp_info["dir_containing_shp"]:
+            shp_dir = Path(os.getcwd())
```
alexrichey, line 183: so `os.getcwd()` is just return the directory where you're working, ie where you invoked Python. Is the intent here to just list the files under `path_to_shp`?

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,322 @@
...
+
+    Returns:
+        bool: True if shapefile has metadata, False if no metadata is found.
+    """
+    shp_info: dict = _parse_path_to_shp(path_to_shp=path_to_shp)
+    file_list: list[str] = _list_files_in_shp_dir(path_to_shp=path_to_shp)
+    xml_filename: Optional[str] = _get_metadata_xml_name(
+        file_list=file_list, shp_name=shp_info["shp_name"]
+    )
+    if xml_filename is not None and xml_filename in file_list:
```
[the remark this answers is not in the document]
jackrosacker: Nice! I like the simplicity of your suggestion. Your snippet is slightly harder for my brain to read, but that's due to unfamiliarity rather than a true preference

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,322 @@
...
+
+    Returns:
+        Optional[str]: Either name of metadata file, or None if no file is found
+    """
+    shp_stem = Path(shp_name).stem
+    file_list_names_only = [Path(item).name for item in file_list]
+    matched_files = [
+        item
+        for item in file_list_names_only
+        if item.startswith(shp_stem) and item.endswith((".xml", ".shp.xml"))
```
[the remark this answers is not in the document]
jackrosacker: Interesting. I think I like your suggestion as it is easier to read and understand what patterns are being matched to, instead of inferring from the starts/endswith. How is this: 
```python
[f for f in file_list if (f == f"{shp_stem}.shp.xml" or f == f"{shp_stem}.xml")]
```
Using the "in" operator was giving false matches since it returns substring matches as well, so both "shapefile.shp" and "shapefile.shp.xml" would be returned, triggering the ValueError exception

dcpy/utils/geospatial/shapefile.py
jackrosacker: Good catch. I haven't dug back into this yet, but it seems like a bug

jackrosacker: Now I'm recalling why I wrote this. Basically, if the function is passed a shapefile name directly with no parent directory (e.g. `shapfile_name.shp`), then the function assumes that the shapefile is in the same directory as the current working directory, and the "dir_containing_shp" value would be empty in the descriptive dict. This made sense to me at the time, but happy to reevaluate. Particularly since this is an internal fn that doesn't necessarily need to handle all potential user input cases. Happy to discuss.

Also realized two other issues with this fn while reviewing this line!

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,322 @@
...
+
+    Returns:
+        list[str]: List of files in same directory as specified shapefile
+    """
+    shp_info: dict = _parse_path_to_shp(path_to_shp=path_to_shp)
+
+    # process zip files ------------------------
+    if shp_info["is_zip"]:
+        with ZipFile(shp_info["path_to_zip"], "r") as archive:
+            return archive.namelist()
```
jackrosacker, line 178: Appears to only handle zipped shapefiles that are at the top level of the zip, and may not handle for parent subdirs within the zip. Evaluate to ensure this is handling nested subdirs as well

dcpy/utils/geospatial/shapefile.py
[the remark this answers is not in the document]
jackrosacker: Ok, that makes sense. I hadn't considered using a set in the list comprehension. Implemented.

alexrichey (approved):
Super clean and focused! I love how the cleanly the `Shapefile` methods read once the zip/not-zipped aspect is abstracted away.

Also, definitely Squash when you hit the merge button

jackrosacker:
Pushed a few more commits. Minor documentation + type hint stuff for the shapefile module, and more substantial changes for the tests.

Thanks for guiding me through all of this @alexrichey! Super valuable experience.

dcpy/utils/geospatial/shapefile.py
```diff
@@ -0,0 +1,265 @@
...
+            _FileManagerZipped(path, zip_subdir)
+            if self.is_zipped
+            else _FileManager(path)
+        )
+        self._post_init()
+
+    def _post_init(self):
+        self.has_metadata = self.metadata_exists()
+
+    def __repr__(self):
```
alexrichey, line 93: One nit / question: why override __repr__?

alexrichey (approved):
good stuff! Had one question/nit unrelated to recent commits, but other than that, ltgm

dcpy/utils/geospatial/shapefile.py
jackrosacker: My understanding was the the default __repr__ method just returns a memory address (e.g. <dcpy.utils.geospatial.shapefile.Shapefile object at 0x1056f4b90>), and that it can or should be overwritten to provide a more meaningful output. Am I misunderstanding how this works? Should I be leaving the __repr__ alone and instead defining my own __str__?

jackrosacker: Gotcha. Makes sense! I just removed the override lines and committed that change. Should be all set to go now

Issue #1752, opened 2025-06-24
CSCL Atomic Polygons - validity/topology issues


fvankrieken:
There are 11 atomic polygons that have empty geometry by the time they're ingested into our processes. I've spot checked 1 of them, it seems to be happening immediately upon reading the gdb into a geopandas GeoDataFrame

ATOMICIDs
```
1000500007
1002901578
1023600116
2006000957
3005601005
3066600116
4019902917
4022001905
4038302512
4107202004
5990100021
```

Of these, I see a few patterns in the underlying gdb data

## Invalid geometries
```
1000500007
1023600116
4107202004
5990100021
```

These all have self-intersections (NOT "ring self-intersection" - there are about 50 or so of those that I noticed, but we seem to ingest those just fine)

Here's the underlying data
```csv
wkt_geom,OBJECTID,BOROUGH,CENSUSBLOCK_2000,CENSUSBLOCK_2000_SUFFIX,CENSUSTRACT_2000,CENSUSBLOCK_2010,CENSUSBLOCK_2010_SUFFIX,CENSUSTRACT_2010,CENSUSTRACT_1990,ADMIN_FIRE_COMPANY,WATER_FLAG,CREATED_BY,CREATED_DATE,MODIFIED_BY,MODIFIED_DATE,ASSEMDIST,ELECTDIST,SCHOOLDIST,COMMDIST,SB1_VOLUME,SB1_PAGE,SB2_VOLUME,SB2_PAGE,SB3_VOLUME,SB3_PAGE,ATOMICID,ATOMIC_NUM,GLOBALID,HURRICANE_EVACUATION_ZONE,CENSUSTRACT_2020,CENSUSBLOCK_2020,CENSUSBLOCK_2020_SUFFIX,COMMERCIAL_WASTE_ZONE,SHAPE_Length,SHAPE_Area,_errors
[...]

Issue #1721, opened 2025-06-11
LION - 2.2.6/Segment locational status


fvankrieken:
Hopefully no changes needed for this, but it deserves its own issue for thoroughness

- [ ] One caveat - the current ETL application erroneously uses 2010 cts instead of 2020. Do we update this to use 2020 cts instead of 2010? If so, when?

162 records (as of 6/5 - 161 as of 6/11) still don’t match
```sql
select 
d."Borough",
d."Face Code",
d."Sequence Number",
d."Segment ID",
d."Segment Locational Status",
p."Segment Locational Status"
from  fvk_lion_offsets.lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Face Code", "Sequence Number", "Segment ID")
where d."Segment Locational Status" <> p."Segment Locational Status";
```

## "9" values
- [ ] 95 have value ‘9’ in POC -> AP issue (missing APs is the only thing that could cause this). See #1711 

## "I" instead of "H"
- [x] 63 ‘I’ instead of ‘H’ -> ~~potentially missing node issue, see #1710~~ actually, it's #1788 

All nodes are being correctly joined now so it looks like this might be either a misinterpretation by me of the ETL DOCs in writing up requirements originally, or a bug in our code. The docs state

> If the ATOMICIDs of the left and right AP’s are identical, the ETL tool will populate L41 with one of the values ‘H’ or ‘I’, as follows.  It will assign the value ‘H’ if there is at least one other segment of any kind connected to each of the given segment’s two endpoints.  If the segment has at least one endpoint to which no other segments are connected, it will assign the value ‘I’. 

Given that upstream, we join segments to nodes, the current implementation
- looks up nodes by endpoints
- looks up segments by nodes
- determines "joined" segments to starting segment based on this node linkage

In the case that no rows are joined then, we should label "I", but apparently this is missing some cases. Might take some digging in qgis (identify relevant segments, pull in node layer, see what isn't getting joined)

## NULL instead of "H" or "I"
- [x] 4 have no status, H or I in prod. H or I is when left and right AP are same -> seems like discrepancy in AP algorithms. Part of #1711, should be resolved by #1712
  - 1 seems resolved by #1711. 3 remaining
  
|Borough|Face Code|Sequence Number|Segment ID|Segment Locational Status|Segment Locational Status|
|-------|---------|---------------|----------|-------------------------|-------------------------|
|2|1049|00010|0342525| |H|
[...]

PR #1788, opened 2025-07-09
LION Segment Locational Status - use all segment feature layers

fvankrieken:
closes portion of #1721 ("H" vs "I" bit - the other bits seem to be tied up with #1752, which we don't have a complete set of inputs/outputs from the fixed data yet so can't quite test)

## Problem

Part of the docs for "Segment Locational Status" state
> If the ATOMICIDs of the left and right AP’s are identical, the ETL tool will populate L41 with one of the values ‘H’ or ‘I’, as follows. It will assign the value ‘H’ if there is at least one other segment of any kind connected to each of the given segment’s two endpoints. If the segment has at least one endpoint to which no other segments are connected, it will assign the value ‘I’.

Originally, this was implemented only using other centerline segments to see if there were "orphaned" nodes, but the actual intent of the docs (and C# code) is that this needs to check if ANY other segments are attached to the ends of a given centerline segment.

## Implementation

Because of the way these tables were organized, I restructured slightly. Links between segments and nodes as a more normalized table were being created downstream of a more centerline-specific table. Given that this seems like a pretty foundational relationship to establish, I thought I'd do this upstream, for all 5 segment source feature layers.

A key point on the wording of some of these tables
- "segments_with_nodes" generally is a wide table - one row per segment, one column for "to" node, one column for "from"
- "segments_to_nodes" - lookup table from segment to node. Many to many

So again, before, we made segments_with_nodes first and made the more generalized lookup table after (since this lookup table was needed downstream (though it never really materialized. It was part of a CTE)

```mermaid
flowchart LR
centerline --> centerline_with_nodes
nodes --> centerline_with_nodes
sectionalmap --> centerline_with_nodes
centerline_with_nodes --> segments_to_nodes
centerline --> seg_loc_status
segments_to_nodes --> seg_loc_status
```

BUT there was a flaw - this lookup table actually needed to be a lookup table for ALL segments, not just centerline

```mermaid
flowchart LR
centerline --> segment_geoms
shoreline --> segment_geoms
rail --> segment_geoms
subway --> segment_geoms
nonstreetfeatures --> segment_geoms
segment_geoms --> segments_to_nodes
nodes --> segments_to_nodes
centerline --> centerline_with_nodes
segments_to_nodes --> centerline_with_nodes
sectionalmap --> centerline_with_nodes
centerline --> seg_loc_status
segments_to_nodes --> seg_loc_status
```

### Performance
Not that we have strict standards here but I was curious
| Table | Build time in main / s | Build time in dev / s |
| - | - | - |
| segment_geoms | - | 0.39 |
| segments_to_nodes | - | 17.81 |
| centerline_segments_with_nodes | 11.24 | 4.23 |
| segment_neighbors | 3.15 | - |
| seg_loc_status | 1.99 | 2.82 |
| total (for relevant tables) | 16.38 | 25.25 |

A little slower in prod, but we are also just joining more segments to nodes than we were before (~220k segments instead of ~180000), so I think this is completely reasonable

### Data
```sql
select 
d."Borough",
d."Face Code",
d."Sequence Number",
d."Segment ID",
d."Segment Locational Status",
p."Segment Locational Status"
from  fvk_lion_seg_loc_status.lion_dat_fields d
inner join production_outputs.citywide_lion p
using("Borough", "Face Code", "Sequence Number", "Segment ID")
where d."Segment Locational Status" <> p."Segment Locational Status";
```

Is down to 96 rows. They all seem to be related to atomic polygons with empty geometries, we no longer have any cases with "I" instead of "H"


products/lion/models/intermediate/2.2.2/int__centerline_segments_with_nodes.sql
```diff
@@ -11,6 +11,10 @@ WITH centerline AS (
         st_startpoint(geom) AS from_geom,
         st_endpoint(geom) AS to_geom
     FROM {{ ref("stg__centerline") }}
+),
+
+segments_to_nodes AS (
```
alexrichey, line 16: Sorry, another performance question, but I'm curious about the CTE impact on performance. When I was doing coincident segment stuff, CTEs had a huge negative performance impact. 

e.g. when I pulled `{{ ref('stg__centerline') }}` into a CTE in `products/lion/models/intermediate/2.6/int__centerline_coincident_subway_or_rail_matches.sql` it took query time from 5 secs to 10

fvankrieken: I had something similar to that in another file in a past PR, though this query is decently quick. I'll check through

alexrichey (approved):
Closer and closer to having a graph database

products/lion/models/intermediate/int__segments_to_nodes.sql
```diff
@@ -0,0 +1,36 @@
...
+        st_endpoint(geom) AS geom
+    FROM segments
+)
+
+SELECT
+    seg.segmentid,
+    seg.direction,
+    nodes.nodeid
+FROM segment_endpoints AS seg
+INNER JOIN nodes ON st_dwithin(seg.geom, nodes.geom, 0.01)
```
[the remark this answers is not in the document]
fvankrieken: Actually given it being just a 10 s difference and this being one of the meatier spatial joins, I think I'll punt this. It's making use of the nodes spatial index which makes a big difference over neither

alexrichey:
image

Issue #1270, opened 2024-11-22
Add sftp `connector`


fvankrieken:
## The problems

- We need to distribute PLUTO major releases to an SFTP server
- We need to ingest source data from SFTP servers
- We want to showcase the potential for various destination types with `dcpy.lifecycle`/metadata

## Acceptance Criteria
- [x] the SFTP connector can push, pull, and list files form SFTP servers (don't push to DOF's ginger server)
- [x] easy to run only unit tests or only integration tests

## Not In Scope

- Deciding exactly what "integration" means. New tests here will require dependencies to exist a server) unlike unit tests.
- Refactoring existing tests that require external dependencies 

### Models

In `dcpy/models/lifecycle/ingest.py` define a model for `SFTPServer`

### Connectors / Lifecycle

In `dcpy/connects/ftp.py` add functionality to `FTPConnector` stub to push and pull from an SFTP server.

[`products/pluto/pluto_build/bash/ssh.sh`](https://github.com/NYCPlanning/data-engineering/blob/main/products/pluto/pluto_build/bash/ssh.sh) is the example of how we currently pull from an SFTP server.

All servers we currently access in production require a private key, so that is the primary profile/credential configuration we need to handle. We currently:
- use 1password to store keys
- load them as environment variables
- save them as a file like `/.ssh/ginger`
- use them in invocations of the `sftp` CLI

The implementation of `FTPConnector` will align with this handling of credentials and use the python package [`pysftp`](https://pysftp.readthedocs.io/)

### Tests

Add tests `dcpy/test/connectors/` to confirm the push and pull of files (using similar patterns as other tests which confirm the existence of files)

Add new/use existing metadata and ingest templates to test the models and connector (e.g. `garlic_sftp` in multuple `dcpy/test/` locations)

Ensure `DistributionSFTPConnector` is as tested as `SocrataPublishConnector`

Use a dockerized SFTP server for tests to interact with instead of mocked responses.

### Nice to have

Run tests which require an SFTP server during PR tests in `test_helper.yml` via docker compose. If we can run relevant tests in a dev container locally, that may be good enough for now.

Use safeguards like [pytest-socket](https://github.com/miketheman/pytest-socket) to prevent undesired server calls.

Handling different types of server profile/credential configuration (e.g. username + password without a key).

---

There are a few needed use cases for this. 
1. distributing pluto to ginger server - see #1226
2. ingesting pluto inputs from same server - see #1261 

[...]

PR #1797, opened 2025-07-14
Implement SFTP Connector

sf-dcp:
Related to #1270.

### What
This PR implements an SFTP connector (finally!). Specifically:
1. Move sftp utils from the `connectors/sftp.py` file to a utils file.
2. Get rid of `SFTPServer` & `SFTPUser` classes that were used to represent sftp attributes. They don't really add any value with the actual connector being implemented.
3. Create the actual connector with the expected methods.
4. Create integration tests for the connector.
5. ~~Register connector in registry.~~ <-- see notes below.

I recommend reviewing the PR commit by commit. First 4 commits are QOL. 

### Testing
I added 5 integration tests: 3 of them for the new sftp util functions, and 3 - testing the connector. I didn't add tests for all 4 connector methods since all of them essentially are wrappers around the sftp utils. All tests passing locally:

image

### Notes
- I originally planned to create unit tests with a mock sftp library, [`paramiko-mock`](https://paramiko-ssh-mock.readthedocs.io/en/latest/), but the library is a pain in a butt. It's bloated with mock objects to the point where it's hard to follow where is a mock response vs the implemented sftp functionality. For example, if you want to push a local object to sftp, you need to create sftp mock client and register a non-existent local object in the mock environment; if you try to push a real local file (like we do with s3 unit tests) to sftp with a mock client, it will error out. 

  Solution: test the connector functionality with integration tests like we did so with sftp utils. 

- I unregistered the connector because I was getting errors [here](https://github.com/NYCPlanning/data-engineering/actions/runs/16381291724?pr=1797) because the connector has required fields.

### Next steps
- Implement sftp model in ingest
- Implement sftp model in distribution (not sure how that works?)
- Create setup code that will pull sftp private and host keys to a local machine
- Migrate scripts/library/distribution code using new connector

### 🚨🚨🚨 Input wanted
The way I defined the sftp connector, it has required fields at the time of initiation like host name, user name, etc. However, having requiring attributes prevents me from registering it in our connector registry. Do I allow `None` values in the attributes so that we can register the connector and don't do connector validation when it's initiated? 

One alternative is using good old env variables that will be default values for the SFTP connector.

### DIFF
--- a/dcpy/test_integration/test_sftp.py
+++ b/dcpy/test_integration/test_sftp.py
@@ -1,6 +1,5 @@
 import pytest
 from pathlib import Path
-from tempfile import TemporaryDirectory
 from datetime import datetime
 import pytz
 from typing import TypedDict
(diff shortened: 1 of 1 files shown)


alexrichey (approved):
Very nice! Happy to see this moving forward. I have one bigger thought here, which I'm happy to leave for later, but I think you could DRY up the utils/sftp methods by implementing those methods on a class, which would look a lot like the SFTP connector that you added. (The SFTPConnector could actually just be a subclass of the new utils connector) 

Happy to chat, or just leave for later. Also happy to talk through the Connector Registry side of things.

sf-dcp:
Per chat with @alexrichey, I will make the following tweaks:

1. Wrap sftp utils into a class
2. Move any helper Connector logic like `SFTPConnector._push()` to utils.
3. Rename connector to SFTPConnectorAdaptor(). Make it inherit from the utils connector
4. For config variables, we will expect them as env vars, get them in `configuration.py`, and pass them to the connector at the time of connector registration.

Reference for myself:
image

dcpy/connectors/sftp.py
```diff
@@ -1,81 +1,66 @@
...
+    username: str | None
+    port: int
+    private_key_path: Path | None
+    known_hosts_path: Path | None
+
+    _connector_cache: SFTPConnector | None = None
+
+    @cached_property
+    def _connector(self) -> SFTPConnector:
+        if not all(
```
alexrichey, line 29: 👩‍🍳💋

dcpy/utils/sftp.py
```diff
@@ -0,0 +1,121 @@
...
+            logger.info(f"Listing files/directories for remote path '{path}' ...")
+            entries = connection.listdir(path=str(path))
+        return entries
+
+    def get_subfolders(self, prefix: str) -> list:
+        with self._connection() as connection:
+            logger.info(f"Listing subfolders for remote path '{prefix}' ...")
+            folder_objects = connection.listdir_attr(prefix)
+            subfolders = [
+                obj.filename for obj in folder_objects if stat.S_ISDIR(obj.st_mode)
```
alexrichey, line 63: out of curiosity, any reason Pathlib doesn't work to do this?

alexrichey (approved):
LGTM! A few comments, but I feel like this is good to go

dcpy/test_integration/test_sftp.py
```diff
@@ -3,64 +3,126 @@
...
+
+class SFTPDefaults(TypedDict):
+    hostname: str
+    port: int
+    username: str
+    known_hosts_path: Path
+    private_key_path: Path
+
+
+SFTP_DEFAULTS: SFTPDefaults = {
```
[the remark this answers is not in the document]
sf-dcp: I will actually pin this for a future iteration. I think this would be more appropriate for a test that validates connectors being registered since this is where we initiate the sftp connector with env vars. The connector itself doesn't use env vars

PR #1806, opened 2025-07-16
distribution tweaks

damonmcc:
related to https://github.com/NYCPlanning/data-engineering/issues/1801
companion PR in metadata repo: https://github.com/NYCPlanning/product-metadata/pull/55

distribution runs related to boolean columns (ignoring other errors):
- validation failing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/16297885195/job/46035355275#step:7:152)
- validation not failing [here](https://github.com/NYCPlanning/data-engineering/actions/runs/16304059179/job/46053188170#step:7:143)


dcpy/lifecycle/package/validate.py
```diff
@@ -88,6 +88,10 @@ def _is_valid_wkb(g):
         return False
 
 
+def _is_bool(s):
+    return bool(s)
```
alexrichey, line 92: what would we be expecting for `s` here? Guessing it's the different casing variants of "true"/"false"?

dcpy/lifecycle/package/validate.py
```diff
@@ -15,7 +15,7 @@
 class ErrorType(Enum):
     INVALID_DATA = "INVALID_DATA"
     NULLS_FOUND = "NULLS_FOUND"
-    COLUMM_MISMATCH = "COLUMN_MISMATCH"
+    COLUMN_MISMATCH = "COLUMN_MISMATCH"
```
alexrichey, line 18: 🙏

dcpy/lifecycle/package/validate.py
damonmcc: yes but this function is _way_ more permissive than I thought it was, I must've really misread something. I just checked and `bool('a_string')` returns `True`

will revise this

alexrichey (approved):

PR #1812, opened 2025-07-16
update high-density residential district label

### LINKED ISSUE #1692: GFT: explicitly handle introduction of new R11 and R12 zoning districts
### Context
Existing processes are written to handle Residence districts up to and including R10. City of Yes changes allow for mapping of R11 and R12 districts as well, though none have yet been mapped as of the creation date of this issue.

Associated experience builder app and web map issue can be found [here](https://github.com/NYCPlanning/fast-tracker-tool/issues/43)

### Request
Update underlying code that classifies GFT zoning categories to allow for R11 + R12 districts. Confirm any label/flag value changes with GIS and CAPS teams.

damonmcc:
resolves #1692

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-gft-new-districts)

The changes to the expected pilot projects are due to a few things:
1. new high-density zoning district label
2. changes in the CATS permits and arterial highways a lot is adjacent to
3. marginal lot geometry changes

All but the first one weren't addressed because the test was set to warn and we haven't been looking at it. If the test becomes noisy on nightly QA and main, we can reconsider raising the severity.

## tests of changes

dbt test named `equality_flags_zoning` [failing](https://github.com/NYCPlanning/data-engineering/actions/runs/16329813169/job/46129314680#step:8:629) before updating the test data in the seed file `test_expected__flags_zoning.csv` and [passing](https://github.com/NYCPlanning/data-engineering/actions/runs/16422112147) after updating the file


fvankrieken:
Looks fine from my end, will let @jackrosacker comment on anything in terms of their side of things

damonmcc:
talked to @jackrosacker IRL and he flagged a few changes to make:
1. add test cases to include R11 and R12 districts
2. investigate why pilot projects test is passing while seed file has old label in two records
3. add R11/R12 to seed file which would map each value to `high_res`

damonmcc:
@jackrosacker I think we're good to go. see comment above about the unfortunate difficulty of testing with mock data like R11s and R12s that don't exist yet

happy to explore dbt unit tests in a follow-up issue, but maybe we'd rather unblock GFT 25v2 first

fvankrieken:
> Sounds good. All clear on my end if you think we're all set with the new changes! Is there anything that's critical to remember to add or check if we get an R11 or R12 district in the future, and if so, do we have a mechanism to remind us to make that change? e.g. a test that fails loudly.

PLUTO will now fail loudly in this case - added a check this month where if any zoning districts aren't matched in the FAR lookup table the build will fail

#1790

fvankrieken:
Oops - actually, they also added R11 and R12 to the lookup table already so we will NOT fail when they're introduced

PR #1844, opened 2025-07-22
setup CPDB dbt project

### LINKED ISSUE #1791: CPDB Refactor: Set up dbt project
Create dbt project with seeds and sources. 

Add dbt build command into the current CPDB script.

damonmcc:
resolves #1791

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-cpdb-dbt)

- created dbt project for CPDB
- migrated all CSV-based tables to be seed models
- added a properties file to describe source models

Some CSVs in `cpddb/data/` aren't as simple to migrate to seed models so I'd rather do those in follow-up work related to actual query changes.

Didn't add any tests for source models and only added minimal tests for seed models. Test will definitely be good to expand later but they already lead to "this csv or query should be changed" which is out of scope for this setup work.

## tests of changes

Build successfully uses the seed models and the two primary output tables (`cpdb_dcpattributes`, `cpdb_projects_shp`) are identical to those in the `nightly_qa` schema.


products/cpdb/bash/config.sh
fvankrieken, line 1: begone!

products/cpdb/sql/_create.sql
```diff
@@ -1,70 +1,5 @@
...
-    origin character varying,
-    parkid character varying,
-    parkname character varying,
-    zipcode character varying
-);
-\COPY dcp_cpdb_agencyverified FROM 'data/dcp_cpdb_agencyverified.csv' DELIMITER ',' CSV HEADER;
-
 DROP TABLE IF EXISTS type_category_patterns CASCADE;
 CREATE TABLE type_category_patterns (
     pattern character varying,
```
fvankrieken, line 5: Looking at the files below that didn't make it into seeds - maybe they're seeds, and they get unioned in a stg (or I guess int) table? Obviously not needed now, just was curious why some were not done so was poking around

fvankrieken (approved):

products/cpdb/sql/_create.sql
fvankrieken: There's a sqldump??

PR #1845, opened 2025-07-22
add dbt-postgres to build-geosupport image requirements

damonmcc:
related to #1791

### DIFF
--- a/admin/run_environment/docker/build-geosupport/requirements.txt
+++ b/admin/run_environment/docker/build-geosupport/requirements.txt
@@ -1,6 +1,7 @@
 boto3
 census
 contextily
+dbt-postgres
 gdal
 geoalchemy2
 geopandas


fvankrieken:
So just a operations/norms thing - you don't need to run the docker build action to build "dev" images (build-base, build-geosupport, dev) for the new environment. They're built automatically in the PR tests for this branch (see [here](https://github.com/NYCPlanning/data-engineering/actions/runs/16455851901/job/46512615841?pr=1845)). Then, you just check the box for using the dev image
image

And the build will use the image specific to your branch.

In this case, not a big deal as there wasn't really anything "risky" about the new image, but in general it makes sense to not update the "prod" images until things are tested, then only build/push the images once this is merged

fvankrieken (approved):

PR #1877, opened 2025-08-01
Run integration tests in CI

fvankrieken:
Pretty self-explanatory! One test was failing before second commit

5% codecov boost let's go


fvankrieken:
> sweet! Now I wonder how many untested connectors I have to add to knock us back down below 70%...

😡

PR #1921, opened 2025-08-28
Finish Developments DB

alexrichey:
Finishing changes to DevDB. Sam and Marcus have signed off, but they need to check with exec, after which we should be good to go. Rationale is in the commit messages

### DIFF
--- a/products/developments/recipe.yml
+++ b/products/developments/recipe.yml
@@ -37,37 +37,23 @@ inputs:
     - name: dcp_mappluto_wi
       version: 25v2
     - name: doe_eszones
-      version: 20240313
     - name: doe_mszones
-      version: 20240313
     - name: doe_school_subdistricts
-      version: 2017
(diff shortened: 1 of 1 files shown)


dcpy/library/templates/dob_cofos.yml
```diff
@@ -6,7 +6,7 @@ dataset:
       path: .library/upload/dob_cofos.csv
       ## This field must be updated manually when archiving. Should eventually be refactored out of here
       ## Do not check in a specific version, leave blank version check in - this has led to issues in the past
-      previous_version: 20240105
+      previous_version: PLACEHOLDER
```
fvankrieken, line 9: Seems like c of os then was archived incorrectly? Are we sure that the issue was fully resolved? Just double-checking

alexrichey:
This should fix failing DevDB on nightly QA, fyi

Issue #1261, opened 2024-11-20
Ingest migration/Ingest v2 - migrate "Pluto Input" datasets


fvankrieken:
These are currently not defined in data library. The templates and logic live in the pluto repo. This needs a couple things

- [ ] a generic sftp connector
- [ ] a configuration for use by this connector so that we have "ginger" established as an sftp endpoint. The actual specification of target host, filename, etc can be left to 3, but we need the code to take a name of a host plus a filename, lookup server configs in env vars, secrets etc, and then retrieve the file
- [ ] a new sftp source type for ingest templates/configs
- [ ] geocode added as a processing step to ingest
- [ ] new templates written and validated

fvankrieken:
All done except for CAMA - next time data is available, it should be tested and migrated

PR #1937, opened 2025-09-08
migrate pluto inputs to ingest

fvankrieken:
Most of #1261 

Big picture - get 3/4 "pluto_input" datasets from library to ingest. CAMA can't be done yet because there's no data on the server at the moment so nothing to test. For the other three, data comparisons are in the comments below (but also links to specific comments in the respective commits).

Ended up touching on a bunch of other little things - getting sftp working in ingest, simplifying the concept of a "source" in ingest since it's really just a config for a connector now, etc

I put some work in making these commits relatively atomic! And there are some in depth descriptions, so go ahead and go through.

### DIFF
--- a/dcpy/lifecycle/ingest/validate.py
+++ b/dcpy/lifecycle/ingest/validate.py
@@ -44,7 +44,7 @@ def validate_against_existing_version(ds: str, version: str, filepath: Path) -> 
         )
 
 
-def validate_source(source: Source) -> dict:
+def find_source_validation_errors(source: Source) -> dict:
     violations: dict = {}
     if source.type not in source_connectors.list_registered():
         violations["invalid source type"] = (
@@ -81,7 +81,7 @@ def _validate_pd_series_func(
     return introspect.validate_kwargs(func, kwargs)  # type: ignore
 
 
-def validate_processing_steps(
+def find_processing_step_validation_errors(
     dataset_id: str, processing_steps: list[ProcessingStep]
 ) -> dict:
     """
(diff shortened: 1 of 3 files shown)


dcpy/models/library.py
```diff
@@ -56,7 +56,7 @@ class Url(BaseModel):
 
         class Socrata(BaseModel):
             uid: str
-            format: socrata.ValidSourceFormats
+            format: Literal["csv", "geojson", "shapefile"]
```
damonmcc, line 59: DRY: could this be `ValidFormat` from `connectors.socrata.configuration`? or are we trying to get rid of imports from `connectors` to `models`?

dcpy/lifecycle/scripts/validate_ingest.py
```diff
@@ -340,3 +373,28 @@ def _get_columns(
 
     for _, row in res.iterrows():
         print(row["t"])
+
+
+@app.command("load_from_s3")
+def _load_from_s3(
```
alexrichey, line 379: I think data_loader will do all this already? Should pick up on your dev_bucket env var.

dcpy/connectors/web.py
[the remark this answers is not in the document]
alexrichey: yeah, I've bumped against this a few times... I think it might be nice to standardize on connector kwargs for output directory and filename, but have the default case that the connector knows where to put stuff on the local machine

dcpy/lifecycle/ingest/validate.py
```diff
@@ -43,6 +44,25 @@ def validate_against_existing_version(ds: str, version: str, filepath: Path) ->
         )
 
 
+def validate_source(source: Source) -> dict:
```
alexrichey, line 47: slight nit: my thought about these types of functions is that I'd naively do the following:
```python
if validate_source(...):
    # assume the source is valid
else: 
    # assume it's not
```
so maybe something something `find_source_errors(...)`

alexrichey (approved):
Good stuff! Appreciate the atomicity

dcpy/lifecycle/ingest/validate.py
fvankrieken: I like that, will edit to something along those lines

PR #1939, opened 2025-09-08
Refactor Recipes to Ingest Datastore and Cloudpath

alexrichey:
# Change Summary
Broadly, this refactors recipes to be platform agnostic. We could store recipes on a disk, S3, Azure, or Google Cloud, and it should be effectively identical. 

## Suggested Review Strategy
Commits are very atomic. However, they don't even remotely resemble the order in which I went about this though... so if there are things that appear to be missing, they may have been inadvertently lumped into another commit.

## Specific Changes
- Implements Cloudpathlib as both 1) underlying storage usable in a connector, 2) a connector on it's own. 
- Refactors _most_ references away from directly calling `connectors.edm.recipes`.
- Adds integration tests for Ingest Datastore for Local, S3, and Azure (Azure creds for our existing Blob are in 1pw)
- Removes the recipes logging. It's not being used. IMO it needs a rethink.
- Enforces a no network/internet callouts rule for unit tests, and moves the Github test to integration tests. (impetus for that: as I was refactoring, I found that certain tests were making callouts, likely due to my changes. It made life so much easier to just enforce no callouts.)

## Notes:
- I didn't attempt to add public-read ACLs for Azure Blobs. Would be worth a chat to see if we want to do that. My assumption is that it will make cloud/architecture review more complicated. 
- We need to figure out longer-term locations of things in Azure. For now, I'm just using the `container` (the Azure equivalent of a bucket) that I created in the sandbox. 

## Open Questions:
- How do we want to handle required env vars for connectors in the context of workflows (e.g. the `Validate Ingest Templates` job) that don't need certain resources like RECIPES_BUCKET? I'd prefer not to provide those vars, but I also do think we should validate env vars when instantiating the connectors. 

## Thoughts Going Forward
- This moves us away from mocks and toward a more Dependency Injection-centric approach. We should standardize on that a little, in particular, we should determine what subset of connectors to make available during integration tests, and then make a collection of mocked connectors for the unit tests. (e.g. we should move the `mock_request_get` stuff to an API connector)
- I added `pytest-xdist` so I could run integration tests in parallel when working locally. However, I think that since these test are mostly network callouts (ie we're mostly just waiting for a response) parallelism should be the default for these tests.

### DIFF
--- a/dcpy/test/lifecycle/ingest/test_extract.py
+++ b/dcpy/test/lifecycle/ingest/test_extract.py
@@ -14,7 +14,8 @@ from dcpy.test.connectors.test_esri import mock_query_layer
 from .shared import TEST_DATASET_NAME, FAKE_VERSION, SOURCE_FILENAMES
 
 pytest.skip(
-    "Skipping this entire file until refactoring is complete.", allow_module_level=True
+    "Skipping this entire file until Ingest 1-many refactoring is complete. This will likely just be deleted.",
+    allow_module_level=True,
 )
 
 web.get_df = mock.MagicMock(return_value=pd.DataFrame())  # type: ignore
--- a/dcpy/test/lifecycle/ingest/test_run.py
+++ b/dcpy/test/lifecycle/ingest/test_run.py
@@ -12,7 +12,8 @@ from dcpy.test.conftest import mock_request_get
 from .shared import FAKE_VERSION, TEMPLATE_DIR
 
 pytest.skip(
-    "Skipping this entire file until refactoring is complete.", allow_module_level=True
+    "Skipping this entire file until Ingest 1-many refactoring is complete. This will likely just be deleted.",
+    allow_module_level=True,
 )
 
 


dcpy/test/conftest.py
```diff
@@ -37,6 +37,16 @@
 UTILS_RESOURCES = Path(__file__).parent / "utils" / "resources"
 
 
+@pytest.fixture(scope="session", autouse=True)
+def ensure_no_callouts():
```
fvankrieken, line 41: Sick

dcpy/lifecycle/config.py
```diff
@@ -22,6 +31,7 @@ def _set_default_conf():
                     "plan": {"local_data_path": "plan"},
                     "load": {"local_data_path": "load"},
                     "build": {"local_data_path": "build"},
+                    "qa": {"local_data_path": "qa"},
```
fvankrieken, line 35: Random note but wonder if I should break up the ingest substages more like this instead of the current approach in #1951 which more just has it all in one folder

dcpy/connectors/hybrid_pathed_storage.py
```diff
@@ -0,0 +1,308 @@
...
+            target.upload_from(self.path)
+
+        return target
+
+    def __str__(self):
+        return str(self.path)
+
+    def __getattr__(self, attr):
+        # Delegate other attributes/methods to the underlying Path
+        return getattr(self.path, attr)
```
fvankrieken, line 75: How does this work with typing? Does mypy, intellisense know that you can treat this like a Path with `p.exists()` and stuff like that?

dcpy/lifecycle/scripts/validate_ingest.py
```diff
@@ -186,13 +205,17 @@ def load_recipe_from_s3(
     client.drop_table(ds_id)
     client.drop_table(target_table)
 
-    ds = recipes.Dataset(id=ds_id, version=s3_version, file_type=file_type)
+    recipes.Dataset(id=ds_id, version=s3_version, file_type=file_type)
```
fvankrieken, line 208: ✂️

dcpy/connectors/hybrid_pathed_storage.py
alexrichey: Unfortunately it does confuse `mypy`, so for mypy and intellisense we need to explicitly declare referenced methods. (They'll still work without it though). Definitely open to a more elegant approach here

dcpy/lifecycle/config.py
```diff
@@ -1,17 +1,26 @@
...
+
 
 def _set_default_conf():
     ## the default is a little sparse at the moment, and a little duplicative
     ## (ie each `local_data_path` is just the `stage` name.)
+    _env_data_path = env.get("DCPY_LIFECYCLE_DATA_DIR")
+    data_path = (
+        Path(_env_data_path)
+        if _env_data_path
+        else Path(__file__).parent.parent.parent / ".lifecycle"  # ugh
```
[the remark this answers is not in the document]
alexrichey: Yeah, good point. Fixed [here]

dcpy/connectors/hybrid_pathed_storage.py
fvankrieken: Seems fine for our purposes, and especially with this being new in our codebase I'd rather not worry too much about little over-optimizations

PR #1943, opened 2025-09-10
CBBR - fix agency_response and agency_category_response

fvankrieken:
@damonmcc I could merge this into your branch as well. This was an older fix that AE needed that somehow I just blanked on getting merged

Prompted by discussion here: https://github.com/NYCPlanning/ae-zoning-api/issues/472

### DIFF
--- a/products/cbbr/sql/export.sql
+++ b/products/cbbr/sql/export.sql
@@ -131,7 +131,7 @@ SELECT
     END AS supporters_2,
     agency_acronym,
     agency,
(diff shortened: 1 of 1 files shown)


fvankrieken:
I think the main conflict with your branch is with this, `additional_comment` no longer exists - I think Tim/David clarified when I was fixing this that should have actually been agency_response.

damonmcc:
gotcha! I'll try getting your branch into mine and resolving the conflict

my change to `additional_comment` wasn't necessary at all so I'll probably drop it

PR #1951, opened 2025-09-17
ingest one to many

fvankrieken:


dcpy/lifecycle/ingest/plan.py
```diff
@@ -0,0 +1,109 @@
...
+def get_version(source: Source) -> str | None:
+    connector = source_connectors[source.type]
+    return (
+        connector.get_latest_version(**source.model_dump())
+        if "get_latest_version" in connector.__dir__()
+        else None
+    )
+
+
+def resolve_config(
```
fvankrieken: rename func

fvankrieken:
Test run [here](https://github.com/NYCPlanning/data-engineering/actions/runs/18789333235) with cscl gdb


Outputs as expected in dev bucket
image
image

fvankrieken:
> Still reading, but so far, looks great. My one over-arching thought (that may dissipate) as I go... It seems like there are stages and maybe substages, but I'm not 100% on what they'd be. It might help with comprehension if that was aligned fully with the file/folder structure. Then if `run` is the top-level thing, it could read a more like
> 
> ```python
> from ingest import plan, extract, process_datasets
> 
> def ingest():
>   plan.run(...)
> 
>   extract.run(...)
>   extract.archive(...)
> 
>   process_datasets.run(e)
>   process_datasets.archive(...)
> ```
> 
> Can chat more tomorrow!

I think I'm with you broadly. My big reading of what you're saying is that some logic currently in `run` could probably be moved into the stage-specific modules and things get a little simpler. Because what you're describing does sort-of happen... but then there's a lot of code in `run` in between those nicely defined stage-specific functions. So it seems like the "stages" are not quite big enough in scope to fully make things flow nicely (and they should probably handle things like dumping json, pushing to datastores, etc so that `run` can stay a bit higher level)

damonmcc (approved):

Issue #1915, opened 2025-08-25
FacilitiesDB 25v2 QA


alexrichey:
The draft build is here [25v2.1-initial-build](https://edm-publishing.nyc3.digitaloceanspaces.com/db-facilities/draft/25v2/1-initial-build/)

caseysmithpgh:
@NYCPlanning/data-engineering Additional flag: still getting a certificate expiration warning for the QAQC apps. Was able to access anyway, so no rush on my end--just a friendly reminder.

fvankrieken:
Uh okay seems like some very odd issues you're seeing. I'll take a look.

fvankrieken:
Okay a couple immediate findings
- diff vs diff_abs -> silly by me. I added diff_abs to the outputs, and they got grabbed into the QA tables without needing code changes... but the QA app tables aggregate on the fly by classification. And `abs(sum(diff)) != sum(abs(diff))`. I can change the qa app code to calculate the abs and pct on the fly
- pct -> my error in sql. The numbers are ints, so just getting rounded down. I can make these decimals and multiply by 100
- "inflated" diffs. This one is funny. All the QA app tables pull from qc_diff, not qc_classification. These differ in a few ways but one weird one -> qc_diff.csv filters to mapped records before comparing. If you noticed, DOE UNIVERSAL PRE-KINDERGARTEN also saw a TON unmapped this time, so this is the "inflation" you see. Couple thoughts
  - Is there a reason we want qc_diff to only be mapped records? I feel like that's what the qc_mapped data is for
  - I should look into why so many of these pre-k sites are unmapped this go-around

caseysmithpgh:
> * "inflated" diffs. This one is funny. All the QA app tables pull from qc_diff, not qc_classification. These differ in a few ways but one that I don't understand the reasoning behind: qc_diff.csv filters to mapped records before comparing. If you noticed, DOE UNIVERSAL PRE-KINDERGARTEN also saw a TON unmapped this time, so this is the "inflation" in diff that you see in the app. Couple thoughts
>   
>   * Is there a reason we want qc_diff to only be mapped records? I feel like that's what the qc_mapped data is for
>   * I should look into why so many of these pre-k sites are unmapped this go-around

[...]

PR #1965, opened 2025-09-25
FacDB 25v2 fixes

fvankrieken:
arising in #1915


products/facilities/facdb/sql/_qaqc.sql
```diff
@@ -243,7 +243,6 @@ FROM
             datasource,
             coalesce(count(*), 0) AS count_new
         FROM facdb
-        WHERE geom IS NOT null
```
fvankrieken, line 246: This was from a discussion with @caseysmithpgh - this output says in the app that it's "total records gained or lost" for the aggregated classification. So it's weird that it would only be mapped records.

We still have the mapped qa output as a separate check. I'll make sure the gde manual says to check this too

alexrichey (approved):
thanks!

PR #1985, opened 2025-10-03
OpenData: Add 2 mins of sleep on publish

alexrichey:
This is a dirty hack to allows us to publish revisions in GHA. We really should poll Socrata that the dataset upload has finished processing, but I don't have the time/motivation to work on that right now.


fvankrieken (approved):

fvankrieken:
Alex this has added about 10 minutes to dcpy tests 😬

PR #1999, opened 2025-10-06
Add Socrata <> Bytes version comparison script

alexrichey:
This should enable us to _potentially_ discard the GIS spreadsheet. It'd be trivial to add other destinations too (e.g. the Ginger sftp)

Output dataframe:
image
image
 (pardon the page-break. I couldn't fix the dataframe header in my notebook)

Note: not all of our socrata pages have the version set in the description yet. So those versions show up blank.

Long-term:
I think that for every connector we should have a workflow to retrieve the available versions for each registered connector and cache them in a datastore. Then we can join them together on a key of product.dataset.

### DIFF
--- a/dcpy/lifecycle/scripts/version_compare.py
+++ b/dcpy/lifecycle/scripts/version_compare.py
@@ -6,8 +6,6 @@ import pandas as pd
 def get_all_open_data_keys():
     """retrieve all product.dataset.destination_ids"""
     return product_metadata.load(version="dummy").query_product_dataset_destinations(
-        # product_ids={"lion"},
-        # dataset_ids={"atomic_polygons", "lion"},
         destination_filter={"types": {"open_data"}},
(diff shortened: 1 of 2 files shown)


dcpy/lifecycle/scripts/version_compare.py
```diff
@@ -0,0 +1,62 @@
+from dcpy.lifecycle.connector_registry import connectors
+from dcpy.lifecycle import product_metadata
+import pandas as pd
+
+
+def get_all_open_data_keys():
+    """retrieve all product.dataset.destination_ids"""
+    return product_metadata.load(version="dummy").query_product_dataset_destinations(
+        # product_ids={"lion"},
+        # dataset_ids={"atomic_polygons", "lion"},
```
damonmcc, line 10: ✂️?

PR #2009, opened 2025-10-15
add LION proto-segments

### LINKED ISSUE #1826: LION coincident segments
Section [2.6](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/_layouts/15/Doc.aspx?sourcedoc=%7BD3194825-3026-4295-8A31-6FB030749AA1%7D&file=ETL_V8_02012024%20-%20Copy.docx&action=default&mobileredirect=true)

damonmcc:
resolves #1826

all builds on this branch [here](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-proto-segments)

**TLDR**: This adds proto-segment records to LION. Resolving remaining diffs between dev and prod should be done in follow-ups.

## background

Proto segments are records from `dcp_cscl_altsegmentdata` that should:
1. have a geoemtry-modeled segment with the same `segmentid`
2. only appear in lion if it has an associated geoemtry-modeled segment
3. inherit some fields from the associated geoemtry-modeled segment
4. use and resolve non-inherited fields

## general approach

1. create the geometry-modeled segments "primary" segments before the full "lion" model
2. union the primary and proto segments with their identifying fields and the fields that proto-segments don't inherit from their primary segment
3. join all other primary fields segments to the full list of segments so proto-segments can inherit fields
4. compute lion fields for all segments

`int__lion` is now the combination of primary and proto segments.

## diffs between dev and prod

The lion record count increased from 212,621 to 214,144. The `production_outputs.citywide_lion` record count is 214,099.

I used segment ID `0035386` as a focus because it has the highest count of relevant records in `dcp_cscl_altsegmentdata`. There are no diffs for those proto segments (yay!). On `main`, there a four diffs because they're `missing in dev`.

There's no impact from these changes on primary segment diffs.

These screenshots compare the `qa__lion_data_infividial_diffs` tables from `main` and this grouped by `_source_table`. The count for `NULL` decreased because lots of prod records are no longer `missing in dev`

### original approach

Screenshot 2025-10-29 at 4 27 14 PM

### latest post-review approach

same as above but less `dcp_cscl_altsegmentdata` diffs!

Screenshot 2025-10-31 at 2 33 51 PM

### DIFF
--- a/products/lion/models/intermediate/_int.yml
+++ b/products/lion/models/intermediate/_int.yml
@@ -1,28 +1,32 @@
 version: 2
 
 models:
+- name: int__primary_segments
+  columns:
+  - name: segmentid
+    tests: [ not_null, unique ]
+
+- name: int__segments
+  columns:
+  - name: lionkey_dev
+    tests: [ unique ]
+  - name: segmentid
+    tests: [ not_null ]
+  - name: feature_type
+    tests: [ not_null ]
+  tests:
+  - dbt_utils.unique_combination_of_columns:
+      name: int__segments__key_ideal
+      combination_of_columns:
+      - boroughcode
+      - face_code
+      - segment_seqnum
+      config:
+        severity: warn
+
 - name: int__lion
   columns:
   - name: segmentid
-    tests: [ not_null ]
-  - name: boroughcode
-    tests:
-      - not_null:
-          config:
-            severity: warn
-  - name: face_code
-    tests:
-      - not_null:
-          config:
-            severity: warn
-  - name: segment_seqnum
-    tests:
-      - not_null:
-          config:
-            severity: warn
-  - name: from_nodeid
-    tests: [ not_null ]
-  - name: to_nodeid
     tests: [ not_null ]
   tests:
   - dbt_utils.unique_combination_of_columns:
@@ -34,26 +38,11 @@ models:
       config:
         severity: warn
   - dbt_utils.unique_combination_of_columns:
-      name: int__lion__key_temp # while nonstreetfeatures do not have segment_seqnum generated
+      name: int__lion__key_temp # while nonstreetfeatures don't have segment_seqnum generated
       combination_of_columns:
       - boroughcode
-      - segmentid
       - face_code
+      - segment_seqnum
+      - segmentid
       config:
         severity: warn
-  - dbt_utils.unique_combination_of_columns:
-      name: int__lion__key # from trying different column combinations, is unique except for duplicate rows
-      combination_of_columns:
-      - segmentid
-      - from_sectionalmap
-      - to_sectionalmap
-      - from_nodeid
-      - to_nodeid
-      - right_atomicid
-      - left_atomicid
-      - r_zip
-      - l_zip
-      - segment_locational_status
-      config:
-        # There are 8 rows caused by 4 pairs of duplicates
-        error_if: ">4"
--- a/products/lion/models/intermediate/int__lion.sql
+++ b/products/lion/models/intermediate/int__lion.sql
@@ -8,6 +8,11 @@
 
 WITH segments AS (
     SELECT * FROM {{ ref("int__segments") }}
+    WHERE is_based_on_primary_segment
+),
+
+segment_locational_status AS (
+    SELECT * FROM {{ ref("int__segment_locational_status") }}
 ),
 
 centerline AS (
@@ -38,10 +43,6 @@ centerline_curve AS (
     SELECT * FROM {{ ref("int__centerline_curve") }}
 ),
 
-street_and_facecode AS (
-    SELECT * FROM {{ ref("int__streetcode_and_facecode") }}
-),
-
 sedat AS (
     SELECT * FROM {{ ref("int__split_election_district") }}
 ),
@@ -54,11 +55,7 @@ zips AS (
     SELECT * FROM {{ ref("int__segment_zipcodes") }}
 ),
 
-segment_locational_status AS (
-    SELECT * FROM {{ ref("int__segment_locational_status") }}
-),
-
-centerline_diff_coincident_segment AS (
+centerline_coincident_subway_or_rail AS (
     SELECT * FROM {{ ref("int__centerline_coincident_subway_or_rail") }}
 ),
 
@@ -67,16 +64,17 @@ noncl_coincident_segment AS (
 )
 
 SELECT
+    segments.lionkey_dev,
     segments.boroughcode,
-    street_and_facecode.face_code,
+    segments.face_code,
     segments.segment_seqnum,
     segments.segmentid,
-    street_and_facecode.five_digit_street_code,
-    street_and_facecode.lgc1,
-    street_and_facecode.lgc2,
-    street_and_facecode.lgc3,
-    street_and_facecode.lgc4,
-    street_and_facecode.boe_lgc_pointer::CHAR(1),
+    segments.five_digit_street_code,
+    segments.lgc1,
+    segments.lgc2,
+    segments.lgc3,
+    segments.lgc4,
+    segments.boe_lgc_pointer,
     nodes.from_sectionalmap,
     nodes.from_nodeid,
     nodes.from_x,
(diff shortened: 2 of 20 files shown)


products/lion/models/intermediate/_int.yml
```diff
@@ -2,3 +2,95 @@ version: 2
...
+      config:
+        severity: warn
+  - dbt_utils.unique_combination_of_columns:
+      name: int__lion__key_temp # while nonstreetfeatures do not have segment_seqnum generated
+      combination_of_columns:
+      - boroughcode
+      - segmentid
+      - face_code
+      config:
+        severity: warn
```
fvankrieken: This one will not be necessarily true (I think) for proto segments sharing the same segment id

products/lion/models/intermediate/int__primary_segments.sql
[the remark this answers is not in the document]
fvankrieken: (or now post-latest changes, mainly from renaming `int__segments`)

products/lion/models/intermediate/int__segments.sql
```diff
@@ -1,62 +1,125 @@
...
+    SELECT
+        segmentid AS segmentid_all,
+        TRUE AS is_proto_segment,
+        source_table,
+        CASE
+            WHEN feature_type_code = '3' THEN 'nonstreetfeatures'
+            WHEN feature_type_code = '8' THEN 'nonstreetfeatures'
+            WHEN feature_type_code = '9' THEN 'centerline'
+            ELSE feature_type_code
+        END AS feature_type,
```
fvankrieken: Let's put this in the staging table?

products/lion/models/intermediate/int__lion.sql
fvankrieken: For this file, let's

1. reorder things to how they were - just easiest in terms of a reference to keep fields in same order as docs, since there are so many. It's also nice to have our non-lion-export fields at the end, keep all of our custom stuff in the same place
2. now that the backbone is no longer unique on segmentid, we need to tweak the joins here. For the most part, things are fine for now given that we haven't introduced lionkey as an identifier yet. BUT for tables like `centerline` which are joined to lookup centerline data, segmentid is all we have (since lionkey doesn't exist for the staging centerline table). So for these featuretype-specific joins, we should filter by feature type
```sql
LEFT JOIN centerline ON segments.segmentid = centerline.segmentid
LEFT JOIN centerline_curve ON centerline.segmentid = centerline_curve.segmentid
-- becomes
LEFT JOIN centerline ON segments.segmentid = centerline.segmentid AND segments.feature_type = 'centerline'
LEFT JOIN centerline_curve ON centerline.segmentid = centerline_curve.segmentid AND segments.feature_type = 'centerline'
```

products/lion/models/intermediate/_int.yml
```diff
@@ -1,4 +1,55 @@
...
+  tests:
+  - dbt_utils.unique_combination_of_columns:
+      name: int__segments__key_ideal
+      combination_of_columns:
+      - boroughcode
+      - face_code
+      - segment_seqnum
+      config:
+        severity: warn
+  - dbt_utils.unique_combination_of_columns:
```
fvankrieken: This is duplicating the unique on lionkey_dev, no?

products/lion/models/intermediate/int__segments.sql
```diff
@@ -1,62 +1,131 @@
...
-        UNION ALL
-    {%- endif %}
-{% endfor %}
+        -- proto-segment fields
+        proto_segments.segmentid,
+        TRUE AS is_proto_segment,
+        primary_segments.segmentid IS NOT NULL AS is_based_on_primary_segment,
+        primary_segments.segmentid AS primary_segment_segmentid,
+        proto_segments.source_table,
+        COALESCE(proto_segments.feature_type, primary_segments.feature_type) AS feature_type,
```
fvankrieken: Is this ever null? Do the docs explicitly say this?

products/lion/models/intermediate/_int.yml
damonmcc: oops, this is left over from the primary-only and pre-lionkey_dev days. will drop (glad it's passing)

products/lion/models/intermediate/int__segments.sql
damonmcc: yup 7 records. the docs say something like "proto-segments without a primary segment can't make it into LION", I'll find the exact language

it's nice to see that result (hence the `left join` rather than `inner join` so then in `int__lion` there's a `WHERE is_based_on_primary_segment` at the top

products/lion/models/intermediate/int__segments.sql
```diff
@@ -1,62 +1,131 @@
...
+        -- primary segment fields
+        primary_segments.lgc6,
+        primary_segments.lgc7,
+        primary_segments.lgc8,
+        primary_segments.lgc9,
+        boe_lgc_pointer,
+        legacy_segmentid,
+        shape_length,
+        from_level_code,
+        to_level_code,
```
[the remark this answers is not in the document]
fvankrieken: Okay I don't think it will work as written looking back at what I wrote in int__primary_segments and int__lion, but that's for sure outside of this PR

products/lion/models/intermediate/int__segments.sql
fvankrieken: I meant the coalesce here specifically. As in, would there be a case where the feature_type of the altsegmentdata is null and we'd want to fall back to the primary segment feature type

products/lion/models/intermediate/int__segments.sql
```diff
@@ -1,62 +1,131 @@
...
+        proto_segments.face_code,
+        proto_segments.segment_seqnum,
+        proto_segments.five_digit_street_code,
+        proto_segments.lgc1,
+        proto_segments.lgc2,
+        proto_segments.lgc3,
+        proto_segments.lgc4,
+        proto_segments.lgc5,
+        -- primary segment fields
+        primary_segments.lgc6,
```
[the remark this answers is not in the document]
fvankrieken: Yeah that's confusing language with the field name too because I feel like that flag means something slightly different in another circumstance in cscl (with primary segments)

products/lion/models/intermediate/int__segments.sql
```diff
@@ -1,62 +1,131 @@
 {{ config(
     materialized = 'table',
     indexes=[
-      {'columns': ['geom'], 'type': 'gist'},
-      {'columns': ['midpoint'], 'type': 'gist'},
       {'columns': ['segmentid']}
```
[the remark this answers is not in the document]
damonmcc: (vocab thought) I went with `lionkey_dev` since I didn't feel like the constructed `CONCAT(segmentid, boroughcode, face_code, segment_seqnum)` was gonna be the true `lionkey` in the future, but if I was wrong happy to drop the `_dev`

products/lion/models/intermediate/int__segments.sql
damonmcc: my bad thought u were talking about `is_based_on_primary_segment ` 

these are the feature code + type counts in stg__altsegmentdata_proto

Screenshot 2025-10-31 at 4 05 00 PM

878 nulls and only 73 `Feature Code` diffs for proto-segments. so seems good to fall back to fall back to the primary segment feature type for now and investigate those diffs

fvankrieken: Looks like NULL 'feature_type_code' should be mapped to 'centerline'.

image

So let's handle that in s[taging table for protosegments](https://github.com/NYCPlanning/data-engineering/pull/2009/files#diff-059a5f9c5a4de87da0e08aed6b3e10b3b05ab7d2f59f70c3e13d2abb57874192R31) and drop this coalesce. Also, it doesn't look like that case when handles all cases in the table shown above. We may not have any of those in altsegmentdata, but per the docs they seem possible so let's handle it in case any are ever added in the future

fvankrieken (approved):
Pending resolution of [this](https://github.com/NYCPlanning/data-engineering/pull/2009#discussion_r2482560751) (and a successful build after), good to go!

Great stuff. @alexrichey this will certainly introduce more rows with wrong coincident segment count. Feel free to ignore diffs with source_table = dcp_altsegmentdata because those will probably need a different approach

PR #2012, opened 2025-10-20
Allow cama failures without failing job

### LINKED ISSUE #2006: Scheduled Action Failure - PLUTO Inputs
[Failed action](https://github.com/NYCPlanning/data-engineering/actions/runs/18653035995)

fvankrieken:
Closes #2006. Very open to feedback, just thought it made sense to get something quick to stop the weekly failing job

## Summary
Before #1964, cama gha had logic to cancel the workflow if the file was not found on the sftp server. We've since moved sftp logic into dcpy/ingest. However, there isn't handling for a case like this in there - if you attempt to "pull" a file that doesn't exist, it's simply an error.

So this seems like a little bit of a bandaid, but basically thought is at the action level, you can maybe just decide that a "failure" - that is, a python ingest error - actually is expected behavior and is fine. I think this makes sense logically - we tried to pull an sftp file, that failed, and that was expected behavior. The main issue is that we're not handling that error specifically: any python error in that ingest step will get ignored. So there is really the bandaid part. If this was handled via an orchestration tool in python, it'd be a little easier to handle this more gracefully.

## Changes
Added error handling to the ingest helper gha - either 'raise' (default behavior), 'ignore' (job succeeds on error), or 'cancel' (cancel the gha). The latter is not used in this PR but it corresponds more to what was done before with CAMA. However, you can't cancel a specific job of a gha with that functionality, just the whole gha. So it wasn't useful in this case where we have 3 datasets that we always want to ingest and 1 that is only occasionally there. But you can see [here](https://github.com/NYCPlanning/data-engineering/actions/runs/18655142998) that it "works"

See job [here](https://github.com/NYCPlanning/data-engineering/actions/runs/18655268516) using `continue-on-error` - we at least get a too-level annotation that there was an error, but not an actual failed action

### DIFF
--- a/.github/workflows/pluto_inputs.yml
+++ b/.github/workflows/pluto_inputs.yml
@@ -36,7 +36,7 @@ jobs:
       datasets: pluto_input_cama_dof
       dev_bucket: ${{ inputs.dev_bucket }}
       dev_image: ${{ inputs.dev_image }}
-      error_behavior: ignore
+      error_behavior: ignore # TODO a bit of a hack. This ignores all errors. Ideally, we just ignore case where the remote file is not found
       latest: true
       label: ${{ github.workflow }}
 


damonmcc (approved):
love the annotation use

`continue-on-error` is a built-in job setting? just wondering cuz I was gonna suggest a different name to align with the "ignore" term but doesn't look like we could

maybe `raise, continue, cancel` would be nice, but `ignore` still makes sense to me

PR #2015, opened 2025-10-21
make default compile behavior not upgrade

fvankrieken:
Not to put anyone on blast but I've been noticing when folks are adding python packages and compiling, we seem to generally bump all (potentially unrelated) packages, which is mostly fine but potentially will lead to changed behavior - sqlfluff tests failing, etc (or even actual python behavior changes).

So with this, keep it easier to
1. add packages and recompile while keeping existing package versions constant without remembering to use a flag
2. save bumping other packages for our monthly compile requirements job

### DIFF
--- a/admin/ops/python_compile_requirements.sh
+++ b/admin/ops/python_compile_requirements.sh
@@ -27,9 +27,9 @@ fi
 
 # if specified, upgrade versions in requirements.txt
 if [[ $* == *--upgrade* ]]; then
-  upgrade=""
-else
   upgrade="--upgrade"
+else
+  upgrade=""
 fi
 
 cd $path


.github/workflows/python_compile_requirements.yml
```diff
@@ -48,6 +49,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         ACTION: ${{ github.workflow }}
-        BUILD_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        BUILD_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{
+          github.run_id }}
```
damonmcc: change due to a line length limit?

alexrichey (approved):
lgtm after flipping the --upgrade. Apologies, didn't know the --no-upgrade flag existed!

PR #2025, opened 2025-10-27
Add pydantic class generator from xml

alexrichey:
Static Python class generator for [pydantic-xml](https://pydantic-xml.readthedocs.io/en/latest/) similar to the code generator for regular [pydantic classes](https://docs.pydantic.dev/latest/integrations/)

The goal is to take an existing XML file, and generate python models for it. 

Given an input like:
```xml
<?xml version="1.0"?>
<metadata xml:lang="en">
    <Esri>
        <CreaDate>19700101</CreaDate>
        <CreaTime>00000000</CreaTime>
        <ArcGISFormat>1.0</ArcGISFormat>
        <SyncOnce>TRUE</SyncOnce>
        <scaleRange>
            <minScale>150000000</minScale>
            <maxScale>5000</maxScale>
        </scaleRange>
        <ArcGISProfile>ISO19139</ArcGISProfile>
    </Esri>
    <mdHrLv>
        <ScopeCd value="005" />
    </mdHrLv>
    <mdDateSt Sync="TRUE">19261122</mdDateSt>
</metadata>
```

`python -m dcpy.utils.code_gen.pydantic_from_xml ./my_file.xml`: Generates `my_file.py` with contents:
```python
"""Autogenerated by generate_pydantic_xml_classes.py
Source XML: shp_metadata_minimum.xml
Do not edit directly.
"""
from __future__ import annotations
from pydantic_xml import BaseXmlModel, element, attr

class Scopecd(BaseXmlModel, tag="ScopeCd"):
    value: str | None = attr(name="value", default=None)

class Mddatest(BaseXmlModel, tag="mdDateSt"):
    sync: str | None = attr(name="Sync", default=None)

class Mdhrlv(BaseXmlModel, tag="mdHrLv"):
    scope_cd: Scopecd | None = element(tag="ScopeCd", default=None)

class Scalerange(BaseXmlModel, tag="scaleRange"):
    min_scale: int | None = element(tag="minScale", text=True, default=None)
    max_scale: int | None = element(tag="maxScale", text=True, default=None)

class Esri(BaseXmlModel, tag="Esri"):
    crea_date: int | None = element(tag="CreaDate", text=True, default=None)
    crea_time: int | None = element(tag="CreaTime", text=True, default=None)
    arc_gis_format: float | None = element(tag="ArcGISFormat", text=True, default=None)
    sync_once: str | None = element(tag="SyncOnce", text=True, default=None)
    scale_range: Scalerange | None = element(tag="scaleRange", default=None)
    arc_gis_profile: str | None = element(tag="ArcGISProfile", text=True, default=None)

class Metadata(BaseXmlModel, tag="metadata"):
    lang: str | None = attr(name="{http://www.w3.org/XML/1998/namespace}lang", default=None)
    esri: Esri | None = element(tag="Esri", default=None)
    md_hr_lv: Mdhrlv | None = element(tag="mdHrLv", default=None)
    md_date_st: Mddatest | None = element(tag="mdDateSt", default=None)
```

This is a very much a peripheral, run-it-once type of thing, so I wouldn't say it's so important to understand the algorithmic side of things. I'd pay more attention to the test cases, honestly.


jackrosacker:
> > Made some edits to how `pydantic_from_xml.py` handles floats without precision loss, which solved the previous issue
> 
> @jackrosacker Awesome! This looks good, though would it be possible to trim out some of the repetition in the pluto xml? (specifically the attrs under eainfo - having a few of them seems good, but most are nearly identical, which doesn't help us test much)

I can definitely take a look tomorrow. Would this just be to cut down on file size? I trimmed out about 4000 lines of image encoding text before committing, but the pluto xml is pretty close to what we publish already so I'd hesitate to cut out too much without making some larger metadata standard changes (outside of code world)

alexrichey:
> > > Made some edits to how `pydantic_from_xml.py` handles floats without precision loss, which solved the previous issue
> > 
> > 
> > @jackrosacker Awesome! This looks good, though would it be possible to trim out some of the repetition in the pluto xml? (specifically the attrs under eainfo - having a few of them seems good, but most are nearly identical, which doesn't help us test much)
> 
> I can definitely take a look tomorrow. Would this just be to cut down on file size? I trimmed out about 4000 lines of image encoding text before committing, but the pluto xml is pretty close to what we publish already so I'd hesitate to cut out too much without making some larger metadata standard changes (outside of code world)

Yep, mostly filesize. But also if there are test issues, it just makes it easier to debug.

jackrosacker:
Out of curiosity I ran the tool on the maximal metadata file I created a while back that has close to every possible field filled out, and the module generation is failing with this error: `RuntimeError: Cycle detected in tag dependencies — cannot emit classes without forward references. Consider simplifying the XML or enabling quoted forward-references.`

I think this is not something we need to solve now, but was curious to see how many XML variations it could handle.

alexrichey:
> Out of curiosity I ran the tool on the maximal metadata file I created a while back that has close to every possible field filled out, and the module generation is failing with this error: `RuntimeError: Cycle detected in tag dependencies — cannot emit classes without forward references. Consider simplifying the XML or enabling quoted forward-references.`
> 
> I think this is not something we need to solve now, but was curious to see how many XML variations it could handle.

Yeah... My guess is it's duplicated elements. Agree, it's not something to solve now.

PR #2039, opened 2025-11-02
fix pathing issue in pull

fvankrieken:
See https://github.com/NYCPlanning/data-engineering/actions/runs/18988480535/job/54237887693?pr=2038#step:9:30 - current implementation pulls instead to the folder name. So a parquet file named "25v3" instead of "25v3/ds_id.parquet".


fvankrieken:
Okay actually closing for now because this is a little funky. Things work for now but it's not quite ideal, certainly not exactly working as intended. We need to figure out our dance around passing destination dirs/filepaths a little

PR #2046, opened 2025-11-07
lion protosegments - face code and sequence numbrer

fvankrieken:
Go by commit!

Main thing is getting actual face codes for protosegments and getting... let's call it "appropriate" seqnum for nonstreetfeature protosegments (as in, unique but not necessarily matching prod)

image

(this is by our proxy key of boro, face code, segmentid still)

## doesn't this mean that lionkey is unique now?
I had thought so, but fun discovery - lionkey is not actually unique in production lion. In cases where you have NSF and some other segments that share a Borough and Face code, you have cases where they share sequence number too. Looks like there are 21 such records. I'll add it to the running list of data issues and bring it up with GR/GSS


products/lion/models/intermediate/segments/int__primary_segments.sql
fvankrieken: Just reording here to get the more "canonical" key first - lion key, then segment id, then other fields

PR #2055, opened 2025-11-12
LION ETL: Address non-centerline Coincident Segment Count

alexrichey:
This brings us down to just a handful of non-centerline coincident segments that differ from the prod ETL. I've emailed GR a list of segments that should probably be consolidated (redrawn) in CSCL - they're nearly identical geometrically, but nonequal, and they cause an issue with prod. 

| segment id | matched segment id |
|---|---|
| 257795 | 288944 |
| 281438 | 281437 |
| 329244 | 125345 |
| 329038 | 123124 |
| 8104016 | 352007 |

So after we merge this, we can wait and see if the next CSCL dump fixes these. 

### Worth flagging
1. t looks like there are still a number of centerline segments with differing coincident segment counts - we'll need to address that.
2. We should note somewhere that the docs are quite a ways off from what the C# ETL was doing. @fvankrieken thoughts on where? Also, that the C# ETL wasn't implementing this: `alt_segdata_type <> 'S'` (nor are we, here)

### DIFF
--- a/products/lion/models/intermediate/coincident_segments/int__noncenterline_coincident_segment_count.sql
+++ b/products/lion/models/intermediate/coincident_segments/int__noncenterline_coincident_segment_count.sql
@@ -10,18 +10,24 @@ WITH spatial_matches AS (
         segmentid,
         COUNT(*) AS sm_count
     FROM {{ ref('int__noncenterline_coincident_segments') }}
-    WHERE distance < 1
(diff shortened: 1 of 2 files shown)


alexrichey:
> Can you run a build? Assume you've been dbting locally

@fvankrieken yes, good point. Kicked one off.

PR #2061, opened 2025-11-13
Add Fuzzy Versions, to compare bytes <> socrata

alexrichey:
This modifies a few things:
- Adds fuzzy version comparisons in the BYTES <> Socrata script so that we can compare version formats like '2025-01-01' with '2025Q1', or "Oct 2025" with "2025-10-01" to see what's potentially out of date. E.g. 
image

- Modifies the output dataframe to make results more actionable. Namely, it sorts:
  1. Products with ANY out of date datasets, where there exists a Socrata version
  2. Products with ANY out of date datasets, but no Socrata version (these may be out of date... or they may just be TODO on the Socrata side, and so have no version)
  3. The rest

This PR is mostly slopped. I took a shot at integrating this with our existing `dcpy.utils.versions` code, but came away thinking that will be a bigger endeavor (and probably refactor of that module).

### DIFF
--- a/dcpy/lifecycle/scripts/version_compare.py
+++ b/dcpy/lifecycle/scripts/version_compare.py
@@ -20,7 +20,7 @@ class FuzzyVersion:
         if not (self.normalized and fuzzy_other.normalized):
             return False
 
-        return (self.normalized == fuzzy_other.normalized)
+        return self.normalized == fuzzy_other.normalized
 
     def _normalize(self):
         """


dcpy/lifecycle/scripts/version_compare.py
```diff
@@ -2,6 +2,169 @@
...
+        Returns:
+            str: Normalized version string in YYYYMM format, or original if no pattern matches
+        """
+        if not self.original:
+            return self.original
+
+        version = self.original.lower().strip()
+
+        # Handle quarter notation (e.g., "25q1", "24q2")
+        quarter_match = re.match(r"^(\d{2})q([1-4])$", version)
```
fvankrieken: can this be q|Q? I feel like we see both

dcpy/test/lifecycle/scripts/test_version_comparison.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class TestIntegrationScenarios:
+    @pytest.mark.parametrize(
+        "version1,version2,expected",
+        [
+            ("September 2025", "202509", True),
+            ("JUNE 2024", "24q2", True),
+            ("march 2025", "20250315", True),
+            ("Q1 2025", "january 2025", False),  # Different months in Q1
```
damonmcc: sorry why aren't these probably equal?

edit: oh direction matters! if something is Q1 it could be any of 3 different months so we'd rather not say these're equalish

Issue #2060, opened 2025-11-13
LION protosegments other issues


fvankrieken:
There seems to be some remaining issues, relating to things like
- level code
- feature type code

especially. But 

```
select * from qa__lion_dat_individual_diffs where "_source_table" = 'dcp_cscl_altsegmentdata'
and 'field' <> 'Sequence Number'
```

Will get you what you need. Ignoring sequence number because that's part of #1825

PR #2062, opened 2025-11-13
LION proto-segments - reverse geoms and fill in other fields

### LINKED ISSUE #2059: LION protosegments reversed geometry and related fields

fvankrieken:
Closes #2059 and #2060 

Do I like how many `CASE ... WHEN`s there are in int__lion? No I do not. Do I currently feel like that's a more holistic issue to refactor later? Yeah. Obviously still quite open to feedback here, but I think regardless, we're due for a whiteboarding and grand refactor to iron out a couple of the structural issues

Commits are descriptive, so go by commit for necessary context!

### DIFF
--- a/products/cscl/models/intermediate/int__lion.sql
+++ b/products/cscl/models/intermediate/int__lion.sql
@@ -288,10 +288,11 @@ LEFT JOIN saf ON segments.segmentid = saf.segmentid AND segments.boroughcode = s
 LEFT JOIN nypd_service_areas ON segments.lionkey_dev = nypd_service_areas.lionkey_dev
 LEFT JOIN sedat ON segments.segmentid = sedat.segmentid AND segments.boroughcode = sedat.boroughcode
 -- centerline only
+LEFT JOIN centerline AS primary_centerline ON segments.segmentid = primary_centerline.segmentid
 LEFT JOIN centerline ON segments.segmentid = centerline.segmentid AND segments.feature_type = 'centerline'
 LEFT JOIN centerline_curve ON centerline.segmentid = centerline_curve.segmentid AND segments.feature_type = 'centerline'
-LEFT JOIN centerline_coincident_subway_or_rail ON centerline.segmentid = centerline_coincident_subway_or_rail.segmentid
-LEFT JOIN centerline AS primary_centerline ON segments.segmentid = primary_centerline.segmentid
+LEFT JOIN centerline_coincident_subway_or_rail
+    ON centerline.segmentid = centerline_coincident_subway_or_rail.segmentid AND segments.feature_type = 'centerline'
 -- shoreline only
 -- rail only
 LEFT JOIN rail ON segments.segmentid = rail.segmentid


products/cscl/models/intermediate/int__lion.sql
```diff
@@ -247,3 +279,4 @@ LEFT JOIN rail ON segments.segmentid = rail.segmentid
 -- other
 LEFT JOIN noncl_coincident_segment ON segments.segmentid = noncl_coincident_segment.segmentid
 LEFT JOIN zips ON segments.lionkey_dev = zips.lionkey_dev AND segments.feature_type != 'centerline'
+LEFT JOIN proto ON segments.source_table = 'altsegmentdata' AND segments.ogc_fid = proto.ogc_fid
```
damonmcc: we can trust `ogc_fid` across `int__segments` and `stg__altsegmentdata_proto`?

looks like it since `int__segments` is a child model of `stg__altsegmentdata_proto`, this just jumped out to me since I feel like we tend to ignore/mistrust `ogc_fid` and use other keys instead

damonmcc:
nice! got any build diffs to show the improvement?

damonmcc (approved):

products/cscl/models/intermediate/int__lion.sql
fvankrieken: ogc_fid is the only unique key for stg__altsegmentdata_proto, since two different protosegments can have the same segmentid and the lionkey field isn't actually accurate. So this is really just using the ogc_fid column as what it is - a guaranteed primary key for the source table.

fvankrieken:
@damonmcc added one extra commit (in addition to the nit) just fyi - slight bug in boe_lgc_pointer, just a one liner

products/cscl/models/staging/segments/stg__centerline.sql
```diff
@@ -5,6 +5,16 @@
 
 SELECT
     *,
+    CASE
```
alexrichey: it may be pipe-dream'y... but it'd be so nice if these lookups were seed tables, then those seed-tables could be used to generate documentation. (like what you did with feature_type_codes from commit 2. Might just be a premature optimization though)

products/cscl/models/intermediate/int__lion.sql
```diff
@@ -128,37 +128,46 @@ SELECT
...
-        WHEN trafdir = 'NV' THEN 'P'
-        -- Two-way
-        WHEN trafdir = 'TW' THEN 'T'
+        WHEN (
+            segments.source_table = 'altsegmentdata'
+            AND (
+                segments.feature_type_code IS NULL
+                OR segments.feature_type_code IN ('6', 'A', 'W')
+            )
+        )
```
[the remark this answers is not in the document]
fvankrieken: I wouldn't say a big goal of this PR was upstreaming logic, and that really is the general tension that hasn't quite been resolved of this whole dbt project. 

For this, mainly just that currently,
- stg__altsegmentdata_proto doesn't have centerline (or primary segment in general) joined in
- int__protosegments is currently a little more intended as sort of more "primary" data for protosegments - not all of these lion attributes joined in, but just a resolved protosegment with it's main identifying fields. None of the non-centerline segments in `int__segments` has traffic_direction or a notion of it, and protosegment table is meant to be essentially the same philosophically as int__segments. Though here's where that tension comes in a little, because we do handle all the face-code related fields in or upstream of int__segments

I moved feature_type_code upstream because 1. it made sense practically, 2. it's only dependent on data within each respective source table - no joins needed, and 3. it felt fairly "identifying" in nature

PR #2070, opened 2025-11-19
cscl - make lionkey unique

fvankrieken:
This does not get us closer to prod, if anything it gets us further. BUT it makes lionkey actually unique in dev

Final build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/19577334885/job/56066095866)

Please go by commit! I think they're fairly atomic.

Last commit shoutout to @damonmcc who last PR asked essentially "isn't there something better to use than ogc_fid"? And I said no... and I was wrong. globalid is apparently meant to be unique across ALL features in cscl (and is used in error logging, so it's gonna be needed in next pr, but also is just more appropriate to use as a unique key for protosegments)

Overall, diffs stay consistent, with 500 added diffs for seqnum since a bunch are offset now from what they were before (see screenshots below)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,12 +16,12 @@ on:
         - checkbook
         - colp
         - cpdb
+        - cscl
         - developments
         - facilities
         - factfinder
         - green_fast_track
         - knownprojects
-        - cscl
         - pluto
         - ztl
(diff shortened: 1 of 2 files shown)


fvankrieken:
Sorry - hold off on review

PR #2072, opened 2025-11-20
qa patch - ingest refactor fallout

fvankrieken:
relatively self-explanatory - I took a silly janky approach to this object that's supposed to return consistent structure for library, legacy ingest code, and current ingest code. This replaces that silly janky approach with pydantics built-in functionality instead.

This was causing issues in the QA app for the GR page - already deployed from this branch and it's looking good

~and migrating dcp_developments to ingest because it fails to ingest with library due to null geoms (that issue in the newer gdals)~ moved this to the pluto pr that already emrged

### DIFF
--- a/products/pluto/models/staging/stg__dcp_developments.sql
+++ b/products/pluto/models/staging/stg__dcp_developments.sql
@@ -1,12 +1,12 @@
 WITH final AS (
     SELECT
-        bbl::decimal::bigint::text,
+        bbl,
         bin,
-        units_co::numeric,
-        classa_prop::numeric,
+        units_co,
+        classa_prop,
         job_type,
         job_status,
-        date_complete::date
+        date_complete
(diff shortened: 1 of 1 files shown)


dcpy/lifecycle/ingest/validate.py
```diff
@@ -124,7 +124,7 @@ def find_definition_file_validation_errors(
 def validate_definition(ds_id: str, definition_dir: Path) -> None:
     errors = find_definition_file_validation_errors(ds_id, definition_dir)
     if errors:
-        raise ValidationError(errors)
+        raise ValueError(errors)
```
fvankrieken: ValidationError needs a little more construction - raised an error trying to raise this error. Just keeps things simple this way

damonmcc (approved):

PR #2075, opened 2025-11-21
update pluto zoning far for 25v3.1, migrate devdb to ingest

fvankrieken:
DevDB is used for our data checks, figured it made sense to migrate especially since it fails with library due to the weird null geom gdal bug (where the python gdal bindings fail translation when null geoms are present). We make it and it's fairly straight forward, so I just went ahead and ingested from this branch - all looks good in the build

[successful run](https://github.com/NYCPlanning/data-engineering/actions/runs/19645390440/job/56259263059)


products/pluto/pluto_build/data/dcp_zoning_maxfar.csv
alexrichey: Are we re-using these anywhere? Wondering if we should add them to ingest.

alexrichey (approved):
Had one question, but otherwise 👍

products/pluto/pluto_build/data/dcp_zoning_maxfar.csv
fvankrieken: They are used nowhere else but potentially yes. I feel like this specific dataset is worth chatting with zoning about to see if we want to handle it a little more holistically - get a better maintenance system than us occasionally emailing them, etc, and wouldn't mind just punting for now and continuing to treat this like a seed file.

products/pluto/models/_sources.yml
```diff
@@ -19,7 +20,14 @@ sources:
...
         arguments:
-          values: [ "Alteration", "Alteration (A2)", "Demolition", "New Building" ]
+          values:
+            [
+              "Alteration",
+              "Alteration (A2)",
+              "Alteration (Non-CO)",
+              "Demolition",
+              "New Building"
+            ]
```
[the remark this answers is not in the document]
fvankrieken: discussed in person

PR #2096, opened 2025-12-03
Use seed file to format text fields

### LINKED ISSUE #1954: Use `seed` table to read in formatting rules
dbt model `lion_dat_fields` does a lot of formatting based on the tables in the docs.

This is a bit duplicated with information in the seed file which has some info about the text format (for now just length/position) of the fields. This seed currently isn't used in the build, but rather for loading a "production" text file (from GR/the current C# application) back into the db after cutting up the text based on the info in the seed file.

But ideally we expand this seed file a little and make it the source of truth on how the output gets formatted. We'll then follow this pattern for the other flat files.

fvankrieken:
closes #1954, though I think it's worth making sure that we actually like this pattern before we move forward

latest [build passing](https://github.com/NYCPlanning/data-engineering/actions/runs/19936136121/job/57161386738)

go by commit. 2 is really the meat of this, 5 shows a little payoff as well

1. I think just makes sense in general, een outside the context of commit 2. Moves a bunch of centerline staging stuff to... the staging table. But also sets the stage for 2 by making the sql in `lion_dat_by_field` simpler
2. put formatting rules in seed file and use macro instead of having all that logic live in `lion_dat_by_field`. I think there's arguments to either approach honestly. Part of the sell here is that the seed file can really mimic these formatting tables that exist in the docs
    image
    This gets useful with things that share schema - see commit 5 which does this for sedat. It also gives us the ability the put all formatting rules (roughly) in one place which seems nice (commit 4 tries to address this a little too - makes seed subfolder to make it very clear what these are). Open to thoughts
3. just some downstream effects of 2
4. move lion dat formatting seed into a subfolder and rename
5. apply logic from 2 for the two sedat product tables

keeps output identical!

image

### DIFF
--- a/products/cscl/models/product/sedat/sedat.sql
+++ b/products/cscl/models/product/sedat/sedat.sql
@@ -1,13 +1,3 @@
 SELECT
-    FORMAT_LION_TEXT(lionkey, 10, ' ') AS lionkey,
-    COALESCE(parity, ' ') AS parity,
-    FORMAT_LION_TEXT(street_name, 32, ' ', FALSE, TRUE) AS street_name,
-    COALESCE(side_of_street, ' ') AS side_of_street,
-    FORMAT_LION_TEXT(lowaddress, 7, ' ') AS lowaddress,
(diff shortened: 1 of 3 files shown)


damonmcc (approved):
sick

Issue #1836, opened 2025-07-22
Special Address File


fvankrieken:
Section [7](https://nyco365.sharepoint.com/:w:/r/sites/NYCPLANNING/itd/edm/Shared%20Documents/PROJECTS/DATA%20ENGINEERING/PROJECT%20PLANNING/LION%20ETL/ETL_V8_02012024%20-%20Copy.docx?d=wd3194825302642958a316fb030749aa1&csf=1&web=1&e=4BSW2O&nav=eyJoIjoiMTUyOTQyODE1NSJ9)

Big one

PR #2098, opened 2025-12-04
CSCL SAF

fvankrieken:
#1836 

A lot of the lines added are docs!! I add them with relevant commits, but easiest spot to view the full file is [here](https://github.com/NYCPlanning/data-engineering/blob/fvk-cscl-saf/products/cscl/design_doc.md#special-address-files-saf)

[latest build](https://github.com/NYCPlanning/data-engineering/actions/runs/20604505342/job/59177244607)

## Background

(most of this is just copied from design_doc.md)

The Special Address File (SAF) contains address information for some street segments supplementary to the address information for the same segments contained in the LION file. There are 9 outputs - 5 different let's call them "output formats", 4 of which have 2 different outputs.

SAF records come from
- altsegmentdata - these are relatively "fully-fledged" and are basically designed to generate these output saf records
- certain CommonPlace records
- certain AddressPoint records
- NamedIntersection records

The different flags are:

| Flag | Source | Corresponds to | Meaning | Output |
| - | - | - | - | - |
| A | altsegmentdata | segment | Alternative address ranges | ABCEGNPX |
| B | altsegmentdata | segment | Alternative street names | ABCEGNPX |
| C | altsegmentdata | segment | Special record for Ruby Street on Brooklyn-Queens border | ABCEGNPX |
| D | altsegmentdata | segment | Duplicate or overlapping address ranges on the same street | D |
| E | altsegmentdata | segment | Neighborhood name usable in place of street name in addresses | ABCEGNPX |
| G | commonplace | segment | Non-Addressable place name of complexes | ABCEGNPX |
| I | namedintersection | intersection | Contain intersection names | I |
| N | commonplace | segment | Non-Addressable placename | ABCEGNPX |
| O | altsegmentdata | segment | Contain out-of-sequence addresses and opposite-parity addresse | OV |
| P | altsegmentdata | segment | Addressable placename | ABCEGNPX |
| S | addresspoint | segment | Suffixed house number(s) occurring at an intersection | S |
| V | addresspoint | segment | Contain Vanity Addresses | OV |
| X | commonplace | segment | Non-Addressable Place name of constituent entities of complexes | ABCEGNPX |

Outputs are really just grouped by what data/fields are expected to be associated with an output, as far as I can tell

## Review
Commits are generally split up by output file. Though for ABCEGNPX, since it's the largest and first output I tackled, I split it up into a few commits for clarity. The other outputs just have a single commit for essentially all their logic.

READ THE DESIGN DOC CHANGES IN EACH COMMIT FIRST (for commits that edit design_doc.md). This is the documentation for those transformations, so if that isn't giving you proper context for review, that's an issue that I need to address.

All feedback welcome on the documentation! Making it more readable, expanding on jargon, structuring it better.

### DIFF
--- a/products/cscl/models/intermediate/saf/int__saf_segments.sql
+++ b/products/cscl/models/intermediate/saf/int__saf_segments.sql
@@ -71,7 +71,6 @@ resolved_segment AS (
         COALESCE(proto.segment_type, primary_segment.segment_type) AS segment_type,
         COALESCE(proto.segment_type, primary_segment.segment_type) AS segment_incex_flag,
         saf.saf_source_table
-        --saf.source_attrs
     FROM with_roadbed_to_generic AS saf
     LEFT JOIN lion AS primary_segment -- TODO error report when none joined?
         ON


products/cscl/docs/roadbed_v_generic.png
fvankrieken: this is used in design_doc.md

products/cscl/macros/qa_compare_by_row.sql
fvankrieken: I originally was using this to create some QA tables. However, it seems overkill to start actually making qa tables for each file output for these smaller files, I've just been running this ad-hoc in dbeaver.

products/cscl/models/product/saf/_saf.yml
```diff
@@ -0,0 +1,44 @@
+version: 2
+
+models:
+# Models with column per field
+- name: saf_abcegnpx_generic_by_field
+  columns: &safa_columns
```
fvankrieken: Started to lose a little faith in this while writing it - I feel like it's a lot of effort to write/maintain tests like these when really, I'd rather just trust the formatting seed and the macro that generates all these columns (and also, these tests take a while to run - for LION, I think it's like a minute). And then, we test that the final text output (the "dat" table with row concatenated as text) has the right length, so that would catch most issues here. 

So I stopped part way through and thought I'd open it up to thoughts.

damonmcc: yea seems like a pain to maintain the column length tests. always doing the final "dat" table testing sounds good to me

products/cscl/models/intermediate/saf/int__address_points.sql
```diff
@@ -0,0 +1,48 @@
...
+        WHEN street_names.snd_feature_type IN ('E', 'F') AND address_points.house_number_suffix IS NOT NULL
+            THEN
+                -- Last character is A, B, etc -> converted to 1, 2, etc
+                (
+                    10000 * COALESCE(ASCII(RIGHT(address_points.house_number_suffix, 1)) - 64, 0)
+                    + address_points.house_number::INT
+                )::TEXT
+        WHEN address_points.hyphen_type = 'R'
+            THEN TRIM(SPLIT_PART(address_points.house_number, '-', 1))
+        ELSE address_points.house_number
```
damonmcc: this is a way to generalize the logic for Edgewater without having to explicitly write a `WHEN` for just that area?

damonmcc: oh sorry I forgot that's how it's flagged in the source data, and that's in the doc!

Issue #1953, opened 2025-09-18
Generalize flat file validation against prod outputs


fvankrieken:
Validation of flat files for LION is currently done as follows
- production outputs are received with working gdb from GR in sharepoint
- user downloads onto computer
- user manually loads them into db-lion.production_outputs (using `products/lion/poc_validation/dat_loader.py`, which also relies on dat structure defined in `seeds`
- during build, dbt compares the `lion_dat_fields` table to these manually loaded tables
- validation against prod outputs in bash using text processing happens manually outside of a build

This doesn't quite make sense in the context of producing 20ish files, and ideally being able to swap out versions easily.

Ideal state is something that needs to be brainstormed a little - I'd rather not cram these things through ingest/recipe to do the comparisons in sql, and that would leave out the actual bash text-based comparisons of the final files themselves. But what would make sense is a few things
- having a norm of where these prod outputs live, say in edm-private
- having a parameterized bash script that makes it easy to both load a version of a specific file output into the db, versioned, as well as generate necessary postgresql comparisons and text-based comparisons
- separating this out from the build (i.e. not having a dbt table in the build that compares to prod outputs)

PR #2103, opened 2025-12-05
cscl - export with dcpy and make loading prod outputs a tad easier

fvankrieken:
Meant to tackle part of #1953 

## Original motivation
Basically, I wanted to generalize things around output format and filenames for the main purpose of loading production outputs into the db for comparison to what we're building. So basically, if we have a folder with prod outputs (and working gdb) in s3

image

That we could just define somewhere how those filenames correspond to table names in the db and formatting

image

Then it would be simple to have some python that just with that can load stuff into the db

image

So for the next output (as in, you're implementing the SAF files), you'd just need to 
1. create a file in `seeds/formatting` that defines the field lengths and names of all the fields (which is also used to format text properly in the build)
2. add the product tables to the recipe
3. run the above command

And there's data in `db-cscl.production_outputs` to compare dev outputs to

## Why it's touching dcpy

Well, if we're mapping table names to desired outputs, that sounds a lot like a `recipe` of sorts. So figured I'd 1. put this info [in the recipe](https://github.com/NYCPlanning/data-engineering/pull/2103/files#diff-dd6a5f274d248957bf283a118b7f75c4b00e2c1637f966a49bc1d72b75609ea8) and then 2. write a quick lifecycle builds function to export.

It [runs](https://github.com/NYCPlanning/data-engineering/actions/runs/20044167495/job/57485639288) and works

## review

Just wanted to brainstorm a little on how we want to handle exports in dcpy - can always punt since that's a broader issue, but it seemed nice to include here. But before I spent too much time on it wanted other eyes on it, maybe agree on some acceptance criteria or whatnot.

### DIFF
--- a/dcpy/lifecycle/builds/build.py
+++ b/dcpy/lifecycle/builds/build.py
@@ -92,7 +92,7 @@ def export(
 
 
 @app.command("export")
-def _cli_wrapper_(
+def _export(
     recipe_lock_path: Path = typer.Option(
         None,
         "--recipe-path",
@@ -104,10 +104,6 @@ def _cli_wrapper_(
         Path(plan.DEFAULT_RECIPE).parent / "recipe.lock.yml"
     )
     export(recipe_lock_path)
-
-
-if __name__ == "__main__":
-    app()
 
 
 @app.command("upload")


fvankrieken:
I agree, that seems like a sometime in 2026 (maybe post orchestration-deployment) revisiting how the different stages connect and how we're defining them.

So with that, down to incorporate some big-picture feedback here if anyone has strong feelings (in addition to usual review), if you/y'all are down I'm sort of pro merge largely as it is here

dcpy/lifecycle/builds/build.py
```diff
@@ -10,6 +15,97 @@
...
+    file_path: Path,
+    format: ExportFormat,
+    pg_client: postgres.PostgresClient,
+) -> None:
+    """Import a recipe to local data library folder and build engine."""
+    logger.info(f"Exporting table {table_name} to {file_path} in format {format}")
+    match format:
+        case ExportFormat.dat:
+            with open(file_path, "wb") as f:
+                psql_process = subprocess.Popen(
```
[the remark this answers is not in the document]
fvankrieken: ... yeah I should just do this in python

dcpy/lifecycle/builds/build.py
```diff
@@ -10,6 +15,97 @@
...
+
+    if not recipe.exports:
+        logger.info("No exports defined in recipe, skipping export step")
+        return
+
+    build_name = metadata.build_name()
+    logger.info(f"Exporting build outputs for {recipe.name} build named {build_name}")
+
+    pg_client = postgres.PostgresClient(schema=build_name)
+    output_folder = Path(recipe.exports.output_folder)
```
alexrichey: I think we should use the `dcpy.lifecycle.config.local_data_path_for_stage(...)` for the default base path, and pass a potential override as an argument to the function. So probably something like `.lifecycle/builds/build/export/product/{{ version + build identifier }}/`. Then might be nice to specify a subpath in the recipe?

dcpy/lifecycle/builds/build.py
```diff
@@ -10,6 +17,99 @@
...
+        help="Path of recipe lock file to use",
+    ),
+):
+    recipe_lock_path = recipe_lock_path or (
+        Path(plan.DEFAULT_RECIPE).parent / "recipe.lock.yml"
+    )
+    export(recipe_lock_path)
+
+
+if __name__ == "__main__":
```
alexrichey: ✂️?

fvankrieken: Oops

PR #2108, opened 2025-12-08
PLUTO CoY Fields

alexrichey:
Adds the following to PLUTO
- Additional FAR values
- MIH Areas for Lots
- Transit Zones for Lots (two addl fields)

Per Matt, all new fields should be at the very end of the file. Explanations are in the commit messages, or comments in the sql. Logic for assignments for MIH Areas and Transit Zones is at the top of those SQL files.

I'm running a build now, and will post a link to the output when it finishes. I'll hold off from merging these changes until confirming logic with GIS and checking that their scripts can gracefully handle the changes.

How to review? Commits are atomic, and there aren't that many changes. The additional FAR values are straightforward (just continuing an existing pattern) and for MIH Area and Transit Zone there are maps below to visualize the logic and edge cases.

# **TODO**
- MIH: AR check with Zoning about a threshold. 
- Transit Zone:
  - Determine the severity of a mis-assignment
  - Determine how / Implement something to hold the tz steady. 
  - Double check pct logic with Zoning

# MIH Areas

# Logic
```sql
-- Similar logic to zoning districts:
-- calculate how much (total area and percentage) of each lot is covered by MIH areas
-- assign the MIH project ID to each tax lot based on which MIH area covers the
-- majority of the lot
-- a MIH area is only assigned if more than 10% of the lot is covered by the MIH area
-- OR more than 50% of the MIH area overlaps with the lot
```

The questionable splits for MIH Areas can be found in the `qaqc_int__mihareas_questionable_assignments` view in my build schema in postgres (`ar_pluto_new_fields_1893`)

## Some Questionable Splits
Overall, only 7 Lots are split between multiple MIH areas. Below are some odd cases

#### The following BBLs are split fairly evenly between two MIH Areas. 
BBL: 3037050014
The southern MIH area covers 50.9% of the BBL... however, you can see that that property is in the northern MIH area. 
**This one might actually be wrong**
image

---

BBL: 1020130029
These are split between an `Option 1 and 2` and `Option 1` area for `One45 Harlem for All`. However, I don't think this one matters - I think these lots are going to be demolished and redrawn into the MIH Areas as part of the `One45 Harlem for All` rezoning.
image

---

BBL: 2042260001
This is just sort of an odd case. This is  PELHAM PARKWAY, and not that it matters terribly, but it probably shouldn't be in either MIH Area. However, it meets the threshold for both, with the slight winner being 1776 Eastchester Road - Option 2, which covers 21% of the lot.
image

# Transit Zones

## Logic
```sql
-- Transit zone assignment logic:
-- Determine whether > 10% of a lot is covered by one of the transit zones
```

Basically all lots have a transit zone, except for ~2k that are either in the water or are large parks. (e.g. Central Park doesn't have a transit zone, understandably)

The questionable splits for MIH Areas can be found in the `qaqc_int__transit_zones_questionable_assignments` table. There are ~800 lots split between multiple zones.

## Questionable Splits (that probably don't matter)
A walkway in Staten Island:
image

A Park in Queens
image

## Questionable Splits (that might have implications)
There are a number of cases where the Transit Zones lines were just drawn poorly, and there's not much we can do about it. 

#### BBL = 4125920057
This lot is split 47% left zone / 53% right zone. 
image

And if you zoom out and look at it's block, that seems to indicate that our choice is correct. 
image

#### BBL =  4036620026 
One more for good measure.
image

We assign cases like this correctly, but there are certainly other cases where we don't. We could try to get clever, and in cases where we find ambiguity then give some preference based on the block. We could also introduce manual corrections.

### DIFF
--- a/products/pluto/pluto_build/sql/miharea.sql
+++ b/products/pluto/pluto_build/sql/miharea.sql
@@ -102,8 +102,7 @@ WHERE perbblgeom >= 10 OR maxpermihgeom >= 50;
 DROP VIEW IF EXISTS mih_distinct_options CASCADE;
 CREATE VIEW mih_distinct_options AS
 WITH split_options AS (
-    SELECT DISTINCT
-        unnest(string_to_array(cleaned_option, ',')) AS option
+    SELECT DISTINCT unnest(string_to_array(cleaned_option, ',')) AS option
     FROM mih_lot_overlap
 )
 SELECT DISTINCT trim(option) AS option
--- a/products/pluto/pluto_build/sql/transitzone.sql
+++ b/products/pluto/pluto_build/sql/transitzone.sql
@@ -174,14 +174,12 @@ WITH ambiguous_bbls AS (
     -- If a lot has ANY zone with priority_rank < 4, exclude "Beyond the Greater Transit Zone" (rank 4)
     SELECT
         ltr.*,
-        CASE
-            WHEN EXISTS (
+        NOT COALESCE(
+            EXISTS (
                 SELECT 1 FROM lot_to_tz_with_rank AS inner_ltr
                 WHERE inner_ltr.bbl = ltr.bbl AND inner_ltr.priority_rank < 4
-            ) AND ltr.priority_rank = 4
-            THEN false
-            ELSE true
-        END AS include_zone
+            ) AND ltr.priority_rank = 4, FALSE
+        ) AS include_zone
     FROM lot_to_tz_with_rank AS ltr
 )
 SELECT
@@ -199,7 +197,7 @@ SELECT
         ORDER BY pct_covered DESC
     ) AS tz_rank
 FROM filtered_zones
-WHERE include_zone = true;
+WHERE include_zone = TRUE;
 ANALYZE transit_zones_bbl_to_tz_ranked;
 
 -- Assign the primary transit zone by


alexrichey:
WIP Comment

### Questionable Splits
Overall, only 7 Lots are split between multiple MIH areas. Below are some odd cases

#### The following BBLs are split fairly evenly between two MIH Areas. 
BBL: 3037050014
The southern MIH area covers 50.9% of the BBL... however, you can see that that property is in the northern MIH area
image

---

BBL: 1020130029
These are split between an `Option 1 and 2` and `Option 1` area for `One45 Harlem for All`
The winner is the MIH Area in the northwest (the top-left)
image

---

BBL: 2042260001
This is just sort of an odd case. This is  PELHAM PARKWAY, and not that it matters terribly, but it probably shouldn't be in either MIH Area. However, it meets the threshold for both, with the slight winner being 1776 Eastchester Road - Option 2, which covers 21% of the lot.
image

ingest_templates/dcp_mih.yml
fvankrieken: What's the motivation for pulling this from bytes rather than edm-publishing?

products/pluto/models/qaqc/intermediate/qaqc_int__mihareas_questionable_assignments.sql
```diff
@@ -0,0 +1,57 @@
...
+    SELECT
+        m.bbl,
+        m.project_name,
+        m.affordability_option,
+        m.perbblgeom AS pct_covered,
+        p.geom AS lot_geometry,
+        p.address,
+        d.wkb_geometry AS mih_geometry,
+        -- Calculate how close the coverage percentage is to 50% (most ambiguous case)
+        ABS(m.perbblgeom - 50.0) AS ambiguity_score,
```
fvankrieken: Should this be maybe `50 - ABS(m.perbblgeom - 50.0)`? I would imagine 0 and 100 should be zero, and 50 should be the maximum ambiguity

fvankrieken: Especially for the `MAX` case below - that's really finding the least ambiguous case right now no?

alexrichey:
> I feel like for some of these cases, we either just leave or manually correct (using the same corrections framework as other corrections), and then if anyone ever flags something just add corrections.
> 
> Like 3037050014 seems like a prime case for correcting.
> 
> Maybe we create a little csv/shp export of "questionable" bbls as part of the build? I feel like this is maybe more of a Matt question, or could kick it to Transportation/Zoning/Sam and see if anyone wants to make tweaks to the official boundaries

@fvankrieken totally agree - I pinged GIS to start the conversation about what type of QA outputs they want. I'd imagined just pointing them at tables in our build schema, but csvs/shapefiles might be ideal. Will probably add something after talking to them.

ingest_templates/dcp_gis_mandatory_inclusionary_housing.yml
alexrichey: Swapped

fvankrieken: Sorry to keep pestering - but the motivation for renaming the dataset? Since we're pulling from the same place we always have for `dcp_mih`

alexrichey: no worries! Not pestering at all. 

re: template name change, and column renamings, this ingest template wasn't actually used anywhere, which is a good thing because it was pulling the wrong dataset. (inclusionary_designated_housing_areas). There's a brief note in the [commit body](https://github.com/NYCPlanning/data-engineering/pull/2108/commits/5c852da43e9ded5ddfa0b6393b9209bb1fb86c1b) and I'd kvetched a little about it in Teams. This is the one where we'd decided to delete what we'd archived in edm-data.

Figured it'd take the time to un-abbreviate the template it to be more inline with our convention while we're at it.

damonmcc (approved):

PR #2113, opened 2025-12-11
add a dcpy script

alexrichey:
~~python -m~~ dcpy lifecycle --help

Add as a script to the pyproject, enabling just:
```bash
dcpy lifecycle ingest --help
```


damonmcc:
would love to not have to do `python -m` in front of everything. shortening `lifecycle` to `lc` though would be a major readability loss for a minor keystroke gain so I don't think we should do that

`lifecycle ingest --help` or `dcpy ingest --help` would be great!

I'm down for `ingest --help` but part of me would rather never worry about conflicting cli tool names. I feel like when I install a cli tool I always expect to use the name of what I installed first and then the names of the things within it. seems odd to basically force `install dcpy`, `ingest --help`?

PR #2131, opened 2025-12-24
Add Tach deps viewer to experimental

alexrichey:
Sample output [here](https://app.gauge.sh/show?uid=f4703670-524b-498f-9486-08cd2bdb3503)

Was walking @jackrosacker through our thinking about dependencies in dcpy, and thought it might be nice to just generate a graph. Then I kinda got going with this a little more than I'd expected - it was a really good exercise in thinking about hierarchies, import directions, etc., a lot of which is fuzzy but sort of implicitly understood. (in general, the arrows in our graph make sense! Go us!).

It does make me think we should aim to reduce the number of arrows on the diagram. 
1. we should work to clarify the hierarchy of utils and models, so they're not importing from each other.
  - low hanging fruit: move some models out of `dcpy.models` and into lifecycle. (e.g. Product metadata models)
2. We might want to move toward exposing things in `__init__` files, to centralize the access points for modules.

## Potential uses
image
We could fail CI in cases like this.

## Alternatively...
Maybe we split out these different layers (utils, lifecycle, etc.) into their on python projects, and enforce import directions that way? Maybe within in dcpy, we've got projects for lifecycle, models, utils, etc. 

Then we don't even have to think about dependency graphs.

### DIFF
--- a/experimental/tach_dep_graph/README.MD
+++ b/experimental/tach_dep_graph/README.MD
@@ -1,0 +1,16 @@
+Generate a graph of the codebase.
+
+Run
+
+``` sh
+# from data-engineering root folder
+cp experimental/tach_dep_graph/tach.toml ./
+tach show --web
+
+# Optionally, to refresh the graph if you make changes
+tach sync
+```
+
+If we wanted to proceed with this type of thing, I'd worry about refactoring. This would break whenever files changed.
+So we'd likely want to generate the tach.toml, which I think would be pretty trivial. (I reckon we could implement that in an hour) 
+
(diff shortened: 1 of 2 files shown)


fvankrieken (approved):
>    we should work to clarify the hierarchy of utils and models, so they're not importing from each other.
>
>    low hanging fruit: move some models out of dcpy.models and into lifecycle. (e.g. Product metadata models)
>
>    We might want to move toward exposing things in __init__ files, to centralize the access points for modules.

Definitely agree with the first two, and probably the third too. The separate models are a bit clunky (and not needed) with the direction we've gone with connectors, so I think it might be time to axe them, not sure it really adds any value.

Any immediate thoughts on axing models entirely? Moving code into respective modules/domains

alexrichey:
@fvankrieken Great point re: ✂️ models. I'm in favor of doing away with it. @damonmcc thoughts?

PR #2165, opened 2026-01-12
drop HPD closings from KPDB

### LINKED ISSUE #2163: ad hoc KPDB build
Housing team has confirmed that removing sensitive HPD source data from KPDB will allow them to share the outputs with more agencies. They have an urgent need to get and use such a build and confirmed that all other source data can remain the same.

damonmcc:
resolves #2163

[all builds on this branch](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-kpdb-no-hpd-closings)

confirmed with Housing team on 1/29 that this should be permanent change to KPDB


alexrichey (approved):
LGTM

PR #2170, opened 2026-01-12
Ignore "rate limit exceeded" GH test failures

alexrichey:
These GH tests are repeatedly failing. I'd almost be in favor of just deleting these, but for now let's just ignore rate limit failures.


fvankrieken (approved):
Not lowered test coverage 😭

PR #2175, opened 2026-01-14
Ar migrate storage builds drafts

alexrichey:
Re-implements the Connectors for builds, drafts, published and gis, but doesn't swap them out just yet. You can read commit-by-commit for localized/grouped changes (e.g. just changes to builds functionality) as well as explanations for the smaller changes in the commit messages. (There's some 🌶️ content in the commit bodies)

Anyhow, the logic in `published.py` is basically distributed into files in a few different locations:
- connectors/edm/{conn_type}.py: The pure connector interfaces for the types above
- dcpy/lifecycle/builds/artifacts/{conn_type}.py: the business logic, including which connector to use, which files to upload, etc.
-  dcpy/lifecycle/builds/_cli.py: all the CLI commands

There will be some continuing commits where we implement this:
- Swapping out the commands ([PR here](https://github.com/NYCPlanning/data-engineering/pull/2179))
- TODO: delete publishing.

@fvankrieken @damonmcc Here's where I could use some input:
Consider a command like publishing a draft. This command needs to 
1. pull the draft (after determining some details about the draft itself)
2. push the draft to published

Ideally this code would live in a module a little bit "above" artifacts/drafts and artifacts/published, almost similar to how lifecycle/scripts lives "above" the various lifecycle stages. 

So I'd love it if you could take a look at the files under dcpy/lifecycle/builds/artifacts/*, and consider these "glue" functions like promote and publish. In the CLI they fit really nicely under namespaced commands like 
`dcpy lc builds artifacts drafts publish -p nypl_libraries -v 24v1 -dn 2 -ip`
which make sense, because we `publishing` is an action that we take on a draft. However, where this glue code should live... I could use some input.

### DIFF
--- a/dcpy/connectors/edm/builds.py
+++ b/dcpy/connectors/edm/builds.py
@@ -149,10 +149,8 @@ class BuildsConnector(VersionedConnector, arbitrary_types_allowed=True):
     ) -> dict:
         build_key = BuildKey(key, version)
 
-        # Construct the source key for the file
         source_key = f"{build_key.path}/{filepath}"
 
-        # Check if the file exists
         if not self.storage.exists(source_key):
             raise FileNotFoundError(f"File {source_key} not found")
 
@@ -168,7 +166,6 @@ class BuildsConnector(VersionedConnector, arbitrary_types_allowed=True):
             f"Downloading {build_key}, {filepath}, {source_key} -> {output_filepath}"
         )
 
-        # Use PathedStorageConnector's pull method
         self.storage.pull(key=source_key, destination_path=output_filepath)
         return {"path": output_filepath}
 


dcpy/connectors/hybrid_pathed_storage.py
```diff
@@ -282,9 +282,6 @@ def pull(self, key: str, destination_path: Path, **kwargs) -> dict:
         if not src_path.exists():
             raise FileNotFoundError(f"Source path {src_path} does not exist")
         if src_path.is_dir():
-            # Recursively copy directory
-            if destination_path.exists():
-                shutil.rmtree(destination_path)
```
fvankrieken: Omg. I'm sure I'm guilty of this too somewhere

dcpy/connectors/edm/builds.py
```diff
@@ -0,0 +1,199 @@
...
+        key: str,
+        version: str,
+        destination_path: Path,
+        *,
+        filepath: str = "",
+        **kwargs,
+    ) -> dict:
+        build_key = BuildKey(key, version)
+
+        # Construct the source key for the file
```
fvankrieken: nit of nits but can we get rid of some of the 

```python
# do the thing
do_the_thing()
```

Comments

dcpy/connectors/edm/builds.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @property
+    def storage(self) -> PathedStorageConnector:
+        """Lazy-loaded storage connector. Only initializes when first accessed."""
+        if self._storage is None:
+            self._storage = PathedStorageConnector.from_storage_kwargs(
+                conn_type="edm.publishing.builds",
+                storage_backend=StorageType.S3,
+                s3_bucket=_bucket(),
+                root_folder=PUBLISHING_BUCKET_ROOT_FOLDER,
```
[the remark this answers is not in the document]
fvankrieken: Ah sorry - that's by data product, so obviously not

dcpy/connectors/edm/builds.py
```diff
@@ -0,0 +1,199 @@
...
+        raise NotImplementedError(
+            "Builds don't have a meaningful 'latest' version. Use list_versions() to see available builds."
+        )
+
+    def version_exists(self, key: str, version: str, **kwargs) -> bool:
+        """Check if a specific build exists."""
+        return version in self.list_versions(key)
+
+    def data_local_sub_path(self, key: str, *, version: str, **kwargs) -> Path:
+        return Path("edm") / "builds" / "datasets" / key / version
```
fvankrieken: Bit of a nit, but this would presumably be so someone else (i.e. the lifecycle code pulling this) knows where to put it right? Can we not do "datasets" then? builds/datasets makes me think of what's going into a build. Even if it's "data_products".

Or would it just make sense to have this match what's in DO? as in, key / builds / version?

alexrichey: I need to review how this even works, honestly... The `local_sub_path` stuff seems really misplaced. Would honestly like to refactor it out if possible. For now, I'll just leave it be

PR #2221, opened 2026-02-06
Ar pluto new fields 1892 p2

alexrichey:
Rewrites MIH Areas and Transit Zone assignments based on our meeting with folks from Transit and Zoning. For logic on the assignments, see the sql files themselves. 

Alas, I don't have a passing build to link here - the last one failed on the dbt QA stuff due to a mistake by me. After our demo, I'm going to re-run the build which should pass. 

For the QA views, they're mostly to help diagnose/demo/visualize the edge cases. The tables look like this:
image

Which allows you to visualize all the relevant geometries in a single row.

*note* I've got this pointed at the previous PR, so you can see what changed between then and now. Feel free to change to point at `main` so you can see all changes. Also, I'm going to wait to merge this until QA is finished so that (if necessary) we can still build pluto without these changes.


products/pluto/tests/assert_only_valid_mih_options.sql
damonmcc: (nit) I think this could be an `accepted_values` dbt test in `_sources.yml`

PR #2241, opened 2026-02-19
DBT'ify pluto staging models

alexrichey:
1. Adds all PLUTO data sources as staging models, so we can now correctly reference them in subsequent models. Also changes all references in subsequent SQL to refer to stg__* models.
2. Moves the "local csvs" to be DBT seeds. 
3. Adds a DBT model to compare the build to `nightly_qa.pluto`. We don't execute this automatically, as it takes a few minutes to run. 
 - This is what I'm using to ensure that this PR doesn't change the PLUTO table. I'm re-running a build now, but last I checked, this model produced no records -> _**This PR doesn't change the resulting PLUTO table at all.**
 - However, it would be really, really cool if we automated this so that any PR against PLUTO produced the diff from nightly_qa and posted it to that PR. Totally feasible.

After this, it should be trivial to keep converting sql scripts to models whenever we touch PLUTO.

Build [here](https://github.com/NYCPlanning/data-engineering/actions/runs/23664276812)


damonmcc (approved):
monumental! left a couple comments. if there aren't any diffs, LGTM!

worth squashing the commits I think

PR #2280, opened 2026-03-19
Add COLP usetypes mapping as dbt seed

alexrichey:
### USE TYPES
Unfortunately (see [here](https://github.com/NYCPlanning/data-engineering/issues/2225#issuecomment-4056722799)) DCAS IPIS usetypes changed from full to truncated versions. We should ask them about this, but in the meantime it was easy enough to just make a translations table from the newer truncated types (e.g. "POLLUTION CONTRL FAC") to the older versions ("POLLUTION CONTROL FACILITY") to maintain stability. 

There are only two newer truncated types that we're seeing now, but haven't seen before:
"EMERGENCY MEDICL STN"
"COURT (SPORTS)"
I'm going to leave these be. All other types are accounted for.

### OTHER REFACTORS
I don't have geosupport locally, so I broke the scripts apart to enable running just the sql part of things.


damonmcc (approved):

damonmcc (approved):
worth squashing the commits

PR #2312, opened 2026-04-08
Glue code for fgdb metadata

jackrosacker:
## What this PR does

Adds a unified `dcpy lifecycle package esri write_metadata` CLI command that writes Esri-format metadata XML into both shapefiles and geodatabases from product-metadata YAML. Previously there was no GDB write path; the shapefile path was a standalone command in `shapefiles.py` that is now removed.

## Key changes

- `write_metadata` CLI: single command for both file types. GDB layer is inferred automatically for single-layer GDBs via `fgdb.resolve_layer`; required explicitly for multi-layer GDBs and shapefiles.
- Per-layer column overrides: `File.layers` / `GdbLayerOverrides` added to the metadata model so GDB entries in `metadata.yml` can declare per-layer column overrides, merged at write time via `calculate_layer_dataset_metadata`.
- Esri type overrides: format-specific types (`String`, `SmallInteger`, etc.) go in `column.custom["fgdb_data_type"]` / `custom["shp_data_type"]` rather than `data_type`.
- Cleanup: old shapefile write command removed; low-level metadata helpers moved to `esri_metadata.py`; tests reorganized to match.

## Also of note
There are a bunch of complementary changes to the pluto YAML in the product-metadata repo. Since that repo has been archived, I plan to get this repo merged, and then will transfer the p-m edits directly in a new PR. Until then, the YAML changes can be seen in the test cases.

resolves #2303

fyi @alexrichey - I may need your help with all of the merge conflicts associated with this PR!


alexrichey (approved):
Looks great!

jackrosacker:
@alexrichey - flagging this PR again for you for two reasons:
- I had to re-add a couple of test fixtures temporarily that you'd removed previously, just so that the tests in this current PR pass. 
  - The immediate next step after merging this PR is to migrate my pending product-metadata edits into this repo, remove those interim fixtures again, and move forward with the clean state that you'd intended when combining repos originally! 
  - Flagging that small change here for your approval
- I was getting a PDF generation pytest failure when pushing to GH. Fixed this by changing a jinja template to allow for a tags value of None. Since this is outside the scope of my initial work, I kept as a separate commit for your review before I rebase/squash again before merging.

alexrichey:
> *
> 
> * I was getting a PDF generation pytest failure when pushing to GH. Fixed this by changing a jinja template to allow for a tags value of None. Since this is outside the scope of my initial work, I kept as a separate commit for your review before I rebase/squash again before merging.

@jackrosacker 
Gotcha - thanks for fixing up. We don't actively use the pdf generation at the moment, so no danger of breakage there. Please merge!

PR #2313, opened 2026-04-08
Implement Packaging All OTI XLSX files

alexrichey:
Enables packaging metadata assets and overridden metadata. 

Still TODO: I'm going to wait to hook this up to distribution until we implement packaging. There's just too much going on in the scripts currently, and the whole packaging + distribution flow would take a fair bit of manual testing.

### Assets
Currently just creates all OTI Excel data dictionaries... there are about ~50 of them

### Product Metadata
Computes the metadata for all destination datasests, e.g. "the PLUTO shapefile that that goes to OpenData"

### Overall 
For PLUTO, it looks like this:
image

We can run this in PMD on merge to main to create a release. Then we when we distribute, we can pull the zip grab the relevant files.

1. Generate metadata for individual dataset (in this example, just zap.bbls)
```
    python -m dcpy lifecycle package metadata generate zap.bbls \
    --output-dir richey_stuff/metadata_outputs \
    --version 26v1

Result:

    - Generated 1 file: zapprojects_datadictionary.xlsx
    - Output location: richey_stuff/metadata_outputs/zap/bbls/zapprojects_datadictionary.xlsx
```

2. Generate metadata for all datasets (no arguments)
```
    python -m dcpy lifecycle package metadata generate \
    --output-dir richey_stuff/metadata_outputs \
    --version 26v1

Result:

    - Auto-discovered 45 datasets with XLSX metadata
    - Generated 47 XLSX files across those 45 datasets
    - Only generates files for datasets with is_metadata=True and type=oti_data_dictionary
    - Output structure: {product}/{dataset}/{filename}.xlsxx
```

### DIFF
--- a/dcpy/lifecycle/package/_cli.py
+++ b/dcpy/lifecycle/package/_cli.py
@@ -96,25 +96,18 @@ def list_versions(
     ),
 ):
     """List all destination versions as JSON."""
-    import json
+    from dcpy.product_metadata.versions import DestinationVersions
 
     org_md = product_metadata.load(version=version)
-    versions = org_md.get_all_destination_current_versions()
-
-    # Build JSON: {"product.dataset.destination": "version"}
-    versions_dict = {}
-    for line in versions:
-        if "|" in line:
-            dest, ver = line.split("|", 1)
-            versions_dict[dest] = ver
-
-    output = json.dumps(versions_dict, indent=2)
+    dest_versions = DestinationVersions.from_org_metadata(org_md)
 
     if output_file:
-        output_file.write_text(output + "\n")
-        typer.echo(f"Wrote {len(versions_dict)} destination versions to {output_file}")
+        dest_versions.to_json_file(output_file)
+        typer.echo(
+            f"Wrote {len(dest_versions.versions)} destination versions to {output_file}"
+        )
     else:
-        typer.echo(output)
+        typer.echo(dest_versions.to_json_string())
 
 
 @metadata_app.command("diff-versions")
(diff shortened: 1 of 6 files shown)


dcpy/lifecycle/package/_cli.py
```diff
@@ -15,6 +18,233 @@
...
+    ),
+    output_file: Path = typer.Option(
+        None, "--output", "-o", help="Output file (default: stdout)"
+    ),
+):
+    """List all destination versions as JSON."""
+    import json
+
+    org_md = product_metadata.load(version=version)
+    versions = org_md.get_all_destination_current_versions()
```
damonmcc: (idea for later) `get_all_destination_current_versions` should probably return a `dict` or something instead of a `list[str]`

dcpy/lifecycle/package/_cli.py
```diff
@@ -15,6 +18,233 @@
...
+
+    if output_file:
+        output_file.write_text(output + "\n")
+        typer.echo(f"Wrote {len(versions_dict)} destination versions to {output_file}")
+    else:
+        typer.echo(output)
+
+
+@metadata_app.command("diff-versions")
+def diff_versions(
```
damonmcc, line 114: are there useful data classes or pydantic models for all the json-ish/dictionary stuff here?

seems easier to catch issue and maintain with models rather than dictionaries

alexrichey: true - will add some models to the product-metadata module, and move this comparison logic there.

alexrichey: @damonmcc Moved some stuff around in the last commit, and cleaned up the Version comparison in particular.

damonmcc (approved):

PR #2314, opened 2026-04-08
add distribution page to QA app

damonmcc:
adds a page to the QA app to show a table of current versions on Bytes and Open Data

having a button to get cached versions may only be useful for local dev work but would like to keep it for now to see if it's needed in the deployed app (e.g. "if I refresh the page, I don't necessarily want to fetch versions again")

## screenshots from local app

Screenshot 2026-04-08 at 4 26 25 PM

Screenshot 2026-04-08 at 4 46 22 PM

Screenshot 2026-04-08 at 4 56 16 PM


alexrichey (approved):
love it

Issue #2315, opened 2026-04-12
CSCL in QA app


damonmcc:
To review and resolve differences between dev and prod outputs, DE and GR need to be able to see overviews and record-level details of the discrepancies.

Sharing/updating the latest state of the discrepancies shouldn't be manual process.

## current state

[stub]

## desired state

[stub]

damonmcc:
added a page to the app. TDB what we want to add to the page next

damonmcc:
GR is starting to use the QA app to review diffs

PR #2321, opened 2026-04-16
add CSCL to QA app

damonmcc:
related to #2315

just adds the page and a single table for now. haven't scoped out ideal page features in the linked issue yet

ran a build on this branch in order to exclude build artifacts from the diffs

Screenshot 2026-04-16 at 10 39 30 AM

### DIFF
--- a/apps/qa/src/pages/cscl/helpers.py
+++ b/apps/qa/src/pages/cscl/helpers.py
@@ -1,0 +1,23 @@
+import pandas as pd
+import streamlit as st
+
+from dcpy.connectors.edm import publishing
+
+PRODUCT = "db-cscl"
+
+
+@st.cache_data(show_spinner=False)
+def get_builds() -> list[str]:
+    return publishing.get_builds(PRODUCT)
+
+
+@st.cache_data(show_spinner=False)
+def get_build_version(build: str) -> str:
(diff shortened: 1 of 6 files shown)


apps/qa/src/pages/cscl/helpers.py
```diff
@@ -0,0 +1,23 @@
+import pandas as pd
+import streamlit as st
+
+from dcpy.connectors.edm import publishing
+
+PRODUCT = "db-cscl"
+
+
+@st.cache_data(show_spinner=False)
```
alexrichey, line 9: Just curious - for the caches, how long are things cached?

alexrichey (approved):
Sweet

apps/qa/src/pages/cscl/helpers.py
damonmcc: by default, forever. but it isn't saved to disk because this doesn't declare `persist="disk"`,  ([docs](https://docs.streamlit.io/develop/api-reference/caching-and-state/st.cache_data))

damonmcc: there's probably no strong reason to cache these, I don't remember them taking a long time

PR #2323, opened 2026-04-17
Add Dagster Ingest and Cron Cleanup Job

alexrichey:
Adds an asset for each ingest template, and a cron (schedule weekly, Sunday at 2am) scheduled to clean up the data directory in dcpy. 

### TODO
- The volume path'ing in the docker container-compose should be relative. This was just the first swing at getting this operational. Same for the INGEST_TEMPLATES_PATH
- apps/dagster/ingest/config_presets.py is a little hacky. We need to find a nice way to interact with job configs in the UI.
- in apps/dagster/ingest/definitions.py, we should store job details in Postgres at some point.


damonmcc (approved):
nice! python formatting test is failing but maybe that's from bumping package version on main

alexrichey:
> nice! python formatting test is failing but maybe that's from bumping package version on main

@damonmcc  Gah... it's a local `ruff check` vs `ruff format` thing. Fixing...

PR #2324, opened 2026-04-19
update packages

damonmcc:


dcpy/utils/geospatial/transform.py
```diff
@@ -54,11 +54,6 @@ def df_to_gdf(df: pd.DataFrame, geometry: file.Geometry) -> gpd.GeoDataFrame:
             "❌ Specified geometry column does not exist in the dataset."
         )
 
-        # replace NaN values with None. Otherwise gpd throws an error
```
alexrichey, line 57: do you know where this was throwing the error?

alexrichey (approved):

dcpy/utils/geospatial/transform.py
damonmcc: ingest. but with the latest geopandas version, it doesn't error anymore

PR #2330, opened 2026-04-23
EDDE migrate AE ETL

alexrichey:
This adds two bits of functionality
- Ports [this repo](https://github.com/NYCPlanning/ose-equity-tool-etl/tree/main) into our EDDE product code. I hand-copied the cells of the notebook into a plain python file (commit 1) then did some refactoring/ruffing/etc.
- There's a script that lives... somewhere... that converts our EDDE outputs into changes over time. This is the dataset referenced here in line one. 
```
Combine data from puma (sort by puma), then boro (sort by boro), then citywide into one csv, or one csv for each vintage

Do whatever you need to do to create 2 files, one containing all of the old vintages, one with all of the new vintages.  

The column headers should be the same for each one, replacing the date with the string "change".

Convert those two csvs to json (https://data.page/csv/json)

Insert those json objects in the script as dataSetOld and dataSetNew

Make sure the "labels" array is up to date with all of the base variable names (generally the ones ending in "_count", "_rate", "_median")

Run "node convert-whateverfile.js"

Take the json file it output and convert it to csv (https://data.page/json/csv)

Go to the bottom of the csv, re-add the boro and citywide
```

Anyhow, I've recreated that under `products/edde/post_processing/change_over_time/`. It matches EXACTLY AE's output from the previous year. (minus 31 cells with NaN/0 discrepancies. I don't think that this matters) 

This isn't in use yet, but will likely be used to generate this year's EDDE.

### ALSO!
dcpy will now yell at you when you have misconfigured your env variables. It'll warn you if, say, you haven't set the var for the Lifecycle data directory (thereby causing your loaded data to go into whatever directory you happened to run the recipe load command in)


pyproject.toml
```diff
@@ -140,6 +140,7 @@ dependencies = [
     "geoalchemy2",
     "geopandas",
     "ijson",
+    "simplejson",
```
damonmcc, line 143: we don't need to add this to `requirements.in` and compile `requirements.txt`?

damonmcc (approved):
nice! love the dcpy config warnings

do those files in `/generated/` cover all the outputs from the previous year?

pyproject.toml
alexrichey: oh, yeah, we actually do. Shoot... had assumed we were just installing the dcpy package, where (in container setup) we're just ignoring its deps:
> python3 -m pip install $editable_install_flag . --constraint ./admin/run_environment/constraints.txt _--no-deps_

Will add. Good catch.

alexrichey:
> previous

@damonmcc yeah, they do. I think we might end up deleting them, or jinja templating them. We'll have to figure it out with AE as we keep going on this.

PR #2339, opened 2026-04-30
improve CSCL QA app page

damonmcc:
related to #2315 

improvements to CSCL diff generation and the QA app CSCL page

## `poc_validation` scripts

- `run_validation.py` (new) — generates diffs by streaming dev and prod files directly from S3, no disk usage for inputs, no full build required
- `validate_outputs.sh` — writes `validation_summary.csv` with per-file prod row counts; `mkdir -p` so it's re-runnable; adds a note that `run_validation.py` is now the primary method
- `summarize_diffs.py` — reads `validation_summary.csv` to add `# of rows in prod` and `% of rows with diffs` columns to the summary CSV

## QA app

- diff row viewer: select a file with diffs, see the dev-only rows
- field-level diffs section: reads from the existing dbt `etl_dev_qa` models (LION DAT, ThinLION, RPL) grouped by output type
  - this is will probably become the primary place to describe and review diffs. a final list of viable tables here is TBD and based on the CSCL dbt project

## stuff to do in follow-ups

- run `run_validation` in the QA app to truly decouple QA from builds
- delete redundant scripts like `validate_outputs` and `summarize_diffs`

## screenshots

Screenshot 2026-05-03 at 9 55 29 PM

Screenshot 2026-05-03 at 9 55 44 PM


alexrichey (approved):
Good stuff!

PR #2347, opened 2026-05-05
Ar cscl diffs1

alexrichey:
Adds (in order of importance):
- Macro'ification of the dbt audit helper + the logic we'd been using to condense its wide output into a single json column with diffs. 
- A framework for accounting for diffs. `products/cscl/models/etl_dev_qa/diffs/thinlion/qa__diffs_thinlion_summary.sql` is a good example: you can see there's some custom logic there to determine whether we've accounted for diffs. In the case of thinlion, the police-precinct level diffs are due to algorithmic differences for geo-joins, and there's not much we can do about that. So I've marked them as accounted for. And a similar case in lion_dat - `segment_seqnum`s can be different from prod. Anyhow, ideally, each output file with diffs has a corresponding qa diff model, which will allow us to track discrepant rows/fields.
- ⚠️New Convention Alert! ⚠️: I've added a folder for `qa_int` with models like `qa_int/flatfiles/qa_int__prod_citywide_lion_dat.sql`. This is for slight modifications on production tables, so that we can easily compare them to our build outputs via the macro. E.g. for the lion_dat prod output, the model adds a `_lion_key` 
- Adds an explicit seed with file groupings, output filenames, file_ids, etc. This just tabularizes the recipe, but we can now use it to group the diffs, allowing us to answer questions like "how many discrepant records are in the District Equivalency files"
- Last, and definitely least, adds some color to the DBT Dag nodes


damonmcc (approved):
can't wait to use em in the QA app! and to find a nice way to set `accounted_for` values for diffs

PR #2382, opened 2026-05-18
Add Melissa Geocoding

alexrichey:
I'm putting this in experimental (as opposed to products) because it's not clear to what degree we'll need this in the future. Maybe we promote it to a real product at some point.

It takes liberally from the old db-melissa repo, as well as other geocoding. It would be nice to have `dcpy` utils to make this a little more `dwim`y. Some day...

### DIFF
--- a/experimental/melissa/neighborhoods_to_boroughs.csv
+++ b/experimental/melissa/neighborhoods_to_boroughs.csv
@@ -1,0 +1,330 @@
+neighborhood,borough
+Allerton,Bronx
+Bathgate,Bronx
+Baychester,Bronx
+Bedford Park,Bronx
+Belmont,Bronx
+Bronx Park South,Bronx
+Bronx River,Bronx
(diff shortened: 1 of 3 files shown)


experimental/melissa/neighborhoods_to_boroughs.csv
damonmcc, line 1: these values do look like neighborhoods but then it's used to map "city" values via `BOROUGH_MAP.get(city_upper, None)`

is it just a bad column name in the source data? or is this just to map Queens cities in a citywide way?

damonmcc (approved):

experimental/melissa/neighborhoods_to_boroughs.csv
alexrichey: yeah, it's just for queens. I didn't dive too deeply into it... if geosupport is happy, I'm happy. (that is, after spot-checking a random sampling of geocoded rows)

PR #2385, opened 2026-05-20
add archive_date to source_data_versions via recipe planning

damonmcc:
adds `archive_date` column to planned recipes and `source_data_versions` tables during load step of a build

having the "archived" dates for loaded source data be easily available during a build would help us quickly determine how stale source data is and decide if we need to get fresher data (e.g. is a passing build ready for QA?)

the date is the `date-created` or `last_modified` from the S3 metadata of the file in the `edm-recipes` bucket. didn't try to make it more clever with something like "use the Open Data metadata" but I think that's in the ingest metadata and technically usable.  seems nice for this to be generalized and not too dependent on data source type

[FacDB build run](https://github.com/NYCPlanning/data-engineering/actions/runs/26138099777)

some of the `source_data_versions` table records from a FacDB build
- Screenshot 2026-05-19 at 10 55 02 PM

### DIFF
--- a/dcpy/connectors/edm/recipes.py
+++ b/dcpy/connectors/edm/recipes.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from pathlib import Path
 
 import pandas as pd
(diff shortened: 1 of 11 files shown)


alexrichey (approved):
Love it

PR #2415, opened 2026-06-03
export parquet files and dbt artifacts

damonmcc:
## why
- we love parquet files and we might as well start generating them at the end of builds as (eventual) primary outputs along with the usual csv, shapefiles, and gdbs
- some of the artifcats generated by dbt builds are useful metadata (compiled SQL, run results, [manifest](https://docs.getdbt.com/reference/artifacts/manifest-json))

### DIFF
--- a/dcpy/lifecycle/builds/__init__.py
+++ b/dcpy/lifecycle/builds/__init__.py
@@ -25,4 +25,8 @@ BUILD_PLAN_ARTIFACTS = [
     "source_data_versions.csv",
 ]
 
+BUILD_ARTIFACT_DIRS = [
+    "target",
+]
+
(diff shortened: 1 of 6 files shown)


dcpy/lifecycle/builds/__init__.py
```diff
@@ -25,4 +25,8 @@
     "source_data_versions.csv",
 ]
 
+BUILD_ARTIFACT_DIRS = [
```
alexrichey, line 28: hrm... at some point, we should move these into a constants file so we can have unidirectional imports from in __init__. (ie __init__ should expose attributes from it's children, but never the other way around)

alexrichey (approved):
sweet! thanks

PR #2416, opened 2026-06-05
add an agency column to CPDB 

damonmcc:
related to https://github.com/NYCPlanning/ae-data-flow/pull/172

adds a column from `dcp_projecttypes_agencies.csv` to the CPDB table `ccp_commitments` which gets exported as `cpdb_planned_commitments.csv`

[builds on this branch](https://github.com/NYCPlanning/data-engineering/actions/workflows/build.yml?query=branch%3Adm-cpdb-budget-agencies)

## todos

**probably will drop this change in favor of the more long-term fix of using `dcp_managing_agencies_lookup` in facdb**

the new column works but the values in it probably don't align with `dcp_managing_agencies_lookup` used by FacDB. should either change the values in `dcp_projecttypes_agencies.csv` or try using `dcp_managing_agencies_lookup` for commitments agency values

## why

`agencyabbrev` has longer acronymns than `agencyacronym`. and those longer acronymns are what all other data products use, espcially those defined in (dcp_managing_agencies_lookup)[https://github.com/NYCPlanning/data-engineering/blob/main/ingest_templates/dcp_managing_agencies_lookup.yml]

this is a temporary fix to give AE a better column to use for commitment agency data. this csv should probably be consolidated into CPDB's `lookup_agency.csv` or into the dataset `dcp_managing_agencies_lookup`

### DIFF
--- a/products/cpdb/seeds/dcp_projecttypes_agencies.csv
+++ b/products/cpdb/seeds/dcp_projecttypes_agencies.csv
@@ -1,41 +1,41 @@
 projecttype,tycs,agencyname,agencyabbrev,agencycode,agencyacronym,agency,projecttypeabbrev
-Correction,CORRECTION                                                  ,New York City Department of Correction,NYCDOC,72,DOC,Department of Correction,C
-Courts,COURTS                                                      ,New York City Office of Court Administration,NYCOCA,0,OCA,Office of Court Administration,CO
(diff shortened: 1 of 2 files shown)


damonmcc:
moved changes to https://github.com/NYCPlanning/data-engineering/pull/2462

PR #2440, opened 2026-06-13
improve docs

damonmcc:
a first pass at consolidating and improving our docs related to development. each dev can still have their own local docs to expand and personalize stuff, but this will hopefully minimize the volume of local docs

for stuff that came from the wiki, the content has already been removed from the wiki

## changes

- in general, move docs from the wiki to the repo if they're useful for development. this makes them more accessible when they need to be and makes them more likely to be kept in sync with code changes
- `docs/dcpy/architecture.md`: layered import flow, known deviations
- `docs/dcpy/README.md`: slim down to be more of a quick reference to other docs
- `docs/testing.md`: suites, how to run, conventions
- `docs/contributing.md` and `docs/conventions.md`: pretty much copy-pasted what we had in the wiki
- [`tach.toml`](https://github.com/tach-org/tach) enforcement of import flow (advisory for now)


docs/testing.md
```diff
@@ -0,0 +1,80 @@
...
+> haven't installed the test extras locally, collection fails; `uv run python -m pytest dcpy/test
+> --noconftest` is the workaround. In CI the test deps are installed, so this isn't needed there.
+
+The **product-metadata tests** (`dcpy/test/product_metadata/`) additionally need a local clone of
+the product-metadata repo via `PRODUCT_METADATA_REPO_PATH` — see
+[dcpy/README.md → Testing](./dcpy/README.md#testing).
+
+### dcpy integration tests (`dcpy/test_integration/`)
+
+Require real services (Postgres, SFTP). We'd like these fully in CI but haven't resolved spinning
```
alexrichey: I think this is out of date. We've got these going in CI. Can delete the test_integration README, I think

tach.toml
```diff
@@ -0,0 +1,119 @@
+# Import-boundary config for dcpy. FIRST PASS — the layer assignments encode the
+# *intended* dependency direction and are meant to be reviewed/confirmed by the team.
+# Run `uvx tach check` to list violations; `uvx tach show --web` to visualize.
+#
+# Layers are ordered highest → lowest. A module may only import from its own layer
+# or a lower one. `utility = true` modules are importable from anywhere.
```
alexrichey, line 6: would be curious to see the graph on this. I'd also created a tach deps in experimental/tach_dep_graph/tach.toml that was fairly well fleshed out. Though it (and this) need to remove models, which should be gone now.

.gitignore
```diff
@@ -141,7 +141,6 @@ dmypy.json
 
 # VS Code local settings
 .vscode
-AGENTS.md
```
alexrichey, line 144: thinking to just leave this in user-level git ignores?

alexrichey (approved):
Good stuff. A few questions, but happy if we want to merge and iterate

docs/testing.md
damonmcc: true! updated, I should've had more faith in our CI tests

tach.toml
damonmcc: that experimental tach.toml was the inspiration for this one. reviewing, refining, and using this seems like the best follow-up work for us

PR #2458, opened 2026-06-23
Implement Dagster Builds for TemplateDB and EDDE

alexrichey:
Goodness gracious. This is a lot. But we now have working Dagster builds for `Template DB` and `EDDE`, and it should be fairly trivial to implement the rest of our product builds. 

## Dagster
image

For each product, we generate the DAG. For Template DB, we can go from plan all the way through distribute 🚀🚀🚀

### 1. Standardizes the build location and folder outputs. 
For a plan / build, we now write outputs to a configured folder, instead of writing the lockfile and build outputs to the source directory, e.g. `products/edde/`. This is also standardizes EDDE and Template to output in our package format, like so:
image

This format is distributable!

### 2. Implements lifecycle build commands
we can now kick off either shell commands, or python commands, as declared in the recipe:
image

For each product, we'll just need to add the build commands to the recipe. That should be enough!

TODO: For each command, we should use [Dagster Ops](https://docs.dagster.io/guides/build/ops) to define the substeps, which would make the individual steps of a build repeatable. Ie if the export step failed because of a typo, you can fix that and rerun, without doing the full build again. 

### 3. Recipe Templating. 
You can now add jinja template vars to the recipe, and dagster will require you to specify these when you plan.


damonmcc:
resolves https://github.com/NYCPlanning/data-engineering/issues/2380

PR #2548, opened 2026-08-07
Scaffold the LIFT project

alexrichey:
- Adds the LIFT project basics. The DCAS LIFT dataset is in edm.private. 
- Adds to the ability to specify duckdb as an alternative for postgres / file recipe loading. When duckdb is specified, it will create a duckdb database file in your build directory then load recipe sources.
- Adds an edm-private connector

New commits:
- Fix the edm-private connector (refactor off ingest datastore)
- add `dcp_duck_my_build` CLI to run the server. (run it from the product's dir, and it'll infer the build path, and find you duckdb file)


products/lift/recipe.yml
```diff
@@ -0,0 +1,42 @@
...
+
+inputs:
+  dataset_defaults:
+    destination: duckdb
+    file_type: parquet
+  missing_versions_strategy: find_latest
+  datasets:
+    # Private CSV file from edm-private bucket
+    - name: dcas_lift
+      source: edm.private
```
damonmcc, line 13: since the new `edm.private` connector seems to be an `ingest_datastore.Connector`, does loading these work even though they aren't archived via ingest?

maybe it's just a matter of avoiding triggering functions that read a datasets's `metadata.yml`

damonmcc (approved):

products/lift/recipe.yml
alexrichey: @damonmcc yeah, it's a good callout. It does loading just fine, but listing versions doesn't work so well. I'll try and refactor this now...

damonmcc: I guess adding ingest templates for `dcas_lift` and `dcp_dri_subindices_indicators` would also mean no need for an `edm.private` connector since they'd end up in the `edm-recipes` bucket

but honestly I like the current approach to quickly use the source files we have and not have to enshrine them as ingest templates/versioned datasets just yet

alexrichey: yeah, that's mostly what I was going for... recipes / ingest datastore has all that machinery. It's good eventually, but if we just have a file from some other agency and want to get something going quickly (and not worry about ACLs, processing steps, etc.) then this is nice.

damonmcc:
related to #2549

PR #2558, opened 2026-08-10
LIFT Joins

alexrichey:
Implements LIFT as a pipeline using DuckDB and python.


admin/run_environment/constraints.txt
```diff
@@ -904,32 +917,32 @@ snowplow-tracker==1.1.0
...
+soupsieve==2.9.2
     # via beautifulsoup4
 sqlalchemy==2.0.51
     # via
     #   -r requirements.in
     #   geoalchemy2
 sqlalchemy-stubs==0.4
     # via -r requirements.in
-sqlfluff==4.2.2
+sqlfluff==4.3.0
```
damonmcc, line 928: pretty sure this is why the linting test is failing. it's failing on files that weren't touched here and `lift` isn't in the [list of tested dbt projects](https://github.com/NYCPlanning/data-engineering/blob/main/.github/workflows/test_helper.yml#L84)

alexrichey: Hrm... I'm gonna go ahead and merge this, and I bet main will be fine (perhaps after another commit to main that doesn't touch the products folder). If it continues to fail, then I'm on the hook to fix!

PR #2579, opened 2026-08-21
improve CI tests and use new issue label

damonmcc:
- **`maintenance` issue label renamed to `platform`** — `Related to tooling and infrastructure`. half the issues it tagged were new work, not upkeep
- **applied automatically now** — title patterns in `issue_metadata.yml`, like the `db-*` labels. 67 issues carry it, up from 21
- **`python` label deleted** — dependabot's, never used on an issue
- **dcpy tests only run when dcpy changes** — the condition was already there, just commented out. the full suite ran on every PR, 5 min of a 6 min run
- **dbt tests only run for changed products** — all 8 projects ran every time

## notes

issues that match a product label don't also get `platform`, so `PLUTO docker image` stays PLUTO's.

dbt test jobs across the last 120 merged PRs: 960 → 85. half of those PRs touch exactly one dbt project.

the new `ceqr` path filter turns on a pytest job that has never run — it may fail.

### DIFF
--- a/.github/issue_metadata.yml
+++ b/.github/issue_metadata.yml
@@ -74,6 +74,32 @@ products:
     label: db-zoningtaxlots
     patterns: ['\bztl\b', '\bzoning[- ]?tax[- ]?lots?\b']
 
+# Non-product labels, keyed by the label name itself. Only applied when no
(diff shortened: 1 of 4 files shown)


alexrichey (approved):
🙏!

PR #2593, opened 2026-09-01
Compile python requirements

damonmcc:


damonmcc:
I think mypy and qa pytests are failing because we use but only happened to get the package git (gitpython) from streamlit. streamlit doesn't require it anymore

so we probably have to explicitly list gitpython now

PR #2595, opened 2026-09-02
Import product-metadata changes for PLUTO metadata project

jackrosacker:
*part 2 of #2312*

### What
Completes the deferred second half of #2312: ports PLUTO GDB metadata overrides into this repo's product-metadata/, corrects a few issues, and deletes the interim test fixture #2312 left in place.

### Why
The product-metadata repo was migrated into data-engineering while I was working on this PLUTO metadata project. This required a second PR to port over the p-m changes for the PLUTO metadata workstream.

### How

`product-metadata/products/pluto/pluto/metadata.yml`:

- Added GDB file entries with layer declarations.
- Added per-format Esri type overrides across all base columns, verified against datasets that we actually publish — which also caught a naming bug where several GDB overrides had inherited shapefile-truncated field names instead of the real ones.
- Removed a handful of stale, no-longer-published columns.

`product-metadata/metadata.yml`:

- Projection switched to the EPSG authority:code format.

### Tests:

- Retired the interim mock fixture and pointed PLUTO's tests at the real repo.
- Brought PLUTO's coverage up to parity with colp's existing standard (both file formats, zip/nonzip, with/without pre-existing metadata).

### What's next

Unfortunately the output metadata from the `write_metadata` CLI command is a little mangled and requires additional work to untangle. I'd still like to merge this PR to fix the p-m YAML changes in place in this repo, and will them come back to improve the outputs.


alexrichey (changes requested):
Is it possible to use the `product-metadata/products/test/` product? (or `product-metadata/products/test_dcpy/` ... hadn't noticed that there were two, but I think we should combine them at some point). It's sort of a pedantic, small thing, but in general we don't want `dcpy` to know anything about our actual products. 

Then bonus points: absolutely not required, but we have the full capability of jinja templating in the product metadata files. Is it possible to use that to remove some of the duplication in the newly added rows?

jackrosacker:
Ready again @alexrichey. Swapped tests over to `product-metadata/products/test_dcpy/`, and reduced field override duplication in the PLUTO `metadata.yml` with a YAML anchor.

Next steps on the following PR will be untangling the output of the `write_metadata` command, cleaning up the PLUTO YAML to serve as a primary reference so we can port the changes to other products, and then wiring all of this into the PLUTO build.

alexrichey (approved):
LGTM!

PR #2602, opened 2026-09-04
Filter out LPC landmarks with no geometry, add data-quality guardrail

alexrichey:
~0.4% of lpc_landmarks records (historic districts, interior landmarks, and multi-site designations) have no geometry in the source archive. This was failing not_null tests on int_spatial__all downstream (historic_resources, historic_resources_adj, shadow_hist_resources), showing up as a nightly build failure. Filter these out in staging, and add a percent_not_null guardrail test so a bigger jump in the null share doesn't go unnoticed.


alexrichey:
@damonmcc FYI, this should fix the nightly build. Though there is one extra issue that may be a race-condition, and may still fail

Issue #2549, opened 2026-08-07
create LIFT data product


damonmcc:
We're adding columns to DCAS's LIFT table:
- DRI Tier
- Capital Projects Total Project count
- Capital Projects Total Spent
- 2020 Census Tract
- NTA

This data is related to the Land Inventory Fast Track (LIFT) task force ([storymap](https://storymaps.arcgis.com/stories/afce9e93d79847608fc846878e17db8e)). it's a list of the "invetory" of city-controlled land that is suitable for housingm development

DCAS's LIFT table has 106 columns and ~15K rows. each row seems to be a lot

we'd love to build the pipeline to produce the entire LIFT table, but for now we're just adding fields to the latest manually created table

# data model

## inputs
- DCAS LIFT
- MapPLUTO
- EDDE DRI
- 2020 Census Tracts to 2020 NTAs and CDTAs Equivalency 
- CPDB

## outputs
- DCP LIFT

# notes

`LIFT_DCAS_Data Fields_20260807` is the last excel file we got from DCAS and is called the "updated DCAS LIFT table". it has 106 columns

`LIFT Master Fields Other Data Sources 20260622` is a list of fields suggested by LIFT agencies earlier this year

since CPDB is a snapshot of the budget, should be clear that any capital project data is related to projects in the latest budget not all-time data

EDDE DRI data is NTA-level and PLUTO doesn't have an NTA column. we need to relate DRIs to tracts to lots

damonmcc:
shared data with DCAS. awaiting DCAS/City Hall follow-up on:
- usability of the new columns
- their need/desire to formalize updates to the dataset via DCP DE pipeline

damonmcc:
we've discussed plans with @croswell81 and DCAS

our short-term goals:
- automate the generation of the LIFT excel file using the source data we currently have (e.g. joined IPIS data and LIFT tracker fields from the LIFT dataset)
- start adding the requested additional fields

our long-term goals
- generate the LIFT excel file using all source data
- develop a dashboard to meet City Hall's needs (e.g. a password protected map of LIFT data)

next steps
1. share plans with City Hall
2. build the short-term pipeline
3. get source data from City Hall (LIFT tracker)
4. get source data from HPD for additional fields
5. get Community District AHFT rankings
6. decide on DCAS data transfer method

dataset details
- source data systems are: DCAS IPIS, DCP datasets (PLUTO, CPDB, DRI, etc.), City Hall LIFT tracker
[...]

PR #2611, opened 2026-09-08
add lift data dictionary and restructure the readme

damonmcc:
related to #2549 

- add latest DCAS LIFT data dictionary excel and convert it to a csv
- update README to improve structure, include latest data product design, and drop artifacts of builds/investigations


alexrichey (approved):
Thanks!

Issue #2613, opened 2026-09-09
Stale library files left in latest/ shadow ingest output


damonmcc:
Ingest writes only `{id}.parquet` and `config.json` to a dataset's `latest/` folder. It doesn't clear what was there before, so every dataset that library ever archived still has that library output sitting in `latest/` next to the current parquet.

81 of 216 ingest datasets have library leftovers in `latest/` (`.sql`, `.csv`, `config.yml`). For 66 of them the leftover `.sql` is older than the parquet, sometimes by most of a year:

- `dcp_developments` — latest `26Q2`, parquet 2026-08-31, `.sql` 2025-11-20
- `dcp_housing` — latest `26Q2`, parquet 2026-08-25, `.sql` 2025-04-01
- `dcp_facilities` — latest `26v1`, parquet 2026-07-17, `.sql` 2025-07-21

That's not only clutter. `RECIPE_FILE_TYPE_PREFERENCE` (`dcpy/lifecycle/builds/plan.py:35`) is `[pg_dump, parquet, csv]`, so a consumer that doesn't set `file_type` explicitly resolves to the stale `.sql`, loads old data, and records the current version against it.

Hit this for real in #2608: pluto reads `doitt_zipcodeboundaries` with no `file_type`, `latest` had just moved to a new ingest archive, and the leftover `.sql` was from a version three years earlier. Worked around it there with an explicit `file_type: parquet`, but the general fix is for ingest to clear non-ingest files from `latest/` when it archives, plus a one-time sweep of the 81.

`admin/ops/library_to_ingest_status_csv.py` already flags this per dataset in the `stale_library_files_in_latest` column.

damonmcc:
**TLDR** — `latest/` accumulates files instead of being replaced, so a library `.sql` left beside a new parquet silently wins file-type resolution. Two one-line-ish fixes in #2618 stop it recurring. For the 81 datasets already in that state, 45 just need `latest/` cleared, 6 need library files moved to `<version>_library/` first, and 30 need a rename plus an ingest run. A script does the first two and verifies every file it deletes exists elsewhere. Avoid `--overwrite` on library-era versions: it destroys library's `config.json` and `.parquet`.

---

Two code changes and a sweep, in that order. Prototyped in #2618.

**1. `latest/` should be replaced, not added to.**

[...]

PR #2618, opened 2026-09-10
replace latest instead of adding to it

damonmcc:
part of #2613

`latest/` accumulates instead of being replaced. `_push` writes two files into it and leaves whatever else is there, so a library-era `.sql` ends up beside the new parquet and wins `RECIPE_FILE_TYPE_PREFERENCE`, which matches on extension alone. Consumers without an explicit `file_type` then silently keep reading the older archive. That bit three times in #2614. Pushing `latest/` as a folder replaces it wholesale, reusing the replace semantics `PathedStorageConnector.push` already has for directories.

`--latest` also needs to work on its own. Archival is skipped when a version already exists with unchanged data, and `latest` was only ever written inside that push, so re-running with `--latest` could leave the pointer on an older version and still report success.

Neither change helps a dataset that isn't re-ingested, so `admin/ops/clear_library_files_from_latest.py` handles the 81 existing folders. It reports by default and sorts them three ways:

- **clear_latest (45)** — `<version>/` is already clean, so only `latest/` needs clearing
- **mixed_version (6)** — `<version>/` is mixed too, so library files move to `<version>_library/` first
- **needs_ingest (30)** — never archived through ingest; needs a rename plus an ingest run, so `--apply` leaves these alone

Before deleting anything from `latest/` it checks each file has a byte-identical copy in a version folder, and skips any that doesn't. `latest/` is a copy of some archive, so that should always hold, but it's the one way this could actually lose data.

All 6 `mixed_version` datasets got that way from `--overwrite`, four of them during #2614. Worth knowing that flag is lossy rather than just messy: it replaces only the files ingest writes, so library's own `config.json` and `.parquet` are overwritten while its `.sql` and `.csv` survive.

### DIFF
--- a/admin/ops/clear_library_files_from_latest.py
+++ b/admin/ops/clear_library_files_from_latest.py
@@ -14,16 +14,17 @@ Datasets fall into three cases, and only the first is safe to fix by clearing:
                  still exists in its own version folder, so clearing loses nothing.
 
   mixed_version  `latest/` holds an ingest config but `<version>/` is mixed too. Caused
-                 by `--overwrite`, which replaces only the files ingest writes. Clearing
-                 `latest/` would leave `<version>/` still resolving to the `.sql`, so
-                 these want a re-ingest at a fresh version instead.
+                 by `--overwrite`, which replaces only the files ingest writes. The
+                 library files move to `<version>_library/` so both folders come out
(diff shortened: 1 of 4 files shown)


alexrichey (approved):
LGTM!

PR #2635, opened 2026-09-16
Fix BYTES connector (add a user agent)

alexrichey:


alexrichey:
@jackrosacker well, this explains my failed demo of version listing. But just FYI, you can invoke 
`dcpy lifecycle connectors get_versions bytes pluto.pluto` to list versions of PLUTO, or other products on Bytes

